PR #15, opened 2015-01-04
CPU load for Mac. Linux and windows needs testing.

kamenitxan:
https://github.com/dblock/oshi/issues/14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,9 @@ Next
 * [#3](https://github.com/dblock/oshi/pull/3): Mavenized project - [@le-yams](https://github.com/le-yams).
 * Added Travis-CI - [@dblock](https://github.com/dblock).
 * Added TODO list and enhanced README documentation - [@ptitvert](https://github.com/ptitvert)
+* [#15](https://github.com/dblock/oshi/pull/15): Added support for CPU load in OSX and Linux [@kamenitxan](https://github.com/kamenitxan)
 * Your contribution here.
-* Added support for CPU load in OSX and Linux[@kamenitxan](https://github.com/kamenitxan)
+
 
 
 1.1 (10/13/2013)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Next
 * Added Travis-CI - [@dblock](https://github.com/dblock).
 * Added TODO list and enhanced README documentation - [@ptitvert](https://github.com/ptitvert)
 * Your contribution here.
-* Added support for CPU load in OSX [@kamenitxan](https://github.com/kamenitxan)
+* Added support for CPU load in OSX and Linux[@kamenitxan](https://github.com/kamenitxan)
 
 
 1.1 (10/13/2013)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Next
 * Added TODO list and enhanced README documentation - [@ptitvert](https://github.com/ptitvert)
 * Your contribution here.
 
+1.2 (2/1/2014)
+================
+* Added support for CPU load in OSX
+
 1.1 (10/13/2013)
 ================
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,8 @@ Next
 * Added Travis-CI - [@dblock](https://github.com/dblock).
 * Added TODO list and enhanced README documentation - [@ptitvert](https://github.com/ptitvert)
 * Your contribution here.
+* Added support for CPU load in OSX [@kamenitxan](https://github.com/kamenitxan)
 
-1.2 (2/1/2014)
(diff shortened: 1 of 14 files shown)


dblock:
The build needs to pass, to start with. I can check the OSX version after that.

src/main/java/oshi/util/ExecutingCommand.java
```diff
@@ -35,6 +33,7 @@
 		} catch (IOException e) {
 			return null;
 		}
+		p.destroy();
```
dblock: Can you explain this?

src/test/java/oshi/SystemInfoTest.java
```diff
@@ -46,6 +46,13 @@ public void testGetMemory() {
         assertTrue(memory.getAvailable() <= memory.getTotal());
     }
 
+    @Test
+    public void testCpuLoad() {
+        SystemInfo si = new SystemInfo();
+        HardwareAbstractionLayer hal = si.getHardware();
+        System.out.println(hal.getProcessorLoad());
```
dblock: This isn't a test, it just prints it out. What we want is something that at least checks that the output is a non-zero % number for example.

dblock:
Please also update CHANGELOG and README.

dblock:
Now, stepping back from the implementation above, getProcessorLoad is too specific, especially if it returns a number. You want to create a structure that returns a lot of different load-related information that we can extend in the future. Top returns load averages, cpu usage for user, system, idle, etc. So something that reflects that closer.

src/main/java/oshi/util/ExecutingCommand.java
kamenitxan: removed p.waitFor(); so it works with top command. 
p.destroy(); just to be safe, that process is dead.

dblock: I don't think you can do this. The process spawns and then just hangs there and you don't stop it. You need to find a  way to run something that returns. Maybe using `top` isn't the best way? On Linux I believe /proc/stat can give you a lot of this, no idea about OSX.

kamenitxan:
Here are changes: https://github.com/kamenitxan/oshi/commit/5a412bd8eb54f8d4cba8c6a1ae9c7edda234a04f
Tests are ok and build too.

What about extending Processor interface? Then this would be called like this
 ```hal.getProcessors()[0].getProcessorLoad() ```

Also CentralProcessor classes implements Processor, but CentralProcessor has more methods. Shouldn't they be in Processor too?

dblock:
Travis runs Linux only, we have to have an implementation, can't have CI go red, so a Linux implementation is needed for anything to be merged.

I think it's a good idea to revisit these classes.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ Next
 * Added TODO list and enhanced README documentation - [@ptitvert](https://github.com/ptitvert)
 * Your contribution here.
 
+1.2 (2/1/2014)
+================
+* Added support for CPU load in OSX
```
dblock: This should be under Next and should just be a bullet with your name (same format as above please).

src/main/java/oshi/util/ExecutingCommand.java
```diff
@@ -21,7 +21,7 @@
 		} catch (IOException e) {
 			return null;
 		} catch (InterruptedException e) {
-			return null;
+			e.printStackTrace();
```
dblock: This is not necessary.

kamenitxan:
I will get Linux and fix it there tomorow.

I will also add methods from CentralProcessor to Processor interface.

kamenitxan:
Fixed and added.

Before I add cpuload to Processor interface I think, it is also good idea to change hal interface.
Now it would be ```hal.getProcessors()[0].getLoad()```

I would like to make ```hal.getProcessor``` which returns first CPU (getProcessors()[0])
Then it would be ```hal.getProcessor().getLoad()```

dblock:
Why would you want a single processor when you have multiple? 

If CPU load is a combination on top of multiple processors, then it should probably not be under getProcessor, no?

dblock:
The code here is much better. I don't think I want an explicit `getProcessorLoad` that returns a float, so lets write a more sensible interface and this will become mergeable.

kamenitxan:
Because I have 1 CPU with more cores. 
I moved  ```getLoad() ``` to Processor interface -> ```hal.getProcessors()[0].getLoad();```
Someone could implement getLoad for cores, but I think it's not so needed.

dblock:
I like this, merging.

dblock:
I just noticed, is the README still correct? Shouldn't the load be per CPU? Please feel free to PR a fix if I am right.

kamenitxan:
Readme is correct. Load is for whole CPU, not for single CPU core.

dblock:
:+1:

PR #18, opened 2015-01-19
CPU load implementation for windows

Sorceror:
Squashed commit:

[f8d2589] Code altered to suit more existing codebase (error management), added english localization flag to wmic command to prevent problems with parsing on non-english localized machines

[890ad66] CPU load implementation for windows

[39b933f] Added ability to read different than just first result line from command output (with proper sanitization) (+1 squashed commits)


dblock:
Good, merging.

Issue #21, opened 2015-04-17
fix the note about sigar on title page


donarus:
Sigar is no longer GPL v2 but is licensed under Apache licence from 1.6.4

see: https://github.com/hyperic/sigar/blob/master/LICENSE

I think you it should be corrected on title page on GitHub

dblock:
Please make a PR, thanks!

dblock:
Closed via https://github.com/dblock/oshi/pull/22.

PR #22, opened 2015-04-20
Fixed Sigar URL and note about Sigar licensing in README.md

donarus:
This PR is fixing #21 

* fixed Sigar URL
* fixed Sigar licensing info


dblock:
Merging, thanks. I think the license is no longer as relevant, when Sigar was GPL that was a big deal for a lot of companies, with their more recent change it's less of a problem.

PR #25, opened 2015-05-18
Include inactive memory in mac/linux available.

### LINKED ISSUE #23: Free memory vs. Available memory
GlobalMemory.getAvailable() only reports "free" memory for Mac and Linux implementations.  Unfortunately, this is rarely useful except immediately upon reboot.  After a short period of usage, free drops to near-zero and much of the memory released by programs resides in a cache for easy access.

Available memory includes both "free" memory (never used or purged) and "inactive" (Mac/Linux) or "standby" (Windows) memory.

The Windows implementation of getAvailable() correctly uses GlobalMemoryStatusEx()'s ullAvailPhys value, which includes both free and standby memory in its total. 

For the mac branch (vm_stat) you would need to add the value of "Pages inactive:" to the existing "Pages free:" calculation.  For linux (proc/meminfo), you need to add the value returned by "Inactive:" to the existing  "MemFree:" or "MemAvailable:".

dbwiddis:
Closes #23.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Next
 * Added Travis-CI - [@dblock](https://github.com/dblock).
 * Added TODO list and enhanced README documentation - [@ptitvert](https://github.com/ptitvert)
 * [#15](https://github.com/dblock/oshi/pull/15), [#18](https://github.com/dblock/oshi/pull/18): Added support for CPU load - [@kamenitxan](https://github.com/kamenitxan), [@Sorceror](https://github.com/Sorceror).
+* [#25](https://github.com/dblock/oshi/pull/25): Include inactive memory amount in GlobalMemory#getAvailable on Mac/Linux - [@dbwiddis](https://github.com/dbwiddis).
 * Your contribution here.
 
 
--- a/src/main/java/oshi/software/os/linux/proc/GlobalMemory.java
+++ b/src/main/java/oshi/software/os/linux/proc/GlobalMemory.java
@@ -14,7 +14,7 @@ import java.util.Scanner;
 import oshi.hardware.Memory;
 
 /**
- * Memory obtained by GlobalMemoryStatusEx.
+ * Memory obtained by /proc/meminfo.
  * 
  * @author alessandro[at]perucchi[dot]org
  */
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Next
 * Added Travis-CI - [@dblock](https://github.com/dblock).
 * Added TODO list and enhanced README documentation - [@ptitvert](https://github.com/ptitvert)
 * [#15](https://github.com/dblock/oshi/pull/15), [#18](https://github.com/dblock/oshi/pull/18): Added support for CPU load - [@kamenitxan](https://github.com/kamenitxan), [@Sorceror](https://github.com/Sorceror).
+* Include inactive memory in mac/linux available - [@dbwiddis](https://github.com/dbwiddis).
```
dblock: Make this something that looks like the line above:

```
[#25](https://github.com/dblock/oshi/pull/25): Include inactive memory amount in GlobalMemory#getAvailable on Mac/LLinux - [ ... ]
```

You can commit with `--amend` and force push up to avoid having to re-squash things.

dblock:
It may be hard, but is there a way to write a test here that would have failed without the code change? Try something?

CHANGELOG.md
dbwiddis: Unfortunately I didn't have the pull request # until I committed (I suppose I could have guessed).   And since I've been committing via svn, I am really unfamiliar with the git command lines to do this. Argh.  I'll do my best, I guess... you're really making me have to learn new tools!

dblock: It's an opportunity!

We force this strange construction where you commit, then push, then make a pull request, then amend. It's weird, but it's the best we have for changelog entries that seems to make the customer (the developer reading the changelog) happy.

dbwiddis: Well, that was painful, until I figured out to use the -f switch on my push... I think this is finally all set.

dbwiddis:
For testing, we can try to parse other memory commands.
* on Linux the code draws from /proc/meminfo.  We could parse the output from "free -m" instead and look at the "free" column. The old version matches the first "Mem:" line, the fixed version would match the second "-/+ buffers/cache"" line.  (We could also pull info from vmstat or top.)
* on Mac the code draws from vm_stat.  We can pull from top (top -l 1 | grep PhysMem) to find the old, incorrect value.  I don't know if there's any other command line version to get the new version, but it's available on system information display.
Unforunately, for both tests, there's no way to guarantee that a system will have nomzero inactive memory.  Right after reboot, the numbers may be similar/equal.

dblock:
Merged, looking forward to more!

PR #26, opened 2015-05-19
Use JNA instead of command parsing to get Mac OS X memory statistics.

dbwiddis:


dbwiddis:
Intended for feedback/comment.  One concern is 32 bit vs. 64 bit Mac, but debating whether to go to all the effort of 32-bit JNA calls vs. just defaulting to text parsing if the 64-bit one doesn't work.

dbwiddis:
One day I'll learn git.  Apparently a hard reset wiped this out.

PR #27, opened 2015-05-19
Replace all Mac OS X command line parsing with JNA or System properties

dbwiddis:


dbwiddis:
For review/feedback.  I'm working on the rest of the Mac port, so if you want them all in this pull request, you can wait.

src/main/java/oshi/software/os/mac/local/GlobalMemory.java
```diff
@@ -8,41 +8,80 @@
...
+							SystemB.HOST_VM_INFO64_COUNT));
+			if (ret == 0) {
+				availableMemory = (vmStats.free_count + vmStats.inactive_count)
+						* pageSize;
+				totalMemory = (vmStats.free_count + vmStats.inactive_count
+						+ vmStats.wire_count + vmStats.active_count + vmStats.compressor_page_count)
+						* pageSize;
+			}
+		} catch (UnsatisfiedLinkError e) {
+			for (String line : ExecutingCommand.runNative("vm_stat")) {
```
dblock: Why would the above fail? I don't think fallback on an old method is necessary just because it was there.

src/main/java/oshi/software/os/mac/local/SystemB.java
```diff
@@ -0,0 +1,78 @@
...
+import com.sun.jna.Structure;
+import com.sun.jna.ptr.IntByReference;
+import com.sun.jna.ptr.LongByReference;
+
+/**
+ * Memory obtained by vm_stat and sysctl
+ * 
+ * @author widdis[at]gmail[dot]com
+ */
+public interface SystemB extends Library {
```
dblock: I want you to contribute this into JNA first, it will be a bit more work, but you get the same code reviewer (me) :)).

dblock: See https://github.com/twall/jna/blob/master/www/Contributing.md.

src/main/java/oshi/software/os/mac/local/GlobalMemory.java
dbwiddis: It would fail if the system did not have the host_statistics64() method, such as a 32 bit system.  I originally had an is64Bit() check instead.

src/main/java/oshi/software/os/mac/local/SystemB.java
dbwiddis: Well I'm continuing to add to that one as I go along so I don't think it's ready for submitting to a more widely used library...

src/main/java/oshi/software/os/mac/local/GlobalMemory.java
dbwiddis: Let me just (properly) do a host_statistics() method for the 32 bit systems.

I'm struggling with what to do with CPU usage.  There's two data points I can use... load factor (system average) or counting ticks, which requires storing an array of the ticks, waiting, and then reporting back.  Which is fine for the end user but a library should not make the user wait an arbitrary amount of time.  We should either return ticks or load average (divide load avg/num CPUs and you get approximate CPU usage.)

Thoughts?

src/main/java/oshi/software/os/mac/local/SystemB.java
dblock: I think *anything* for OSX would be great in JNA, there's a lot of demand and nobody is taking the time to develop it.

src/main/java/oshi/software/os/mac/local/GlobalMemory.java
dblock: Definitely wouldn't want the call to sit there waiting. Load/num CPUs seems reasonable to me.

src/main/java/oshi/software/os/mac/local/SystemB.java
dblock: Btw, I would take the code here, but add a TODO: move this to JNA.

src/main/java/oshi/software/os/mac/local/CentralProcessor.java
```diff
@@ -185,9 +192,35 @@ public void setFamily(String _family) {
...
+						SystemB.HOST_CPU_LOAD_INFO_COUNT));
+		if (ret == 0) {
+			for (int i = 0; i < SystemB.CPU_STATE_MAX; i++) {
+				prevTicks[i] = cpuLoadInfo.cpu_ticks[i];
+			}
+		}
+		try {
+			Thread.sleep(1000);
+		} catch (InterruptedException e) {
+		}
```
dblock: I think it's crazy, but what do I know :)

src/main/java/oshi/software/os/mac/local/GlobalMemory.java
```diff
@@ -8,41 +8,68 @@
...
+		int machPort = SystemB.INSTANCE.mach_host_self();
+
+		LongByReference pPageSize = new LongByReference();
+		int ret = SystemB.INSTANCE.host_page_size(machPort, pPageSize);
+		if (ret == 0) {
+			pageSize = pPageSize.getValue();
+		}
+
+		if (Platform.is64Bit()) {
+			VM_STATISTICS64 vmStats = new VM_STATISTICS64();
```
dblock: Isn't `VM_STATISTICS64` an extension of `VM_STATISTICS`? In that case you could inherit or compose one with the other, then the `availableMemory` math could be the same.

src/main/java/oshi/software/os/mac/local/GlobalMemory.java
```diff
@@ -8,41 +8,68 @@
...
+			ret = SystemB.INSTANCE.host_statistics64(machPort,
+					SystemB.HOST_VM_INFO64, vmStats, new IntByReference(
+							SystemB.HOST_VM_INFO64_COUNT));
+			if (ret == 0) {
+				availableMemory = (vmStats.free_count + vmStats.inactive_count)
+						* pageSize;
+				totalMemory = (vmStats.free_count + vmStats.inactive_count
```
dblock: This is unused.

src/main/java/oshi/software/os/mac/local/GlobalMemory.java
```diff
@@ -8,41 +8,68 @@
...
 /**
- * Memory obtained by GlobalMemoryStatusEx.
+ * Memory obtained by host_statistics (vm_stat) and sysctl
  * 
- * @author alessandro[at]perucchi[dot]org
+ * @author widdis[at]gmail[dot]com
  */
 public class GlobalMemory implements Memory {
 
 	private long totalMemory = 0;
```
dblock: I see you piggy back on `totalMemory` in `getTotal`, but I think it would be clearer if both methods just make the system calls, there's no need to do premature optimization here IMO.

src/main/java/oshi/software/os/mac/local/SystemB.java
dbwiddis: Fair enough.

src/main/java/oshi/software/os/mac/local/CentralProcessor.java
dbwiddis: I could display load factor / nCPUs for a decent approximation without waiting 1000 milliseconds to count cpu ticks...

src/main/java/oshi/software/os/mac/local/GlobalMemory.java
dbwiddis: As you noted later, it sets the class variable for premature optimization.  Removing it would remove the need for the 64 bit VM_STATISTICS variant which will simplify oshi code (but I could leave the 64 bit version in the SystemB class with an eye towards later use in JNA library.

src/main/java/oshi/software/os/mac/local/GlobalMemory.java
dbwiddis: The class variable and conditional in getTotal were already there, so I figured that was the intent.  I'll simplify.

dbwiddis:
Fun fact. If your project is intended as a primary interface with the JNA package, you probably shouldn't duplicate JNA class names like Memory, making importing awkward...

dbwiddis:
Got all the memory stuff and everything asseccible with sysctl.  Next project is sw_vers ...

src/main/java/oshi/software/os/mac/local/CentralProcessor.java
```diff
@@ -32,8 +37,14 @@
 	 * @return Processor vendor.
 	 */
 	public String getVendor() {
-		if (_vendor==null)
-			_vendor=ExecutingCommand.getFirstAnswer("sysctl -n machdep.cpu.vendor");
+		if (_vendor == null) {
+			int[] mib = { SystemB.CTL_MACHDEP, SystemB.MACHDEP_CPU,
+					SystemB.MACHDEP_CPU_VENDOR };
+			IntByReference size = new IntByReference(100);
```
dblock: Is this a documented max constant? I am not familiar with the Mac API, Is there a way to do the call in 2 phases where you get the size first, alloc memory then do it again?

src/main/java/oshi/software/os/mac/local/CentralProcessor.java
```diff
@@ -32,8 +37,14 @@
...
 	 */
 	public String getVendor() {
-		if (_vendor==null)
-			_vendor=ExecutingCommand.getFirstAnswer("sysctl -n machdep.cpu.vendor");
+		if (_vendor == null) {
+			int[] mib = { SystemB.CTL_MACHDEP, SystemB.MACHDEP_CPU,
+					SystemB.MACHDEP_CPU_VENDOR };
+			IntByReference size = new IntByReference(100);
+			Pointer p = new Memory(size.getValue());
+			SystemB.INSTANCE.sysctl(mib, mib.length, p, size, null, 0);
```
dblock: Does this not ever fail?

dblock:
Other than the checking for the return values of those system calls this would be good to go.

src/main/java/oshi/software/os/mac/local/CentralProcessor.java
dbwiddis: Yes, I believe you can call once sending null for the pointer to get the size, and call again.  I'll address this.  I found that constant 100 in some other Apple code, which isn't exactly documentation but it's an example... moot point if I do two calls.

src/main/java/oshi/software/os/mac/local/CentralProcessor.java
dbwiddis: All the API calls we're using return an integer, 0 if successful and 1 if not.  I will put in checks to provide sensible defaults.

dblock: When the call fails you want to raise an exception.

dbwiddis:
Return values checked, and calling with null to get string lengths.

dbwiddis:
Exceptions added. I think we can stick a fork in this.

dblock:
Looks good, merging.

dbwiddis:
Thanks for patience and instruction as I learned git, re-learned C and pointers, etc.

dbwiddis:
Hey @dblock, found a bug in this PR exactly 11 years later.  How much I've learned since then!

dblock:
> Hey @dblock, found a bug in this PR exactly 11 years later. How much I've learned since then!

Crazy to think that we have AI now.

PR #29, opened 2015-05-24
Use sysinfo for linux total memory

dbwiddis:
Submitted for comment/discussion.  No feelings will be hurt if it's rejected. :)   Some comments:
- Only Linux function which is easily exposed to JNA is sysinfo, and that's really only good for Total ram.  You can also get Free ram from it, but as previously noted, that doesn't account for the true amount available.
- I tried various methods to get the other components of available.  I managed inactive using the meminfo() function (updates global variables) and then reading the kb_inactive global variable.  Unfortunately the necessary library for these was "libprocps.so.3".  I am not sure we should specify the version (version 4 may be out soon) and my system I set up as a testbed (Ubuntu 14.04 LTS) did not have a link from the non-versioned extension (so just plain "procps" didn't work to locate the library), so I'm not sure how portable this would be.
- Even when I got the inactive, the sum of free+inactive didn't equal available on my system. I also had "reclaimable".  Digging into the source code, I also found various minimum thresholds before swap space was used.   
The bottom line, parsing /proc/meminfo is fast, and it's really a virtual filesystem intended to provide access to all the kernel info we're trying to get, so it (and the other /proc file reading elsewhere in the linux patch) is really probably the way to go.  And it's reasonable to just go back to that for total as well.

The only method that might benefit from changing is the total ram call, which is shown in this pull request.  You'll notice a few things:
- I had to write two versions of the structure, because on a 64-bit system, the structure copied from the linux source includes a zero-length array that JNA chokes on.  (And I can't even extend one to get the other because the fields are loaded from children first and then parent, so the mapping to the C structure would be wrong.)
- I finally realized the reason for the class variable totalmemory, and the if (totalmemory==0) check.  It, like all the fields in the CentralProcessor class, keeps the old value rather than always fetching the new value.   So I should, perhaps, put that conditional back in the Mac OS version of things.  

I did find a file not closed (CentralProcessor line 181) so there may be at least one line that needs to be kept from this PR even if I roll back everything else.

Finally, lots of format changes creating diffs.  Do you have a coding conventions/formatting conventions page to prevent this from making diffs unreadble?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Next
 * [#15](https://github.com/dblock/oshi/pull/15), [#18](https://github.com/dblock/oshi/pull/18): Added support for CPU load - [@kamenitxan](https://github.com/kamenitxan), [@Sorceror](https://github.com/Sorceror).
 * [#25](https://github.com/dblock/oshi/pull/25): Include inactive memory amount in GlobalMemory#getAvailable on Mac/Linux - [@dbwiddis](https://github.com/dbwiddis).
 * [#27](https://github.com/dblock/oshi/pull/27): Replace all Mac OS X command line parsing with JNA or System properties - [@dbwiddis](https://github.com/dbwiddis).
+* [#29](https://github.com/dblock/oshi/pull/29): Replace Linux totalMemory calculation with JNA and revised MemAvailable backup - [@dbwiddis](https://github.com/dbwiddis).
 * Your contribution here.
 
 
--- a/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
+++ b/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
@@ -9,21 +9,22 @@ package oshi.hardware;
 
 /**
  * A hardware abstraction layer.
+ * 
  * @author dblock[at]dblock[dot]org
  */
 public interface HardwareAbstractionLayer {
 
 	/**
 	 * Get CPUs.
-	 * @return
-	 *  An array of Processor objects.
+	 * 
+	 * @return An array of Processor objects.
 	 */
 	Processor[] getProcessors();
-	
+
 	/**
 	 * Get Memory information.
-	 * @return
-	 *  A memory object.
+	 * 
+	 * @return A memory object.
 	 */
 	Memory getMemory();
 
--- a/src/main/java/oshi/hardware/Memory.java
+++ b/src/main/java/oshi/hardware/Memory.java
@@ -8,24 +8,27 @@
 package oshi.hardware;
 
 /**
- * Memory refers to the state information of a computing system, as it is kept active in some physical 
- * structure. The term "memory" is used for the information in physical systems which are fast (ie. RAM),
- * as a distinction from physical systems which are slow to access (ie. data storage). By design, the term
- * "memory" refers to temporary state devices, whereas the term "storage" is reserved for permanent data.
+ * Memory refers to the state information of a computing system, as it is kept
+ * active in some physical structure. The term "memory" is used for the
+ * information in physical systems which are fast (ie. RAM), as a distinction
+ * from physical systems which are slow to access (ie. data storage). By design,
+ * the term "memory" refers to temporary state devices, whereas the term
+ * "storage" is reserved for permanent data.
+ * 
  * @author dblock[at]dblock[dot]org
  */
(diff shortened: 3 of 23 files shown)


dblock:
As always I think using JNA native is better than running command line tools. The 32/62 problem can likely be fixed with a bit of inheritance I think, but haven't tried.

Check the build, it's failing on Travis-CI for something that looks legit.

For formatting I've always used default Eclipse's.

dbwiddis:
The problem with inheritance is that the 32-bit version has the extra 0-length padding variable, so would need to inherit from the 64-bit version.  But internally when mapping structures to fields, children are read before parents, so the padding would end up at the beginning of the structure rather than the end of it.   I'm still trying to think of ways it can be done...

Travis-CI is complaining about me including a method with a body (isSysinfo64) inside an interface.  I suppose that's a valid complaint.  

I'll make sure I'm using the eclipse default.

dbwiddis:
I did a project-wide code format per Eclipse built-in (what I had been using)  and left it as a separate commit so the remaining diffs are more applicable to this discussion.

So for totalmemory -- code works, but open question if there's a better way to do it. I've got a question out on the mailing list about the zero-length array.

For availablememory -- to be most portable I'd want to try a few libs in sequence:  "procps" first, and then if we get the UnsatisfiedLinkError try libprocps.so.3 (although that would break when a .4 comes out.)  Does that sound reasonable?   (I'm not sure I can put a try/catch block in an interface initialization, though, without Travis-CI complaining.)

Of course, then there's still the decision on what to add up.   /proc/meminfo's MemAvailable does a lot of heavy lifting (when it's there, on more recent distributions).  When it's not there, you have to add various other components, specifically free, inactive and reclaimable.   

That gives a different response than "free -m" which gives you free + buffers + cached.  This is also what "top" produces.  I guess it's all an estimate, anyway, so it doesn't matter too much.

The author of this StackOverflow page seems to have done a lot more work than me researching this: http://stackoverflow.com/questions/63166/how-to-determine-cpu-and-memory-consumption-from-inside-a-process .  Like me, he appears to have concluded that other than the system calls (sysinfo is the only useful one for our purposes), reading the /proc virtual filesystem is the best and most portable option.

dbwiddis:
I was able to work around the 32/64-bit problem by just adding the 8 bytes of padding anyway. That will ensure enough memory is allocated, and in 64 bit systems the structure will just leave the last 8 bytes untouched.

The longer I work on trying to avoid using the /proc filesystem the more I see trying to shortcut it is going to cause problems.
* trying to load the procps library, as noted above, isn't consistent across distributions.  The Travis CI build is failing because it can't find libprocps.so.3.
* Even when it does load on my system, the initialization routine is producing "2.4+ kernel w/o ELF notes? -- report this\n".  I don't see any way around this.
* There are insufficient functions available in non-procps library systems to get the same information.
I believe Linux has designed the /proc filesystem for exactly the implementation that I'm trying to avoid, and we should just use it, except where a system call can be made to get the information.  sysinfo() really only gets us total ram.  uname() is available for a few naming formats but is inconsistent with what we're already pulling so we'll have to decide what we want it to display.

dbwiddis:
This commit message, by Linus Torvalds himself, is pretty compelling reason to use /proc/meminfo for available memory.  https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773

It also helpfully identifies the other components we should add up as a fallback for systems prior to this MemAvailablel patch: MemFree, Active(file), Inactive(file), and SReclaimable.  (I'd figured as much by reading the source and trying in vain to replicate it.)

dbwiddis:
While I think I'm throwing in the towel for any other system calls, I have discovered an inefficiency with our /proc filesystem reads.  It appears that every time any /proc file is read, even if it's a single byte, the kernel regenerates the data.  This can take a (relatively, in microseconds) long time for meminfo with all the pages it has to count (and could in theory lead to inconsistent results.)   Advice is to read the entire file into Java memory once, and then parse it.  So we'll need to change the current implementation with Scanner reading line-by-line.

dbwiddis:
This is actually ready for your final review/acceptance now.  Let me know if you want anything squished or removed from this pull.  I intentionally kept the reformat commit separate, but the two little fixes could be squished.

dblock:
This is good, merging.

Issue #28, opened 2015-05-21
API is inconsistent on CPU vs. Processor


dbwiddis:
Most modern machines have multi-core processors.  However, the API mixes terminology/code.  The Processor class has system-wide statistics, so should really only be instantiated once.  However, the HardwareAbstractionLayer refers to Processors in the plural, and implementations set up a loop to instantiate multiple (identical) copies of these virtual CPUs.

The multiple copies seem unnecessary with the current implementation.  However, there are some statistics, such as CPU load, which can be measured per-virtual CPU (although in the tick-counting between timestamps way, so iteratively calling all processors and waiting to count ticks is not a good model.)

Are we stuck with the current API or can we revise the structure?  And while we're at it, any chance we can avoid using the same class name as major JNA classes like "Memory"?

dblock:
We can revise the structure/API by bumping the major version of the library and writing a proper UPGRADING doc.

Underbalanced:
Additionally the CPU getProcessors()[val].getLoad() is highly inaccurate on multicore systems as this will only show that specific processor; 4 cores or more one core may register higher then the rest as such happens on my 32 core programming machine when testing this Library.

Further an issue, is that the current getLoad() already takes to much time; this would be alright if not having to loop through the processors and obtain an average CPU load correctly.

I am interested in contributing where possible as I find ways to improve it.

dbwiddis:
Yes, the com.sun.management.OperatingSystemMXBean is pretty capable, and I only found oshi in my search to more accurately represent the available memory, which that MXBean (and oshi until my contribution) didn't model well.  If I were designing oshi from scratch I'd extend that MXBean...

Interestingly if you use that MXBean when you first start up the JVM, it doesn't accurately represent CPU usage. It needs a few microseconds to start collecting those tick averages.  We could consider a model where oshi retained tick state and updated the CPU % with each successive call.  Or we could calculate the load average and divide by number of cores and get pretty close, too.   Unfortunately, these are implementation choices and it should not fall on us to specify which version the user should use; we should provide a means to all available information.

Of course, Windows doesn't have all the same info that Linux and Mac do.  The bottom line is we should decide how we want to represent these things, and do so consistently and usefully (particularly in the way of providing functionality that does not exist elsewhere.)

Underbalanced:
[...]

PR #30, opened 2015-06-03
Support for processor frequency

alessiofachechi:
Added the `getFreq` and `setFreq` methods to the `Processor` interface and for its implementations for Linux, Windows and Mac.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 # Output
 /target/
+/bin/
+/.project
+/.settings/
+/.classpath
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Next
 * [#15](https://github.com/dblock/oshi/pull/15), [#18](https://github.com/dblock/oshi/pull/18): Added support for CPU load - [@kamenitxan](https://github.com/kamenitxan), [@Sorceror](https://github.com/Sorceror).
 * [#25](https://github.com/dblock/oshi/pull/25), [#29](https://github.com/dblock/oshi/pull/29): Included inactive/reclaimable memory amount in GlobalMemory#getAvailable on Mac/Linux - [@dbwiddis](https://github.com/dbwiddis).
 * [#27](https://github.com/dblock/oshi/pull/27): Replaced all Mac OS X command line parsing with JNA or System properties - [@dbwiddis](https://github.com/dbwiddis).
+* [#30](https://github.com/dblock/oshi/pull/30): Added processor vendor frequency information - [@alessiofachechi](https://github.com/alessiofachechi).
 * [#32](https://github.com/dblock/oshi/pull/32): Added battery state information - [@dbwiddis](https://github.com/dbwiddis).
 * Your contribution here.
 
--- a/src/main/java/oshi/hardware/Processor.java
+++ b/src/main/java/oshi/hardware/Processor.java
@@ -44,6 +44,18 @@ public interface Processor {
 	 *            Name.
 	 */
 	void setName(String name);
+	
+	/**
+	 * Vendor frequency (in Hz).
+	 * @return Processor frequency.
+	 */
+	long getVendorFreq();
+	
+	/**
+	 * Set processor vendor frequency (in Hz).
+	 * @param freq Frequency.
+	 */
+	void setVendorFreq(long freq);
 
 	/**
 	 * Identifier, eg. x86 Family 6 Model 15 Stepping 10.
--- a/src/main/java/oshi/software/os/linux/proc/CentralProcessor.java
+++ b/src/main/java/oshi/software/os/linux/proc/CentralProcessor.java
@@ -11,9 +11,12 @@ import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.util.ArrayList;
 import java.util.Scanner;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import oshi.hardware.Processor;
 import oshi.util.FormatUtil;
+import oshi.util.ParseUtil;
 
 /**
  * A CPU as defined in Linux /proc.
@@ -27,7 +30,8 @@ public class CentralProcessor implements Processor {
 	private String _stepping;
 	private String _model;
 	private String _family;
-	private boolean _cpu64;
+	private Long _freq = null;
+	private Boolean _cpu64;
 
 	/**
 	 * Vendor identifier, eg. GenuineIntel.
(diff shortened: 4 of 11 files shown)


alessiofachechi:
@dbwiddis ups, I probably was sleeping when I wrote 1024 instead of 1000! :sweat_smile: 
Anyway yes, it should be better to rename the `getFreq` and `setFreq` in `getMaxFreq` and `setMaxFreq`. I will do this soon!
Thanks for your time! :smiley:

dbwiddis:
I hope you're still working on this.  I'm looking into making a 1.2 release and would be great if we can add this first.

If you can merge the latest version here with your version it will help out a lot with your development: you'll see the necessary code to do the sysctl() calls on Mac OS X, and the frequency one should be an easy copy/paste from the other examples.  Also for windows, the PowrProf library and method has been added so you'd just need to add the new informationlevel (11) and structure.  See the new PowerSupply code for example and see https://msdn.microsoft.com/en-us/library/windows/desktop/aa372675(v=vs.85).aspx for more details.

Also squashing your commits will make it easier to review!

alessiofachechi:
Yeah @dbwiddis, sure. I will fix  this weekend and improve with your advices the code (sorry again for the bug, I coded it very fastly for a project I work on).

alessiofachechi:
Hi @dbwiddis, did you notice that last upstream commits are not 1.5 compatible?

dbwiddis:
Nobody uses 1.5 anymore:  https://plumbr.eu/blog/java/java-version-statistics-2015-edition

Java 1.6 reached end of supported life in February 2013.  I personally don't find any compelling reason to code for anything earlier than 1.7 (which itself has just reached end of supported life but is still widely used), and have not seen anything in the project documentation to constrain me otherwise.  

All that said... what did I use that's post 1.5?  If it's an easy change to make I'll consider it, but most open source projects I work with have already announced 1.7+.

@dblock may want to weigh in on this.

dbwiddis:
I think the 1.7 dependency is the File.readallbytes()... there are ways to do that in older versions if we want to support pre-1.7, I suppose.  Do we?

dblock:
I think we can stop with 1.5 at this point :)

dbwiddis:
How about 1.6, then?   I can easily replace the Files.readAllBytes() call with the exact same source code the Java 7 version calls.  So this is less about this particular change as what policy we want to set moving forward.

alessiofachechi:
Ehm... I use 1.5 for the project I work on ahah! Joking aside it's because I saw in pom.xml:
```
<maven.compiler.source>1.5</maven.compiler.source>
<maven.compiler.target>1.5</maven.compiler.target>
```

dbwiddis:
So I went to try to compile under 1.5 and it's not complaining about readAllBytes(), possibly because the underlying source code doesn't do anything crazy ... the main incompatibility seems to be using @Override on methods implementing an interface.  Which appears to be a 1.6 thing (and Eclipse adds it for me automatically.)

alessiofachechi:
Yes, exactly. Also `FormatUtilTest` needs to be fixed changing
```
@BeforeClass
public static void setUpClass() {
	// use decimal separator according to current locale
	DECIMAL_SEPARATOR = DecimalFormatSymbols.getInstance().getDecimalSeparator();
}
```
with 
```
@BeforeClass
public static void setUpClass() {
	// use decimal separator according to current locale
	DecimalFormatSymbols syms = new DecimalFormatSymbols();
	DecimalFormatSymbols syms = new DecimalFormatSymbols();
}
```

alessiofachechi:
Yes, if you configure the project to be 1.6 compliant Eclipse adds them for you. Probably it's because you have not configured the project as a Maven project. In the latter case Eclipse should read the target from the pom.xml.

dbwiddis:
Aha.... and once Maven's properly configured it's complaining about that readAllBytes().   So, @dblock, do you want 1.6 or 1.7?

alessiofachechi:
If there's no big effort I'd suggest to keep 1.5 compliance (removing bad `@Override`s, etc).

dbwiddis:
`@Override`s aren't bad!  That's why they're included in less ancient versions!  :smirk:

But the point is taken.  I've removed them and replaced the `readAllBytes()` method with the code that method actually executes in Java 7 (and your suggested change to FormatUtilTest.)

alessiofachechi:
`@Override` should be applied when you override a method behaviour, not when you implement an interface method... But hey, that's only phylosophy! :stuck_out_tongue_winking_eye:

dblock:
So 1.5 compat is fixed in https://github.com/dblock/oshi/pull/33 (thanks), what do we do with this PR?

alessiofachechi:
I think it's better if I close this PR, refork the project, rebase my local changes to the upstream and then redo a cleaner PR when I'll complete them. Is it?

dbwiddis:
@alessiofachechi: another philosophy is to use `@Override` whenever possible to avoid mistakes with misspellings or incorrect method signatures (e.g., using a more specific type than the interface).  Personally, I have no philosophy, only an assumption that a newer version of Java/my IDE is probably much wiser than I am.  What we really needs is an `@Implements` annotation!  Maybe that'll come in Java 9.

I think if you merge upstream, rebase, and then reorder the commits to put the merge first, and then squash everything else to a second commit, you get the same effect as your suggestion (which was my default choice on two earlier PRs as I was still learning git.)

dblock:
@alessiofachechi You can always do this:

```
git checkout master
git checkout -b save-my-code # creates a new branch from your current master
git reset --hard HEAD~10 # rewind 10 commits
git pull upstream master # sync with upstream master
git push origin master -f # force push master
git checkout -b my-new-topic-branch
git merge save-my-code --squash
```

alessiofachechi:
Thanks to both! :)

dbwiddis:
Travis CI build is failing on `testCpuVendorFreq()` in `oshi.SystemInfoTest`.  It may be the vendor string doesn't return a parseable Hz value.

dbwiddis:
While you're making minor edits, can you add OS X 10.11 "El Capitan" ? :grin:

alessiofachechi:
Yes @dbwiddis, it was because of a not parseable Hz value (I'd like to know the CPU name string of the Travis machine :smile:).

Anyway I coded the `ParseUtil` class as you suggested to avoid code repetition.

dbwiddis:
Looks great.  Add an entry to the CHANGELOG and we're all set!

alessiofachechi:
Done! :)

alessiofachechi:
Next is

alessiofachechi:
Next feature will be, as you suggested, che current proc freq. :)

dbwiddis:
Take a look at Issue #28 before working on current freq.  Many of the elements currently in the Processor class are for the entire machine, while some (current frequency, and potentially CPU usage) can be measured per-virtual processor.  I'd like to develop a more consistent way of displaying that (similar to how I did the power sources with overall summary (state, time left) and individual (percent capacity).

alessiofachechi:
Perfect! I'd like to contribute when possible also in revising the library architecture if you need! :)

dbwiddis:
Please do. If you have a strong opinion, take charge.  I haven't even started because I'm mulling over a naming convention on just what a Processor or CPU or Core is these days!

PR #32, opened 2015-06-04
Battery Information

dbwiddis:
Submitted for review/discussion of API structure before continuing Linux and Windows ports.

Sample output: 
Power: InternalBattery-0 (Charging) @ 96.01%, 0:55 remaining.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,16 @@
 Next
 ====
 
-* [#13](https://github.com/dblock/oshi/pull/13): Support for Windows 8.1 and Windows Server 2008 R2 - [@NagyGa1](https://github.com/NagyGa1).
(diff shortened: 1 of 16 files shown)


dbwiddis:
I can't get the isCharging boolean to work... convinced it's just not implemented properly anymore.  And the time remaining is inconsistent and useless.  So just the name and % capacity now... 
Sample output: 
Power: InternalBattery-0 @ 96.0%

dblock:
Looks good to me.

dbwiddis:
OK, I've added the windows version and written a Junit test which passes all 3 platforms both plugged-in and on battery.  Ready to merge, with your approval.

dbwiddis:
OK, fixed a few more issues with NPEs (null vs. empty array since Travis CI has no battery) and unsigned 32-bit ints.  Good to merge now, I think!

dblock:
Merged.

PR #33, opened 2015-06-08
Java 1.5 compatibility fixes.

dbwiddis:


dbwiddis:
Ugh.... test did fine but it's failing on linux due to virtual filesystem showing 0 length....

dbwiddis:
there...

src/main/java/oshi/util/FileUtil.java
```diff
@@ -21,10 +22,24 @@
...
+		BufferedReader br = null;
+		try {
+			br = new BufferedReader(new FileReader(filename));
+			for (;;) {
+				String line = br.readLine();
+				if (line == null)
+					break;
+				result.add(line);
+			}
+		} catch (FileNotFoundException e) {
```
dblock: Minor: hiding a `FileNotFoundException` seems unnecessary here.

trejkaz:
New code uses the default platform charset, whereas the old code hard-coded UTF-8. Intentional?

src/main/java/oshi/util/FileUtil.java
dbwiddis: That was residual from implementing one step at a time, and catching that before later deciding to just throw IOExceptions.

dbwiddis:
@trejkaz Good question.  Not intentional, but after researching I can't find where /proc filesystem differs in encoding from the system default. For our purposes (reading /proc and /sys) the information is all ASCII which is common across charsets.

Issue #38, opened 2015-06-15
Add disk usage information


dbwiddis:
Cross platform, Java 1.6:
```java
    File[] roots = File.listRoots();
    for (File root : roots) {
      double usedSpace = (double)(root.getTotalSpace() - root.getUsableSpace())/root.getTotalSpace();
      System.out.format("%s: %.1f used of %s%n", 
          root.getAbsolutePath(), usedSpace, FormatUtil.formatBytes(root.getTotalSpace()));
    }
```

dbwiddis:
See http://stackoverflow.com/questions/462110/acquiring-drive-names-as-opposed-to-drive-letters-in-java for more useful drive names and descriptions (probably use root.toString() as a fallback for an empty name).  This works well for Windows, but only gives "/" in Mac/Linux so will need platform-specific implementations.

dbwiddis:
Mac solution is easy: just list /Volumes.
Unix possibly search for /dev/disk*

dbwiddis:
I've got a working cross-platform (see note) implementation with a FileSystem class that can be extended by the platform-specific versions. Windows will use the default implementation, Mac/Linux will hide the `getFileSystems()` method to parse /Voluumes and /proc/mounts, respectively.

Note: Mac/Linux work now with the "/" directory but should be able to improve on this:
```
File System:
 / (file system) 121.9 GB of 697.5 GB free (17.5%)
```
And I need a bigger hard drive, or to do some cleanup maintenance :grin:

dbwiddis:
Mac is looking good... 
```
File System:
 Data (/Volumes/Data) 19.2 GB of 1.8 TiB free (1.0%)
 MacData (/Volumes/MacData) 3.4 GB of 4.4 GB free (77.4%)
 Macintosh HD (/Volumes/Macintosh HD) 122.1 GB of 697.5 GB free (17.5%)
 MobileBackups (/Volumes/MobileBackups) 0 bytes of 697.5 GB free (0.0%)
 Time Machine Backups (/Volumes/Time Machine Backups) 19.1 GB of 1.8 TiB free (1.0%)
```

dbwiddis:
Windows...
```

File System:
 Floppy Disk Drive (A:) (Floppy Disk Drive) 1.1 MB of 1.4 MB free (82.4%)
 Local Disk (C:) (Local Disk) 27.3 GB of 64.0 GB free (42.7%)
 D:\ (CD Drive) 0 bytes of 0 bytes free (NaN%)
[...]

PR #41, opened 2015-06-17
Bootstrap Site, Maven Magic, License Patch, Java 7

hazendaz:
~~License renamed so that expected license header can be used.~~ Completed on #48 
~~All license headers updated.  Similar to Waffle for the most part.~~ Completed on #49, #50, and Correct Assembly location here
~~Update to JNA 4.1.0 - added field order~~ Completed on #47 
~~Change assembly reference to license~~ Completed on #48 
Add bootstrapped site page
Maven Magic
    Bring up-to-date with Waffle and/or simlar to my base-parent project
~~Upgraded to Java 1.7~~ Completd on #50

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,34 @@ Next
 * [#50](https://github.com/dblock/oshi/pull/50): Added file store information - [@dbwiddis](https://github.com/dbwiddis).
 * Your contribution here.
 
+1.3 (in-progress)
+-----------------
+* See site page for individual commits going forwards as only major changes will be mentioned manually in change log.
+* Upgraded to java 7 base support
+* Upgraded JNA to 4.1.0
+* Bringing all lessons learned from waffle over to oshi in one go.
+  * Additionally raised to java 1.6.
+  * Assembly is now performed on all builds.
+  * Use sonatype parent
+  * Added inception year
+  * Added organization
+  * Require Maven 3.2.3 or better
+  * Added issues management
+  * Added ci management
+  * Added distribution for site pages to github gh pages
+  * Added copyright
+  * Added maven properties for java version (source and test side)
+  * Added maven build timestamp
+  * Set project and reporting encoding to UTF-8
+  * Added entire default plugin set to control the build level of each even if not otherwise used.  This is intended to override mavens hidden base parent
+  * Set incremental compilation at false due to maven bug and note the jira ticket
+  * Set site page to allow markdown usage in order to show readme
+  * Switch from legacy emma code coverage to jacoco code coverage
+  * Add all the various reports such as javadocs, checkstyles, findbugs, pmd, nvd scan for security, etc
+  * Add patch for git commit id for manifest along with jira to same issue
+  * Add build time, copyright, os informaton, compiler information to manifest for all resulting builds
+  * Remove rather manual pieces of release to nexus
+  * Keep gpg signing in case we need to use deploy when maven release plugin fails so we can get to sonatype regardless
 
 1.2 (6/13/2014)
 ================
--- a/pom.xml
+++ b/pom.xml
@@ -1,8 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" 
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.sonatype.oss</groupId>
+        <artifactId>oss-parent</artifactId>
+        <version>9</version>
+        <relativePath />
+    </parent>
     <groupId>com.github.dblock</groupId>
     <artifactId>oshi-core</artifactId>
     <version>1.3-SNAPSHOT</version>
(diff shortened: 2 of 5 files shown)


hazendaz:
I did not update changelog.  I really don't think that is necessary as it is manual.  I've added maven-changelog-plugin as part of this.  On the site page, it adds the changelog for last month at default level.  That can be adjusted but for the most part we can control the length of that between releases if needed thus automating the change log

hazendaz:
Building site

   mvn site
   mvn site:deploy --> Will push to gh pages (https://dblock.github.io/oshi)

Building code

   mvn clean install

Releasing

   mvn release:clean
   mvn release:prepare
   mvn release:perform

hazendaz:
Future releases will pick up the following for distribution to maven central.

oshi-core-*.jar
oshi-core-*-javadoc.jar
oshi-core*-oshi-assembly.tar.bz2
oshi-core*-oshi-assembly.tar.gz
oshi-core*-sources.jar
oshi-core*-test-javadoc.jar
oshi-core*-test-sources.jar

hazendaz:
For a general idea on the POM.  Please see this project https://github.com/hazendaz/base-parent

dblock:
Someone in this thread should write a hands on guide on implementing Maven for a project and sell it for a lot of $.

hazendaz:
I created 2 separate safe PRs for tonight that can be pulled.  One is the bump to jna 4.1.0 along with junit to 4.12 and the other is to address setting the license files somewhat different than was done here.  Next up, I'll perform the license piece only since it touches so much.  After than I will bump to java 1.6, then finally, address the overall pom.  So basically lot of easy smaller PRs in order to make it easier to review and pull things in.  As soon as things merge in, I will start addressing the rest.  Thanks.

As to implementing maven book, sounds great but my current employer frowns upon making money outside the firm...although I think I'd love to do that one day...

dbwiddis:
I've merged the two PRs (JNA 4.1/Junit 4.12 and licenses).
When you bump Java, bump to 1.7.  I've got #38 all coded up waiting on your PRs, and it's got a 1.7 dependency (`java.nio.file.FileSystems`). So save me the trouble of bumping the version twice. :grin:

dbwiddis:
@hazendaz Just noticed the licence link at the bottom of the README was broken with the PR I merged.  Should be updated to wherever the new licence is going to live or to https://www.eclipse.org/legal/epl-v10.html ... not sure what magic you're doing with licenses so I'll let you make that call.

hazendaz:
Sure thing.  I'll get to this tonight.

--- Original Message ---

hazendaz:
We probably can point to the local copy of that licence in the Readme and the pom so if it changed ever to say version 2.0 we would only need to replace the copy we store leaving pom and Readme out of it.

--- Original Message ---

dbwiddis:
OK, this is merged but I have a few errors in my pom.xml now and need some help resolving them.
* `git-commit-id-plugin` gives me an error on line 418: ".git directory could not be found! Please specify a valid [dotGitDirectory] in your pom.xml"
* `license-maven` plugin gives me an error on line 600: "Plugin execution not covered by lifecycle configuration"
And, finally, the `animal-sniffer-maven-plugin` at line 370 is causing the Travis CI build for my latest PR #51 to fail because it can't validate the signature for the `com.sun.management.OperatingSystemMXBean`.  The build tells me "Signature errors found. Verify them and ignore them with the proper annotation if needed." but I cannot find any reference anywhere for the "proper annotation" to ignore these errors.

hazendaz:
I’ve got the license issue resolved.  Now that I have commit access, I’ll create a PR and merge that shortly so you can confirm that it has been resolved from being an error in your pom.  What I did on my end was to uninstall the m2e connector for the plugin which obviously isn’t portable for all users, then restarted.  Upon restart, I had the same error.  I thing applied the standard fix which is portable and the error went away.  I then ran a build without issue.

 

I’ll be looking at the other two after I get this one applied.

 

Thanks,

 

Jeremy

hazendaz:
Looks like you got com.sun issue resolved.  Problem there is that it is highly advisable to not use com.sun packaging from java because that is only in the oracle jvm.  Others won’t have it and it is subject to change.  What probably should be done is to find a dependency that also implements that but not using com.sun.*.  I’ll look around but otherwise you have it ignored so things are ok.  That is why it has issues though :)

 

The .git file should be in the same directory as the pom.xml.  I’m unsure why it would be saying you are missing that.  When you cloned the repo you should have gotten that.  If you don’t see it there, try changing your folder options so you can see hidden folders and see if that exists.  Maybe if it is hidden that is causing an issue.  I’m not 100% sure there.  I do know that plugin is more of a problem than a solution but the others out there last I looked are not really any better.

 

Thanks,

 

Jeremy

dbwiddis:
I've resolved the other two already!

hazendaz:
Awesome!  I did put a note out on another project I work on ‘psi-probe’ which is a drop in place replacement for tomcat management.  I suggested other people on the project check this out as it could be included there :)  So far feedback is that this library shows far more information than that project was gathering.  Only holdup is that project getting to java 7.  They will eventually but are going slow at the moment.  So I think in the next few months this might start seeing a larger use case :)

 

Thanks

 

Jeremy

dbwiddis:
Is there a way to programmatically test whether that method is available and fall back to another method instead?

hazendaz:
Yes…something like this…

 

try {

Class.forName( "your.fqdn.class.name" );

} catch( ClassNotFoundException e ) {

//my class isn't there!

}

 

So I think you could move out the questionable class into another class then have the first one proxy this way, if there it would then perform whatever it does using that class, otherwise it would fall back to something else.  Definitely a good way to make it cross platform for those still running other jdks :)  Of course we pre-compile it in our builds and it will then work on all jvms and will run the better logic you have that needs that.  That is to say if there was another way to do something similar.  At least that way we don’t have to add any more dependencies.

dbwiddis:
The com.sun version extends/ is instantiated as a typecast of the java.lang version, so I can do this test on instantiation and cast appropriately, and set a boolean to switch in the method.

PR #51, opened 2015-06-20
Add CPU ticks and use OperatingSystemMXBean for CPU load

dbwiddis:


dbwiddis:
Travis CI doesn't seem to like the com.sun.management version of the OperatingSystemMXBean...

PR #57, opened 2015-06-21
jautodoc remainder of non javadoc'd items.

hazendaz:
Formatting is off a bit in a couple of places.  I missed setting spaces back to tabs.  However, I plan to do another pull sometime soon enforcing formatting which is space based and generally the accepted cross platform usage.  For now this is ok as these are at least aligned together in the right spaces.  Now everything have javadocs.

### DIFF
--- a/src/main/java/oshi/hardware/package-info.java
+++ b/src/main/java/oshi/hardware/package-info.java
@@ -1,0 +1,4 @@
+/**
+ * Provides...
+ */
+package oshi.hardware;
--- a/src/main/java/oshi/package-info.java
+++ b/src/main/java/oshi/package-info.java
@@ -1,0 +1,4 @@
+/**
+ * Provides...
+ */
+package oshi;
--- a/src/main/java/oshi/software/os/linux/package-info.java
+++ b/src/main/java/oshi/software/os/linux/package-info.java
@@ -1,0 +1,4 @@
+/**
+ * Provides...
+ */
+package oshi.software.os.linux;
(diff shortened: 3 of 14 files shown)


dbwiddis:
We've been using the eclipse built in conventions for formatting. If that is where the tabs come from we should commit a coding conventions xml to the repository so everyone is consistent.

hazendaz:
There is a plugin already in the POM for formatting needs.  Another project I’m on works the formatting angle.  It’s based on Eclipse Luna but does not directly require eclipse.  All I need to do is copy over the formatting xml file from waffle here and it will perform that in one go.  I’ve got some other things coming over shortly though.  I have eclipse checks 90%+ turned on so a lot of warnings were in my workspace.  I worked all those which was quick but now trying to break up the commits so it’s easy to see.  The format file that will be used can also be imported into eclipse easily enough to ensure it formats correctly.  However, I do not generally suggest that since that affects all formatting and could be a problem for folks working on multiple projects as I do.  I do have mine setup that way and tabs/spaces always ends up being a sore point.  That is until the formatting can be mentioned and ultimately accepted.  After I get a chance to get this next set up and get some feedback from you or Db regarding the changes, I can quickly dump that in place.  It’s pretty intrusive though so when we do it, we would want to make sure all outstanding changes are up and merged into the code base as to avoid any merge conflicts as it certainly will cause them that first go.  After that, every build will ensure the formatting so we can forget about it going forwards.

 

Jeremy

PR #58, opened 2015-06-21
Language Cleanups

hazendaz:
Lots of little language cleanups.  Please review and comment.  No rush to pull this in unless it all is acceptable.  After this PR, I'll put formatting onto oshi pulled from waffle.  There then will be a few other minor language issues but those get a little more questionable and there is enough here I wanted to hold back some.  I'll be out a few hours but will get back in here later this evening.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -593,9 +593,9 @@
                     </execution>
                 </executions>
                 <configuration>
-                  <ignores>
-                    <ignore>com.sun.management.OperatingSystemMXBean</ignore>
-                  </ignores>
+                    <ignores>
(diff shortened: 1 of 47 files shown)


dbwiddis:
Merged this, because I only have a few minor nits that I can address later.   There are a few emtpy methods by intention that don't need the //TODO, but perhaps I can replace the todo with a comment that it's intentionally empty.

I like the addition of this.* for class variables, and it will let me get rid of those pesky underscores that aren't Java convention. :)

hazendaz:
Comment would be good on ones that need to be empty.  I didn’t really look at context there so put the todo’s :)

PR #59, opened 2015-06-21
Test for Oracle JVM on getSystemCpuLoad()

dbwiddis:


dbwiddis:
Please review and merge if this addresses the Oracle JVM issue and/or suggest a better way to do it!

dbwiddis:
In addition to the new Oracle JVM test (and backup calculation) I've updated all the systemwide class variables to be static so they aren't recalculated with each Processor instantiation. (Future) variables associated with the logical CPUs will not be static.  I've kept the two commits separate for ease of review.

dbwiddis:
After further consideration I've restored instance variable status.  Windows and Linux are actually set separately from registry/cpuinfo and even though Mac OS X uses the same values the fact that we have setters means the user could change them and should expect the change to not apply globally.  So this should be good to go for the new system cpu load code.

PR #66, opened 2015-06-29
Update RELEASING.md doc to reflect new release process

dbwiddis:
And other minor fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 
-2.0 (in-progress)
------------------
+1.4 (in-progress)
+================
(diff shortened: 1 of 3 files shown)


hazendaz:
Looks great!  Only possible change, if any, would be that reviewing version number not really needed.  mvn release will let us change to whatever.  So say it's 1.2.0-SNAPSHOT, when running that, we get the option to change to whatever, so rather than 1.2.0 as the release, we could say 2.0.0.  But at least noting that makes us think about it as that is really important.  Anyway, looks great and merging now...

Issue #73, opened 2015-09-02
Get uptime


Miloune:
Hello,

Is it possible to get system uptime ? I found this:
https://github.com/dblock/oshi/blob/master/src/main/java/oshi/software/os/linux/Libc.java

Someone can help me on this please ?

Regards.

dbwiddis:
Take a look at the code in the linux GlobalMemory class for total memory, which pulls totalram from that same structure. Just grab 'uptime' instead of 'totalram'.

Without checking/debugging, this should do the trick:
```
    import oshi.software.os.linux.Libc;
    import oshi.software.os.linux.Libc.Sysinfo;
    
    Sysinfo info = new Sysinfo();
    if (0 != Libc.INSTANCE.sysinfo(info))
        throw new LastErrorException("Error code: " + Native.getLastError());
    System.out.println("Uptime in milliseconds: "+info.uptime.longValue());
```

I think we can add uptime to Oshi, using the above code for Linux (or by parsing /proc/uptime if the above proves problematic.)

Mac OS X has a systctl call using "kern.boottime" that should work..

Windows can get uptime using the Performance Data Helper (PDH) code similar to the tick-counting in CentralProcessor, using the path "\\\\.\\System\\System Up Time"

dbwiddis:
Working on implementing this now.  The above code works, except that the value (as documented) is in seconds, not milliseconds as I wrote.  This matches the first value in /proc/uptime as well, with the advantage in the latter that it includes 2 decimal places.

Given that we probably don't need less-than-a-second resolution, I'll just use the above result.

Miloune:
Thanks a lot ! :)

dbwiddis:
Code is in #75.  It'll be part of 1.4 which I'll try to get to this month, along with some other issues.

PR #75, opened 2015-09-03
Add uptime capability. #73

dbwiddis:


dbwiddis:
Travis is failing because JAVA_HOME is not defined correctly... no idea how to fix that!

dbwiddis:
OK, seems a relatively new thing that Travis CI has implemented OS X support.  Previously our builds were only running against Linux.

But OS X needs JAVA_HOME set and I don't know what to set it to for the travis VM.

We can allow errors in travis.yml on OS X for the time being, while this gets resolved...

dblock:
You can set the env variables in Travis via `env:`, http://docs.travis-ci.com/user/environment-variables. So figure out what JAVA_HOME should be and just hard-code it in there?

dblock:
Note that Travis claims that JAVA_HOME is correctly set in http://docs.travis-ci.com/user/environment-variables/. Maybe it's a bug in their infrastructure?

Issue #89, opened 2015-09-24
Small language cleanups


hazendaz:
In spirit of complying with sonar and other similar tools, enhance code to follow suggested styling.

hazendaz:
Merged code.

PR #92, opened 2015-09-24
Clearing warnings in eclipse and small language cleanups

hazendaz:
Addresses #89


hazendaz:
I took out the formatter piece setup on my PR.  As soon as this passes, I'll merge.  I reviewed again in GIT and I'm pretty good on this.  If you have any questions as to why some things where changed let me know.  If for some reason after some of this cleanup is done you want to release again, since there is no real gain with this, it probably should be something like 1.5.1.

PR #117, opened 2016-02-22
Properly escape JSON strings

dbwiddis:


src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
```diff
@@ -587,16 +587,16 @@ public int getPhysicalProcessorCount() {
     @Override
     public String toJSON() {
         StringBuilder sb = new StringBuilder("{");
```
dblock: There's got to be a JSON parser where you can turn a Dictionary into JSON which takes care of all of this ...

dbwiddis: There are dozens. Jackson is the most common. 

The overhead to create the dictionaries is more lines of code than building the strings. And this way avoids another dependency.

dblock: Just think about it :)

dbwiddis: Already working on porting existing code over to `javax.json`. :)

Issue #119, opened 2016-02-22
OS temperature


Amjad121092:
Hi,

Would it be possible to add a method to get the PC temperature? I think this is crucial and it's missing. Other than that the project is amazing thank you very much!

Regards

dblock:
I wanted to label this 🔥, but github doesn't support Emoji in issue labels!!!
https://twitter.com/dblockdotorg/status/701913056931135488

dbwiddis:
This sounds like a great suggestion... question is how hard it is to get the info.

OS X looks the easiest.  [This project](https://github.com/lavoiesl/osx-cpu-temp) has C code that I can port to JNA.

Linux appears to require installation of a package lm-sensors, plus root permissions. 

Windows appears to have it (possibly) in WMI, info [here](http://stackoverflow.com/questions/23314886/get-cpu-temperature).

dbwiddis:
Update... once lm-sensors is installed and sensors-detect is run as root, it does provide some basic info without root permissions.  So, it looks like this is a possibility!

Amjad121092:
This is great news i hope you can add it in as it would be amazing to have!

Amjad121092:
Hi,

Just wondering if there's any updates on this?

Regards

dblock:
@Amjad121092 There's no update or ETA or anyone doing anything other than best effort to implement features. Please feel free to contribute, see https://github.com/dblock/oshi/blob/master/CONTRIBUTING.md.

dbwiddis:
@Amjad121092 as @dblock said, this is a best effort approach.  I started on the OS X port and hit a snag (C pointers drive me nuts).  I haven't even looked at Linux and Windows yet.  The information is posted in this issue if you, or anyone you know, wants to take a crack at contributing a solution.

dbwiddis:
Here's a rundown of current research/challenges.

Linux - This is looking the best so far.  If lm-sensors is installed by default (it is on Ubuntu) then the `sensors` command can be easily parsed.  By reverse engineering the lm-sensors code it appears that it gets the value from the path `/sys/class/hwmon/hwmon0/` and reading temp1_input.  There are other files in this directory structure that give information on fan speeds as well.  I haven't written any code here yet but it looks easy enough once the other OS's are taken care of.

[...]

PR #121, opened 2016-02-29
CPU temperature, fan speeds, and voltage

dbwiddis:
OS X and Linux CPU temperature and fan speeds for #119

### DIFF
--- a/src/main/java/oshi/hardware/platform/linux/LinuxSensors.java
+++ b/src/main/java/oshi/hardware/platform/linux/LinuxSensors.java
@@ -1,0 +1,162 @@
+/**
+ * Oshi (https://github.com/dblock/oshi)
+ * 
+ * Copyright (c) 2010 - 2016 The Oshi Project Team
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ * 
+ * Contributors:
+ * dblock[at]dblock[dot]org
+ * alessandro[at]perucchi[dot]org
+ * widdis[at]gmail[dot]com
+ * https://github.com/dblock/oshi/graphs/contributors
+ */
+package oshi.hardware.platform.linux;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.json.Json;
+import javax.json.JsonArrayBuilder;
+import javax.json.JsonBuilderFactory;
+import javax.json.JsonObject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import oshi.hardware.Sensors;
+import oshi.util.FileUtil;
+
+public class LinuxSensors implements Sensors {
+    private static final Logger LOG = LoggerFactory.getLogger(LinuxSensors.class);
+
+    private JsonBuilderFactory jsonFactory = Json.createBuilderFactory(null);
+
+    // Possible sensor types. See sysfs documentation for others, e.g. current
+    private static final String[] SENSORS = { "temp", "fan", "in" };
+
+    // Base HWMON path, adds 0, 1, etc. to end for various sensors
+    private static final String HWMON = "/sys/class/hwmon/hwmon";
+
+    // Map from sensor to path
+    private Map<String, String> hwmonMap = new HashMap<String, String>();
+
+    public LinuxSensors() {
+        // Iterate over all hwmon* directories and look for sensor files
+        // e.g. /sys/class/hwmon/hwmon0/temp1_input
+        int i = 0;
+        while (Files.isDirectory(Paths.get(HWMON + i))) {
+            for (String sensor : SENSORS) {
+                String path = HWMON + i + "/" + sensor;
+                File sensorFile = new File(path + "1_input");
+                if (sensorFile.exists()) {
+                    hwmonMap.put(sensor, path);
+                }
+            }
+            i++;
+        }
+    }
+
+    @Override
+    public double getCpuTemperature() {
+        if (hwmonMap.containsKey("temp")) {
+            String hwmon = hwmonMap.get("temp");
+            List<String> tempInfo = null;
+            try {
+                tempInfo = FileUtil.readFile(hwmon + "1_input");
+            } catch (IOException e) {
+                LOG.error("Problem with {}{}: {}", hwmon, "1_input", e.getMessage());
+                return 0d;
+            }
+            // Should return a single line of millidegrees Celsius
+            if (tempInfo.size() > 0) {
+                int millidegrees = 0;
+                try {
+                    millidegrees = Integer.parseInt(tempInfo.get(0));
+                } catch (NumberFormatException e) {
+                    LOG.error("Invalid format for temperature: {}", tempInfo.get(0));
+                }
+                return millidegrees / 1000d;
+            }
+        }
+        return 0d;
+    }
+
+    @Override
+    public int[] getFanSpeeds() {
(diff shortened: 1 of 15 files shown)


src/main/java/oshi/hardware/platform/linux/LinuxSensors.java
```diff
@@ -0,0 +1,103 @@
...
+
+import oshi.hardware.Sensors;
+import oshi.util.FileUtil;
+
+public class LinuxSensors implements Sensors {
+    private static final Logger LOG = LoggerFactory.getLogger(LinuxSensors.class);
+
+    private JsonBuilderFactory jsonFactory = Json.createBuilderFactory(null);
+
+    private static final String HWMON = "/sys/class/hwmon/hwmon0/";
```
dbwiddis: The final number after hwmon indexes different sensors.  hwmon0 may be temp and hwmon1 may be fans, or they may be in a different order.  This needs rewriting to account for temp*_input and fan*_input being in any potential hwmon* path.

src/main/java/oshi/hardware/platform/mac/MacSensors.java
```diff
@@ -0,0 +1,305 @@
...
+     * Read a key from SMC
+     */
+    private int smcReadKey(String key, SMCVal val, int retries) {
+        SMCKeyData inputStructure = new SMCKeyData();
+        SMCKeyData outputStructure = new SMCKeyData();
+
+        inputStructure.key = strtoul16(key, 4);
+        int result;
+        do {
+            Util.sleep(8);
```
dblock: This can't be right :) I would do it in a loop until it succeeds if that is more reliable.

dblock:
Hope to see some of this CoreFoundation stuff in JNA proper sometime ;)

src/main/java/oshi/hardware/platform/mac/MacSensors.java
dbwiddis: Unfortunately, the errors I'm getting in that method appear to be associated with overrunning buffers.  The more rapidly I query, the worse the performance gets.  With no delay it takes an average of 300+ tries.  And that's the average... individual queries are longer.  Adding a few millisecond delay in this loop helps tremendously. Once we get to 6-10 ms the average drops to about 8 to 10 tries (average... individual attempts still take more tries).

"Try until it fails" is therefore a problem with the delay.  100 tries takes 800ms, nearly a second.  And I didn't see much difference having a longer number of retries on initialization vs. during runs, basically if it took more than 50 tries it was going to take hundreds.  Very little middle ground without a longer delay.

Once it succeeds the first time and caches the IOData structure, the repeated queries are much more reliable.

The only way I can think "try until it succeeds" would be successful is if we started it in a separate thread, possibly keeping the value updated until the user queries it at which time immediate response would be available.

dbwiddis:
Maybe sometime. There's a difference between "write what you need" and "write the entire library and develop test cases for it" :)

src/main/java/oshi/hardware/platform/mac/MacSensors.java
dbwiddis: Here's an example of how the current code runs.  It takes 4 polls at a 1-second interval to populate all the fields once:
```
{"cpuTemperature":63.0,"fanSpeeds":[]}
{"cpuTemperature":61.875,"fanSpeeds":[0,0]}
{"cpuTemperature":63.125,"fanSpeeds":[0,2347]}
{"cpuTemperature":63.125,"fanSpeeds":[2283,2260]}
```
It gets temperature successfully the first time, number of fans the second, and each of the 2 fan speeds the third and fourth times.  So you'd think extending the retries from 50 to 200 would help, right?  Nope, it doesn't... looks exactly the same, just takes 4 times as long.

I still don't understand why the delay seems to be required in JNA, when C code making the same calls doesn't appear to have any problems.

dblock: It looks to me like you have a typical buffer overflow bug. I would look for a bug in your code.

dbwiddis: Believe me, I've looked, and looked again. And slept. And drank. And looked.

You are welcome to compare... many examples out there.  I have gone line by line and tried to match function/method names.
In C:
  https://github.com/Chris911/iStats/blob/master/ext/osx_stats/smc.c (I have actually installed this ruby gem on my machine and it works flawlessly)
  https://github.com/lavoiesl/osx-cpu-temp/blob/master/smc.c
  https://github.com/beltex/libsmc/blob/master/src/smc.c
In Swift:
  https://github.com/beltex/SMCKit/blob/master/SMCKit/SMC.swift
In Objective C:
  https://github.com/FergusInLondon/SMCWrapper/blob/master/SMCWrapper/SMCWrapper.m

The only code I "skipped" is the memset() commands to zero the structure memory.  It is my understanding that JNA auto-clears structures when it creates them.  I tried manually using `.clear()` on my structures with no appreciable change.

If the buffer is overflowing it'd possibly be because the native structures and my JNA structures don't match.  I've compared them numerous times.  I'd appreciate a second set of eyes on my IOKit.java headers vs. the various smc.h structure declarations.

dbwiddis: Ummm.... so I see what you meant. I forgot the `break;` at the bottom of that while loop that I had intended!  *facepalm*

Also my caching wasn't working as I forgot my `.put()` ... 

Here's another look at what's happening with the loop break and caching working:
```
100|fail
 98|  1| 14| 16|{"cpuTemperature":62.625,"fanSpeeds":[2454,2458]}
 18|  2|  2|{"cpuTemperature":62.625,"fanSpeeds":[2468,2466]}
  1| 14|  0|{"cpuTemperature":62.75,"fanSpeeds":[2469,2476]}
  0|  2|  0|{"cpuTemperature":62.75,"fanSpeeds":[2488,2479]}
  8|  0| 10|{"cpuTemperature":62.75,"fanSpeeds":[2492,2493]}
  9|  8|  6|{"cpuTemperature":62.625,"fanSpeeds":[2495,2486]}
 19|  1|  0|{"cpuTemperature":62.75,"fanSpeeds":[2482,2492]}
  7|  9|  1|{"cpuTemperature":62.625,"fanSpeeds":[2495,2496]}
  4|  0|  8|{"cpuTemperature":62.5,"fanSpeeds":[2498,2486]}
  0|  0|  0|{"cpuTemperature":62.5,"fanSpeeds":[2502,2495]}
  0|  1|  7|{"cpuTemperature":62.75,"fanSpeeds":[2500,2498]}
```
First call (during initialization) exceeded 100 tries on temp and gave up.
Second call succeeded after 98 tries on temp, 1 on #fans, and 14, 16 on each fan (successfully caching all of the initial dataInfo variables)
Once cached, the subsequent calls were a lot more reliable.

dbwiddis: And compare the above with a delay of 1ms to this, with a delay of 4ms:
```
100|fail
 94|  1|  4| 47|{"cpuTemperature":63.375,"fanSpeeds":[2524,2527]}
  1|  9|  6|{"cpuTemperature":62.5,"fanSpeeds":[2531,2527]}
  0|  2|  0|{"cpuTemperature":63.375,"fanSpeeds":[2542,2539]}
  0|  4|  2|{"cpuTemperature":63.375,"fanSpeeds":[2547,2541]}
  0|  1|  0|{"cpuTemperature":62.5,"fanSpeeds":[2544,2542]}
  0|  0|  0|{"cpuTemperature":63.375,"fanSpeeds":[2548,2550]}
  0|  0|  0|{"cpuTemperature":63.375,"fanSpeeds":[2542,2554]}
  0|  1|  1|{"cpuTemperature":63.375,"fanSpeeds":[2555,2544]}
  1|  0|  0|{"cpuTemperature":63.375,"fanSpeeds":[2552,2548]}
  0|  0|  0|{"cpuTemperature":63.25,"fanSpeeds":[2548,2550]}
  2|  0|  2|{"cpuTemperature":63.25,"fanSpeeds":[2551,2548]}
  ```

PR #125, opened 2016-03-10
Rework SystemInfoTest removing system outs

hazendaz:
I reworked this class a bit to leverage the loggers.  I've been otherwise copying for the most part this specific main contents to use psi probe.  I assume this is the most complete output portion and as such probably should eventually be retrievable via a single call if user so desires.  For now, this helps me keep my duplication changes low between this and what I copied over :)  I don't think this hurts but it might be on the road to what I'm looking for.  Let me know...

### DIFF
--- a/src/test/java/oshi/SystemInfoTest.java
+++ b/src/test/java/oshi/SystemInfoTest.java
@@ -24,7 +24,9 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.FileStore;
 import java.nio.file.Files;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 
 import org.junit.Test;
 import org.slf4j.Logger;
@@ -50,6 +52,9 @@ import oshi.util.Util;
  * @author dblock[at]dblock[dot]org
  */
 public class SystemInfoTest {
+
+    /** The Constant logger. */
+    private static final Logger logger = LoggerFactory.getLogger(SystemInfoTest.class);
 
     /**
      * Test get version.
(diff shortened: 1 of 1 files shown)


dbwiddis:
No great objections to reformatting the test output, but...
- Given this is in the test package, it's not intended to be user-retrievable.  The main purpose of this package is to give users examples for code they should write in their own code which calls oshi. 
- Updating `toString()` methods in the main project would be preferable if you want a consistent format for output consumption (that isn't JSON), and then the `SystemInfoTest` can just print out the various `toString()` outputs.
- The "single call" expected to be used by other code is presently the `toJSON()` outputs.

hazendaz:
Idea here is moving that direction.  When I first tried this code json failed with null pointer.  So used the other output which works fine when monitor connected.  Tried on a Linux instance and had similar null pointer in monitor section.  I'm just in demo mode currently...I ran initial tests with earier version so I need to repeat all this.


Sent from Outlook Mobile






On Wed, Mar 9, 2016 at 8:26 PM -0800, "dbwiddis" <notifications@github.com> wrote:





No great objections to reformatting the test output, but...
- Given this is in the test package, it's not intended to be user-retrievable.  The main purpose of this package is to give users examples for code they should write in their own code which calls oshi.
- Updating `toString()` methods in the main project would be preferable if you want a consistent format for output consumption (that isn't JSON), and then the `SystemInfoTest` can just print out the various `toString()` outputs.
- The "single call" expected to be used by other code is presently the `toJSON()` outputs.

---

dbwiddis:
RE: Logger vs. LOG:  Wasn't aware of Google exception. Just going on personal habit and Sun conventions.  I used LOG everywhere else in oshi.

RE: Extra newlines:  I think all the extra ones you see are when we use String.format with the %n character in the output itself.  Suggesting that removing those %n's from the format will prevent you from having to treat them specially in the list / log later.  Although as I mentioned, this may be too much in the weeds if you care about the format.

RE: Null pointer.  I found/fixed that error (and others) and have completely reworked the JSON from when you last tried it, so it ought to work now.  If not, let me know where you're still getting a null error.

hazendaz:
Thanks.  I'll give it a try this weekend.


Sent from Outlook Mobile






On Fri, Mar 11, 2016 at 10:10 AM -0800, "dbwiddis" <notifications@github.com> wrote:





RE: Logger vs. LOG:  Wasn't aware of Google exception. Just going on personal habit and Sun conventions.  I used LOG everywhere else in oshi.

RE: Extra newlines:  I think all the extra ones you see are when we use String.format with the %n character in the output itself.  Suggesting that removing those %n's from the format will prevent you from having to treat them specially in the list / log later.  Although as I mentioned, this may be too much in the weeds if you care about the format.

RE: Null pointer.  I found/fixed that error (and others) and have completely reworked the JSON from when you last tried it, so it ought to work now.  If not, let me know where you're still getting a null error.

---

dbwiddis:
I'm going to target a release for this coming weekend, so if you wanted to do more on this to get it in the new release, great.  Or if there are a few specific changes you need me to make, I'm happy to do those as well.

hazendaz:
I think I am good from my perspective at this time for how and why I needed this.  I believe there were some changes you wanted here so if you have the time great.  I won't have time myself to look again this week.


Thanks,


Jeremy


Sent from Outlook Mobile






On Tue, Apr 5, 2016 at 9:13 PM -0700, "dbwiddis" <notifications@github.com> wrote:





I'm going to target a release for this coming weekend, so if you wanted to do more on this to get it in the new release, great.  Or if there are a few specific changes you need me to make, I'm happy to do those as well.

---

PR #132, opened 2016-04-07
FIX: NullPointer when try to extract display

pcollaog:
+ NullPointer control when OS doesn't have command xrandr when try to iterate over null value.
+ Controll over null value in foreach.


dbwiddis:
Looks good.  I'll merge both of these PRs and do a combined CHANGELOG entry for them....

Issue #131, opened 2016-04-07
NullPointer when try to extract info about the cpu's architecture on raspberry-pi


pcollaog:
NullPointer when try to extract info about the cpu's architecture on raspberry-pi
<pre>
java.lang.NullPointerException: null
        at oshi.hardware.platform.linux.LinuxCentralProcessor.isCpu64bit(LinuxCentralProcessor.java:302)
        at oshi.hardware.platform.linux.LinuxCentralProcessor.toJSON(LinuxCentralProcessor.java:618)
        at oshi.hardware.platform.linux.LinuxHardwareAbstractionLayer.toJSON(LinuxHardwareAbstractionLayer.java:158)
        at oshi.SystemInfo.toJSON(SystemInfo.java:119)
</pre>

pcollaog:
Raspberry pi doesn't have a *flags* attribute, so... the method *processor.setCpu64(found)* never was invoke. This cause the null pointer when you execute the getter.

```
processor       : 0
model name      : ARMv7 Processor rev 4 (v7l)
BogoMIPS        : 38.40
Features        : half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm crc32
CPU implementer : 0x41
CPU architecture: 7
CPU variant     : 0x0
CPU part        : 0xd03
CPU revision    : 4
```

PR #133, opened 2016-04-07
FIX: initialize cpu64 attribute in false by default

pcollaog:
+ avoid NullPointer exception with default value in false in cpu64 attribute in LinuxCentralProcessor

issue #131


dbwiddis:
Looks good. Can you add a CHANGELOG entry referencing this PR and squash the commit?

dbwiddis:
I'll just do the CHANGELOG and reference both these NPE fixes.

pcollaog:
I have some weird errors with raspberry-pi, so... I'll try to fix it ;)

PR #136, opened 2016-04-12
Modified method for getting CPU load average

henryx:
With this pull request, for getting the CPU Load average is used the Linux proc file as fallback of the native Java method

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * [#124](https://github.com/dblock/oshi/pull/124): Read Windows Sensor info from Open Hardware Monitor as fallback - [@dbwiddis](https://github.com/dbwiddis).
 * [#129](https://github.com/dblock/oshi/pull/129): Improve Linux version and code name parsing - [@dbwiddis](https://github.com/dbwiddis).
 * [#132](https://github.com/dblock/oshi/pull/132), [#133](https://github.com/dblock/oshi/pull/133): Fix NPEs on Raspberry Pi - [@pcollaog](https://github.com/pcollaog).
+* [#136](https://github.com/dblock/oshi/pull/136): Modified method for getting CPU load average- [@henryx](https://github.com/henryx).
 * Your contribution here.
 
 2.2 (3/1/2016)
--- a/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/src/main/java/oshi/hardware/CentralProcessor.java
@@ -1,17 +1,18 @@
 /**
  * Oshi (https://github.com/dblock/oshi)
- * 
+ *
  * Copyright (c) 2010 - 2016 The Oshi Project Team
- * 
+ *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
- * 
+ *
  * Contributors:
  * dblock[at]dblock[dot]org
  * alessandro[at]perucchi[dot]org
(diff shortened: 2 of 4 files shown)


dbwiddis:
Great contribution!   Can you please:

1. Modify the javadoc for the CentralProcessor interface to note this fallback, at least in general terms, and
2. Make an entry in the CHANGELOG summarizing this PR?

And of course, squash the commits.

hazendaz:
Github allows squashing commits now on merges of PR.  Maybe something dB can turn on.


Sent from Outlook Mobile






On Tue, Apr 12, 2016 at 3:32 PM -0700, "dbwiddis" <notifications@github.com> wrote:





Great contribution!   Can you please:

1. Modify the javadoc for the CentralProcessor interface to note this fallback, at least in general terms, and
2. Make an entry in the CHANGELOG summarizing this PR?

And of course, squash the commits.

---

dbwiddis:
Cool feature!  I hope dB turns it on.  (I can also merge and then rebase 
and squash manually later.)

hazendaz:
It might already be on by default.  When you go to merge.  There is a new drop down box.  Click on that and it should show you ability to squash commits and you get to brand it.  The default looks like it takes all commit messages and adds those to the description.  I just tried it on another repo and it worked great!

dbwiddis:
Actually, on further review, I'm not sure this is right.  The `/proc/loadavg` file is more appropriate to the `getSystemLoadAverage()` method.  And this one is part of the official Java MXBean so we don't need a fallback for it.

henryx:
@dbwiddis I'm sorry, I've realized now that I have modified the wrong method. Please do not consider this pull request for these reasons (for the future, I will do more attention when I will write patches). Before reject it, please, can you link me how is explained how works the Java MXBean `getSystemLoadAverage()` on Linux systems? I've searched on Internet, but I doesn't have found if it uses `/proc/loadavg` or other methods for calculating the cpu load average

dbwiddis:
The MXBean method is documented [here](https://docs.oracle.com/javase/7/docs/api/java/lang/management/OperatingSystemMXBean.html#getSystemLoadAverage()) but only returns a singular number. The Java implementation uses the C API call `getLoadAvg()` (see `man getloadavg` for more) and calls `getLoadAverage(loadavg, 1)` and returns `loadavg[0]`.  This implies it would return the 1-minute (first) field of the load average.

It would be reasonable to add that direct implementation to OSHI to return multiple load averages for multiple time periods, per the API.  Just overload and add an optional `int` argument to `getSystemLoadAverage()` and return up to the 3 values stored. But rather than parsing `/proc/loadavg` we should use the system call in Libc using JNA.

dbwiddis:
Here's the Java implementation for reference.
```
public double getSystemLoadAverage() 
{
        if (unsafe.getLoadAverage(loadavg, 1) == 1) {
             return loadavg[0];
        } else {
             return -1.0;
        }
}
```

henryx:
Ok, I've restored the `getSystemCpuLoad()` method and I've modified the `getSystemLoadAverage()` method for using the JNA call. For me, implementing a method that return the three values for load average is out of scope for this pull request (my goal is remove as much possible the Java MXBean calls for ensuring that works with other JVMs)

dbwiddis:
Looks good.  I'll accept this and then make some of my own tweaks (extending it to OS X and adding the "return 3 values" version.)

Fair warning, I think when we build the release it overwrites the headers and you may lose the specific credit you added on these files.  If you keep contributing a lot more I'll add you to the template. :)

dbwiddis:
Also, FYI @hazendaz I do have the "Squash and commit" option available and will be using it here!

PR #141, opened 2016-04-18
Rewritten LinuxGlobalMemory class for using JNA system call instead of /proc/meminfo parsing

henryx:


henryx:
I'm sorry if the merge has created problems, if you want, I recreate the pull request after sanitize my fork

dbwiddis:
I'm very glad to have an enthusiastic contributor! :)

Unfortunately, as tempting as it is to use sysinfo for the memory information, it's wrong.

For background, this exact "free memory" issue is what got me starting to contribute to oshi in the first place, about a year ago, so I studied this in way too much detail.  I spent a lot of time debugging what I thought was a memory leak in my program when it turned out it was simply displaying the "free" memory on Linux which drops close to zero over time; plenty of memory is available, but it's "cached".

While adding free and cached can be a decent estimate, it's a bit more fine grained than that, and  `/proc/meminfo` is the correct place to get Available Memory, [per Linus Torvalds himself](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773).

The `/proc` filesystem isn't actually real files and incurs no disk reading delay to read from it; it's become the de facto interface to the Linux kernel.  We can try to reverse engineer the source code (which I tried to do to get the available number in meminfo) but in this case it was rather complex to try to follow.

henryx:
Mmm... I understand... So, feel free to reject my pull request

2016-04-19 1:07 GMT+02:00 dbwiddis <notifications@github.com>:

PR #142, opened 2016-04-19
Added methods for getting swap (total and used) in the system

henryx:


henryx:
Small note: I've used the `proc` file system for uniforming to the rest of the file

src/main/java/oshi/hardware/platform/linux/LinuxGlobalMemory.java
```diff
@@ -75,7 +76,7 @@ public long getAvailable() {
                 availableMemory += parseMeminfo(memorySplit);
             } else if (checkLine.startsWith("Active(file):")) {
                 String[] memorySplit = checkLine.split("\\s+");
-                availableMemory += parseMeminfo(memorySplit);
+//                availableMemory += parseMeminfo(memorySplit);
```
dbwiddis: Why is this commented out?   Left over from a debug?  [Active(file) is part of MemAvailable](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773).

src/main/java/oshi/hardware/platform/linux/LinuxGlobalMemory.java
```diff
@@ -121,6 +122,51 @@ public long getTotal() {
...
         return this.totalMemory;
     }
+    
+    @Override
+    public long getSwapAvailable() {
+        long swapAvailable = 0;
+        List<String> memInfo = null;
+        
+        try {
+            memInfo = FileUtil.readFile("/proc/meminfo");
```
dbwiddis: You can actually get swap free and total from the sysinfo structure (that's where /proc/meminfo gets it).  See `getTotal()` for a template; you can make the swap total a peristent class variable to save repeated reads.

henryx:
Yes, I've viewed the `sysinfo()` structure, but for consistency with the rest of the class, I've preferred to use the `/proc/meminfo` file  (in this regard, I ask if I can rewrite the class to make it more uniform and maintainable)

src/main/java/oshi/hardware/platform/mac/MacGlobalMemory.java
```diff
@@ -86,6 +87,18 @@ public long getTotal() {
     }
 
     @Override
+    public long getSwapTotal() {
+        // TODO: for now, it returns 0
```
dbwiddis: On Mac, use a `sysctl()` call with the name `vm.swapusage`.
It will return a string like this:
total = 1024.00M  used = 440.75M  free = 583.25M  (encrypted)

Note that unlike Linux, total can increase (up to free space on drive) if more swap is needed.

henryx:
I've preferred to don't implement Windows and Mac `getSwapTotal()` and `getSwapAvailable()` because I haven't these systems, so I didn't the ability to test it

src/main/java/oshi/hardware/platform/windows/WindowsGlobalMemory.java
```diff
@@ -68,6 +69,18 @@ public long getTotal() {
     }
 
     @Override
+    public long getSwapTotal() {
+        // TODO: for now, it returns 0
+        return 0;
+    }
+
+    @Override
+    public long getSwapAvailable() {
```
dbwiddis: Amount used from Performance Monitor counters (see `WindowsCentralProcessor`) under "\Paging File(_Total)\% Usage".  Will have to reference that percent used to total and subtract used from total to get available.

src/main/java/oshi/hardware/platform/windows/WindowsGlobalMemory.java
```diff
@@ -68,6 +69,18 @@ public long getTotal() {
     }
 
     @Override
+    public long getSwapTotal() {
+        // TODO: for now, it returns 0
```
dbwiddis: Windows swap info is part of the [MEMORYSTATUSEX](https://msdn.microsoft.com/en-us/library/windows/desktop/aa366770(v=vs.85).aspx) structure.  Copy code from `getAvailable()` and change which structure element you're returning.

dbwiddis:
OK, looks good.   

Feel free to rewrite the whole LinuxGlobalMemory class for readability/maintainability.

Here's my thoughts on `sysinfo()` vs. `/proc/meminfo`:
- Internally, the /proc/meminfo code calls `si_meminfo()` and `si_swapinfo()` to populate relevant portions of the `sysinfo` structure.  
- Internally, the `sysinfo()` code eventually references the same lower level calls that `/proc/meminfo` does, but also populates the other fields of the `sysinfo` structure.

Given that only the needed information is being populated, it may be more efficient to simply call `/proc/meminfo` in all cases, unless we want to implement lower level calls to `si_meminfo()`.   Since the Linux Kernel maintainers have indicated in the `MemAvailable` commit that they want this to be the "single point" it's reasonable to simply read all the memory information from `/proc/meminfo` and ignore the `sysinfo()` call.

No need for you to implement the Mac port, I'll do that after merging this one, my notes were mostly to myself on how to do it so I had a reference.  Windows is a trivial copy/paste/edit that I can do if you don't.. I'll be sure to test it in any case.

dbwiddis:
For Mac, we also have a philosophical question to answer...

OS X uses all of the free space on the drive as potential swap file space.  So when calculating "available" you'd actually get free space on that drive (`statfs("/", &stats);` and `free = `stats.f_bsize * stats.f_bfree;`.  But calculating used requires parsing `sysctl` for `vm.swapusage` which displays its own total and used values.

So which should we report as "available" for OS X?  I think it'd be silly on my current system to pretend I have 1.5TB of available virtual memory (even though it's true).  I'd much rather have oshi report similarly to how `vm.swapusage` does, and report used, and total (with the knowledge that total can increase).  This at least pushes off the question of "available" to the user.

So I'm leaning toward having a `swapUsed()` method in the interface rather than a `swapAvailable()` method.  Thoughts?

dbwiddis:
Also, the Windows `MEMORYSTATUSEX` is apparently limited to the current process, so it's better to call [GetPerformanceInfo()](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683210(v=vs.85).aspx) for the systemwide statistics.  This also reports a committed (used) value rather than an available value.

dbwiddis:
This is good, unless you were going to rewrite the class?  Let me know if I should merge or not.

src/main/java/oshi/hardware/platform/linux/LinuxGlobalMemory.java
```diff
@@ -124,7 +124,7 @@ public long getTotal() {
     }
     
     @Override
-    public long getSwapAvailable() {
+    public long getSwapUsed() {
         long swapAvailable = 0;
```
dbwiddis: You changed the method name but not the internals.  Also, need to change (or commit if you changed already) the updated SystemInfoTest.

henryx:
My fault, I've corrected. For now, I ask you to wait before merge, I'll try to implement the Windows methods

dbwiddis:
Here's the code you need for OS X.  Drop it in. :)
```

    public long getSwapUsed() {
        IntByReference size = new IntByReference();
        if (0 != SystemB.INSTANCE.sysctlbyname("vm.swapusage", null, size, null, 0)) {
            LOG.error("Failed to get Swap Usage. Error code: " + Native.getLastError());
            return 0L;
        }
        Pointer p = new Memory(size.getValue() + 1);
        System.out.println("Size: " + size.getValue());
        if (0 != SystemB.INSTANCE.sysctlbyname("vm.swapusage", p, size, null, 0)) {
            LOG.error("Failed to get Swap Usage. Error code: " + Native.getLastError());
            return 0L;
        }
        // p points to an array of four longs
        // first 3 elements total, free, used
        return p.getLong(16);
    }

    public long getSwapTotal() {
        IntByReference size = new IntByReference();
        if (0 != SystemB.INSTANCE.sysctlbyname("vm.swapusage", null, size, null, 0)) {
            LOG.error("Failed to get Swap Usage. Error code: " + Native.getLastError());
            return 0L;
        }
        Pointer p = new Memory(size.getValue() + 1);
        if (0 != SystemB.INSTANCE.sysctlbyname("vm.swapusage", p, size, null, 0)) {
            LOG.error("Failed to get Swap Usage. Error code: " + Native.getLastError());
            return 0L;
        }
        // p points to an array of four longs
        // first 3 elements total, free, used
        return p.getLong(0);
    }
```

Don't forget to add the new fields to the `toJSON()` method as well, and update README with the new sample output from `SystemInfoTest`.

henryx:
All done, I think that is finished :)

dbwiddis:
Looks good. I'll merge this.   As I noted earlier, the [MEMORYSTATUSEX ](https://msdn.microsoft.com/en-us/library/windows/desktop/aa366770(v=vs.85).aspx) docs for ullTotalPageFile state "The current committed memory limit for the system or the current process, whichever is smaller, in bytes. To get the system-wide committed memory limit, call [GetPerformanceInfo](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683210(v=vs.85).aspx)."   Since I just implemented that function, I'll go ahead and switch the Windows code over.   I may tweak the Linux code a bit to improve efficiency.  Good work!

henryx:
I haven't used [GetPerformanceInfo](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684824%28v=vs.85%29.aspx) because it doesn't report the total (physical + page file) memory size. I've reported in the `getSwapTotal()` method a better way to extract the information (note this: the Python and the Golang equivalent of this library they uses the [MEMORYSTATUSEX](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684824%28v=vs.85%29.aspx) structure for getting the swap size)

dbwiddis:
GetPerformanceInfo includes the commit limit, which is the physical + pagefile total pages.  (It also has page size in the structure for easy multiplication.)

Getting used is a bit more complicated. The structures you had from MEMORYSTATUSEX (and the equivalent from GetPerformanceInfo) report the total (pagefile + physical) committed/available values which are not directly relatable to the physical used/available.   I've ended up using the Performance Monitor counter for % Page file usage to get what we need.

Done with the Windows tweaks, now working on reducing redundant reads in Linux.

dbwiddis:
Committed the updates...

PR #147, opened 2016-04-23
Added disk information and statistics

henryx:
For creating this pull request, I've used the same logic for File System information extract

### DIFF
--- a/src/main/java/oshi/jna/platform/linux/Udev.java
+++ b/src/main/java/oshi/jna/platform/linux/Udev.java
@@ -35,34 +35,58 @@ public interface Udev extends Library {
 
     public static final class BlockDevStats extends Structure {
 
+        public BlockDevStats(long read_ops,
+                long read_merged,
+                long read_512bytes,
+                long read_waits_ms,
+                long write_ops,
+                long write_merged,
+                long write_512bytes,
+                long write_waits_ms,
+                long in_flight,
+                long active_ms,
+                long waits_ms) {
+            this.read_ops = read_ops;
+            this.read_merged = read_merged;
+            this.read_512bytes = read_512bytes;
+            this.read_waits_ms = read_waits_ms;
+            this.write_ops = write_ops;
+            this.write_merged = write_merged;
+            this.write_512bytes = write_512bytes;
+            this.write_waits_ms = write_waits_ms;
+            this.in_flight = in_flight;
+            this.active_ms = active_ms;
+            this.waits_ms = waits_ms;
+        }
+
         // Number of read I/Os processed
-        long read_ops;
+        public final long read_ops;
         // Number of read I/Os merged with in-queue I/O
-        long read_merged;
+        public final long read_merged;
         // Number of sectors read
-        long read_512bytes;
+        public final long read_512bytes;
         // Total wait time for read requests, milliseconds
-        long read_waits_ms;
+        public final long read_waits_ms;
         // Number of write I/Os processed
-        long write_ops;
+        public final long write_ops;
         // Number of write I/Os merged with in-queue I/O
-        long write_merged;
+        public final long write_merged;
         // Number of sectors written
-        long write_512bytes;
+        public final long write_512bytes;
         // Total wait time for write requests, milliseconds */
-        long write_waits_ms;
+        public final long write_waits_ms;
         // Number of I/Os currently in flight
-        long in_flight;
+        public final long in_flight;
         // Total active time, milliseconds
-        long active_ms;
+        public final long active_ms;
         // Total wait time, milliseconds
-        long waits_ms;
+        public final long waits_ms;
 
         @Override
         protected List getFieldOrder() {
             return Arrays.asList(new String[]{"read_ops", "read_merged", "read_512bytes",
-                "read_waits_ms", "write_ops", "write_merged", "write_512bytes", "write_watis_ms",
-                "in_flight", "actice_ms", "watis_ms"});
+                "read_waits_ms", "write_ops", "write_merged", "write_512bytes", "write_waits_ms",
+                "in_flight", "active_ms", "waits_ms"});
         }
     };
 
--- a/src/main/java/oshi/jna/platform/linux/Udev.java
+++ b/src/main/java/oshi/jna/platform/linux/Udev.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) 2016 com.github.dblock.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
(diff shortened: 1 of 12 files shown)


henryx:
Mmm... I don't know if this pull request has some commits of @dbwiddis, I don't know if it is a problem...

henryx:
Ok, I've fixed the pull request's history, now shouldn't have problems

dbwiddis:
Looks great.   

Unfortunately, I'm not sure if you noticed, but I've been doing some major refactoring over the last 48 hours in #145 ... pretty much touched every file.   The good news is all this is new code from you so it shouldn't conflict too much (just a few inserted lines in the `SystemInfoTest` and the `HardwareAbstractionLayer` classes.)  So we need to decide whose code to merge first!

I can probably merge this and then continue refactoring your code, or I can ask you to refactor to match what I'm doing, merge my code, and let you adjust.   Not sure if either way is much easier.  Given that you've done some things that I'd like to change my OSFileStore code to match, I think I'll go with merging yours and then hoping it doesn't make my code a disaster. :)

Two quick changes you (and/or I) need:
* You define a `Disks` interface but never implement it.   Were the individual platform Disks objects intended to implement Disks?
* The `HWDiskStore` object is at the top level along with all the other interfaces.  I realized in my refactoring that the `OSFileStore` was similarly situated, and separated it into an interface and then an abstract class extended by the individual file store objects (which only contained static methods).  See [this commit](https://github.com/dblock/oshi/pull/145/commits/2139dfea6bc51a5ad32c6788c195b298bac5632b) for how I split it out.   Unfortunately, the static method precludes the use of an interface or abstract method (like you've done with the `Disks` interface).  I'm open to ideas on how to better implement this.

src/main/java/oshi/jna/platform/linux/Udev.java
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Interface for talking with Udev.
+ *
+ * @author enrico[dot]bianchi[at]gmail[dot]com
+ */
+public interface Udev extends Library {
+
+    public static final NativeLibrary JNA_NATIVE_LIB = NativeLibrary.getInstance("udev");
+    public static final Udev INSTANCE = (Udev) Native.loadLibrary("udev", Udev.class);
```
dbwiddis: I'm not sure you need both of the above lines.  In all my JNA Library code I've only included the `INSTANCE` version.  I don't see you accessing the `JNA_NATIVE_LIB` anywhere else in your code.

src/main/java/oshi/jna/platform/linux/Udev.java
```diff
@@ -0,0 +1,168 @@
...
+import com.sun.jna.Structure;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Interface for talking with Udev.
+ *
+ * @author enrico[dot]bianchi[at]gmail[dot]com
+ */
+public interface Udev extends Library {
```
dbwiddis: Please consider contributing this library to the [JNA](https://github.com/java-native-access/jna) project!

src/main/java/oshi/jna/platform/linux/Udev.java
```diff
@@ -0,0 +1,168 @@
...
+
+    public static final NativeLibrary JNA_NATIVE_LIB = NativeLibrary.getInstance("udev");
+    public static final Udev INSTANCE = (Udev) Native.loadLibrary("udev", Udev.class);
+
+    public static final class BlockDevStats extends Structure {
+
+        public BlockDevStats(long read_ops,
```
dbwiddis: I think JNA automatically does reads of structures and this constructor is unnecessary.

src/test/java/oshi/SystemInfoTest.java
```diff
@@ -225,6 +226,27 @@ public void testFileSystem() throws IOException {
...
+     * Test disks extraction.
+     *
+     * @throws IOException
+     *             Signals that an I/O exception has occurred.
+     */
+    @Test
+    public void testDisks() throws IOException {
+        SystemInfo si = new SystemInfo();
+        HardwareAbstractionLayer hal = si.getHardware();
+        if (hal.getDisksStores().length > 1) {
```
dbwiddis: Change this conditional to iterate with a counter from 0 to length, and check the counter rather than 0 in the lines below.

src/main/java/oshi/util/PdhUtil.java
```diff
@@ -0,0 +1,69 @@
...
+import com.sun.jna.ptr.PointerByReference;
+
+import oshi.jna.platform.windows.Pdh;
+
+/**
+ * Provides access to open pdh queries and add counters
+ * 
+ * @author widdis[at]gmail[dot]com
+ */
+public class PdhUtil {
```
dbwiddis: This is an older version than the currently committed PdhUtil.  The newer one is in a platform-specific subdirectory.   You can probably just delete this version of the file from your local branch to make it go away from this PR.

henryx:
My fault, I've created the `Disks` interface but for distraction I haven't implemented it. I've correct this

src/main/java/oshi/util/PdhUtil.java
henryx: I've rebased the my repository because for some reasons pull request was ported old commits before my code. I'll correct it

dbwiddis:
Figured that's what you intended.  :)  I think I'll use this template for the `OSFileStores` as well.  (I still may separate your interface and implementation into separate classes.)

henryx:
Ok, I've removed all unusual commits for this PR

dbwiddis:
Looks good enough to merge, and I can tweak.  Are you still working on any of my comments?

src/main/java/oshi/jna/platform/linux/Udev.java
henryx: It's a mistake, the code that populate this structure isn't the JNA library, because the `udev_device_get_sysattr_value();` syscall returns always a string. So, I converted in a POJO the structure but I don't moved in a correct place

dbwiddis:
Note that your tests are failing as of your "fixed check if device is a physical disk" commit.

henryx:
Mmm... It seems that the CI doesn't work with the `assertNotNull(disk.getName())`, so I don't know how continue it (note: `disk.getName()` return the name of the physical device, probably the CI uses a device mapper for working and for this it doesn't work - with a7dd463 I've excluded device mappers because there aren't real disks)

dbwiddis:
Merged this, resolved conflicts with my refactoring, abstracted your code to match my patterns, and committed my mass refactoring in #145.  You'll want to update before doing any more coding.

A few outstanding issues on this PR, see #148.  Need to add JSON to the HAL and filter out loopback and ram disks (or use some other method to exclude 24 spammy non-disks from my output.)

Issue #138, opened 2016-04-14
Add network information


dbwiddis:
Add a new interface for network information to include MAC addresses, IPv4 and IPv6, cumulative bandwidth, etc.

dbwiddis:
Looks like most (all?) of this is [already available in core java methods](https://docs.oracle.com/javase/tutorial/networking/nifs/index.html).

dbwiddis:
Name, IP(s), MAC are all in core Java.
Bandwidth
- OS X using `sysctl`, see [here](http://stackoverflow.com/questions/1126790/how-to-get-network-adapter-stats-in-linux-mac-osx).
- Linux `/sys/class/net/eth0/statistics/` or `/proc.net/dev`
- Windows [GetIfEntry](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365939(v=vs.85).aspx) or WMI (`wmic path Win32_PerfRawData_Tcpip_NetworkInterface get BytesReceivedPersec,BytesSentPersec,BytesTotalPersec`)

dbwiddis:
Linux done in #155 
OS X done in #157 
Windows in progress, will add to #157

PR #155, opened 2016-04-29
 Added network interfaces information and statistics

henryx:
This code implements (partially for all systems and extensively for Linux) issue #138

### DIFF
--- a/src/main/java/oshi/hardware/Networks.java
+++ b/src/main/java/oshi/hardware/Networks.java
@@ -33,5 +33,5 @@ public interface Networks extends OshiJsonObject {
      * 
      * @return Array of {@link HWNetworkStore} objects
      */
-    public HWNetworkStore[] getNetworks();
+    HWNetworkStore[] getNetworks();
 }
--- a/src/main/java/oshi/hardware/Networks.java
+++ b/src/main/java/oshi/hardware/Networks.java
@@ -1,0 +1,37 @@
+/**
+ * Oshi (https://github.com/dblock/oshi)
+ *
+ * Copyright (c) 2010 - 2016 The Oshi Project Team
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Maintainers:
+ * dblock[at]dblock[dot]org
+ * widdis[at]gmail[dot]com
+ * enrico[dot]bianchi[at]gmail[dot]com
+ *
+ * Contributors:
+ * https://github.com/dblock/oshi/graphs/contributors
+ */
+package oshi.hardware;
+
+import oshi.hardware.stores.HWNetworkStore;
+import oshi.json.OshiJsonObject;
+
+/**
+ * Networks refers to network interfaces installed in the machine.
+ *
+ * @author enrico[dot]bianchi[at]gmail[dot]com
+ */
+public interface Networks extends OshiJsonObject {
+
+    /**
+     * Get network interfaces on this machine
+     * 
+     * @return Array of {@link HWNetworkStore} objects
+     */
+    public HWNetworkStore[] getNetworks();
+}
--- a/src/main/java/oshi/hardware/common/AbstractNetworks.java
+++ b/src/main/java/oshi/hardware/common/AbstractNetworks.java
@@ -43,6 +43,7 @@ public abstract class AbstractNetworks implements Networks {
 
     /**
      * Set network parameters in store
+     *
      * @param netstore Store which set network parameters
      * @param netint Network interface examined
      * @throws SocketException
@@ -63,9 +64,13 @@ public abstract class AbstractNetworks implements Networks {
         addresses = netint.getInetAddresses();
         for (InetAddress address : Collections.list(addresses)) {
             if (address.getHostAddress().contains(":")) {
-                netstore.setIpv6addr(address.getHostAddress().split("%")[0]);
+                netstore.setIpv6addr(
+                        (address.getHostAddress().split("%")[0] == null) ? "Unknown"
+                        : address.getHostAddress().split("%")[0]);
             } else {
-                netstore.setIpaddr(address.getHostAddress());
+                netstore.setIpaddr(
+                        (address.getHostAddress() == null) ? "Unknown"
+                                : address.getHostAddress());
             }
         }
     }
--- a/src/main/java/oshi/hardware/common/AbstractNetworks.java
+++ b/src/main/java/oshi/hardware/common/AbstractNetworks.java
@@ -55,22 +55,21 @@ public abstract class AbstractNetworks implements Networks {
 
         mac = netint.getHardwareAddress();
         netstore.setName(netint.getDisplayName());
+
         sb = new StringBuilder();
         for (int i = 0; i < mac.length; i++) {
             sb.append(String.format("%02X%s", mac[i], (i < mac.length - 1) ? ":" : ""));
         }
         netstore.setMacaddr(sb.toString());
 
+        netstore.setIpaddr("Unknown");
+        netstore.setIpv6addr("unknown");
         addresses = netint.getInetAddresses();
         for (InetAddress address : Collections.list(addresses)) {
             if (address.getHostAddress().contains(":")) {
-                netstore.setIpv6addr(
-                        (address.getHostAddress().split("%")[0] == null) ? "Unknown"
-                        : address.getHostAddress().split("%")[0]);
+                netstore.setIpv6addr(address.getHostAddress().split("%")[0]);
             } else {
-                netstore.setIpaddr(
-                        (address.getHostAddress() == null) ? "Unknown"
-                                : address.getHostAddress());
+                netstore.setIpaddr(address.getHostAddress());
             }
         }
     }
--- a/src/main/java/oshi/hardware/common/AbstractNetworks.java
+++ b/src/main/java/oshi/hardware/common/AbstractNetworks.java
@@ -54,7 +54,8 @@ public abstract class AbstractNetworks implements Networks {
         byte[] mac;
 
         mac = netint.getHardwareAddress();
-        netstore.setName(netint.getDisplayName());
+        netstore.setName(netint.getName());
+        netstore.setDescription(netint.getDisplayName());
 
         sb = new StringBuilder();
         for (int i = 0; i < mac.length; i++) {
--- a/src/main/java/oshi/hardware/common/AbstractNetworks.java
+++ b/src/main/java/oshi/hardware/common/AbstractNetworks.java
@@ -63,7 +63,7 @@ public abstract class AbstractNetworks implements Networks {
         addresses = netint.getInetAddresses();
         for (InetAddress address : Collections.list(addresses)) {
             if (address.getHostAddress().contains(":")) {
-                netstore.setIpv6addr(address.getHostAddress());
+                netstore.setIpv6addr(address.getHostAddress().split("%")[0]);
             } else {
                 netstore.setIpaddr(address.getHostAddress());
             }
--- a/src/main/java/oshi/hardware/common/AbstractNetworks.java
+++ b/src/main/java/oshi/hardware/common/AbstractNetworks.java
@@ -55,7 +55,7 @@ public abstract class AbstractNetworks implements Networks {
 
         mac = netint.getHardwareAddress();
         netstore.setName(netint.getName());
-        netstore.setDescription(netint.getDisplayName());
+        netstore.setDisplayName(netint.getDisplayName());
 
         sb = new StringBuilder();
         for (int i = 0; i < mac.length; i++) {
--- a/src/main/java/oshi/hardware/common/AbstractNetworks.java
+++ b/src/main/java/oshi/hardware/common/AbstractNetworks.java
@@ -1,0 +1,84 @@
+/**
+ * Oshi (https://github.com/dblock/oshi)
+ *
+ * Copyright (c) 2010 - 2016 The Oshi Project Team
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Maintainers:
+ * dblock[at]dblock[dot]org
+ * widdis[at]gmail[dot]com
+ * enrico[dot]bianchi[at]gmail[dot]com
+ *
+ * Contributors:
+ * https://github.com/dblock/oshi/graphs/contributors
+ */
+package oshi.hardware.common;
+
+import java.net.InetAddress;
+import java.net.NetworkInterface;
+import java.net.SocketException;
+import java.util.Collections;
+import java.util.Enumeration;
+import javax.json.Json;
+import javax.json.JsonArrayBuilder;
+import javax.json.JsonBuilderFactory;
+import javax.json.JsonObject;
+
+import oshi.hardware.Networks;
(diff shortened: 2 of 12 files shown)


src/main/java/oshi/hardware/Networks.java
```diff
@@ -0,0 +1,37 @@
...
+ * @author enrico[dot]bianchi[at]gmail[dot]com
+ */
+public interface Networks extends OshiJsonObject {
+
+    /**
+     * Get network interfaces on this machine
+     * 
+     * @return Array of {@link HWNetworkStore} objects
+     */
+    public HWNetworkStore[] getNetworks();
```
dbwiddis: Standardizing per PMD that I just finished updating elsewhere, no need for `public` (or `static` or `final`) modifiers in interfaces.

src/main/java/oshi/hardware/stores/HWNetworkStore.java
```diff
@@ -0,0 +1,164 @@
...
+import javax.json.JsonObject;
+import oshi.json.NullAwareJsonObjectBuilder;
+import oshi.json.OshiJsonObject;
+
+/**
+ * Store object of network interfaces attributes
+ *
+ * @author enrico[dot]bianchi[at]gmail[dot]com
+ */
+public class HWNetworkStore implements OshiJsonObject {
```
dbwiddis: I'm not sure the "store" is applicable to networks... was ok for FileStores (software) and DiskStores (hardware).  Probably want `NetworkInterface`

dbwiddis:
Thanks for doing this!  Saving me a lot of time and effort. :)  

CI tests are failing, haven't looked to see why, but hopefully when you get those fixed this will be good to go.   Let me know if you need me to do more for the Mac and Windows portions.

src/main/java/oshi/hardware/platform/linux/LinuxNetworks.java
```diff
@@ -0,0 +1,88 @@
...
+        List<String> read;
+
+        txBytesPath = "/sys/class/net/" + netstore.getName() + "/statistics/tx_bytes";
+        rxBytesPath = "/sys/class/net/" + netstore.getName() + "/statistics/rx_bytes";
+        txPacketsPath = "/sys/class/net/" + netstore.getName() + "/statistics/tx_packets";
+        rxPacketsPath = "/sys/class/net/" + netstore.getName() + "/statistics/rx_packets";
+
+        try {
+            read = FileUtil.readFile(txBytesPath);
+            netstore.setBytesSent(Long.parseLong(read.get(0)));
```
dbwiddis: `FileUtil.readFile()` could potentially return an empty array.  You need to test `.size()` before using `.get(0)`.

src/main/java/oshi/hardware/platform/linux/LinuxNetworks.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+public class LinuxNetworks extends AbstractNetworks {
+
+    private static final Logger LOG = LoggerFactory.getLogger(LinuxNetworks.class);
+
+    private void setNetworkStats(HWNetworkStore netstore) {
+        String txBytesPath, rxBytesPath, txPacketsPath, rxPacketsPath;
+        List<String> read;
+
+        txBytesPath = "/sys/class/net/" + netstore.getName() + "/statistics/tx_bytes";
```
dbwiddis: These strings should probably be constants.  You might try a setup like:
  `private static final String TX_BYTES_FORMAT = "/sys/class/net/%s/statistics/tx_bytes";
`And then, to use,
  `txBytesPath = String.format(TX_BYTES_FORMAT, netstore.getName());`

src/main/java/oshi/hardware/platform/linux/LinuxNetworks.java
```diff
@@ -0,0 +1,88 @@
...
+            netstore.setPacketsRecv(Long.parseLong(read.get(0)));
+        } catch (IOException ex) {
+            LOG.error("Error when retrieving network statistics for interface " + netstore.getName());
+            LOG.debug("Error message: " + ex.getMessage());
+        }
+
+    }
+
+    @Override
+    public HWNetworkStore[] getNetworks() {
```
dbwiddis: This method looks common to all 3 classes. Shouldn't it be in the parent with only the `setNetworkStats()` method in the platform-specific versions?

src/main/java/oshi/hardware/platform/mac/MacNetworks.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ *
+ * @author enrico[dot]bianchi[at]gmail[dot]com
+ */
+public class MacNetworks extends AbstractNetworks {
+
+    private static final Logger LOG = LoggerFactory.getLogger(MacNetworks.class);
+
+    private void setNetworkStats(HWNetworkStore netstore) {
+        // TODO: set network stats on Mac
```
dbwiddis: So, on Mac, the code to get the stats returns an array of structures with the data, so it's inefficient to read the entire array every time just to match the particular interface name.  To fit that code into this structure we'd probably read it once and cache the data in a map with the netstore.getName() as the key.

src/main/java/oshi/hardware/stores/HWNetworkStore.java
```diff
@@ -0,0 +1,164 @@
...
+
+    private String name;
+    private String macaddr;
+    private String ipaddr;
+    private String ipv6addr;
+    private long bytesRecv;
+    private long bytesSent;
+    private long packetsRecv;
+    private long packetsSent;
+
```
dbwiddis: Let's add, as a minimum, MTU (maximum transmission unit) and speed/baudrate (Bits per second) to these values as I believe they're available on all 3 platforms.

For reference to see if there are any other stats you want (dropped packets, errors, etc.) compare the linux available data with the data available on OS X in [if_data64](http://fxr.watson.org/fxr/source/bsd/net/if_var.h?v=xnu-792.6.70#L178) and on Windows in [MIB_IFROW](https://msdn.microsoft.com/en-us/library/windows/desktop/aa366836(v=vs.85).aspx)

src/main/java/oshi/hardware/platform/linux/LinuxNetworks.java
```diff
@@ -0,0 +1,88 @@
...
+
+            read = FileUtil.readFile(rxBytesPath);
+            netstore.setBytesRecv(Long.parseLong(read.get(0)));
+
+            read = FileUtil.readFile(txPacketsPath);
+            netstore.setPacketsSent(Long.parseLong(read.get(0)));
+
+            read = FileUtil.readFile(rxPacketsPath);
+            netstore.setPacketsRecv(Long.parseLong(read.get(0)));
+        } catch (IOException ex) {
```
dbwiddis: You also need to catch `NumberFormatException` if the `parseLong()` format is invalid.

src/main/java/oshi/hardware/platform/windows/WindowsNetworks.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ *
+ * @author enrico[dot]bianchi[at]gmail[dot]com
+ */
+public class WindowsNetworks extends AbstractNetworks {
+
+    private static final Logger LOG = LoggerFactory.getLogger(WindowsNetworks.class);
+
+    private void setNetworkStats(HWNetworkStore netstore) {
+        // TODO: set network stats on Windows
```
dbwiddis: Similar to mac, there's no easy lookup of stats by name. You have to iterate over all devices and get all the info, so we'll read and cache it the first time (per polling interval) so subsequent calls within the interval can just read from the map.

src/main/java/oshi/hardware/common/AbstractNetworks.java
```diff
@@ -0,0 +1,89 @@
...
+    protected void setNetworkParameters(HWNetworkStore netstore, NetworkInterface netint) throws SocketException {
+        Enumeration<InetAddress> addresses;
+        StringBuilder sb;
+        byte[] mac;
+
+        mac = netint.getHardwareAddress();
+        netstore.setName(netint.getDisplayName());
+        sb = new StringBuilder();
+        for (int i = 0; i < mac.length; i++) {
+            sb.append(String.format("%02X%s", mac[i], (i < mac.length - 1) ? ":" : ""));
```
dbwiddis: Stuff like this makes me want to upgrade to Java 8. :-)

src/main/java/oshi/hardware/stores/HWNetworkStore.java
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * Store object of network interfaces attributes
+ *
+ * @author enrico[dot]bianchi[at]gmail[dot]com
+ */
+public class HWNetworkStore implements OshiJsonObject {
+
+    private JsonBuilderFactory jsonFactory = Json.createBuilderFactory(null);
+
+    private String name;
```
dbwiddis: In addition to name (such as eth0), add the DisplayName field (such as Wireless Network Connection) from Java.

src/main/java/oshi/hardware/stores/HWNetworkStore.java
henryx: I've intended "store" as generic term for reporting data. In fact, extended version can be FileStoreObject or NetworkStoreObject or, if you want use another term, FileData or NetworkData

src/main/java/oshi/hardware/platform/linux/LinuxNetworks.java
henryx: `String.format()` can be an idea, but I'm not sure if `TX_BYTES_FORMAT` should be a per class static variable. Maybe a good way can be `txBytesPath = String.format("/sys/class/net/%s/statistics/tx_bytes", netstore.getName()); `?

src/main/java/oshi/hardware/platform/linux/LinuxNetworks.java
henryx: I agree, but I haven't any idea for transfer it in `AbstractNetwoks` without break all...

src/main/java/oshi/hardware/platform/mac/MacNetworks.java
henryx: For now, this code is a simple placement (I have no idea how implement it on Mac)

src/main/java/oshi/hardware/platform/windows/WindowsNetworks.java
henryx: I'll view it on Monday, sadly, for now I haven't a Windows machine for working on it

src/main/java/oshi/hardware/platform/linux/LinuxNetworks.java
```diff
@@ -50,17 +50,19 @@ private long readFile(String file) {
...
 
     private void setNetworkStats(HWNetworkStore netstore) {
-        String txBytesPath, rxBytesPath, txPacketsPath, rxPacketsPath;
+        String txBytesPath, rxBytesPath, txPacketsPath, rxPacketsPath, speed;
 
         txBytesPath = String.format("/sys/class/net/%s/statistics/tx_bytes", netstore.getName());
         rxBytesPath = String.format("/sys/class/net/%s/statistics/rx_bytes", netstore.getName());
         txPacketsPath = String.format("/sys/class/net/%s/statistics/tx_packets", netstore.getName());
         rxPacketsPath = String.format("/sys/class/net/%s/statistics/rx_packets", netstore.getName());
+        speed = String.format("/sys/class/net/%s/speed", netstore.getName());
```
henryx: It's ugly, because it isn't a statistic, but I haven't other ideas how to place it

src/main/java/oshi/hardware/platform/mac/MacNetworks.java
dbwiddis: I know. I'll be implementing the Mac version.  This was more a comment about thinking higher level that looking up the Java information and iterating one by one to get this information may not be the best approach.

And after more research I found a way to look it up by interface (setting the `mib[5]` value to the index number found in a lookup by name function).

src/main/java/oshi/hardware/platform/windows/WindowsNetworks.java
dbwiddis: Don't worry about implementing the Windows port.  Just know that looking up "by name" isn't as easy as you might think.  I can still make it work, though.

src/main/java/oshi/hardware/stores/HWNetworkStore.java
dbwiddis: That makes sense, but we've named all the first-level interfaces and objects corresponding to physical names. CentralProcessor, GlobalMemory, PowerSource, Sensors, Display, FileStore, and DiskStore (where Store = Storage).  Network interfaces (and Java `interface`s) are broken if they store anything. They're just supposed to send stuff from one place to another. :)

Since Java has already claimed the `NetworkInterface` name, we should modify it slightly.  Mac and Windows methods make a lot of use of the IF abbreviation (so dos Linux `ifconfig`) so we should definitely use IF or Interface in the name. NWInterface or NetworkIF or HWNetworkInterface or HWNetInterface or NetIF...

src/main/java/oshi/hardware/platform/linux/LinuxNetworks.java
dbwiddis: Sure, that works.  My personal preference is to not hardcode strings in the source, but make them private static variables in their appropriate class.   Not sure if that's official coding style/best practice but it's what I learned in school a long time ago.   Although I've made exceptions for file paths which this is, so it's probably just a matter of preference here.

src/main/java/oshi/hardware/platform/linux/LinuxNetworks.java
dbwiddis: Just move this method as is to `AbstractNetworks` and put in an abstract method for `setNetworkStats()` which you will implement class-specific.

src/main/java/oshi/hardware/stores/HWNetworkStore.java
dbwiddis: Didn't realize Description already had that... but I like matching the Java name as you've already done.

dbwiddis:
So, it looks like this mostly works?  Other than the renaming/refactoring that I've suggested?   I might merge this and work on it myself over the weekend to rename/refactor and implement the Mac and Linux ports.  Let me know if you're at a good stopping point.

dbwiddis:
Also a brief google search for other projects includes IPNetworkInterface or WANNetworkInterface.  Some of them actually extend the Java NetworkInterface class and then add their own info... I wonder if we should go that route.

henryx:
I think is done. I've factorized some code, but I don't think if the other code need it. For extending the Java `NetworkInterface`, I'm not convinced at all. If you want, before merge, I'll try to do it

dbwiddis:
I'm leaning toward extending it but still thinking about it.  I think there's value in letting users have access somehow to all the methods on the `NetworkInterface` object.

No need for you to do anything else unless you want. I'll merge and start working later this evening.

dbwiddis:
Can't extend the `NetworkInterface` because it has no constructor.  But I'm getting rid of most of the redundant `HWNetworkIF` instance variables in favor of keeping that object with a getter to allow access to all its methods.   So the instance variables will look like:
```
    private NetworkInterface networkInterface;
    private long bytesRecv;
    private long bytesSent;
    private long packetsRecv;
    private long packetsSent;
    private long speed;
```
I'll keep some convenience getter methods (name, display name) and parse some others (have a MAC address that parses the `byte[]` hardware address; return an array of IP addresses, etc.)

dbwiddis:
OK, I've refactored and changed names.  Check out the latest code and test it out.

Could use more detailed javadocs, needs CHANGELOG and updated sample output in README.

I'll get to the Windows and Mac ports later this weekend.

PR #163, opened 2016-05-04
Reviewing getFileStores() method

henryx:
Another two things:

 - I'm not sure if `store.name()` can be replaced if `path` equals "/"
 - Why informations are fetched from `/proc/self/mounts` and ` FileSystems.getDefault().getFileStores()` ?

### DIFF
--- a/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
+++ b/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
@@ -51,6 +51,7 @@ public class LinuxFileSystem extends AbstractFileSystem {
      *         volumes. May return disconnected volumes with
      *         {@link OSFileStore#getTotalSpace()} = 0.
      */
+    @Override
     public OSFileStore[] getFileStores() {
         // Parse /proc/self/mounts to map filesystem paths to types
         Map<String, String> fstype = new HashMap<>();
@@ -63,8 +64,8 @@ public class LinuxFileSystem extends AbstractFileSystem {
                 if (split.length < 6) {
                     continue;
                 }
-                fstype.put(split[1].replaceAll("\\\\040", " "), split[2]);
-            }
+                    fstype.put(split[1].replaceAll("\\\\040", " "), split[2]);
+                }
         } catch (IOException e) {
             LOG.error("Error reading /proc/self/mounts. Can't detect filetypes.");
         }
@@ -76,8 +77,8 @@ public class LinuxFileSystem extends AbstractFileSystem {
             // parentheses e.g., "/ (/dev/sda1)" and "/proc (proc)"
             String path = store.toString().replace(" (" + store.name() + ")", "");
             // Exclude special directories
-            if (path.startsWith("/proc") || path.startsWith("/sys") || path.startsWith("/run") || path.equals("/dev")
-                    || path.equals("/dev/pts"))
+            if (path.startsWith("/proc") || path.startsWith("/sys") || path.startsWith("/run")
+                    || path.startsWith("/dev"))
                 continue;
             String name = store.name();
             if (path.equals("/"))
--- a/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
+++ b/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
@@ -69,7 +69,7 @@ public class LinuxFileSystem extends AbstractFileSystem {
         // NOTE: FUSE's fuseblk is not evalued because used as file system representation of a FUSE block storage
         //"fuseblk" // FUSE block file system
     });
-
+    
     /**
      * Gets File System Information.
      *
@@ -116,7 +116,7 @@ public class LinuxFileSystem extends AbstractFileSystem {
                 type = fstype.get(path);
             }
             try {
-                fsList.add(new OSFileStore(name, description, type, store.getUsableSpace(), store.getTotalSpace()));
+                fsList.add(new OSFileStore(name, path, description, type, store.getUsableSpace(), store.getTotalSpace()));
             } catch (IOException e) {
                 // get*Space() may fail for ejected CD-ROM, etc.
                 LOG.trace("", e);
--- a/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
+++ b/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
@@ -11,6 +11,7 @@
  * Maintainers:
  * dblock[at]dblock[dot]org
  * widdis[at]gmail[dot]com
+ * enrico[dot]bianchi[at]gmail[dot]com
  *
  * Contributors:
  * https://github.com/dblock/oshi/graphs/contributors
@@ -46,7 +47,7 @@ public class LinuxFileSystem extends AbstractFileSystem {
     private static final Logger LOG = LoggerFactory.getLogger(LinuxFileSystem.class);
 
     // Linux defines a set of virtual file systems
-    private final String[] pseudofs = new String[]{
+    private final List<String> pseudofs = Arrays.asList(new String[]{
         "sysfs", // SysFS file system
         "proc", // Proc file system
         "devtmpfs", // Dev temporary file system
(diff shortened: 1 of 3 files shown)


src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -63,8 +64,8 @@
                 if (split.length < 6) {
                     continue;
                 }
-                fstype.put(split[1].replaceAll("\\\\040", " "), split[2]);
-            }
+                    fstype.put(split[1].replaceAll("\\\\040", " "), split[2]);
+                }
```
dbwiddis: Check your indentation/formatting settings. FYI I'm using Eclipse with Sun Java Conventions.

dbwiddis:
Not sure what the question is.  

The `/proc/self/mounts` call is simply a way to grab the filesystem type (third column) which is stored in a map for lookup, while the `.getFileStore()` call is just plain Java, which I trust to handle the Linux filesystem parsing better than I personally could...

src/main/java/oshi/software/os/linux/LinuxFileSystem.java
henryx: Mmm... It's strange, I don't remember that I've touched this part of code (maybe I've reformatted automatically and I dont' remember). However, I use Netbeans with default settings for code formatting

dbwiddis:
So, continuing this discussion:
- I see that `/proc/mounts` may be better than `/proc/self/mounts` but I think that's the best way to get file system type. Open to other ideas.
- The original "name" of the `/` filestore is `/dev/sda1`.  I changed that to be consistent with other OSHI names (e.g., Local Disk vs. Mount Point, etc.)

src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -76,9 +101,8 @@
             // FileStore toString starts with path, then a space, then name in
             // parentheses e.g., "/ (/dev/sda1)" and "/proc (proc)"
             String path = store.toString().replace(" (" + store.name() + ")", "");
-            // Exclude special directories
-            if (path.startsWith("/proc") || path.startsWith("/sys") || path.startsWith("/run")
-                    || path.startsWith("/dev"))
+            // Exclude pseudo file systems
+            if (Arrays.asList(this.pseudofs).contains(store.name()))
```
henryx: change of the approach is necessary because in some distributions (e.g. Fedora) external storage (such as USB key) are mounted in `/run/media/$user/mountpoint`

henryx:
Sore of `/` is not always `/dev/sda1`. For example, in a system that uses LVM, `/` can be `/dev/mapper/os-root`

src/main/java/oshi/software/os/linux/LinuxFileSystem.java
dbwiddis: I like this approach!

Can you put the `Arrays.asList()` call in the constant definition to begin with so we don't have to repeatedly call it inside this loop for every fs.

src/main/java/oshi/software/os/linux/LinuxFileSystem.java
henryx: Also, now reformatting has restored old style (stupid Netbeans) :expressionless:

henryx:
I'm not close this PR now because I want to try how Java report data with Btrfs or ZFS file systems and I want to view if is possible to add a mountpoint attribute (on Windows, I can have also file systems mounted a la Unix way)

henryx:
Another thing. Currently, the method reports network file systems (NFS, CIFS) too. I'm not sure if it is correct, because they refers to resource outside of the local machine. I think that they are removed from  list of the filestores, but I'm not sure if it is a good idea. What do you think?

dbwiddis:
For the software (OSFileStore) the intent is to include all possible places of storing data.  Local disk (possibly multiple partitions treated separately), ram disk, SD card, USB drive, network drive, etc.  From a Linux perspective, anything with a mount point.  From an OS X perspective, anything called a Volume.  From a Windows perspective, any drive with a named letter.  Specifically, the "usable space/total space" is what is of interest, so if this is a location you could move a file to, it should be listed.

For the hardware (HWDiskStore) the intent is only physical hard drives attached to the machine. Local disk (the physical disk, not partitions), possibly USB drives.  Exclude network drives, virtual drives, partitions, ram drives.

henryx:
Mmm... In this case, a filter for the tmpfs pseudo file system is wrong, because it can be used as ramdisk... I need think how correct the PR for this detail, because the tmpfs is used also for internal kernel usage (e.g. I have the `/dev/shm` mountpoint which is a tmpfs file system)

henryx:
Ok, I think that I've founded a good solution for checking if mountpoint's tmpfs refers a system path or not. Also, I've added a field and a constructor for setting the mountpoint. Now, I test what reports Windows if I define a partition as mountpoint and after, I declare the PR closed

src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -71,10 +114,19 @@
             // FileStore toString starts with path, then a space, then name in
             // parentheses e.g., "/ (/dev/sda1)" and "/proc (proc)"
             String path = store.toString().replace(" (" + store.name() + ")", "");
-            // Exclude special directories
-            if (path.startsWith("/proc") || path.startsWith("/sys") || path.startsWith("/run") || path.equals("/dev")
-                    || path.equals("/dev/pts"))
-                continue;
+
+            // Exclude pseudo file systems
+            if (this.pseudofs.contains(store.name())) {
```
dbwiddis: Two things stopping me from merging this is this conditional. One: I don't understand what it's doing or why.  It may just need some comments to explain what's going on, but the three nested if statements are also confusing (could be a single if with "or" conditions?

src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -37,20 +39,61 @@
...
+    });
+    
+    // System path mounted as tmpfs
+    private final List<String> tmpfsPaths = Arrays.asList(new String[]{
+        "/dev/shm",
+        "/run",
+        "/sys/fs/cgroup",
+    });
+
+    private Boolean listElementStarts(List<String> aList, String charSeq) {
```
dbwiddis: Need a javadoc explaining what this method does and when it returns true.   Also I think there's an error in `match.startsWith(match)`... shouldn't something there be `charSeq`?

src/main/java/oshi/software/os/linux/LinuxFileSystem.java
henryx: I've used three nested if statements (and a private method) because tmpfs
is used for system paths (e.g. in Fedora, /dev/shm) and ram disks (e.g. in
Debian, /tmp), so, if I want only ram disks I need to exclude these paths,
which differs between distributions (e.g. in Fedora I have /dev/shm
meanwhile in Debian 7 I have /run/shm). It's ugly, and I don't think if it
is the correct way, because the df(1) command shows all tmpfs (and
devtmpfs) file systems mounted in the os. What do you think? Is a good
choice to rollback this commit or is a good way to continue with this
direction?

2016-05-10 6:17 GMT+02:00 dbwiddis <notifications@github.com>:

src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -37,20 +39,61 @@
...
+        for (String match : aList) {
+            if (match.startsWith(match)){
```
dbwiddis: Is this line (`match.startsWith(match)`) correct? You aren't ever checking `charSeq`.

src/main/java/oshi/software/os/linux/LinuxFileSystem.java
dbwiddis: My default position is just to use the filesystems that Java returns.  I figure whoever wrote those did all the hard thought work required.   I generally would prefer to return more information to OSHI users and let them do the filtering on their end, rather than arbitrarily choose how to filter.

That said, I think adding the mount point field is a good idea.  I think OSHI already parses that internally to get the path, but we don't display that.

PR #165, opened 2016-05-08
Implement system-wide file descriptors

cholland1989:
Added methods to retrieve the current number of open file descriptors and the maximum number of open file descriptors on Linux and OS X. A Windows implementation was omitted, as there does not appear to be a direct equivalent of file descriptors. Windows [handles](https://blogs.technet.microsoft.com/markrussinovich/2009/09/29/pushing-the-limits-of-windows-handles/) are conceptually similar to file descriptors, but may also refer to a number of non-I/O related [objects](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724515(v=vs.85).aspx). Additionally, while there is a per-process maximum for open handles, there does not appear to be a system-wide maximum.

### DIFF
--- a/src/main/java/oshi/software/os/OperatingSystem.java
+++ b/src/main/java/oshi/software/os/OperatingSystem.java
@@ -48,4 +48,18 @@ public interface OperatingSystem extends OshiJsonObject {
      * @return Version.
      */
     OperatingSystemVersion getVersion();
+
+    /**
+     * Open file descriptors.
+     *
+     * @return long
+     */
+    long getOpenDescriptors();
+
+    /**
+     * Maximum file descriptors.
+     *
+     * @return long
+     */
+    long getMaxDescriptors();
 }
--- a/src/main/java/oshi/software/os/OperatingSystem.java
+++ b/src/main/java/oshi/software/os/OperatingSystem.java
@@ -48,18 +48,4 @@ public interface OperatingSystem extends OshiJsonObject {
      * @return Version.
      */
     OperatingSystemVersion getVersion();
-
-    /**
-     * Open file descriptors.
-     *
-     * @return long
-     */
-    long getOpenDescriptors();
-
-    /**
-     * Maximum file descriptors.
-     *
-     * @return long
-     */
-    long getMaxDescriptors();
 }
--- a/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
+++ b/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
@@ -46,6 +46,32 @@ public class LinuxOperatingSystem extends AbstractOperatingSystem {
         this.version = (osRelease != null) ? new LinuxOSVersionInfoEx(osRelease) : new LinuxOSVersionInfoEx();
     }
 
+    @Override
+    public long getOpenDescriptors() {
+        return getDescriptors(0);
+    }
+
+    @Override
+    public long getMaxDescriptors() {
+        return getDescriptors(2);
+    }
+
+    private long getDescriptors(int index) {
+        if (new File("/proc/sys/fs/file-nr").exists()) {
+            try {
+                List<String> osDescriptors = FileUtil.readFile("/proc/sys/fs/file-nr");
+                for (String line : osDescriptors) {
+                    line = line.replaceAll("[^0-9]+", " ");
+                    String [] splittedLine = line.split(" ");
+                    return Long.parseLong(splittedLine[index]);
+                }
+            } catch (Exception e) {
+                LOG.trace("", e);
+            }
+        }
+        return -1L;
+    }
+
     private String parseFamily() {
         if (this.family == null) {
             String etcOsRelease = getReleaseFilename();
--- a/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
+++ b/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
@@ -46,32 +46,6 @@ public class LinuxOperatingSystem extends AbstractOperatingSystem {
         this.version = (osRelease != null) ? new LinuxOSVersionInfoEx(osRelease) : new LinuxOSVersionInfoEx();
     }
 
-    @Override
-    public long getOpenDescriptors() {
-        return getDescriptors(0);
-    }
-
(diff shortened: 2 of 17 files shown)


src/main/java/oshi/software/os/OperatingSystem.java
```diff
@@ -48,4 +48,18 @@
      * @return Version.
      */
     OperatingSystemVersion getVersion();
+
+    /**
+     * Open file descriptors.
+     *
+     * @return long
+     */
+    long getOpenDescriptors();
```
dbwiddis: This method should probably be `getOpenFileDescriptors()` and I think it would be better placed in the `FileSystem` interface, rather than the `OperatingSystem`.

src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -46,6 +46,32 @@ public LinuxOperatingSystem() {
...
+        return getDescriptors(0);
+    }
+
+    @Override
+    public long getMaxDescriptors() {
+        return getDescriptors(2);
+    }
+
+    private long getDescriptors(int index) {
+        if (new File("/proc/sys/fs/file-nr").exists()) {
```
dbwiddis: I like doing the exists check rather than relying on catching exceptions.  I wonder if we should modify the `FileUtil.readFile()` method to include this check, and catch the exceptions in the util method so all the other implementations can simply call the one method...

src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -46,6 +46,32 @@ public LinuxOperatingSystem() {
...
+    public long getOpenDescriptors() {
+        return getDescriptors(0);
+    }
+
+    @Override
+    public long getMaxDescriptors() {
+        return getDescriptors(2);
+    }
+
+    private long getDescriptors(int index) {
```
dbwiddis: Needs a javadoc telling me what the index is.  (0, 1, 2?)

src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -46,6 +46,32 @@ public LinuxOperatingSystem() {
...
+        if (new File("/proc/sys/fs/file-nr").exists()) {
+            try {
+                List<String> osDescriptors = FileUtil.readFile("/proc/sys/fs/file-nr");
+                for (String line : osDescriptors) {
+                    line = line.replaceAll("[^0-9]+", " ");
+                    String [] splittedLine = line.split(" ");
+                    return Long.parseLong(splittedLine[index]);
```
dbwiddis: Need to do argument check for index between 0 and 2 before this point, where an array index out of bounds will be much more confusing to debug.

src/main/java/oshi/software/os/OperatingSystem.java
```diff
@@ -48,4 +48,18 @@
      * @return Version.
      */
     OperatingSystemVersion getVersion();
+
+    /**
+     * Open file descriptors.
```
dbwiddis: Need a much more verbose javadoc at the higher level.  Explain what these are and specifically note that they are not implemented on Windows (because of the way Windows treats file handles.)

src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -46,6 +46,32 @@ public LinuxOperatingSystem() {
...
+    public long getMaxDescriptors() {
+        return getDescriptors(2);
+    }
+
+    private long getDescriptors(int index) {
+        if (new File("/proc/sys/fs/file-nr").exists()) {
+            try {
+                List<String> osDescriptors = FileUtil.readFile("/proc/sys/fs/file-nr");
+                for (String line : osDescriptors) {
+                    line = line.replaceAll("[^0-9]+", " ");
```
dbwiddis: Curious why we can't just do \D+ here?  (Actually, "\\D+"..)

src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -46,6 +46,32 @@ public LinuxOperatingSystem() {
...
+                    String [] splittedLine = line.split(" ");
```
dbwiddis: On second thought, rather than the replaceall above, probably easiest to just use \\\\D+ as the split character here.

src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
dbwiddis: Need to catch NumberFormatException on `parseLong()`.  Unless we are 100% sure the regexp split guarantees a parseable long....

src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -26,4 +27,14 @@ public MacOperatingSystem() {
         this.family = System.getProperty("os.name");
         this.version = new MacOSVersionInfoEx();
     }
+
+    @Override
+    public long getOpenDescriptors() {
+        return SysctlUtil.sysctl("kern.num_files", -1);
```
dbwiddis: Need the default to be -1L to use the overloaded method that returns a long, otherwise you're returning an int and casting to long.

dbwiddis: Disregard above comment. The type that sysctl returns is int, so this is correct

src/test/java/oshi/SystemInfoTest.java
```diff
@@ -240,6 +240,8 @@ public void testOSVersion() {
         assertNotNull(version.getVersion());
         assertNotNull(version.getCodeName());
         assertNotNull(version.getBuildNumber());
+        assertTrue(os.getOpenDescriptors() >= -1L);
```
dbwiddis: This is probably worthy of testing >=0 with a platform check for mac or linux.  See test for `getSystemLoadAverage()` for example.

dbwiddis:
Looks good. I'm going to merge this and then move around the JSON a bit.

PR #194, opened 2016-06-12
Added volume information when getting Linux and Windows file system

henryx:
First of all: I've added a volume property in `OSFilestore` object, because it can permit an association of current file system evaluated (value for now is enabled only for Windows and Linux).
Second: I've rewritten the `LinuxFileSystem.getFileStores()` method, now uses `/proc/self/mounts/` for evaluation.
Third: I've reviewed check of pseudo file systems


dbwiddis:
I'm not sure what the intent of the new volume property is.  For Linux you use exactly the same text you're already using for the name property, and for Windows, the ProviderName WMI property selected (Not sure why it was leftover in the query string) is usually blank.  I'm not even sure what I would put in for the OS X property to match as I don't know what it even is.

The new version doesn't properly escape the spaces in the names, and I'm not sure why, but the current version has:
```
Time Machine Backups (Mount Point) [prl_fs] 5.0 TiB of 5.3 TiB free (93.7%) mounted at /media/psf/Time Machine Backups
```
And your new version has
```
Time\040Machine\040Backups (Mount Point) [prl_fs] 5.0 TiB of 5.3 TiB free (93.7%) mounted at /media/psf/Time Machine Backups
```
Also here's how my main hard drive used to look:
```
/ (Local Disk) [ext4] 51.1 GiB of 60.9 GiB free (84.0%) mounted at /
```
And now it's duplicated with rootfs (which the kernel uses to bootstrap but isn't really a filestore) and the less friendly `/dev/sda1` name (which I suppose I can live with).
```
 rootfs (Mount Point) [rootfs] 51.1 GiB of 60.9 GiB free (84.0%) mounted at /
 /dev/sda1 (Local Disk) [ext4] 51.1 GiB of 60.9 GiB free (84.0%) mounted at /
```

dbwiddis:
The \\040 stuff is in the name field (split[0]) so you need to replace there as well as split[1], the path.

henryx:
Rationale is simple. Assume this code (is a modification to the line 428 and line 431 in the `SystemInfoTest` class):
`
System.out.format(" %s (%s) [%s] %s of %s free (%.1f%%) is %s and is mounted at %s%n", fs.getName(),  fs.getDescription().isEmpty() ? "file system" : fs.getDescription(), fs.getType(), FormatUtil.formatBytes(usable), FormatUtil.formatBytes(fs.getTotalSpace()), 100d * usable / total, fs.getVolume(), fs.getMount());
`
Output is:
`
Disco rigido locale (C:) (Fixed drive) [NTFS] 361,4 GiB of 464,8 GiB free (77,7%) is \\?\Volume{c09b78b6-0000-0000-0000-500600000000}\ and is mounted at C:\
`
As you can see, `getName()` and `getVolume()` reports different informations, because in Windows `C:\` is the mountpoint of the volume. In the same way, on Linux `getName()` isn't the correct way to report the volume information. So, to uniform the output, I've added a `volume` property which reports this data

dbwiddis:
You are still using exactly the same name as volume on Linux so I'm missing the benefit of adding this volume information which appears to be Windows only, with the only new information being provided is essentially the Volume UUID (all mounted devices on Windows will have a volume name of `\\?\Volume{UUID goes here}\`

So how about instead of a new volume field, we have a UUID field?  This is accessible on Linux (follow symlinks from `/dev/disk/by-uuid`) and OS X (in the IOKit registry accessible a few different ways).  

I see you made the changes to escaping and / drive, but are you intentionally keeping rootfs?

dbwiddis:
The WMI `ProviderName` property is blank on my system for my drives.   I suggest using the [GetVolumeNameForVolumeMountPoint](https://msdn.microsoft.com/en-us/library/windows/desktop/aa364994(v=vs.85).aspx) function (it's already in JNA) to fetch the info.

dbwiddis:
Here's a patch for the OS X code to get UID from the bsd name (e.g., disk0s2):
```
Index: src/main/java/oshi/software/os/mac/MacFileSystem.java
===================================================================
--- src/main/java/oshi/software/os/mac/MacFileSystem.java	(revision 405)
+++ src/main/java/oshi/software/os/mac/MacFileSystem.java	(working copy)
@@ -33,10 +33,20 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.sun.jna.ptr.IntByReference;
+
+import oshi.jna.platform.mac.CoreFoundation;
+import oshi.jna.platform.mac.CoreFoundation.CFMutableDictionaryRef;
+import oshi.jna.platform.mac.CoreFoundation.CFStringRef;
+import oshi.jna.platform.mac.CoreFoundation.CFTypeRef;
+import oshi.jna.platform.mac.IOKit;
+import oshi.jna.platform.mac.IOKit.MachPort;
 import oshi.jna.platform.mac.SystemB;
 import oshi.jna.platform.mac.SystemB.Statfs;
 import oshi.software.common.AbstractFileSystem;
 import oshi.software.os.OSFileStore;
+import oshi.util.platform.mac.CfUtil;
+import oshi.util.platform.mac.IOKitUtil;
 import oshi.util.platform.mac.SysctlUtil;
 
 /**
@@ -56,6 +66,9 @@
     // Regexp matcher for /dev/disk1 etc.
     private static final Pattern localDisk = Pattern.compile("/dev/disk\\d");
 
+    // For matching IORegistry entry
+    private static final CFStringRef cfUUID = CFStringRef.toCFString("UUID");
+
     /**
      * Gets File System Information.
      * 
@@ -108,6 +121,39 @@
                     if (fstype.containsKey(path)) {
                         type = fstype.get(path);
                     }
+                    // Use fs.name() to find registry entry and get UUID
+                    String uuid = ""; // Default
+                    String bsdName = fs.name().replace("/dev/disk", "disk");
+                    MachPort masterPort = IOKitUtil.getMasterPort();
+                    if (bsdName.startsWith("disk") && masterPort != null) {
+                        CFMutableDictionaryRef matchingDict = IOKit.INSTANCE.IOBSDNameMatching(masterPort.getValue(), 0,
+                                bsdName);
+                        // search for all IOservices that match the bsd name
+                        IntByReference fsIter = new IntByReference();
+                        IOKitUtil.getMatchingServices(matchingDict, fsIter);
+                        // getMatchingServices releases matchingDict
+                        // Should only match one logical drive
+                        int fsEntry = IOKit.INSTANCE.IOIteratorNext(fsIter.getValue());
+                        if (fsEntry != 0) {
+                            // Should be an IOMedia object
+                            if (!IOKit.INSTANCE.IOObjectConformsTo(fsEntry, "IOMedia")) {
+                                LOG.error("Unable to find IOMedia device for ", fs.name());
+                                break;
+                            // Now get the UUID
+                            CFTypeRef uuidAsCFString = IOKit.INSTANCE.IORegistryEntryCreateCFProperty(fsEntry, cfUUID,
+                                    CoreFoundation.INSTANCE.CFAllocatorGetDefault(), 0);
+                            if (uuidAsCFString != null) {
+                                uuid = CfUtil.cfPointerToString(uuidAsCFString.getPointer());
+                            IOKit.INSTANCE.IOObjectRelease(fsEntry);
+                        }
+                        // TODO: For debug/info.
+                        // Remove when uuid added to OSFileStore constructor
+                        System.out.println("UUID for " + fs.name() + " is " + uuid);
+                        IOKit.INSTANCE.IOObjectRelease(fsIter.getValue());
+                    }
+
                 } catch (IOException e) {
                     LOG.trace("", e);
                     continue;
```
Sample output of the above:
```
UUID for /dev/disk4s2 is 0E23889D-37CE-434D-B80C-D612A875972B
UUID for /dev/disk0s2 is 9510983A-04F4-4189-A4A9-7336A8579027
UUID for /dev/disk3s2 is 
UUID for /dev/disk2s2 is 23CC1A7E-559F-460D-B0B3-7D6B2E977C53
````

**Comment by henryx** [comment]
Motivation of I've added the `volume` property is this. As I've understand, currently `getName()` returns a generic label that identifies the `OSFIleStore`. This is because, I've three types of data in it:

 - If I'm on Windows, `getName()` returns a string that contains a `String.format("%s (%s)", strName, strMount)`, e.g. `Disco rigido locale (C:)`;
 - If I'm on Linux, `getnName()` returns:
   - The volume if it is different to the root file system (e.g. `/dev/sda5`);
   - The string `/` if I examine the root file system.

So, while I don't have the same result, I cannot use this value for manage purpose. Also, at this moment I don't have the `volume` information, because it isn't reported (for Linux, in some cases is in `getName()`, but for the motivation as I've described before, isn't possible to use; for Windows, isn't reported because is extracted only the volume label - `Disco rigido locale` - and the mountpoint - C: - ). With the `volume` property I have a place for reporting this information.

Return null of the WMI `ProviderName` property is strange, because I've tested on Windows 7 and 10 and it returns the correct value. Also, it is evaluated only for non local disks (the data returned from `getWmiVolumes()` method are added only if the `getLocalVolumes()` method doesn't return mounted volumes) and, on the network file systems, returns the remote share (e.g. \\server\path). I'll check this

For the UUID, is a good idea, but I prefer to use a Udev method for getting it (and I prefer to use a dedicated property for the reasons described before). I haven't merged the patch for OSX because I have problems to do it (the patch command and the git subcommands returns errors)

**Comment by dbwiddis** [comment]
Let's do both volumeName and UUID.  For OS X I will use fa.nane() for the volume to get the /dev/disk0s2 path. I can merge the patch myself after you are done.  

I do still think the GetVolumeNameForVolumePath may be preferable to WMI.

**Comment by henryx** [comment]
Ok, I've tested the WMI call and result is quite strange. Suppose this scenario:

 - C:\ is a fixed drive;
 - D:\ is a CD-ROM drive;
 - H:\ is a network drive and point to \\server\share.

If I use the WMI `ProviderName`, I have this result:

 - C:\ returns `unknown`;
 - D:\ returns `unknown`;
 - H:\ returns `\\server\share`.

If i use `GetVolumeNameForVolumeMountPoint()`, I have this result:

 - C:\ returns `\\?\Volume{VolumeID}`;
 - D:\ returns `\\?\Volume{VolumeID}`;
 - H:\ doesn't return any value (empty string).

A possible solution is to get the `DriveType` and check if is different to 3 (local drives) for using `GetVolumeNameForVolumeMountPoint()` or `ProviderName`

**Comment by henryx** [comment]
Small note: the WMI `ProviderName` reports only the network path to the logical drive

**Comment by dbwiddis** [comment]
Merged this, but when I make the OS X changes I think I'll remove the `rootfs` listing:
```
 / (Mount Point) [rootfs] 50.6 GiB of 60.9 GiB free (83.2%) is rootfs and is mounted at /
 / (Local Disk) [ext4] 50.6 GiB of 60.9 GiB free (83.2%) is /dev/sda1 and is mounted at /
```

PR #195, opened 2016-06-16
Fix reading multiple (external) displays on Macs

dpagano:
Looping over the different children of IODisplayConnect nodes did not work. Only the first one was returned, because the value given to IOIteratorNext in the while loop was wrong.


dbwiddis:
You know, I noticed that error when I copied the code for the USB Devices and it didn't work, and I fixed it there.  Unfortunately I didn't fix it before using it in the `MacDisks` code!

Since you're fixing this can I ask you to do two more things?

- Fix the identical error (copy/paste) in `MacDisks.java`.
- Add the necessary IOObjectRelease() statement immediately before the iteration.  The `MacUSBDevice.java` file is done correctly for reference.

dbwiddis:
Actually, the MacDisks fix is a little more complex.   If you can just do the IOObjectRelease I'll merge this and I'll handle the MacDisks one.

dpagano:
Ok, no problem

PR #212, opened 2016-06-25
Clean up build

hazendaz:
@dbwiddis I didn't fix the scm issues here just ran sort to get the spacing proper per maven standards, updated slf4j, updated some plugins, and fixed the relative path issue.  I think at this point you have two tasks to complete.

- Create separate 'oshi-assembly' and move things there for that.
- Add scm to the underlying modules (all of them and they are identical)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,5 @@
 # Output
-/target/
-/bin/
-/.project
-/.settings/
-/.classpath
-/oshi-*/target/
-/oshi-*/bin/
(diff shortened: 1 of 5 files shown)


hazendaz:
I"m rebasing...be back shortly.

hazendaz:
btw- I get test failures on my machine.  I'll get details here shortly.

hazendaz:
ok - build is running now.

PR #245, opened 2016-08-07
[javadoc] Fixed param name in javadoc to match param used in method call

hazendaz:


oshi-core/src/main/java/oshi/jna/platform/windows/IPHlpAPI.java
```diff
@@ -155,7 +155,7 @@
      * The GetIfEntry function retrieves information for the specified interface
      * on the local computer
      * 
-     * @param pIfRow
+     * @param pIfRow2
```
dbwiddis: Actually this whole javadoc is wrong. I copied from the old one and edited the function/method but forgot to update the doc from MSDN. I'll do that on master....

PR #259, opened 2016-09-17
Fixed test in Windows environment

henryx:
Rationale: this test doesn't work on Windows because `echo` is an internal command of cmd.exe

### DIFF
--- a/oshi-core/src/test/java/oshi/util/ExecutingCommandTest.java
+++ b/oshi-core/src/test/java/oshi/util/ExecutingCommandTest.java
@@ -24,13 +24,15 @@ import static org.junit.Assert.assertTrue;
 import java.util.List;
 
 import org.junit.Test;
+import oshi.PlatformEnum;
+import oshi.SystemInfo;
 
 /**
  * Test command line and returning the result of execution.
  */
 public class ExecutingCommandTest {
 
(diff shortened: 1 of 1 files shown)


dbwiddis:
Looks good. Please add a changelog entry (and then you should be able to merge it!)

oshi-core/src/test/java/oshi/util/ExecutingCommandTest.java
```diff
@@ -30,7 +30,7 @@
  */
 public class ExecutingCommandTest {
 
-    private static final String ECHO = "echo Test";
+    private static final String ECHO = System.getProperty("os.name").startsWith("Windows") ? "cmd.exe /C echo Test" : "echo Test";
```
dbwiddis: You might consider using `SystemInfo.getCurrentPlatformEnum().equals(PlatformEnum.WINDOWS)` as the conditional here.  Ultimately it will do the same thing, so it's mostly a matter of style whether to use an enum that's already checked the `os.name` property.

henryx:
I've forgot the changelog :-(

dbwiddis:
It's ok. I decided it was probably not needed since I doubt I'll cut a release for this. :-)

Issue #260, opened 2016-10-03
macOS branding


dbwiddis:
With its latest OS release last week, OS X has been rebranded as macOS.

OSHI uses `System.getProperty("os.name")` which still returns "Mac OS X".  This is consistent with the commandline `sw_vers`:

```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.12
BuildVersion:   16A323
```

But `system_profiler` returns a different story:

```
$ system_profiler SPSoftwareDataType
Software:

    System Software Overview:

      System Version: macOS 10.12 (16A323)
```

Should OSHI do a test on the OS version and edit the name for versions 10.12 and higher?

PR #262, opened 2016-10-07
[OSHI] Add bytesRead and bytesWritten to OSProcess

plamenko:
Summary:
This change adds bytesRead and bytesWritten fields to OSProcess.
An implementation is provided for MacProcess which uses `proc_pid_rusage(pid, RUSAGE_INFO_V2, ...)`.
Note that RUSAGE_INFO_V2 is only available on MacOSX 10.9+.

Test Plan:
Successfully obtained process io stats on MacOSX 10.10 and 10.11.
Verified that `mvn test` succeeds.

### DIFF
--- a/oshi-core/src/main/java/oshi/jna/platform/mac/SystemB.java
+++ b/oshi-core/src/main/java/oshi/jna/platform/mac/SystemB.java
@@ -55,6 +55,9 @@ public interface SystemB extends com.sun.jna.platform.mac.SystemB {
     int MNT_NOWAIT = 0x0010;
     int MNT_DWAIT = 0x0100;
 
+    // resource.h
+    int RUSAGE_INFO_V2 = 2;
+
     class ProcTaskAllInfo extends Structure {
         public ProcBsdInfo pbsd;
         public ProcTaskInfo ptinfo;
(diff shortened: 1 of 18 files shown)


oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -107,12 +108,18 @@ public OSProcess getProcess(int pid) {
                 }
             }
         }
+        long bytesRead = 0, bytesWritten = 0;
+        RUsageInfoV2 rUsageInfoV2 = new RUsageInfoV2();
+        if (0 == SystemB.INSTANCE.proc_pid_rusage(pid, SystemB.RUSAGE_INFO_V2, rUsageInfoV2)) {
```
dbwiddis: Should we check for OS version >= 10.9 here?  What happens if this is run on an earlier version?  I suspect an exception (if the method doesn't exist) or possibly it populates a smaller structure and would give memory access errors/segfault on the next two lines.

We could use a utility method to find OS X version (need to check for 12+ for #260) so might as well dual-purpose that.

oshi-core/src/main/java/oshi/jna/platform/mac/SystemB.java
```diff
@@ -388,6 +422,19 @@ public IFmsgHdr2(Pointer p) {
...
+     *
+     * @param pid
+     *            the process identifier
+     * @param flavor
+     *            the type of information requested
+     * @param buffer
+     *            holds results
+     * @returns 0 on success; or -1 on failure, with errno set to indicate the specific error.
+     */
+    int proc_pid_rusage(int pid, int flavor, Structure buffer);
```
dbwiddis: Should use type `RUsageInfoV2` here.  Or, if preferred, `RUsageInfoT` (define to extend `Structure`) which the V2 class can subsequently extend.

oshi-core/src/main/java/oshi/jna/platform/mac/SystemB.java
```diff
@@ -127,6 +130,37 @@
         }
     }
 
+    class RUsageInfoV2 extends Structure {
+        public byte ri_uuid[] = new byte[16];
```
dbwiddis: Should be `public byte[] ri_uuid =` ...

dbwiddis:
Forgot to note in the code review ... you'll also need to modify the oshi-json code (and config file) to output the additional fields.

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
plamenko: I added a version check.

plamenko:
Yeah, I'll try to make a pull request for Linux and possibly windows.

plamenko:
I amended the commit and you can see the new changes, but for some reason GitHub didn't comment here that the pull request got updated and it still shows `plamenko committed 23 days ago` instead of showing the time of the last update.

plamenko:
Okay I added proc io stats for Linux too and verified it works:

My test app that consumes io/cpu/mem resources:
```
mem size: 401
bytes read: 42991616
bytes written: 42991616
^C
real  0m41.241s
user  0m2.007s
sys 0m10.478s
```

OSHI with these two commits:
```
pid: 1619028, MEM Resident: 411 MB, MEM Size: 829 MB, CPU Real: 41.299 s, CPU User: 2.0 s, CPU Sys: 10.43 s, CPU Total: 12.43 s, IO Read: 42 MB, IO Write: 41 MB, IO Total: 83 MB
```


I also found out a bug in reporting RSS on Linux as the number OSHI reported were suspicious. Turns out `stats` gives RSS in pages, not in bytes, so I fixed that too:

http://man7.org/linux/man-pages/man5/proc.5.html
```
(23) vsize  %lu
          Virtual memory size in bytes.

(24) rss  %ld
          Resident Set Size: number of pages the process has
          in real memory.  This is just the pages which count
          toward text, data, or stack space.  This does not
          include pages which have not been demand-loaded in,
          or which are swapped out.
```

PR #286, opened 2016-12-22
Reduce required calculations for LinuxProcess initialization

### LINKED ISSUE #265: LinuxProcess should use its own process to calculate Jiffies per second
In order to calculate Jiffies per second, [LinuxProcess.java](https://github.com/oshi/oshi/blob/e51f8e84f20bc4b132aa02601a4760b20bd4a198/oshi-core/src/main/java/oshi/software/os/linux/LinuxProcess.java) in its static initialization lists all the processes, finds the youngest one and divide its Jiffies since boot with its time since boot.

However, listing all the processes and finding the youngest one seems excessive. It seems unlikely that the process that uses OSHI is during its initialization much younger than what would the youngest process at that time be, and therefore OSHI could just use its own process to calculate this.

A concern might be if the hosting JVM started very soon after the boot, and for example a [Nailgun](https://github.com/martylamb/nailgun) server is used to run the programs. However, meeting both of these conditions seems unlikely.

dbwiddis:
Fixes #265

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxProcess.java
@@ -19,6 +19,7 @@
 package oshi.software.os.linux;
 
 import java.io.File;
+import java.util.Arrays;
 import java.util.List;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


oshi-core/src/main/java/oshi/software/os/linux/LinuxProcess.java
```diff
@@ -103,25 +104,36 @@ public LinuxProcess(String name, String path, char state, int processID, int par
...
 
-        // Iterate /proc/[pid]/stat checking the creation time (field 22,
-        // jiffies since boot) for the largest value
+        // To avoid having to check all processes we can just pick the highest
+        // PID. This will either be the youngest one or at least big enough.
+
+        // Get all the pid files (guaranteed to be digit-only filenames)
         File[] pids = ProcUtil.getPidFiles();
+        // Sort descending "numerically"
+        Arrays.sort(pids, (f1, f2) -> Integer.valueOf(f2.getName()).compareTo(Integer.valueOf(f1.getName())));
```
henryx: This is Java8, right? I'm not sure if is a good idea, because today Java7 is widespread yet

plamenko: Much of OSHI is Java8 anyways. Making this Java7 compatible won't make a difference.

dbwiddis: I initially wrote that without lambdas, and Sonar yelled at me.  Ok, more a gentle nudge. ;)

But we switched from Java7 to Java8 with version 2.6, notably in the need to use the improved JSR310 Date/Time functions.  Given the choice between requiring Java 8 or adding a dependency, I chose the former.  That also opened up a lot more flexibility in the Map methods that I've leveraged almost everywhere.

dbwiddis: Now it would be fair to note that this method is inefficient; it requires the String to integer coversion for every comparision, rather than converting the String array to integer, sorting, and then converting back only the needed integers.  I'll take the one-liner readability though.

Issue #269, opened 2016-11-01
Feature request: Routing information & DNS settings


Ryan421:
Hello,

Is it possible to add more network information features like routing (at least default gateways), dns settings, hostname into OSHI?

dbwiddis:
I think OSHI already has most of the easily accessible information leveraging Java's `NetworkInterface` class.  You can get the hostname with `InetAddress.getHostName()`.  

Routing information is probably available somewhere from the routing tables but that's a whole lot of command-line parsing for not very much info, and a lot more complexity.  Open to a PR if you've got an idea on how you'd like to add it.

dbwiddis:
I see you've gotten started on this in #294.   FYI:

FreeBSD:
- Hostname, Domain name.  There are commands for these but they simply call the `libc` functions `gethostname()` and `getaddrinfo()` which would be preferred.  (Should try to use this approach for Linux as well.)  
- Gateways: use [route](https://www.freebsd.org/cgi/man.cgi?query=route) and search for the line in the output:
```
$ route -4 get default
... <snip> ...
    gateway: 10.211.55.1
... <snip>

$ route -6 get default
... <snip> ...
    gateway: fe80::21c:42ff:fe00:18%em0
... <snip>
```
- DNS Servers in `resolv.conf` as for Linux.

Solaris:
- Hostname, Domain name.  Same `gethostname()` and `getaddrinfo()`-based commands as FreeBSD and Linux.
- Gateways:  `route get -inet default` produces the same output on Solaris as `route -4 get default` on FreeBSD, and `route get -inet6 default` produces the same as `route -6 get default`.
- DNS Servers in `resolv.conf` as for Linux.

dbwiddis:
OS X:
- Hostname, Domain name:  same approach as described above for FreeBSD, Solaris, and suggested change for Linux.  (Yay standard C library.)
- Gateway:  `route -n get default` gets the same as above for FreeBSD/Solaris ipv4 but doesn't offer a way to get ipv6.   Instead use `netstat -nr` and look for the `G` flag  or the `default` destination:
```
<danielwiddis> [01-03 14:16] ~/git/oshi $ netstat -nr
Routing tables

Internet:
Destination        Gateway            Flags        Refs      Use   Netif Expire
default            192.168.1.1        UGSc           51        0     en5
default            192.168.1.1        UGScI          12        0     en1
... <snip> ...
Internet6:
Destination                             Gateway                         Flags         Netif Expire
[...]

PR #294, opened 2017-01-03
Host/domain name, dns servers and default gateways information for Win/Linux

chikei:
Hi,

here is interface for OS networking parameters and implementation for Linux/Win (sorry I don't have machine for other OS :().
Now it contains host name/dns domain name/dns servers/default gateway of IPv4/6.

Domain suffix search list is also considered but I think I'll wait JNA release `VT_ARRAY` support for this.

### DIFF
--- a/oshi-core/src/main/java/oshi/software/common/AbstractNetworkParams.java
+++ b/oshi-core/src/main/java/oshi/software/common/AbstractNetworkParams.java
@@ -1,5 +1,24 @@
+/**
+ * Oshi (https://github.com/dblock/oshi)
+ *
+ * Copyright (c) 2010 - 2017 The Oshi Project Team
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Maintainers:
+ * dblock[at]dblock[dot]org
+ * widdis[at]gmail[dot]com
+ * enrico.bianchi[at]gmail[dot]com
+ *
+ * Contributors:
+ * https://github.com/dblock/oshi/graphs/contributors
+ */
 package oshi.software.common;
 
+import com.sun.jna.ptr.PointerByReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -8,6 +27,7 @@ import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.List;
 
+import oshi.jna.platform.unix.LibC;
 import oshi.software.os.NetworkParams;
 import oshi.util.FileUtil;
 
@@ -20,6 +40,41 @@ public abstract class AbstractNetworkParams implements NetworkParams {
 
     private static final Logger LOG = LoggerFactory.getLogger(AbstractNetworkParams.class);
     private static final String NAMESERVER = "nameserver";
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getDomainName() {
+        LibC.Addrinfo hint = new LibC.Addrinfo();
+        hint.ai_flags = LibC.AI_CANONNAME;
+        String hostname = null;
+        try {
+            hostname = InetAddress.getLocalHost().getHostName();
+        } catch (UnknownHostException e) {
+            LOG.error("Unknown host exception when getting address of local host: " + e);
+            return "";
+        }
+        PointerByReference ptr = new PointerByReference();
+        doGetaddrinfo(hint, hostname, ptr);
+        LibC.Addrinfo info = new LibC.Addrinfo(ptr.getValue());
+        String canonname = info.ai_canonname;
+        doFreeaddrinfo(ptr);
+        int dot = canonname.indexOf('.');
+        if (dot == -1) {
+            return "";
+        } else {
+            return canonname.substring(dot + 1);
+        }
+    }
+
+    protected void doFreeaddrinfo(PointerByReference ptr) {
+        LibC.INSTANCE.freeaddrinfo(ptr.getValue());
+    }
+
+    protected void doGetaddrinfo(LibC.Addrinfo hint, String hostname, PointerByReference ptr) {
+        LibC.INSTANCE.getaddrinfo(hostname, null, hint, ptr);
+    }
 
     /**
      * {@inheritDoc}
@@ -62,10 +117,10 @@ public abstract class AbstractNetworkParams implements NetworkParams {
         return servers.toArray(new String[servers.size()]);
     }
 
-    static protected String searchGateway(List<String> lines){
-        for(String line: lines){
+    static protected String searchGateway(List<String> lines) {
+        for (String line : lines) {
             String leftTrimmed = line.replaceFirst("^[ \t]+", "");
-            if(leftTrimmed.startsWith("gateway:")){
+            if (leftTrimmed.startsWith("gateway:")) {
                 return leftTrimmed.split("[ \t]", 2)[1];
             }
         }
--- a/oshi-core/src/main/java/oshi/software/common/AbstractNetworkParams.java
+++ b/oshi-core/src/main/java/oshi/software/common/AbstractNetworkParams.java
@@ -61,4 +61,14 @@ public abstract class AbstractNetworkParams implements NetworkParams {
         }
         return servers.toArray(new String[servers.size()]);
     }
+
+    static protected String searchGateway(List<String> lines){
+        for(String line: lines){
+            String leftTrimmed = line.replaceFirst("^[ \t]+", "");
+            if(leftTrimmed.startsWith("gateway:")){
+                return leftTrimmed.split("[ \t]", 2)[1];
+            }
+        }
+        return "";
+    }
 }
--- a/oshi-core/src/main/java/oshi/software/common/AbstractNetworkParams.java
+++ b/oshi-core/src/main/java/oshi/software/common/AbstractNetworkParams.java
@@ -1,0 +1,64 @@
+package oshi.software.common;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.ArrayList;
+import java.util.List;
+
+import oshi.software.os.NetworkParams;
+import oshi.util.FileUtil;
+
+/**
+ * Common NetworkParams implementation.
+ */
+public abstract class AbstractNetworkParams implements NetworkParams {
+
+    private static final long serialVersionUID = 1L;
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractNetworkParams.class);
+    private static final String NAMESERVER = "nameserver";
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getHostName() {
+        try {
+            String hn = InetAddress.getLocalHost().getHostName();
+            int dot = hn.indexOf('.');
+            if (dot == -1) {
+                return hn;
+            } else {
+                return hn.substring(0, dot);
+            }
+        } catch (UnknownHostException e) {
+            LOG.error("Unknown host exception when getting address of local host: " + e);
+            return "";
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String[] getDnsServers() {
+        List<String> resolv = FileUtil.readFile("/etc/resolv.conf");
+        String key = NAMESERVER;
+        int maxNameServer = 3;
+        List<String> servers = new ArrayList<>();
+        for (int i = 0; i < resolv.size() && servers.size() < maxNameServer; i++) {
+            String line = resolv.get(i);
+            if (line.startsWith(key)) {
+                String value = line.substring(key.length()).replaceFirst("^[ \t]+", "");
+                if (value.length() != 0 && value.charAt(0) != '#' && value.charAt(0) != ';') {
+                    String val = value.split("[ \t#;]", 2)[0];
+                    servers.add(val);
+                }
+            }
+        }
+        return servers.toArray(new String[servers.size()]);
+    }
+}
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxNetworkParams.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxNetworkParams.java
@@ -18,15 +18,13 @@
  */
 package oshi.software.os.linux;
 
-import java.util.ArrayList;
 import java.util.List;
 
-import oshi.software.os.NetworkParams;
+import oshi.software.common.AbstractNetworkParams;
 import oshi.util.ExecutingCommand;
-import oshi.util.FileUtil;
 import oshi.util.ParseUtil;
 
-public class LinuxNetworkParams implements NetworkParams {
+public class LinuxNetworkParams extends AbstractNetworkParams {
 
     private static final long serialVersionUID = 1L;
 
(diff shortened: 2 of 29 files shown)


dbwiddis:
Looks like you missed committing one of the changed files:

> [ERROR] COMPILATION ERROR : 
> [INFO] -------------------------------------------------------------
> [ERROR] /home/travis/build/oshi/oshi/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java:[49,8] oshi.software.os.linux.LinuxOperatingSystem is not abstract and does not override abstract method getNetworkParams() in oshi.software.os.OperatingSystem

oshi-core/src/main/java/oshi/software/os/linux/LinuxNetworkParams.java
```diff
@@ -0,0 +1,139 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public String[] getDnsServers() {
+        List<String> resolv = FileUtil.readFile("/etc/resolv.conf");
+        String key = "nameserver";
+        int maxNameServer = 3;
+        List<String> servers = new ArrayList<>();
+        for(String line: resolv){
+            if(!line.startsWith(key)) continue;
```
dbwiddis: Style: avoid the `continue` and `break` in the same loop.  Just test for a true `startsWith()` here and put the rest of the method inside the `if()`.

oshi-core/src/main/java/oshi/software/os/linux/LinuxNetworkParams.java
```diff
@@ -0,0 +1,139 @@
...
+
+            String value = line.substring(key.length())
+                .replaceFirst("^[ \t]+", "");
+            if(value.length() == 0 ||
+                value.charAt(0) == '#' || value.charAt(0) == ';') continue;
+            String val = value.split("[ \t#;]", 2)[0];
+            servers.add(val);
+            if(servers.size() >= maxNameServer) break;
```
dbwiddis: Use brackets and separate lines.  If you're using Eclipse IDE, use its code style for "Sun Java Conventions".  Also change your "cleanup" style to remove unused imports that I see the codacy-bot has flagged.

oshi-core/src/main/java/oshi/software/os/linux/LinuxNetworkParams.java
```diff
@@ -0,0 +1,139 @@
...
+
+        String gateway = "";
+        int minMetric = Integer.MAX_VALUE;
+
+        for(int i = 2; i < routes.size(); i++){
+            String[] fields = routes.get(i).split("\\s+");
+            if(fields.length > 3 && fields[0].equals("::/0")) {
+                try{
+                    boolean isGateway = fields[2].indexOf('G') != -1;
+                    int metric = Integer.parseInt(fields[3]);
```
dbwiddis: Use `ParseUtil.parseIntOrDefault()` which already handles the exception and will simplify your code.

oshi-core/src/main/java/oshi/software/os/windows/WindowsNetworkParams.java
```diff
@@ -0,0 +1,148 @@
...
+
+        return list.toArray(new String[0]);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getIpv4DefaultGateway() {
+        String dest = "0.0.0.0/0";
```
dbwiddis: Put this (and other similar throughout the PR) string as a constant at the top of the file.

oshi-core/src/main/java/oshi/software/os/windows/WindowsNetworkParams.java
```diff
@@ -0,0 +1,148 @@
...
+        return getNextHop(dest);
+    }
+
+    private String getNextHop(String dest) {
+        Map<String, List<Object>> vals = WmiUtil.selectObjectsFrom("ROOT\\StandardCimv2", "MSFT_NetRoute",
+            "NextHop,RouteMetric", "WHERE DestinationPrefix=\"" + dest + "\"", GATEWAY_TYPES);
+        List<Object> metrics = vals.get("RouteMetric");
+        if (vals.get("RouteMetric").size() == 0) {
```
dbwiddis: Style/readability: use `!isEmpty()` and go straight to what's currently under `else`.  Then just put the `return "";` after the if.

oshi-core/src/main/java/oshi/software/os/linux/LinuxNetworkParams.java
```diff
@@ -0,0 +1,139 @@
...
+    private static final long serialVersionUID = 1L;
+
+    private static final Logger LOG = LoggerFactory.getLogger(LinuxNetworkParams.class);
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getHostName() {
+        String hn = FileUtil.getStringFromFile("/proc/sys/kernel/hostname");
```
dbwiddis: Prefer native calls to reading files or parsing command lines.  In this case, there is a `gethostname()` function in `libc`.

oshi-core/src/main/java/oshi/software/os/linux/LinuxNetworkParams.java
```diff
@@ -0,0 +1,139 @@
...
+            return hn.substring(0, dot);
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getDomainName() {
+        return ExecutingCommand.getFirstAnswer("dnsdomainname");
```
dbwiddis: Prefer native calls to reading files or parsing command lines. In this case, `dnsdomainname` is the result of calling `getaddrinfo()` on the result of `gethostname()`.

dbwiddis:
See comments on #269 for code which works on other OS's.

dbwiddis:
Some more thoughts:

Why aren't we using built-in Java methods that already do all the cross-platform heavy lifting (probably using the same system calls I already suggested we use)?  For example:
- Host Name: just use `InetAddress.getHostName()`
- Domain Name: just use `InetAddress.getCanonicalHostName()`

(As an aside, my initial objection to adding these to OSHI was because they were already available to users in core java.  It makes sense to add them as part of an overall implementation including other things, though.)

For the gateway, `netstat -nr` is cross-platform (OSX, Linux, Unix, and even Windows, although the Windows output format is different) so you could implement that in a parent class, and override on Windows only with the WMI implementation, although the `route` command alternatives also work.

There is a non-API cross-platform Sun way of getting the nameservers:
```
        List<String> nameservers = (List<String>) sun.net.dns.ResolverConfiguration.open().nameservers();
```
But we shouldn't use non-API methods.  Still, digging into the source code of their implementations may be instructive although the `resolv.conf` approach for *nix, WMI for WIndows, and `scutil --dns` for OSX are probably sufficient.

hazendaz:
That sun package would not likely work with ibm jdk

Get Outlook for Android<https://aka.ms/ghei36>

oshi-json/src/main/java/oshi/json/software/os/impl/NetworkPramsImpl.java
```diff
@@ -0,0 +1,111 @@
...
+import javax.json.JsonObject;
+import javax.json.JsonObjectBuilder;
+
+import oshi.json.json.AbstractOshiJsonObject;
+import oshi.json.json.NullAwareJsonObjectBuilder;
+import oshi.json.software.os.NetworkParams;
+import oshi.json.software.os.OSFileStore;
+import oshi.json.util.PropertiesUtil;
+
+public class NetworkPramsImpl extends AbstractOshiJsonObject implements NetworkParams{
```
dbwiddis: Class name typo.

oshi-json/src/main/java/oshi/json/software/os/impl/NetworkPramsImpl.java
```diff
@@ -0,0 +1,111 @@
...
+    @Override
+    public String getIpv6DefaultGateway() { return this.networkParams.getIpv6DefaultGateway(); }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public JsonObject toJSON(Properties properties) {
+        JsonObjectBuilder json = NullAwareJsonObjectBuilder.wrap(this.jsonFactory.createObjectBuilder());
+        if (PropertiesUtil.getBoolean(properties, "operatingSystem.networkParams.hostName")) {
```
dbwiddis: Need to add all these property strings to the sample oshi.json.properties file in src/test/resources.

dbwiddis:
Agreed we should not use the sun package.  I was only suggesting it from a 'look at the source' standpoint.  Which may be overkill.   I did find a [`dnsjava` project](http://www.dnsjava.org/) that does a lot of DNS-related stuff.  But digging into the [resolving code](https://sourceforge.net/p/dnsjava/code/HEAD/tree/trunk/org/xbill/DNS/ResolverConfig.java) they don't have a significant improvement on command line parsing. They check `/etc/resolv.conf`, check the `dns.server` and `dns.search` system properties, or parse the output of `ipconfig` and platform-specific alternatives.

oshi-core/src/main/java/oshi/software/os/windows/WindowsNetworkParams.java
```diff
@@ -0,0 +1,148 @@
...
+            String addr = new String(dns.IpAddress.String);
+            int nullPos = addr.indexOf(0);
+            if(nullPos != -1) {
+                addr = addr.substring(0, nullPos);
+            }
+            list.add(addr);
+            dns = dns.Next;
+        }
+
+        return list.toArray(new String[0]);
```
dbwiddis: This should be `return list.toArray(new String[list.size()]);` to avoid creating/discarding the empty String array.

chikei:
I was also hesitated on add host/dns domain name at beginning, it was then added because it's kind of weird having a network parameter but couldn't provide these two things. But like you said, maybe I should just make them as a `InetAddress.getXXXName` wrapper.

As for the resolving code, actually it was following how the sun `ResolverConfigurationImpl` did, so it should also work for freebsd and solaris (I guess it works for mac too ?).

dbwiddis:
Glad to hear you dug into the Sun class.  As for the host/dns names, I'd default to the`InetAddress` wrappers. The Canonical name, especially, gives you a lot more "best effort" attempts than it'd be worth to write yourself.

`resolv.conf` should work on all *nix systems, including OSX.  On OSX, while it is not used, it does look like the system keeps the file updated so it'd be available for information:
```
<danielwiddis> [01-03 17:06] ~/git/oshi $ cat /etc/resolv.conf
#
# Mac OS X Notice
#
# This file is not used by the host name and address resolution
# or the DNS query routing mechanisms used by most processes on
# this Mac OS X system.
#
# This file is automatically generated.
#
nameserver 8.8.8.8
nameserver 8.8.4.4
```

dbwiddis:
My suggestion based on everything discussed so far:
- Use what you have for Windows.
- Make a common *nix parent class that 
  - Implements the `InetAddress` approach for host and dns name, or uses the libc system calls for `getHostName()` and `getAddrInfo()`; depending on how good the canonical name actually performs
  - uses `/etc/resolv.conf` for nameservers
  - parses the output of `netstat -nr` for default gateway.
- Override the *nix parent in a few cases (e.g., when the `dnshostname` command is available, and the `route` command variants for gateway discussed in #269)

dbwiddis:
[This StackOverflow post](http://stackoverflow.com/questions/34842698/inetaddress-getcanonicalhostname-returns-ip-instead-of-hostname) discusses why `InetAddress.getCanonicalHostName` sometimes returns numeric.    Meanwhile, [the source code for `dnsdomainname`](http://www.gnu.org/software/shishi/coverage/shishi/gl/xgethostname.c.gcov.frameset.html) shows how easy it is to get with system calls.

chikei:
@dbwiddis PTAL

Changes:
* Using `InetAddress.getHostName()` for hostname
* Parse `/etc/resolv.conf` for *nix
* Using `route` for other *nix, (Linux stays the same because it outputs the same format as `/proc/net/route`)
* Using `getaddrinfo` for *nix

dbwiddis:
I'll take a look at this later this week.  Can you write some tests?

dbwiddis:
A few quick comments... codacy found an unused `LOG` variable, and my IDE is flagging a few serializable classes that don't have the serial version ID.

oshi-core/src/main/java/oshi/software/os/windows/WindowsNetworkParams.java
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getIpv4DefaultGateway() {
+        return getNextHop(IPV4_DEFAULT_DEST);
+    }
+
+    private String getNextHop(String dest) {
+        Map<String, List<Object>> vals = WmiUtil.selectObjectsFrom("ROOT\\StandardCimv2", "MSFT_NetRoute",
```
dbwiddis: The `ROOT\\StandardCimv2` requires Windows 8+.  This call produces errors on Windows 7.   

We need to provide a fallback method for finding the default gateway on Win7.  My suggestion would be to parse the output of `netstat -nr` and return the gateway for `0.0.0.0` (IPv4) and `::/0` (IPv6).

oshi-core/src/main/java/oshi/software/os/windows/WindowsNetworkParams.java
```diff
@@ -0,0 +1,136 @@
...
+     */
+    @Override
+    public String getDomainName() {
+        char[] buffer = new char[256];
+        IntByReference bufferSize = new IntByReference(buffer.length);
+        if (!Kernel32.INSTANCE.GetComputerNameEx(Kernel32.ComputerNameDnsDomain, buffer, bufferSize)) {
+            LOG.error("Failed to get dns domain name. Error code: {}", Kernel32.INSTANCE.GetLastError());
+            return "";
+        }
+        return new String(buffer);
```
dbwiddis: On my test VM, buffer is filled with null characters, so the `String` results in `"\u0000\u0000\u0000\  ... 256 times ... \u0000"`.   You should `return new String(buffer).trim();`.

oshi-core/src/main/java/oshi/software/common/AbstractNetworkParams.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractNetworkParams.class);
+    private static final String NAMESERVER = "nameserver";
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getDomainName() {
+        LibC.Addrinfo hint = new LibC.Addrinfo();
```
dbwiddis: These `LibC` calls here are specific to Unix, and this code will only work on FreeBSD and Solaris.  Either this Unix-specific should be implemented in a Unix class that accept the unix lass you imported, or you should do an OS check and import the appropriate LibC to use (the Linux LibC is a different class, and SystemB is the equivalent for the OS X case).

oshi-core/src/main/java/oshi/software/common/AbstractNetworkParams.java
```diff
@@ -0,0 +1,129 @@
...
+        hint.ai_flags = LibC.AI_CANONNAME;
+        String hostname = null;
+        try {
```
dbwiddis: Why not just call this class's own `getHostName()` here, where you've implemented this code?

oshi-core/src/main/java/oshi/software/os/mac/MacNetworkParams.java
```diff
@@ -0,0 +1,74 @@
...
+    private final String IPV6_ROUTE_HEADER = "Internet6:";
+
+    private final String DEFAULT_GATEWAY = "default";
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getDomainName() {
+        return "";
```
dbwiddis: This overrides the default methods we put in the abstract class.  If you get the correct `LibC`/`SystemB` class to create an instance of in that class, you can just leave this blank and OS X will use the C functions.

oshi-core/src/main/java/oshi/software/os/mac/MacNetworkParams.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String[] getDnsServers() {
```
dbwiddis: You're overriding the default method of reading `/etc/resolv.conf`.  Remove this override.

oshi-core/src/main/java/oshi/software/os/mac/MacNetworkParams.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getIpv6DefaultGateway() {
+        List<String> lines = ExecutingCommand.runNative("netstat -nr");
+        boolean v6Table = false;
+        for(String line: lines){
+            if(v6Table && line.startsWith(DEFAULT_GATEWAY)){
+                String[] fields = line.split("[ \t#;]");
```
dbwiddis: You need a `+` after the regex here to split on the multiple whitespace.   You could also just use `"\\s+"`.

dbwiddis: Also if you do decide to use the `\t` you would need a second backslash to escape the escape.

PR #297, opened 2017-01-11
Fix #295: Extend OSProcess (AbstractProcess.java) class is not easy

### LINKED ISSUE #295: Extend OSProcess (AbstractProcess.java) class is not easy
I need to return an OSProcess instance through my api with custom properties.  
But it is not easy to extend any OSProcess implementation since the AbstractProcess.java does not allow to access setters (implemented only in WindowsProcess.java for example).

Is there a way to do this ?

michaeldesigaud:


dbwiddis:
Looks great.  Can you please:

1.  Add tests for each setter method.  You can either update the existing test in `OperatingSystemTest.java` or create a new test file for the OSProcess (and move the getter tests from the existing ones over).

2. (Optional, if you want credit) Add a changelog entry.

3. (Optional, should eventually be done) Update the respective class-specific OperatingSystem classes to use a default constructor and call the setters, rather than calling one setter with too many arguments to count, which are easily skipped/reordered, etc.  Sonar, coverity, and other CI tools have told me it's bad practice but I have ignored them.  In general, we should probably move completely to a POJO model to make the json stuff easier but that's a huge task to do all at once; but doing it one class at a time as we add setters like in this PR seems a reasonable piecemeal approach.

PR #303, opened 2017-01-17
Fix 302: Missing setters and constructors in GlobalMemory implementations

michaeldesigaud:


oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
```diff
@@ -36,14 +36,14 @@
      *
      * @return Total number of bytes.
      */
-    long getTotal();
+    long getMemTotal();
```
dbwiddis: We can't change the API without a major version bump.  If you want to change the getter for readability you need to keep the old name around (deprecated) until the next major version bump.

oshi-core/src/main/java/oshi/software/common/AbstractProcess.java
```diff
@@ -50,12 +50,11 @@
 
     public AbstractProcess() {}
 
-    public AbstractProcess(String name, String path, State state, int processID, int parentProcessID, int threadCount,
+    public AbstractProcess(String name, String path, int processID, int parentProcessID, int threadCount,
                            int priority, long virtualSize, long residentSetSize, long kernelTime, long userTime,
                            long startTime, long upTime, long bytesRead, long bytesWritten) {
         this.name = name;
         this.path = path;
-        this.state = state;
```
dbwiddis: Why are we removing the state?

oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
```diff
@@ -18,14 +18,11 @@
...
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
-
 import org.junit.Test;
-
 import oshi.SystemInfo;
-import oshi.software.common.AbstractProcess;
+import oshi.software.os.windows.WindowsProcess;
```
dbwiddis: Can't use a platform-specific class in a test that will run on any OS.  Running this test on Linux results in the below:

> testGlobalMemoryCopy(oshi.hardware.GlobalMemoryTest)  Time elapsed: 0.024 sec  <<< ERROR!
> 
> java.lang.UnsatisfiedLinkError: Unable to load library 'Psapi': Native library (linux-x86-64/libPsapi.so) not found in resource path

oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
trejkaz: Whether it's an improvement for readability is also questionable. The class is already called GlobalMemory, so it's pretty obvious that a method called getTotal() would be getting the total memory.  And even then, it isn't saying it's getting the total memory, it's saying it's getting the total "mem". Whatever "mem" is.

dbwiddis: It's an abbreviation on a private variable that an automated getter/setter-maker produced. :)

But seriously, I was leaving the option open of changing variable names. "Physical" and "Virtual" would represent the two types of memory this class covers.   But unless we do something significant like that, there's no great need to change.

oshi-core/src/main/java/oshi/software/common/AbstractProcess.java
michaeldesigaud: oups! bad merge

oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
michaeldesigaud: Yep, i totally missed that. I added a default AbstractProcess implementation for unit tests. That should do the trick.

oshi-core/src/test/java/oshi/hardware/DefaultAbstractGlobalMemory.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Contributors:
+ * https://github.com/dblock/oshi/graphs/contributors
+ */
+
+package oshi.hardware;
+
+import oshi.hardware.common.AbstractGlobalMemory;
+
+public class DefaultAbstractGlobalMemory extends AbstractGlobalMemory {
```
dbwiddis: I'm not sure of the purpose of this class.  Is this just for testing?  (See other comment)

oshi-core/src/test/java/oshi/hardware/GlobalMemoryTest.java
```diff
@@ -49,4 +47,27 @@ public void testGlobalMemory() {
...
+     * Test OSProcess setters and getters
+     */
+    @Test
+    public void testGlobalMemoryCopy() {
+        SystemInfo si = new SystemInfo();
+        HardwareAbstractionLayer hal = si.getHardware();
+        GlobalMemory oldMemory = hal.getMemory();
+        assertNotNull(oldMemory);
+
+        DefaultAbstractGlobalMemory newMemory = new DefaultAbstractGlobalMemory();
```
dbwiddis: You've jumped through some hoops to create a new empty class so you can test the setters and getters, but why not just test arbitrary values like 42 or 12345.  You can test those in the same method you use to test for sane values for the other memory attributes.

oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
```diff
@@ -18,14 +18,10 @@
...
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
-
 import org.junit.Test;
-
 import oshi.SystemInfo;
-import oshi.software.common.AbstractProcess;
+
+import static org.junit.Assert.*;
```
dbwiddis: FYI, my source formatter/IDE will put this back the way it was (individual imports) whenever I eventually edit this.

oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
```diff
@@ -79,23 +75,11 @@ public void testOSProcessSetters(){
...
-        newProcess.setPriority(oldProcess.getPriority());
-        newProcess.setProcessID(oldProcess.getProcessID());
-        newProcess.setResidentSetSize(oldProcess.getResidentSetSize());
-        newProcess.setStartTime(oldProcess.getStartTime());
-        newProcess.setState(oldProcess.getState());
-        newProcess.setUpTime(oldProcess.getUpTime());
-        newProcess.setUserTime(oldProcess.getUserTime());
-        newProcess.setVirtualSize(oldProcess.getVirtualSize());
-        newProcess.setBytesWritten(oldProcess.getBytesWritten());
+        DefaultAbstractProcess newProcess = new DefaultAbstractProcess(oldProcess.getName(),oldProcess.getPath(),
```
dbwiddis: Why are we doing this?  We're removing tests of the setters, going in the opposite direction of what we're explicitly adding for Memory.    

And a more general question, other than a minor change (`abstract`), why are we mucking around with OSProcess in a PR for memory?

oshi-core/src/test/java/oshi/hardware/DefaultAbstractGlobalMemory.java
michaeldesigaud: Yes, to have a default implementation of AbstractGlobalMemory which allows unit tests on any platform

dbwiddis: You can just use SystemInfo.getHardware().getMemory() to get an object for your platform to run tests on.  Take a look at `NetworksTest.java` for the pattern I'd like to see.

michaeldesigaud: I wonder if setters have any sense in GlobalMemory.java because the values (swap, available, etcc) will always be overriden when calling getters. For example, setAvailable will be ignored since getAvailable call the updateMeminfo() which override the memAvailable property

dbwiddis: Ah, good insight.  That's why it didn't originally have them to begin with.  And that would make testing set/get not very repeatable!

At the same time, I'm interested in eventually migrating most of the objects to POJO to make serialization, JSON/XML, etc.  But that's probably a bigger rewrite/restructuring of everything that might not make sense to do piecemeal.

michaeldesigaud: So i wonder if it could be interesting to retrieve the memAvailable value without updating it first ? For example, like you said, to make POJO serialization and we just want to copy the value not necessarly to retrieve the new value

dbwiddis: If we do this we need to timestamp it; return the old value if current timestamp < old timestamp + 900 ms.

michaeldesigaud: Ok. Ayway i think we can delete this PR,  because like you said, adding setters on GlobalMemory is a much more complicated subject. Sorry for the inconvenience

dbwiddis:
Closing per request since this pulls dynamic information and setters aren't very applicable.  We may or may not revisit this as part of a project-wide rewrite.

Issue #296, opened 2017-01-11
Processor Serial Number should return the CPU's SN, not the BIOS' SN


YoshiEnVerde:
While the current impementation of **_getSystemSerialNumber_** from the **CentralProcessor** interface is solid on every O.S., I find myself disconcerted on how the values returned are not the values I expected when I first called the method.

From following the logical path _System > Hardware > Processor > Serial Number_, I'm expecting to recieve the **_Processor's SN_**.
Instead, I get the **_BIOS' SN_**.

This gets worse when I read the source code, and I find that this value is kept in memory as "**cpuSerialNumber**".
Even worse, the processor's SN doesn't ever even come close to the implementation of these methods.

Then, when I think something lie "OK, so _System Serial Number_ will target the **BIOS**, but at least I have a _Processor Identifier_, that might have the **Processor's ID**", and knowing that for all purposes the Processor's ID **_is_** its SN, I call that method, and I get a string made by grabbing a few other values and concatenating them; a string closer to the Processor's Name that to the ID.

As an example, running a simple java app to test these values, I get:
> Processor Vendor: GenuineIntel
> Processor x64: true
> Processor Family: 6
> Processor Model: 69
> Processor Stepping: 1
> 
> Processor ID: Intel64 Family 6 Model 69 Stepping 1
> Processor Name: Intel(R) Core(TM) i3-4005U CPU @ 1.70GHz
> Processor SSN: G7N0GP000709308

Meanwhile, running **dmidecode**, I get the following:
> Processor ID: 51 06 04 00 FF FB EB BF
> Processor Version: Intel(R) Core(TM) i3-4005U CPU @ 1.70GHz
> Processor Signature: Type 0, Family 6, Model 69, Stepping 1
> System Serial Number: G7N0GP000709308

Those are values from an Ubuntu Linux distro.
Same thing, from a Win7 VM (and using **wmic** instead of **dmidecode**, of course):
OSHI:
> Processor Vendor: GenuineIntel
> Processor x64: false
> Processor Family: 6
> Processor Model: 30
> Processor Stepping: 5
> 
> Processor ID: x86 Family 6 Model 30 Stepping 5
> Processor Name: Intel(R) Xeon(R) CPU           X3440  @ 2.53GHz
> Processor SSN: VMware-56 4d da b8 62 63 33 16-67 b0 20 ae 74 33 36 95

WMIC:
> Processor ID: 0FABFBFF000106E5
[...]

PR #307, opened 2017-02-03
Deprecate CentralProcessor's getSystemSerialNumber method

dbwiddis:
It duplicated ComputerSystem's getSerialNumber method.  See #296

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ================
 * [#294](https://github.com/oshi/oshi/pull/294), [#305](https://github.com/oshi/oshi/pull/305): Add NetworkParams for network parameter of OS - [@chikei](https://github.com/chikei), [@dbwiddis](https://github.com/dbwiddis).
 * [#295](https://github.com/oshi/oshi/pull/295): Make OSProcess (AbstractProcess.java) more easily extendable - [@michaeldesigaud](https://github.com/michaeldesigaud).
+* [#307](https://github.com/dblock/oshi/pull/307): Deprecate CentralProcessor's getSystemSerialNumber method that duplicated ComputerSystem's getSerialNumber method. - [@dbwiddis](https://github.com/dbwiddis).
 * Your contribution here.
 
 3.3 (12/31/2016)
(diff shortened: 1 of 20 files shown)


neilyoung:
What is supposed to be used instead?

Qiu1024:
I think it's:
"
 SystemInfo sys = new SystemInfo();
 ComputerSystem cs = sys.getHardware().getComputerSystem();
 String serialNum = cs.getSerialNumber();
"

neilyoung:
@Qiu1024 Indeed. Thanks

neilyoung:
@Qiu1024 The problem with cs.getSerialNumber() is, that it returns "unkown" on two different Ubuntu systems, hence it cannot be used for generating keys. What else do you have? :)

YoshiEnVerde:
@neilyoung On which Ubuntu versions is this failing? Are you running the code under root privileges?

neilyoung:
@YoshiEnVerde 

Ubuntu 17.10 on a 10 years old 32 bit Samsung R509 laptop
And no, I'm not (g)root :)

YoshiEnVerde:
@neilyoung Then **that's** where it's failing. For older versions of Linux, the code defaults to HAL (and equivalent) tables. For newer versions, we check values that can only be accessed by root.

Sadly, hardware info at this level can only be accessed by root users in Linux

dbwiddis:
> Sadly, hardware info at this level can only be accessed by root users in Linux

Sadly for OSHI.  Happily for privacy advocates who won this particular battle in Linux.

But yes... for full functionality on Linux (and many Unix features) part of your implementation plan needs to be running OSHI (or the program you import it into) with elevated privileges.

neilyoung:
@dbwiddis Strange enough, the same code doesn't need to have root on MacOSX.

dbwiddis:
There are workarounds besides root.... for example, installing `hal` package.   OSHI has several fallbacks that it attempts but if the user doesn't cooperate, there's not much we can do!

Not sure why Apple makes it easy. :)

neilyoung:
@dbwiddis What are the other fallbacks? There is no HAL installer any longer for Ubuntu since 10.something

Issue #275, opened 2016-11-22
Support getting User/Group/Principal data per process


CodingFabian:
In Sigar you can do getProcCredName/getProcCred to get ids and names for the user/group somtimes called principal executing a process.
It would be great to have such an API in Oshi. Ideally which returns this data in a single call.

dbwiddis:
If the information is available in the extended format of the ps -O command, this seems doable.  I think getting userid/username is easy.  Not sure about the group but maybe a *nix expert can chime in here.

PR welcome.

dbwiddis:
I've been looking into this.  I'd definitely like to add user/group information.  It's easily doable on the *nix-based systems:
- FreeBSD, Solaris:  the existing `ps` command line `-O`/`-o` switch has `uid`, `user`, `gid`, and `group` fields. Easy peasy
- Linux: we currently read the `/proc` directory for process information.  We can use `stat()` followed by `getpwuid()` and `getgrgid()` to get ownership of the `/proc/pid` directory for each process.  Or, we can simply parse the `ps` output like the other *nixes.
- OS X: we currently programmatically get the `proc_bsd_info` structure which contains the uid and gid.  We can then use `getpwuid()` and `getgrgid()` as in Linux.  Or, we can simply parse the `ps` output like the other *nixes.

That leaves Windows.  Which doesn't make anything easy.

The existing WMI implementation uses the [Win32_Process](https://msdn.microsoft.com/en-us/library/aa394372(v=vs.85).aspx) class. Unfortunately, this doesn't have user/group/domain info.  There's a `getOwner` method on the class, but this is not accessible via the WMI queries we use.  We would have to get creative parsing a powershell commandline.

The "proper" programmatic way to do it is to [enumerate the process list using PSAPI](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682629(v=vs.85).aspx), then for each process,  use `OpenProcessToken` to get the token, `GetTokenInformation` with the `TokenOwner` flag to get the SID of the owner, then `LookupAccountSid`.  [See here](http://stackoverflow.com/questions/2686096/c-get-username-from-process).  And all that's just for the owner.  I'm not even sure Windows has a `group` equivalent.

dbwiddis:
[...]

PR #314, opened 2017-02-15
WMI ExecMethod implementation

dbwiddis:


dbwiddis:
Windows implementation to support #275

PR #322, opened 2017-02-26
Extend PowerSource classes

ejaszewski:
Extends PowerSource classes on all systems but FreeBSD. Converts PowerSources to getter/setter system.

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/PowerSource.java
+++ b/oshi-core/src/main/java/oshi/hardware/PowerSource.java
@@ -25,6 +25,7 @@ import java.io.Serializable;
  * of charge/discharge
  *
  * @author widdis[at]gmail[dot]com
+ * @author ethanjaszewski[at]yahoo[dot]com
  */
 public interface PowerSource extends Serializable {
     /**
@@ -48,4 +49,31 @@ public interface PowerSource extends Serializable {
      *         or -2.0 (unlimited)
      */
     double getTimeRemaining();
+    
+    /**
+     * Power source health as a fraction of the battery's designed capacity.
+     * 
+     * @return A value between 0.0 (no capacity) and 1.0 (at designed capacity)
+     */
+    double getHealth();
+    
+    /**
+     * Maximum charge of the power source, in milliwatt hours.
+     * 
+     * @return If positive, milliwatt hours of charge. If negative, the maximum charge is unknown.
+     */
+    long getMaximumCharge();
+    
+    /**
+     * Remaining charge of the power source, in milliwatt hours.
+     * 
+     * @return If positive, milliwatt hours of charge. If negative, the Remaining charge is unknown. 
+     */
+    long getRemainingCharge();
+    
+    /**
+     * Current power supplied by the power source, in milliwatts.
+     * @return If positive, milliwats of power. If negative, the power is unknown.
+     */
+    long getPower();
 }
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
@@ -35,7 +35,15 @@ public abstract class AbstractPowerSource implements PowerSource {
     protected double remainingCapacity;
 
     protected double timeRemaining;
+    
+    protected double health;
 
+    protected long maximumCharge;
+    
+    protected long remainingCharge;
+    
+    protected long power;
+    
     /**
      * Super constructor formerly used by platform-specific implementations
      * of PowerSource
@@ -87,6 +95,38 @@ public abstract class AbstractPowerSource implements PowerSource {
     @Override
     public double getTimeRemaining() {
         return this.timeRemaining;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public double getHealth() {
+        return health;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public long getMaximumCharge() {
+        return maximumCharge;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public long getRemainingCharge() {
+        return remainingCharge;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public long getPower() {
+        return power;
     }
 
     /**
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
@@ -69,8 +69,12 @@ public abstract class AbstractPowerSource implements PowerSource {
      */
     public AbstractPowerSource() {
         this.name = "unknown";
-        this.remainingCapacity = -1;
+        this.remainingCapacity = 0;
         this.timeRemaining = -1;
+        this.health = 1d;
+        this.maximumCharge = 0;
+        this.remainingCharge = 0;
+        this.power = 1;
     }
 
     /**
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
@@ -24,6 +24,7 @@ import oshi.hardware.PowerSource;
  * A Power Source
  *
  * @author widdis[at]gmail[dot]com
+ * @author ethanjaszewski[at]yahoo[dot]com
  */
 public abstract class AbstractPowerSource implements PowerSource {
 
@@ -36,8 +37,8 @@ public abstract class AbstractPowerSource implements PowerSource {
     protected double timeRemaining;
 
     /**
-     * Super constructor used by platform-specific implementations of
-     * PowerSource
+     * Super constructor formerly used by platform-specific implementations
+     * of PowerSource
      *
      * @param newName
      *            The name to assign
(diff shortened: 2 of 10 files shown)


oshi-core/src/main/java/oshi/hardware/PowerSource.java
```diff
@@ -48,4 +49,31 @@
      *         or -2.0 (unlimited)
      */
     double getTimeRemaining();
+    
+    /**
+     * Power source health as a fraction of the battery's designed capacity.
+     * 
+     * @return A value between 0.0 (no capacity) and 1.0 (at designed capacity)
+     */
+    double getHealth();
```
dbwiddis: Is "Health" a reasonably consistent term?  I googled and found [State of Health (SOH)](https://en.wikipedia.org/wiki/State_of_health) which looks like exactly what you're trying to represent here.  I think `getHealth()` is probably ok but if SOH is really a more common/standard term I might prefer a `getStateOfHealth()` method.

oshi-core/src/main/java/oshi/hardware/PowerSource.java
```diff
@@ -48,4 +49,31 @@
...
+     * @return A value between 0.0 (no capacity) and 1.0 (at designed capacity)
+     */
+    double getHealth();
+    
+    /**
+     * Maximum charge of the power source, in milliwatt hours.
+     * 
+     * @return If positive, milliwatt hours of charge. If negative, the maximum charge is unknown.
+     */
+    long getMaximumCharge();
```
dbwiddis: At the time I'm reviewing this I've only looked at the interface and not the implementation. I see we have a health/state of health measure which ratios current max vs. design max, and it's not clear to me which of those two this method fetches.   I recall in the issue we discussed including both design max capacity and current max capacity.   I gather with the health stat we can compute one from the other, but if we go that route we need to be clear which one this is.

If we're only including two stats I'd prefer to have design max and current max and let the user calculate the health themselves.

I also think `getMaximumCapacity()` would be better than `getMaximumCharge()` -- however, I also see we already have `getRemainingCapacity()` as a fraction of the max, and it's necessary to use different terms.  I was probably wrong to use capacity to represent the fraction (and not the mWh statistic) and I'd want the correct terms implemented on a major version change.  I'm wondering if there's a better way of presenting this info here.

oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
```diff
@@ -46,11 +55,27 @@
      * @param newTimeRemaining
      *            Seconds of time remaining
      */
+    @Deprecated
```
dbwiddis: Good effort but not necessary. The OSHI API is documented as being the interfaces and classes in the `oshi.hardware` and `oshi.software.os` packages.  Constructors in the class-specific (or abstract) classes aren't API and it's okay to remove this one (and the class-specific variants) without documentation.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxPowerSource.java
```diff
@@ -73,13 +79,20 @@ public LinuxPowerSource(String newName, double newRemainingCapacity, double newT
...
                 continue;
             }
+            
+            LinuxPowerSource pSource = new LinuxPowerSource();
+            
             // Initialize defaults
             boolean isPresent = false;
             boolean isCharging = false;
-            String name = "Unknown";
+            String name = "unknown";
```
dbwiddis: Earlier you changed a lowercase unknown to an uppercase Unknown, and here you do the opposite.  No big deal which one you use, but should try to be consistent! :)

oshi-core/src/test/java/oshi/hardware/PowerSourceTest.java
```diff
@@ -40,6 +40,10 @@ public void testPowerSource() {
             double epsilon = 1E-6;
             assertTrue(ps.getTimeRemaining() > 0 || Math.abs(ps.getTimeRemaining() - -1) < epsilon
                     || Math.abs(ps.getTimeRemaining() - -2) < epsilon);
+            assertTrue(ps.getHealth() >= 0 && ps.getHealth() <= 1);
```
dbwiddis: Not sure health is always <=1.  I've seen batteries with current capacity > design capacity.

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisPowerSource.java
```diff
@@ -110,8 +127,12 @@ public SolarisPowerSource(String newName, double newRemainingCapacity, double ne
 
         // Set up single battery in array
         SolarisPowerSource[] ps = new SolarisPowerSource[1];
-        ps[0] = new SolarisPowerSource("BAT0", (double) energyNow / energyFull,
-                isCharging ? -2d : powerNow > 0 ? 3600d * energyNow / powerNow : -1d);
+        ps[0] = new SolarisPowerSource();
+        ps[0].setName("BAT0");
+        ps[0].setRemainingCapacity((double) energyNow / energyFull);
+        ps[0].setTimeRemaining(isCharging ? -2d : 3600d * energyNow / powerNow);
```
dbwiddis: Need to check for the case if powerNow = 0.

oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
```diff
@@ -46,11 +55,27 @@
...
+    /**
+     * Super constructor used by platform specific implementations of
+     * PowerSource. Initializes all values to defaults, which are then set
+     * using the setter methods.
+     */
+    public AbstractPowerSource() {
+        this.name = "unknown";
+        this.remainingCapacity = 0;
+        this.timeRemaining = -1;
+        this.health = 1d;
```
dbwiddis: OK, I'm really being type A and this is probably so trivial I shouldn't comment.  But I'm on a roll. ;)   This declaration is properly a `double`.  The other defaults above and below should also match their defined type to save a nanosecond or two of int-to-double casting.  Or not.  Not a big deal, really.  ;)

ejaszewski:
I don't have a problem with pushing this back to 4.0.0, given the scope of the changes. The getCapacity() should definitely return a mWh value, and I still don't have any idea if the code works on Win/Mac/Solaris, and BSD doesn't even have an implementation. Moving back to 4.0.0 would give us adequate time to test on all of the systems, as well as allowing for the renaming. We could also put the new information in in 3.4.0 as getMaximumEnergy, getRemainingEnergy, etc. which would be a more technically correct name, then we could refactor in 4.0.0, along with redoing the PowerSource system to use an update() method. Let me know what you think is best.

dbwiddis:
No need to major version bump for a minor renaming.  The fact is that `getCapacity()` is vague and we just need to create a set of `getFooCapacity()` methods appropriately.    For example, deprecate `getCapacity()` and point it to `getPercentCapacity()`.  Create new `getMaxEnergyCapacity()` and `getDesignEnergyCapacity()` and `getCurrentEnergyCapacity()` etc.

dbwiddis:
I'll test win/mac tonight... and there may not be a solution on FreeBSD so just create some sensible defaults.

dbwiddis:
macOS test....
Fully charged, plugged in:
```
Name  :InternalBattery-0
Health:25655.0
Power :0
Maximum Charge    :0
Remaining Charge  :0
Remaining Capacity:1.0
Time Remaining    :-2.0
```
Discharging, at 100%:
```
Name  :InternalBattery-0
Health:25655.0
Power :0
Maximum Charge    :0
Remaining Charge  :0
Remaining Capacity:1.0
Time Remaining    :8460.0
```
Something tells me those aren't supposed to be all 0's :)

dbwiddis:
Looked over the OSX code.  Seems right, but the keys we need are only "recommended" and apparently aren't implemented on my system.   You should default the design capacity to the max capacity for the numbers to work.   

There is a lot of info in the registry (command line ioreg but we can get it programmatically) for the "AppleSmartBattery" which could be an alternate backup source of info -- and more cool stuff like cycle count: 
screen shot 2017-02-26 at 6 08 08 pm

So I'm thinking we do push this back to the next version and take the time to dig into alternative info sources when the IOPS calls we currently use don't give the results.

dbwiddis:
Windows test.
Fully charged, plugged in:
```
Name  :System Battery
Health:1.0
Power :0
Maximum Charge    :0
Remaining Charge  :0
Remaining Capacity:100000.0
Time Remaining    :-2.0
```


[WMI](https://msdn.microsoft.com/en-us/library/aa394074(v=vs.85).aspx) looked promising at first, but there are a lot of blanks:
```
C:\>wmic path Win32_Battery
Availability  BatteryRechargeTime  BatteryStatus  Caption           Chemistry  ConfigManagerErrorCode  ConfigManagerUserConfig  CreationClassName  Description       DesignCapacity  DesignVoltage  DeviceID    ErrorCleared  ErrorDescription  EstimatedChargeRemaining  EstimatedRunTime  ExpectedBatteryLife  ExpectedLife  FullChargeCapacity  InstallDate  LastErrorCode  MaxRechargeTime  Name     PNPDeviceID  PowerManagementCapabilities  PowerManagementSupported  SmartBatteryVersion  Status  StatusInfo  SystemCreationClassName  SystemName       TimeOnBattery  TimeToFullCharge
3                                  1              Internal Battery  2                                                           Win32_Battery      Internal Battery                  12835          SMPbq20z45                                  100                       362                                                                                                                   bq20z45               {1}                          FALSE                                          OK                  Win32_ComputerSystem     JACQUELINEWA909
```
Looks like there are [some other API calls](https://msdn.microsoft.com/en-us/library/windows/desktop/aa372693(v=vs.85).aspx) with more power related info.

ejaszewski:
Agreed on pushing back to 3.5. Probably for the best since it will give me time to get BSD working as well.

PR #327, opened 2017-02-27
Java 7 Compatibility

### LINKED ISSUE #326: Java 7 support
I just found your great library as I was searching for a way to get better information for troubleshooting issues in [UniversalMediaServer](https://github.com/UniversalMediaServer/UniversalMediaServer). Universal Media Server is an UPnP/DLNA server that serves media files to DLNA capable devices like mobile phones, "smart" TV's, game consoles etc. We often have to covert content from one format to another on the fly to make it available in a format that the "client" can understand. For this we use mostly non-Java software like FFmpeg. That means that the amount of memory (and the CPU capabilities, load etc) available to the OS is interesting to us, not just what the JVM has been assigned.

Although we want to ditch Java 7 as well, it doesn't seem like we can for quite some time yet without excluding a lot of users. The last time I checked there were still Linux distros out there where Java 8 isn't yet available. Oracle has decided (with good help from Microsoft I suspect) to withdraw their Windows XP support for Java 8 (it was supported in the beginning). [Windows XP is still the 3. most used OS today](https://www.netmarketshare.com/operating-system-market-share.aspx?qprid=10&qpcustomd=0), much larger than any Linux distro or OS X.

I'm guessing that there won't be long until the remaining Linux distros offer OpenJDK8, but the Windows XP situation isn't likely to be solved. 

I just found that Red Hat is offering [OpenJDK 8 for Windows](https://developers.redhat.com/blog/2016/06/27/openjdk-now-available-for-windows/), but I haven't been able to find any information about which Windows versions are supported. Registration is required for download, so I haven't checked it any further.

I'm currently testing out version 2.6.1-Java7, and it seems to provide us with what we need. But, the same consideration might be relevant in your decision-making.

dbwiddis:
Fixes #326

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ script: mvn test
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
-  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+  - if [ "$AFTER_SUCCESS" == "jacoco" ]; then
       mvn clean test jacoco:report coveralls:report;
       mvn clean org.jacoco:jacoco-maven-plugin:prepare-agent package sonar:sonar -Dsonar.host.url=https://sonarqube.com -Dsonar.login=df16219a00545aae71f880423a6f23191330744f;
     fi
@@ -28,5 +28,12 @@ matrix:
   include:
     - os: linux
       jdk: oraclejdk8
+      env: AFTER_SUCCESS="jacoco"
+    - os: linux
+      jdk: oraclejdk7
+    - os: linux
+      jdk: openjdk7
     - os: osx
       osx_image: xcode8
+    - os: osx
+      env: JAVA_HOME=$(/usr/libexec/java_home)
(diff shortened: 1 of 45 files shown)


Nadahar:
I just did a test on my Windows XP VM, and it worked fine 👍 

I should be mentioned that my use in this test is limited to basic OS, RAM and CPU information.

Wouldn't it be beneficial to change the name of ```getOrDefault()``` so that it will also compile on Java 8? I can see that there are 42 references to the method, so that might be considered too many to maintain? 

The way we work is that we're in reality using Java 8, but we make sure that the code is Java 7 compatible (I use Eclipse's compliance checking to enforce Java 7 syntax). If something slips, Travis will let us know. 

That way, we only actually use Java 7 to compile the Java 7 release.

It would be much more hazzle if we have to switch dependencies when compiling with Java 7 or Java 8.

dbwiddis:
Yes, as mentioned on the Issue, I totally missed the impact of the name conflict. You're absolutely right that it should work on Java 8... I was trying to keep the same naming conventions as before. Now I'm trying to decide what to name the methods...

Nadahar:
```getWithDefault()```?

dbwiddis:
I eventually went back to a Util class that takes a `Map` as an argument, rather than trying to extend a particular map implementation. This restores the flexibility to use on either `HashMap` or `ConcurrentHashMap` (we had both and used features of both) and lets me use the original method names.  This is all done except for some travis configuration.

PR #361, opened 2017-05-14
Adding in mapped volume attribute to file system (only implemented in…

darinhoward:
Adding in mapped volume attribute to file system (only implemented in linux). This fixes a gap where some filesystems via LVM cannot be linked to drives with the volume or mount values.

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/OSFileStore.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSFileStore.java
@@ -35,6 +35,8 @@ public class OSFileStore implements Serializable {
 
     private String volume;
 
+    private String mappedVolume;
+
     private String mount;
 
     private String description;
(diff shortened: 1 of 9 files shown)


dbwiddis:
Rather than adding a new constructor with a new parameter, can we just go with a no-arg constructor and a bunch of setters?  It'll make the code cleaner and yes, I know that's a lot more work but I think it's the right thing to do going forward and rewrite as we expand/fix particular classes.

Also you need to write tests in FileSystemTest.

I'm still musing whether "mapped volume" is the best name choice and thinking "logical volume" may be better. I'd like something that is consistent across other OS's, particularly the OS X fusion drive/FileVault, and possibly ZFS.

darinhoward:
@dbwiddis I've made a separate pull request with your suggestions - https://github.com/oshi/oshi/pull/362

PR #362, opened 2017-05-15
Adding in logical volume attribute to file system (only implemented i…

darinhoward:
Adding in logical volume attribute to file system (only implemented in linux). This fixes a gap where some filesystems via LVM cannot be linked to drives with the volume or mount values.


dbwiddis:
This looks great.  Three minor requests:

1. In the `SystemInfoTest` code, use a conditional on whether to display the logical volume.  At present you'll mostly get empty brackets `[]`. 
2. Please elaborate slightly on the Javadoc for the logical volume description with a "for example" type statement.  (I plan to rewrite the macOS code to move the FileVault / Fusion Drive logical volumes to this field.)
3. Make a changelog entry! :)

darinhoward:
I've updated the pull request - let me know if you would like any other alterations.

PR #367, opened 2017-05-24
Update maven wrapper to 0.2.1 and change missed mvn to ./mvnw

hazendaz:


hazendaz:
One of the java 7 builds is failing.  Not sure why.  All others are passing.

dbwiddis:
Openjdk build is a known bug in Travis. Previously fixed but regressed. There's a workaround that I'm too lazy to put in, hoping they'll fix it first. :)

dbwiddis:
Workaround is to add the following in travis.yml (name is irrelevant, having any hostname is the fix):
```
+addons:
 +  hosts:
 +    - foobar
 +  hostname: foobar
```
While you're updating versions can you update pom dependencies?

hazendaz:
ok - I'll live with the broken issue for time being.  Interesting enough it built entirely on my fork without issues.

I'll take a look at the dependencies as well.  That will at least get another build kicking to see if it fixes the issues.  I'll merge this for now as it's good to go otherwise.

PR #381, opened 2017-07-09
Performance improvement on regex reuse

igreenfield:
Create pattern for whitespace and use it in Linux implementation.

### DIFF
--- a/oshi-core/pom.xml
+++ b/oshi-core/pom.xml
@@ -47,11 +47,6 @@
 			<version>4.4.0</version>
 		</dependency>
 		<dependency>
-			<groupId>org.apache.commons</groupId>
-			<artifactId>commons-lang3</artifactId>
-			<version>3.6</version>
-		</dependency>
-		<dependency>
 			<groupId>org.threeten</groupId>
 			<artifactId>threetenbp</artifactId>
 			<version>1.3.4</version>
--- a/oshi-core/pom.xml
+++ b/oshi-core/pom.xml
@@ -47,6 +47,11 @@
 			<version>4.4.0</version>
 		</dependency>
 		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-lang3</artifactId>
+			<version>3.6</version>
+		</dependency>
+		<dependency>
 			<groupId>org.threeten</groupId>
 			<artifactId>threetenbp</artifactId>
 			<version>1.3.4</version>
--- a/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
@@ -18,7 +18,6 @@
  */
 package oshi.hardware.platform.linux;
 
-import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import oshi.hardware.common.AbstractCentralProcessor;
@@ -63,7 +62,7 @@ public class LinuxCentralProcessor extends AbstractCentralProcessor {
         String[] flags = new String[0];
         List<String> cpuInfo = FileUtil.readFile("/proc/cpuinfo");
         for (String line : cpuInfo) {
-            String[] splitLine = ParseUtil.whitespaceCloneWhitespace.split(line);
+            String[] splitLine = ParseUtil.whitespaceColonWhitespace.split(line);
             if (splitLine.length < 2) {
                 break;
             }
@@ -302,9 +301,9 @@ public class LinuxCentralProcessor extends AbstractCentralProcessor {
                 String edx = "";
                 for (String register : ParseUtil.whitespace.split(checkLine)) {
                     if (register.startsWith("eax=")) {
-                        eax = StringUtils.replace(register, "eax=0x", "");
+                        eax = ParseUtil.removeMatchingString(register, "eax=0x");
                     } else if (register.startsWith("edx=")) {
-                        edx = StringUtils.replace(register, "edx=0x", "");
+                        edx = ParseUtil.removeMatchingString(register, "edx=0x");
                     }
                 }
                 return edx + eax;
--- a/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
@@ -67,35 +67,35 @@ public class LinuxCentralProcessor extends AbstractCentralProcessor {
                 break;
             }
             switch (splitLine[0]) {
-                case "vendor_id":
-                    setVendor(splitLine[1]);
-                    break;
-                case "model name":
-                    setName(splitLine[1]);
-                    break;
-                case "flags":
-                    flags = splitLine[1].toLowerCase().split(" ");
-                    boolean found = false;
-                    for (String flag : flags) {
-                        if ("lm".equals(flag)) {
-                            found = true;
-                            break;
-                        }
+            case "vendor_id":
+                setVendor(splitLine[1]);
+                break;
+            case "model name":
+                setName(splitLine[1]);
+                break;
+            case "flags":
+                flags = splitLine[1].toLowerCase().split(" ");
+                boolean found = false;
+                for (String flag : flags) {
+                    if ("lm".equals(flag)) {
+                        found = true;
+                        break;
                     }
-                    setCpu64(found);
-                    break;
-                case "stepping":
-                    setStepping(splitLine[1]);
-                    break;
-                case "model":
-                    setModel(splitLine[1]);
-                    break;
-                case "cpu family":
-                    setFamily(splitLine[1]);
-                    break;
-                default:
-                    // Do nothing
-            }
+                }
+                setCpu64(found);
+                break;
+            case "stepping":
+                setStepping(splitLine[1]);
+                break;
+            case "model":
+                setModel(splitLine[1]);
+                break;
+            case "cpu family":
+                setFamily(splitLine[1]);
+                break;
+            default:
+                // Do nothing
+        }
         }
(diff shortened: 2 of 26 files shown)


oshi-core/pom.xml
```diff
@@ -47,6 +47,11 @@
 			<version>4.4.0</version>
 		</dependency>
 		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-lang3</artifactId>
```
dbwiddis: Have you looked into any documentation requirements for including an Alv2 dependency into an EPL product?

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
```diff
@@ -1,37 +1,37 @@
 /**
  * Oshi (https://github.com/oshi/oshi)
- *
+ * <p>
```
dbwiddis: These comment headers are automatically rewritten by maven.  If you need to change them, you should change the [LICENSE_HEADER](https://github.com/oshi/oshi/blob/master/LICENSE_HEADER) file.

oshi-core/src/main/java/oshi/util/ParseUtil.java
```diff
@@ -83,6 +83,10 @@
 
     private static final Map<String, Long> multipliers;
 
+    public static final Pattern whitespaceCloneWhitespace = Pattern.compile("\\s+:\\s");
```
dbwiddis: Should this be "Colon" rather than "Clone"?

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
```diff
@@ -58,43 +58,44 @@ public LinuxCentralProcessor() {
...
-                setName(splitLine[1]);
-                break;
-            case "flags":
-                flags = splitLine[1].toLowerCase().split(" ");
-                boolean found = false;
-                for (String flag : flags) {
-                    if ("lm".equals(flag)) {
-                        found = true;
-                        break;
+                case "vendor_id":
```
dbwiddis: This looks like mostly formatting changes. Please use the [same Eclipse formatting](https://github.com/oshi/oshi/blob/master/oshi-dist/src/main/resources/eclipse/format.xml) as the project standard.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
```diff
@@ -299,11 +300,11 @@ private String getProcessorID(String stepping, String model, String family, Stri
             if (checkLine.contains(marker) && checkLine.trim().startsWith("0x00000001")) {
                 String eax = "";
                 String edx = "";
-                for (String register : checkLine.split("\\s+")) {
+                for (String register : ParseUtil.whitespace.split(checkLine)) {
                     if (register.startsWith("eax=")) {
-                        eax = register.replace("eax=0x", "");
+                        eax = StringUtils.replace(register, "eax=0x", "");
```
dbwiddis: This looks like the primary performance improvement, replacing `String.replace()`'s inefficient regex implementation with `StringUtils.replace()` which uses a `StringBuffer`.   However, it would be even faster with a `StringBuilder` implementation.    If we want efficiency, can we just write our own instead of introducing an entirely new dependency simply to "delete this character sequence from our string" ?

YoshiEnVerde: Just as a small help, the following is the code from the Apache Commons StringUtils.replace(String, String, String) method simplified, to avoid performance hits from the varied uses available, to just replacing all the occurences of the search string in the original value:

`/**
* <p>Replaces all occurrences of a String within another String.</p>
*
* <p>A {@code null} reference passed to this method is a no-op.</p>
*
* <pre>
* StringUtils.replace(null, *, *)        = null
* StringUtils.replace("", *, *)          = ""
* StringUtils.replace("any", null, *)    = "any"
* StringUtils.replace("any", *, null)    = "any"
* StringUtils.replace("any", "", *)      = "any"
* StringUtils.replace("aba", "a", null)  = "aba"
* StringUtils.replace("aba", "a", "")    = "b"
* StringUtils.replace("aba", "a", "z")   = "zbz"
* </pre>
*
* @param text  text to search and replace in, may be null
* @param searchString  the String to search for, may be null
* @param replacement  the String to replace it with, may be null
* @return the text with any replacements processed,
*  {@code null} if null String input
*/
private static String replace(final String text, String searchString, final String replacement, int max, final boolean ignoreCase) {
    if (text == null || text.isEmpty() || searchString == null || searchString.isEmpty() || replacement == null) {
        return text;
    }

    int start = 0;
    int end = text.indexOf(searchString, start);

    if (end == -1) {
        return text;
    }

    final int replLength = searchString.length();
    int increase = replacement.length() - replLength;
    increase = increase < 0 ? 0 : (increase * 16);

    final StringBuilder buf = new StringBuilder(text.length() + increase);
    while (end != -1) {
        buf.append(text.substring(start, end)).append(replacement);
        start = end + replLength;
        end = text.indexOf(searchString, start);
    }
    buf.append(text.substring(start));
    return buf.toString();
}`

The only thing missing would be any needed reference to the code being leveraged from the Apache Commons v3.6

dbwiddis: > The only thing missing would be any needed reference to the code being leveraged from the Apache Commons v3.6

That code is licenced under the Apache License v. 2.0, so we need to be very careful about licensing a derivative work if we base our method on theirs.

YoshiEnVerde: This is the only case we would apply this new method?
We could just as well write a util method: **removeMatch(String original, String search)**
And then implement something almost as efficient as this **replace()**, but for our specific case.

dbwiddis: Yep.  This is probably the way to go.

YoshiEnVerde: I just added a new code snippet to my last answer, with a possible implementation of the new method

oshi-core/src/main/java/oshi/util/ParseUtil.java
igreenfield: All the split remove the regex compile on each split. 
also, the replace of String also compile regex.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
igreenfield: Done

oshi-core/src/main/java/oshi/util/ParseUtil.java
```diff
@@ -489,4 +493,31 @@ public static int getNthIntValue(String line, int n) {
...
         return 0;
     }
+
+    /**
+     * Removes all sub strings from the string
+     * @param original source String to remove from
+     * @param toRemove the sub string to be removed
+     * @return
+     */
+    public static String removeMatchingString(final String original, final String toRemove) {
```
dbwiddis: In the interest of efficiency, let's do one last change. Add a boolean argument `removeAll`.  If true, behaves as written (`String.replaceAll()`). If false, only removes the first occurrence (`String.replace()`).

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
dbwiddis: I'll merge like this, but unless we change the license header before the next release (snapshot) they'll get reverted.  I guess I can add these paragraph breaks to the license header though.

oshi-core/src/main/java/oshi/util/ParseUtil.java
YoshiEnVerde: Well, the original code (both Apache and JDK) has two extra params I've removed to simplify the code: an int to specify the amount of times the match is searched for, and a boolean to set if you wanted case to be ignored.

We can add the boolean, move the method to private access, and then give two public methods, for removeAllMatchingStrings and removeFirstMatchingString, to keep from duplicating code.

Otherwise, we could implement:
```
String removeFirstMatchingString(final String original, String toRemove) {
    if (original == null || original.isEmpty() || toRemove == null || toRemove.isEmpty())
        return original;

    int matchIndex = original.indexOf(toRemove, 0);
    if (matchIndex == -1)
        return original;
    
    return original.substring(0, matchIndex) + original.substring(matchIndex + toRemove.length());
}
```

Anyways, I just realized there's a small error in my previous code. Replace
```
    ...
    StringBuilder buffer = new StringBuilder(original.length());
    int currIndex = 0;
    do {
        buffer.append(original.substring(currIndex, matchIndex));
        currIndex = matchIndex;
        matchIndex = original.indexOf(toRemove, currIndex);
    } while (matchIndex != -1);
    ...
```
with
```
    ...
    StringBuilder buffer = new StringBuilder(original.length());
    int currIndex = 0;
    int removedLength = toRemove.length();
    do {
        buffer.append(original.substring(currIndex, matchIndex));
        currIndex = matchIndex + removedLength;
        matchIndex = original.indexOf(toRemove, currIndex);
    } while (matchIndex != -1);
    ...
```
Or it won't step forward correctly

dbwiddis: After thinking about it, I think I'm fond of your first, simplest, `removeFirstMatchingString()` for this specific application.

For the bigger picture, I'd like a more general key/value parsing method that handles all these specific cases with options to handle how to find the key, delimiter, whitespace, quotes around value, etc.

igreenfield:
I have fixed all the reuse in all the other implementations. Do you mind I will add it to this PR?

dbwiddis:
Sure, go ahead and add it

igreenfield:
Done

PR #399, opened 2017-08-17
No warning when PID file is no longer avail

muryoh:
When initializing LinuxOperatingSystem class, a lookup
is done through PID files to find the more recent one
However some processes may have been already killed
between the PID list is obtained and the time the PID
files are accessed
So explicitly say that no error should be printed when
this happens

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
@@ -119,7 +119,7 @@ public class LinuxOperatingSystem extends AbstractOperatingSystem {
         long youngestJiffies = 0L;
         String youngestPid = "";
         for (File pid : pids) {
-            List<String> stat = FileUtil.readFile(String.format("/proc/%s/stat", pid.getName()));
+            List<String> stat = FileUtil.readFile(String.format("/proc/%s/stat", pid.getName()), false);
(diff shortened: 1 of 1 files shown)


SangheeKim:
Hi. I have a similar issue.
```
java.lang.ArithmeticException: / by zero
	at oshi.software.os.linux.LinuxOperatingSystem.getProcess(LinuxOperatingSystem.java:249) ~[oshi-core-3.4.2.jar:3.4.2]
```
It seems that there is a pid file, but it is an error from split [14]

dbwiddis:
Is this a new issue that wasn't fixed by this PR?

YoshiEnVerde:
We would be better with a full stack trace, but as far as I can tell, this error comes from a misshandling of a possible error within the `init()` method of `LinuxOperativeSystem`:

If the **youngestJiffies** field is set to 0 (which marks an OS-level error when trying to read the list of running processes, we log the error and cancel the **init** method halfway through.

That leaves a few fields set at 0, when they should have some value.
Mainly **bootTime** and **hz**.

Then, at `LinuxOperatingSystem.getProcess()`, there's a division by **hz** that fails.

I'd suggest a small check in that method, where bootTime and hz are used (one time for bootTime, three for hz), to avoid errors (and to log method use when this situation arises)

YoshiEnVerde:
I'll create a separate issue for this and do the PR myself

Issue #413, opened 2017-09-26
Division by zero error when getting process info on Linux


YoshiEnVerde:
There seems to be the possibility of an error when gathering some values from the PID file in Linux.
This ends up with some code in LinuxOperatingSystem, dependant on these values, throwing an ArithmeticException (Division by Zero).

This was reported by @SangheeKim at #399

YoshiEnVerde:
We would be better with a full stack trace, but as far as I can tell, this error comes from a misshandling of a possible error within the `init()` method of `LinuxOperativeSystem`:

If the **youngestJiffies** field is set to 0 (which marks an OS-level error when trying to read the list of running processes, we log the error and cancel the **init** method halfway through.

That leaves a few fields set at 0, when they should have some value.
Mainly **bootTime** and **hz**.

Then, at `LinuxOperatingSystem.getProcess()`, there's a division by **hz** that fails.

I'd suggest a small check in that method, where bootTime and hz are used (one time for bootTime, three for hz), to avoid errors (and to log method use when this situation arises)

YoshiEnVerde:
PR opened on #414

SangheeKim:
Thank you for quickly change.

PR #414, opened 2017-09-26
Add validation for uninitialized fields in LinuxOperatingSystem

YoshiEnVerde:
Added validation on use of fields that might be zero after initialization error for the LinuxOperatingSystem class

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
@@ -247,9 +247,13 @@ public class LinuxOperatingSystem extends AbstractOperatingSystem {
         proc.setPriority(ParseUtil.parseIntOrDefault(split[17], 0));
         proc.setVirtualSize(ParseUtil.parseLongOrDefault(split[22], 0L));
         proc.setResidentSetSize(ParseUtil.parseLongOrDefault(split[23], 0L) * this.memoryPageSize);
-        proc.setKernelTime(ParseUtil.parseLongOrDefault(split[14], 0L) * 1000L / hz);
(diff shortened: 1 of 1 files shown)


dbwiddis:
Looks good to me!

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -247,9 +247,13 @@ public OSProcess getProcess(int pid) {
...
         proc.setVirtualSize(ParseUtil.parseLongOrDefault(split[22], 0L));
         proc.setResidentSetSize(ParseUtil.parseLongOrDefault(split[23], 0L) * this.memoryPageSize);
-        proc.setKernelTime(ParseUtil.parseLongOrDefault(split[14], 0L) * 1000L / hz);
-        proc.setUserTime(ParseUtil.parseLongOrDefault(split[13], 0L) * 1000L / hz);
-        proc.setStartTime(bootTime + ParseUtil.parseLongOrDefault(split[21], 0L) * 1000L / hz);
+        if(hz != 0) {
+            proc.setKernelTime(ParseUtil.parseLongOrDefault(split[14], 0L) * 1000L / hz);
+            proc.setUserTime(ParseUtil.parseLongOrDefault(split[13], 0L) * 1000L / hz);
+            proc.setStartTime(bootTime + ParseUtil.parseLongOrDefault(split[21], 0L) * 1000L / hz);
+        } else {
```
dbwiddis: Actually, on second thought.... if hz == 0 we should explicitly re-set it to its default value of 1000L.

YoshiEnVerde: Easily done! PR updated!

dbwiddis:
Fixes #413

PR #433, opened 2017-12-13
performance improvements for getProcesses() on Linux

bildechinger:
getProcesses() called "stat" to get the userId, user, groupId and group. This call is pretty expensive - especially when done for all processes.
Instead of using stat I changed it to use the userId/groupId from /proc/[pid]/status. 
Unfortunately for username and groupname is not included in status. I added code to get it from passwd and group databases instead. 
This could be a additional feature to add for the future to query information about users (username, userid, home-dir, real name, etc.). I will add a issue to track this.

On my workstation LinuxOperatingSystem.getProcesses(int, ProcessSort) improved from ~1.7s to ~130ms after these changes. 

Let me know what you think and if you want me to change anything.


oshi-core/src/main/java/oshi/util/FileUtil.java
```diff
@@ -69,7 +69,7 @@ private FileUtil() {
      *         list if file could not be read or is empty
      */
     public static List<String> readFile(String filename, boolean reportError) {
-        if (new File(filename).exists()) {
+        if (new File(filename).canRead()) {
```
henryx: Is preferable to split this control in two parts, for a better comprehension of the error

oshi-core/src/main/java/oshi/software/os/linux/LinuxUserInfo.java
```diff
@@ -0,0 +1,37 @@
+package oshi.software.os.linux;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
```
henryx: Instead of a `Map<String, String>`, is not preferrable an object that return the username and the userid? For example:

```
public class User {
    private String name;
    private int id;

    public User(String name, int id) {
        this.name = name;
        this.id = id;
    }
    public String getName() {
        return this.name;
    }

    public int getId() {
        return this.id;
    }
}
```

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -191,9 +191,14 @@ public FileSystem getFileSystem() {
     public OSProcess[] getProcesses(int limit, ProcessSort sort) {
         List<OSProcess> procs = new ArrayList<>();
         File[] pids = ProcUtil.getPidFiles();
+        
+        LinuxUserInfo userInfo = getUserInfo();
```
dbwiddis: Calling the code this way will re-parse the user and group list every time.  However, this information rarely changes in reality.  Is there a way to check when this information is last updated (e.g., timestamp on `/etc/shadow` (if available) or `/etc/passwd`)?  If so, this information could be stored/cached as a class variable and then updated only if the timestamp is changed.   

Alternately you could still cache and trigger a re-read upon an error (finding a UID not in the map) or after a longer time period (e.g., 5 minutes).

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -191,9 +191,14 @@ public FileSystem getFileSystem() {
     public OSProcess[] getProcesses(int limit, ProcessSort sort) {
         List<OSProcess> procs = new ArrayList<>();
         File[] pids = ProcUtil.getPidFiles();
+        
+        LinuxUserInfo userInfo = getUserInfo();
+        Map<String, String> users = userInfo.getUsers();
```
dbwiddis: Style: name this `usersMap` or something more self-documenting.  Also, is the key necessarily a `String` or an `Integer`?  I guess if we're saving ourselves a parsing step a String is okay...

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -256,18 +265,19 @@ public OSProcess getProcess(int pid) {
...
-                proc.setUserID(split[0]);
-                proc.setUser(split[1]);
-                proc.setGroupID(split[2]);
-                proc.setGroup(split[3]);
-            }
+        
+        Map<String, String> status = FileUtil.getKeyValueMapFromFile(String.format("/proc/%d/status", pid), ":");
+        proc.setUserID(ParseUtil.whitespaces.split(MapUtil.getOrDefault(status, "Uid", ""))[0]);
+        proc.setGroupID(ParseUtil.whitespaces.split(MapUtil.getOrDefault(status, "Gid", ""))[0]);
+        if (users == null) {
```
dbwiddis: In the single-PID case, both `users` and `groups` would be `null` and you would end up calling `getUserInfo()` and incurring the object creation overhead twice.  Call it once (if either variable is `null`) similar to how you did in  LinuxOperatingSystem line 195.  Or better yet, don't call this method with null, and call `getUserInfo()` before calling this method.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -207,6 +212,10 @@ public FileSystem getFileSystem() {
      */
     @Override
     public OSProcess getProcess(int pid) {
+    	return getProcess(pid, null, null);
```
dbwiddis: all this really does is pass the `getUserInfo()` call into the next method and require you to null check to figure out that you need to update it.  Just get that info here and pass it so you don't have to null check later.  Better yet, have that info cached by the UserInfo class.

oshi-core/src/main/java/oshi/software/os/linux/LinuxUserInfo.java
dbwiddis: If we wanted to expose the user/group database to the user I'd agree. But as I read the code, this is just an internally cached lookup of UID->user and GID->group, and a `HashMap` is the most efficient way to store that lookup.  Otherwise you'd have to iterate through all your `User` objects looking for a matching `User.id` to match your `uid`.

oshi-core/src/main/java/oshi/software/os/linux/LinuxUserInfo.java
```diff
@@ -0,0 +1,37 @@
...
+public class LinuxUserInfo {
+	
+	protected Map<String, String> getUsers() {
+		Map<String, String> users = new HashMap<>();
+		List<String> passwd = ExecutingCommand.runNative("getent passwd");
+		// see man 5 passwd for the fields
+		for (String entry : passwd) {
+			String[] split = entry.split(":");
+			// it is allowed to have multiple entries for the same userId, we use the first one
+			if (!users.containsKey(split[2])){
```
dbwiddis: do a length check on `split` before you get here.

oshi-core/src/main/java/oshi/software/os/linux/LinuxUserInfo.java
```diff
@@ -0,0 +1,37 @@
...
+		return users;
+	}
+
+	protected Map<String, String> getGroups() {
+		Map<String, String> groups = new HashMap<>();
+		List<String> group = ExecutingCommand.runNative("getent group");
+		// see man 5 group for the fields
+		for (String entry : group) {
+			String[] split = entry.split(":");
+			if (!groups.containsKey(split[2])) {
```
dbwiddis: length check `split`.

oshi-core/src/main/java/oshi/software/os/linux/LinuxUserInfo.java
```diff
@@ -0,0 +1,37 @@
+package oshi.software.os.linux;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import oshi.util.ExecutingCommand;
+
+public class LinuxUserInfo {
```
dbwiddis: I'm on the fence whether we need a whole new class for this.  And if we do, I'm thinking maybe something more like `LinuxDatabaseEntries` would be more appropriate (with an eye toward possibly using this class for other `getent` lookups).

If we're just looking up the data every time, we probably don't need the class, but if this class is going to do some sort of timestamp-based tracking of the cache, it's probably best as its own class.

oshi-core/src/main/java/oshi/software/os/linux/LinuxUserInfo.java
```diff
@@ -0,0 +1,37 @@
...
+	
+	protected Map<String, String> getUsers() {
```
dbwiddis: why `protected`?

oshi-core/src/main/java/oshi/util/FileUtil.java
dbwiddis: I'm actually okay with the combined error here.  This should be an unusual occurrence and the error is verbose enough to permit troubleshooting.

oshi-core/src/main/java/oshi/software/os/linux/LinuxUserInfo.java
bildechinger: Yes, for now this is just a lookup. I have created issue (https://github.com/oshi/oshi/issues/434) to work on more the more general UserInformation feature.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
bildechinger: I like the idea of caching. I was thinking about it when I implemented but I don't know how to find out when the cache gets dirty. However, I think making it dirty after a few minutes is an ok approach for a users/group database that changes rarely and anyways sometimes needs some time to propagate (NIS setup with cached information). I will change it to this approach.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
bildechinger: I used String because in OSProcess userId is a String as well.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
bildechinger: Yes, that makes sense.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
dbwiddis: It's easy to find a dirty cache with a new user -- they won't be in the cache when you look up by UID.  Simply refresh and you're done.  As for a deleted user -- doesn't matter if the extra user hangs around in the cache if you're never looking them up.   The only issue you'll have is a renamed user (is that even possible?) and I think a few-minute cache won't hurt there.

oshi-core/src/main/java/oshi/software/os/linux/LinuxUserInfo.java
dbwiddis: Good that you're thinking ahead.  Whatever class you design for the lookup, perhaps you might think about the structure you'll have for #434 and whether it makes sense to build the skeleton now.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
henryx: For future expansion of class, use of String is correct because in Windows we don't have an integer, but a SID in form "S-1-5-21-3623811015-3361044348-30300820-1013"

oshi-core/src/main/java/oshi/software/os/linux/LinuxUserGroupInfo.java
```diff
@@ -0,0 +1,98 @@
...
+				String gid = split[2];
+				if (!groupsIdMap.containsKey(gid)) {
+					groupsIdMap.put(gid, groupName);
+				}
+			}
+			lastRefreshedTimestamp = System.currentTimeMillis();
+		}
+	}
+	
+	private UsersCache usersCache = new UsersCache();
```
dbwiddis: Style: These two variable declarations need to be at the top of the file.  Might be worth making them final.

bildechinger: Just did it. Thanks for integrating the change!

Issue #442, opened 2018-01-28
Process parsing breaks on Linux when names include spaces


newk5:
I noticed that I can sort the running processes by memory with:

`OperatingSystem.ProcessSort.MEMORY`

I'm under the impression this sorting is done by shared memory (SHR) correct me if I'm wrong about this, but is it possible to sort by Resident Set Size (RSS) instead ? 

Thanks!

dbwiddis:
The `MEMORY` sort is done by RSS.  See the [AbstractOperatingSystem](https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java#L48) class.

newk5:
Oh, I see. Hmm, maybe I'm doing something wrong. I'm currently running ElementaryOS 0.4.1 Loki and I'm using:` OSProcess topMem = os.getProcesses(1, OperatingSystem.ProcessSort.MEMORY)[0]` to to get the top process using the most memory but for some reason it doesn't seem to pick up the right process. Using `htop` if I sort by RES I can see it shows Netbeans at the top. I also  tried using a task manager like utility program and it too shows NetBeans at the top:
![screenshot from 2018-01-29 00 58 00](https://user-images.githubusercontent.com/32560510/35489538-bb0aba3e-048f-11e8-9c75-ce79e583ccdc.png)

But when using oshi, I always get firefox as the top process. Is there something I'm doing wrong?

dbwiddis:
On Linux, OSHI pulls the RSS value from the 24th field of `/proc/pid/stat`.  (The value is in pages.)  Can you check this value for the NetBeans and FireFox processes and see what it says, compared to `htop`?

dbwiddis:
So OSHI should report RSS of 4096 x 408420 = 1.53 GiB for pid 4086 (java/netbeans), and 4096 x 100352 = 392 MiB for pid 2913 (firefox).  

Is it returning the correct memory size for firefox?   Does it even see the netbeans process?

newk5:
I've tried this:

```
List<OSProcess> procs = Arrays.asList(os.getProcesses(10, ProcessSort.MEMORY));

        System.out.println("   PID  %CPU %MEM       VSZ       RSS Name");
        for (int i = 0; i < procs.size() && i < 20; i++) {
            OSProcess p = procs.get(i);
            System.out.format(" %5d %5.1f %4.1f %9s %9s %s%n",
                    p.getProcessID(),
[...]

Issue #440, opened 2018-01-11
Network speed unit wrong on Linux


detyo:
Hi,
  I have been looking into network statistics reported by OSHI on different OS-es and I noticed that on Linux, the network speed is assumed to be in bits per second, when read from `/sys/class/net/<iface>/speed`, but according to Linux [kernel documentation](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-net) the value is in Mbits/sec. 
I tested this on Red Hat Enterprise Linux Server 6.0 (build 2.6.32-71.el6.x86_64) VM, but it seems that on a physical machine the units are the same.
For example, on the mentioned VM the file contains 1000 and OSHI displays this as:

```
Network interfaces:
 Name: eth1 (eth1)
   MAC Address: 00:50:56:9d:79:cd 
   MTU: 1500, Speed: 1 Kbps 
```

while `ethtool` returns: Speed: 1000Mb/s

Is this a bug or perhaps the interface speed unit might differ based on kernel version?

Detelin

dbwiddis:
Probably a bug.  Should be an easy fix!

YoshiEnVerde:
Added the fix for this in my latest PR #443

detyo:
Thanks for fixing it! I was going to do this myself but got distracted with some high-prio tasks.

dbwiddis:
Fixed in #443

PR #443, opened 2018-01-30
Fix Linux OS issues: #440 and #442

YoshiEnVerde:
#440:
Multiplied the network speed by 2^20 to normalize it from MiBs to Bytes.
Also added an overflow check (It shouldn't ever happen though, as that would mean network speeds above 7 Exabytes per second)

#442:
Instead of building a whole new method inside a util class to allow for the correct **/proc/[PID]/stat** file parsing (where the first two delimiters are " (" and ") ", instead of whitespaces), I just used the current file splitting, then concatenated any resulting token that belonged to the **name** field.

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxNetworks.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxNetworks.java
@@ -52,6 +52,7 @@ public class LinuxNetworks extends AbstractNetworks {
         netIF.setPacketsRecv(FileUtil.getUnsignedLongFromFile(rxPacketsPath));
         netIF.setOutErrors(FileUtil.getUnsignedLongFromFile(txErrorsPath));
         netIF.setInErrors(FileUtil.getUnsignedLongFromFile(rxErrorsPath));
-        netIF.setSpeed(FileUtil.getUnsignedLongFromFile(speed));
+        long netSpeed = FileUtil.getUnsignedLongFromFile(speed) * 1024 * 1024;
+        netIF.setSpeed(netSpeed < 0 ?0 :netSpeed);
     }
 }
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
@@ -226,9 +226,16 @@ public class LinuxOperatingSystem extends AbstractOperatingSystem {
         long now = System.currentTimeMillis();
         OSProcess proc = new OSProcess();
         // See man proc for how to parse /proc/[pid]/stat
-        proc.setName(split[1].replaceFirst("\\(", "").replace(")", ""));
+        int nameOffset = 0;
+        StringBuilder name = new StringBuilder(
+                split[1].replaceFirst("\\(", ""));
+        while (!split[1 + nameOffset].endsWith(")")) {
+            nameOffset++;
+            name.append(' ').append(split[1 + nameOffset]);
+        }
+        proc.setName(name.substring(0, name.length() - 1));
         proc.setPath(path);
-        switch (split[2].charAt(0)) {
+        switch (split[2 + nameOffset].charAt(0)) {
         case 'R':
             proc.setState(OSProcess.State.RUNNING);
             break;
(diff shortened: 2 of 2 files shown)


oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -226,9 +226,16 @@ public OSProcess getProcess(int pid) {
...
         // See man proc for how to parse /proc/[pid]/stat
-        proc.setName(split[1].replaceFirst("\\(", "").replace(")", ""));
+        int nameOffset = 0;
+        StringBuilder name = new StringBuilder(
+                split[1].replaceFirst("\\(", ""));
+        while (!split[1 + nameOffset].endsWith(")")) {
+            nameOffset++;
+            name.append(' ').append(split[1 + nameOffset]);
+        }
+        proc.setName(name.substring(0, name.length() - 1));
```
dbwiddis: This all looks good, other than the part of me that thinks we should do some sort of error checking.  While we can probably rely on an actual `/proc/pid/stat` result to be formatted as required, we should check a "just in case" it's an empty string.  Perhaps test the existence of the `(` character in the string?

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -249,14 +256,14 @@ public OSProcess getProcess(int pid) {
...
-        proc.setUserTime(ParseUtil.parseLongOrDefault(split[13], 0L) * 1000L / hz);
-        proc.setStartTime(bootTime + ParseUtil.parseLongOrDefault(split[21], 0L) * 1000L / hz);
+        proc.setParentProcessID(ParseUtil.parseIntOrDefault(split[3 + nameOffset], 0));
+        proc.setThreadCount(ParseUtil.parseIntOrDefault(split[19 + nameOffset], 0));
+        proc.setPriority(ParseUtil.parseIntOrDefault(split[17 + nameOffset], 0));
+        proc.setVirtualSize(ParseUtil.parseLongOrDefault(split[22 + nameOffset], 0L));
+        proc.setResidentSetSize(ParseUtil.parseLongOrDefault(split[23 + nameOffset], 0L) * this.memoryPageSize);
+        proc.setKernelTime(ParseUtil.parseLongOrDefault(split[14 + nameOffset], 0L) * 1000L / hz);
+        proc.setUserTime(ParseUtil.parseLongOrDefault(split[13 + nameOffset], 0L) * 1000L / hz);
+        proc.setStartTime(bootTime + ParseUtil.parseLongOrDefault(split[21 + nameOffset], 0L) * 1000L / hz);
```
dbwiddis: similar to above comment, perhaps ensure `split` length exceeds `21 + nameOffset`

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxNetworks.java
```diff
@@ -52,6 +52,7 @@ public static void updateNetworkStats(NetworkIF netIF) {
         netIF.setPacketsRecv(FileUtil.getUnsignedLongFromFile(rxPacketsPath));
         netIF.setOutErrors(FileUtil.getUnsignedLongFromFile(txErrorsPath));
         netIF.setInErrors(FileUtil.getUnsignedLongFromFile(rxErrorsPath));
-        netIF.setSpeed(FileUtil.getUnsignedLongFromFile(speed));
+        long netSpeed = FileUtil.getUnsignedLongFromFile(speed) * 1024 * 1024;
+        netIF.setSpeed(netSpeed < 0 ?0 :netSpeed);
```
dbwiddis: can you run this through a code formatter?  Wouldn't ask if it was the only change, but I expect some spaces would be added in the next totally unrelated change....

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
YoshiEnVerde: These are god points.
I'll add some better array length and string content checking, plus run the code through a code formatter.

dbwiddis: Other thing I've been thinking is that it may not be necessary to check for the ") " on the name.  We know there will be 52 fields, so if the split length is >52 we can subtract 52 to know how big the offset is (and how many fields to append to the name.)   This could wrap the length check in with a useful determination of nameOffset.  (And then you could piece together the name and strip off first and last characters in one go.)

dbwiddis:
How's this PR coming? :)

PR #446, opened 2018-02-05
Adding getChildProcesses to OperatingSystem #446

jsimomaa:


oshi-core/src/main/java/oshi/software/os/OperatingSystem.java
```diff
@@ -92,6 +92,26 @@
...
+     * @param limit
+     *            Max number of results to return, or 0 to return all results
+     * @param sort
+     *            If not null, determines sorting of results
+     * @return An array of {@link oshi.software.os.OSProcess} objects presenting the
+     *         specified number (or all) of currently running child processes of the provided PID, sorted
+     *         as specified. The array may contain null elements if a process
+     *         terminates during iteration. 
+     */
+    OSProcess[] getChildProcesses(int parentPid, int limit, ProcessSort sort);
```
dbwiddis: Similar to `getProcesses()` can we have a version of this that only takes the `parentPid` and calls this method with a default (non)limit and default (CPU?) sort?

jsimomaa: > Looks great. Can you please add a junit test? I'd suggest getting the regular process list and building a map to find the processes with the most and least child processes; then testing whether the list matches.

I don't fully understand what you mean by _testing whether the list matches_?

YoshiEnVerde: I'm guessing the test would:

1. Use `getProcesses()` to fetch the full list of processes
2. Manually map all child processes to their parents' PID
3. Identify the one process with the least children, and the one process with the most children
4. Call this new code for those two processes
5. See if our manually mapped list of children processes match the ones we get from this method

Is that what you were asking for @dbwiddis?

dbwiddis: Yes... although more of a suggestion than directive of how to do it.   But essentially, my thought is to add a new method to the [OperatingSystemTest](https://github.com/oshi/oshi/blob/master/oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java) class.  This method should:
* find at least one process using the existing `getProcesses()` method with no children, if there are any; then test using the new method to make sure the `getChildProcesses()` call for the same PID returns no child processes.  If this initial test fails due to a race condition (a child process starts or stops between method calls) try again with a different process.
* find at least one process with exactly one child process, if there are any; test using the new method to ensure you get the same child process.   If this initial test fails due to a race condition (a child process starts or stops between method calls) try again with a different process.
* find at least one process with two or more child processes, if there are any; test using the new method to ensure the child process PIDs in the new method match the same child process PIDs from the old method.  If this initial test fails due to a race condition (a child process starts or stops between method calls) try again with a different process.

PR #447, opened 2018-02-13
Add context switches & interrupts

jpbempel:
Add 2 methods in CentralProcessor interface
only Linux implementation provided for now


dbwiddis:
Is this not applicable/available for other OS's?   I'm worried about committing code for only a single OS without a path to trying the others (if possible).  Should also file an issue to track this if possible.

You need to implement the new methods in the JSON version as well.

jpbempel:
Added windows implementation, but cannot test on other systems like Mac/Solaris/FreeBSD
However tips for this can be found in https://github.com/giampaolo/psutil/blob/master/psutil/_psutil_osx.c#L1774

Thanks

Issue #365, opened 2017-05-16
Consolidating all the differing parts of OSHI into a better model


YoshiEnVerde:
I've been thinking on this strongly for a while.
Between issues like #310, and the kind of issues that keep cropping up (specially when asking for new features), I can't help but think that OSHI's model is starting to become a bit of a Frankenstein's monster...

Right now, it's not strange to have somebody ask for a feature only available in a single platform, and OSHI ending up with an extra feature on that platform alone.
In direct opposition, however, the current model will abstract or rename things in some platforms to have them fall in line with another platform.

**As far as I can see, we have a few model/design details that need to be addressed:**

**1. We need a consolidated model across all platforms**
**2. We need to keep the API as far away from the non-consolidated parts as possible**
**3. We need a better way of recovering and updating the system info**

PR #455, opened 2018-02-25
Feature/#454 open files/handles support

spyhunter99:


oshi-core/src/main/java/oshi/software/os/OSProcess.java
```diff
@@ -523,4 +524,24 @@ public void setBytesRead(long bytesRead) {
...
+     * @param count 
+     */
+    public void setOpenFiles(long count){
+        this.handles = count;
+    }
+    
+    /**
+     * Sets the number of open file handles (or network connections) that 
+     * belongs to the process
+     * @since 3.5.0
```
dbwiddis: While this may be good documentation we've not used it at all in the past; also this may be a 3.4.x patch release....

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -262,6 +262,15 @@ public OSProcess getProcess(int pid) {
         proc.setBytesRead(ParseUtil.parseLongOrDefault(MapUtil.getOrDefault(io, "read_bytes", ""), 0L));
         proc.setBytesWritten(ParseUtil.parseLongOrDefault(MapUtil.getOrDefault(io, "write_bytes", ""), 0L));
         
+        //gets the open files count
+        try {
+            String openFilesString = ExecutingCommand.getFirstAnswer(String.format("ls -f /proc/%d/exe | wc -l", pid));
+            if (openFilesString!=null)
+                proc.setOpenFiles(Long.parseLong(openFilesString));
```
dbwiddis: Please use the Util's `parseLongOrDefault()` method to catch any errors here (and elsewhere, won't repeat this comment).  I believe that will avoid having to catch exceptions at this level.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -262,6 +262,15 @@ public OSProcess getProcess(int pid) {
         proc.setBytesRead(ParseUtil.parseLongOrDefault(MapUtil.getOrDefault(io, "read_bytes", ""), 0L));
         proc.setBytesWritten(ParseUtil.parseLongOrDefault(MapUtil.getOrDefault(io, "write_bytes", ""), 0L));
         
+        //gets the open files count
+        try {
+            String openFilesString = ExecutingCommand.getFirstAnswer(String.format("ls -f /proc/%d/exe | wc -l", pid));
```
dbwiddis: Shouldn't this be `/proc/%d/fd`?  See [proc man page](http://man7.org/linux/man-pages/man5/proc.5.html).

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOperatingSystem.java
```diff
@@ -143,9 +145,17 @@ public OSProcess getProcess(int pid) {
...
-        }
+            //gets the open files count
+            try {
+                String openFilesString = ExecutingCommand.getFirstAnswer(String.format("lsof -p %d | wc -l", pid));
+                if (openFilesString!=null)
+                    fproc.setOpenFiles(Long.parseLong(openFilesString));
+            } catch (Exception ex) {
+                LOG.warn("Couldn't find get open file count for pid {}: {}", pid, ex);
+            }
+                // 'top -bm io' gives read/write counts, not bytes
```
dbwiddis: this comment looks out of place, should it be located elsewhere?

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
```diff
@@ -56,11 +56,11 @@
     // For WMI Process queries
     private static String processProperties = "Name,ExecutablePath,CommandLine,ExecutionState,ProcessID,ParentProcessId"
             + ",ThreadCount,Priority,VirtualSize,WorkingSetSize,KernelModeTime,UserModeTime,CreationDate"
-            + ",ReadTransferCount,WriteTransferCount,__PATH,__PATH";
+            + ",ReadTransferCount,WriteTransferCount,__PATH,__PATH,HandleCount";
```
dbwiddis: purely preference, I'd prefer to move the HandleCount before the two process queries, to leave them at the end...

oshi-core/src/main/java/oshi/software/os/OSProcess.java
spyhunter99: ok, what should this return if it's not supported or could not be acquired? -1? 0?

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
spyhunter99: so after `WriteTransferCount`?

dbwiddis: Yep... again, purely preference to make it easier to find those processes at the end.

oshi-core/src/main/java/oshi/software/os/OSProcess.java
dbwiddis: I would suggest either deleting the @since, or updating all the rest of the code to add the @since everywhere it's needed (which is pretty much everywhere).

oshi-core/src/main/java/oshi/software/os/OSProcess.java
```diff
@@ -523,4 +524,22 @@ public void setBytesRead(long bytesRead) {
...
+     * @param count 
+     */
+    public void setOpenFiles(long count){
+        this.handles = count;
+    }
+    
+    /**
+     * Sets the number of open file handles (or network connections) that 
+     * belongs to the process
+     * @return open files or -1 if
```
dbwiddis: -1 if... ? :)

oshi-core/src/main/java/oshi/software/os/OSProcess.java
```diff
@@ -53,6 +53,7 @@
     private long upTime;
     private long bytesRead;
     private long bytesWritten;
+    private long handles;
```
dbwiddis: I think I'd prefer the internal variable name match the getter name.

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -195,6 +196,14 @@ public OSProcess getProcess(int pid) {
         proc.setBytesRead(bytesRead);
         proc.setBytesWritten(bytesWritten);
         proc.setCommandLine(getCommandLine(pid));
+        //gets the open files count
+        try {
+            String openFilesString = ExecutingCommand.getFirstAnswer(String.format("lsof -p %d | wc -l", pid));
+            if (openFilesString!=null)
+                proc.setOpenFiles(Long.parseLong(openFilesString));
```
dbwiddis: another `ParseUtil.parseLongOrDefault()` needed here.

oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
```diff
@@ -73,6 +73,7 @@ public void testOperatingSystem() {
         assertTrue(proc.getStartTime() >= 0);
         assertTrue(proc.getBytesRead() >= 0);
         assertTrue(proc.getBytesWritten() >= 0);
+        assertTrue(proc.getOpenFiles()>= -1);
```
dbwiddis: my code formatter will add a space here.  Wouldn't hold up the commit for this but if you're making other changes, may as well prevent the space being added in the next unrelated commit :)

dbwiddis:
Also don't forget your changelog entry!

dbwiddis:
Annnnd.... did you add the code to retrieve the stat in the JSON version?

spyhunter99:
why is there this much code duplication? probably simpler to have a json utils class or just use jackson's data bind to convert the objects to/from json

dbwiddis:
Yes, we know.  [Sometime in the unspecified future we'd like this fixed](https://github.com/oshi/oshi/issues/365#issuecomment-302506211).  Until then, you're stuck with what I (a non-SDE) hacked together. :)

Issue #459, opened 2018-03-10
Get process metrics by a list of pids


spyhunter99:
It may be more efficient (on windows) to query for a list of known pids instead of one by one. This would add a new method to support querying a list of process related metrics from the os. I believe right now the options are

 - ask for everything up to a given limit with sorting
 - ask for a single pid's metrics

dbwiddis:
Great idea

spyhunter99:
on it. windows is the only case where a specialized implementation seems feasible

dbwiddis:
It seems reasonable to just have an extra class that blends in seamlessly with the other "get all" and "get PID" class; we can put the functionality in the abstract class and just override it in the Windows version.

dbwiddis:
Fixed in #460

PR #460, opened 2018-03-11
#459 adds new methods for querying for a list of specific pids

spyhunter99:


oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
```diff
@@ -150,6 +150,9 @@ public String getManufacturer() {
                 throw new IllegalArgumentException("Unimplemented enum type: " + sort.toString());
             }
         }
+        if (limit == 0)
```
dbwiddis: Good idea here to avoid copying the same list we'll return, but you also want to do this in the case that the list size is less than or equal to the limit, so you might consider moving this down and integrating with the other list logic.   (Also we should consider using a sublist rather than a new list and a bunch of adds.)

oshi-core/src/main/java/oshi/software/os/OperatingSystem.java
```diff
@@ -91,6 +92,17 @@
...
+    /**
+     * Gets information on a currently running processes. This was primarily written
+     * to provide an optimized mechanism for windows based operating systems.
+     *
+     * @param pids
+     *            A collection of process IDs
+     * @return An {@link oshi.software.os.OSProcess} object for the specified
+     *         process ids if it is running
+     */
+    Collection<OSProcess> getProcesses(Collection<Integer> pids);
```
dbwiddis: I don't mind using Collection as a parameter but I'm not sure I like it as a return type rather than Set or List -- which we use most other places in the code.

PR #465, opened 2018-03-13
Shaded jar that includes it's dependencies for Oshi core

lprimak:
This is also needed for Payara because SLF4J isn't OSGi compliant,
This deploys an additional -shaded JAR that includes the transitive dependencies

### DIFF
--- a/oshi-core-shaded/pom.xml
+++ b/oshi-core-shaded/pom.xml
@@ -72,6 +72,7 @@
                                     <includes>
                                         <include>org.slf4j:slf4j-api:</include>
                                         <include>org.slf4j:slf4j-jdk14:</include>
+                                        <include>org.threeten:threetenbp:</include>
                                     </includes>
                                 </artifactSet>
                             </configuration>
--- a/oshi-core-shaded/pom.xml
+++ b/oshi-core-shaded/pom.xml
@@ -1,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Oshi (https://github.com/oshi/oshi)
+
+    Copyright (c) 2010 - 2017 The Oshi Project Team
+
+    All rights reserved. This program and the accompanying materials
+    are made available under the terms of the Eclipse Public License v1.0
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+
+    Maintainers:
+    dblock[at]dblock[dot]org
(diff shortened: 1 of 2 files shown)


dbwiddis:
@lprimak I need your help.  I can't release oshi to nexus with this new module because it doesn't build sources and javadocs.

It looks like there's a switch for the shade plugin to build sources (`createSourcesJar`) but I don't see a switch for Javadocs.  Any idea how to fix this?

lprimak:
Does shaded JAR need sources or javadoc?  Is it not creating ANY javadocs anymore?

dbwiddis:
It is a [Central Repository Requirement to include sources and javadocs](http://central.sonatype.org/pages/requirements.html#supply-javadoc-and-sources), even if they are just dummy files.   Would a basic `package-info.java` file generate a javadoc on build?  Coupled with the `createSourcesJar` that may work.  But I'm out of my depth here.

lprimak:
I do not know to be honest.  Google search yielded nothing.  Is sonatype kicking you out?
I did this some time ago and never had any problems with missing javadoc or sources,
especially for non-primary JARs

dbwiddis:
Sonatype does not let me "close" the staging repository because it fails the check.  I could manually upload dummy files, but that's a pain to do for every release.

[This looks like the same question](http://maven.40175.n5.nabble.com/shade-plugin-and-aggregated-javadocs-td3409253.html) but I don't see if the answer is effective or if anyone else answered it.
I believe turning on the `createSourcesJar` switch in the plugin should solve that part of it, we'd just need to create a javadoc jar.  

Do you have a sample project on Central that has successfully done this that I can look at?

dbwiddis:
This looks like a[nother attempt to answer the question.](https://stackoverflow.com/questions/23563857/including-a-synthetic-file-as-javadoc-artifact-in-a-maven-build)

lprimak:
That looks helpful. If you hit a brick wall I can certainly help you with that.  Maybe a new plugin is helpful or something like that.  Keep me posted

dbwiddis:
Well after a lot of trying to hit it with a sledgehammer like the above links, the answer was elegantly simple.

Including `createSourcesJar` was obviously the trick for sources.   Javadocs ended up easily included by using the javadoc plugin's `includeDependencySources` (limiting to just oshi dependencies).

lprimak:
Thanks for taking this on. I appreciate it.

Issue #488, opened 2018-05-11
Processor count retrieval on Solaris


Kald0:
The Processor count retrieval on Oracle Solaris 11.3 SPARC doesn't work as expected.
When the **CentralProcessor** is retrieved with the following command
_CentralProcessor processor = new SystemInfo().getHardware().getProcessor();_
the following error is displayed in the output:
ERROR oshi.hardware.platform.unix.solaris.SolarisCentralProcessor - Couldn't find logical processor count. Assuming 1.

The issue happens in the _calculateProcessorCounts_ method of the **SolarisCentralProcessor** class.
The command **psrinfo -pv** doesn't return any string that is matched with the _PSRINFO_ pattern.
Here's an example of what the function actually returns:

> The physical processor has 2 cores and 16 virtual processors (0-15)
> The core has 8 virtual processors (0-7)
> The core has 8 virtual processors (8-15)
> sun4v-cpu (chipid 0, clock 2998 MHz)  

The issue can be fixed by using the following pattern:
`.*physical processor has (\\d+) cores and (\\d+) virtual processors.*`

and by changing the matching block a bit:
```
for (String cpu : ExecutingCommand.runNative("psrinfo -pv")) {
    Matcher m = PSRINFO.matcher(cpu.trim());
    if (m.matches()) {
        this.physicalProcessorCount = Integer.parseInt(m.group(1));
        this.logicalProcessorCount = Integer.parseInt(m.group(2));
    }
}
```

Do you have a different output of the psrinfo command on other versions of Solaris ?

dbwiddis:
Thanks for the report, and good to hear this might be the only bug coming from Solaris (so far).  I've developed entirely on a virtual machine so have limited ability to test different hardware.

I based the parsing on the example I saw [here](https://www.thegeekdiary.com/how-to-find-number-of-physicallogical-cpus-cores-and-memory-in-solaris/):
```
# psrinfo -pv
The physical processor has 64 virtual processors (0-63)
  UltraSPARC-T2+ (cpuid 0 clock 1165 MHz)
The physical processor has 64 virtual processors (64-127)
  UltraSPARC-T2+ (cpuid 64 clock 1165 MHz)
```
However, the counting of physical and virtual processors using text parsing was written before I added the kstat code which counts `chip_id` to determine packages, and it's probably a better solution to eliminate all the text parsing and simply use the existing loop to count `chip_id`, `core_id`, and `device_id`.

Can you submit a PR for that?

[...]

PR #494, opened 2018-05-16
Solaris Processor count method change

Kald0:
Updated the SolarisCentralProcessor class to retrieve the processor counts with the kstat command only.
This will fix the #488 issue since we no longer rely on parsing the _psrinfo_ command, whose result depends on the architecture of the machine.


dbwiddis:
Looks great! Thanks!

PR #498, opened 2018-05-23
Added OSFamily enum

cjbrowne:
Sometimes it's useful to be able to distinguish between Windows and Unix operating systems.  This adds that capability to OSHI.

Feedback welcome, this is an early draft of the interface that includes everything I need for my specific use-case, but I'm happy to add or change anything :)

### DIFF
--- a/oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
+++ b/oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
@@ -25,6 +25,7 @@ import java.util.Comparator;
 import java.util.LinkedList;
 import java.util.List;
 
+import oshi.software.os.OSFamily;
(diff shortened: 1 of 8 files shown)


dbwiddis:
Thanks for the contribution.

However, I don't see that this has any improved functionality over [`SystemInfo.getCurrentPlatformEnum()`](http://oshi.github.io/oshi/apidocs/oshi/SystemInfo.html#getCurrentPlatformEnum--).

cjbrowne:
Oh, I didn't know getCurrentPlatformEnum existed.  Thanks!

dbwiddis:
Yeah, it's sneaky up there on the top level, needed to determine which of the other classes is hidden underneath the interface.  There's probably a good argument to send that as an argument when instantiating the `OperatingSystem` class and exposing it there.

PR #501, opened 2018-05-25
Added HWDiskStore.updateDiskStats

cjbrowne:
I used a similar approach to NetworkIF.updateNetworkStats, and implemented the update methods for each platform to the best of my ability.

NB: untested, but the changes are really just *moving* existing code rather than adding anything especially new, so in most cases should be fairly low risk.  I can test it on Windows, Linux, and FreeBSD fairly easily but don't have access to Mac or Solaris for testing those.

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/platform/mac/MacDisks.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/mac/MacDisks.java
@@ -18,30 +18,16 @@
  */
 package oshi.hardware.platform.mac;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.sun.jna.Pointer;
 import com.sun.jna.ptr.IntByReference;
 import com.sun.jna.ptr.PointerByReference;
-
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import oshi.hardware.Disks;
 import oshi.hardware.HWDiskStore;
 import oshi.hardware.HWPartition;
 import oshi.jna.platform.mac.CoreFoundation;
-import oshi.jna.platform.mac.CoreFoundation.CFBooleanRef;
-import oshi.jna.platform.mac.CoreFoundation.CFDictionaryRef;
-import oshi.jna.platform.mac.CoreFoundation.CFMutableDictionaryRef;
-import oshi.jna.platform.mac.CoreFoundation.CFNumberRef;
-import oshi.jna.platform.mac.CoreFoundation.CFStringRef;
+import oshi.jna.platform.mac.CoreFoundation.*;
 import oshi.jna.platform.mac.DiskArbitration;
 import oshi.jna.platform.mac.DiskArbitration.DADiskRef;
 import oshi.jna.platform.mac.DiskArbitration.DASessionRef;
@@ -53,6 +39,8 @@ import oshi.util.MapUtil;
 import oshi.util.ParseUtil;
 import oshi.util.platform.mac.CfUtil;
 import oshi.util.platform.mac.IOKitUtil;
+
+import java.util.*;
 
 /**
  * Mac hard disk implementation.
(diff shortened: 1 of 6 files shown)


oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsDisks.java
```diff
@@ -70,6 +70,31 @@
 
     private static final int BUFSIZE = 255;
 
+    public static void updateDiskStats(HWDiskStore diskStore) {
+        Map<String,List<Object>> vals = WmiUtil.selectObjectsFrom(null, "Win32_DiskDrive",
+                "Name,Manufacturer,Model,SerialNumber,Size,Index",
+                "WHERE SerialNumber = " + diskStore.getSerial(),DRIVE_TYPES);
+        // TODO: null checks
```
dbwiddis: Is this still a TODO?

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsDisks.java
```diff
@@ -70,6 +70,31 @@
 
     private static final int BUFSIZE = 255;
 
+    public static void updateDiskStats(HWDiskStore diskStore) {
+        Map<String,List<Object>> vals = WmiUtil.selectObjectsFrom(null, "Win32_DiskDrive",
```
dbwiddis: I'm not sure we need to read anything from WMI for this update.  None of these values change.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsDisks.java
```diff
@@ -70,6 +70,31 @@
...
+                "Name,Manufacturer,Model,SerialNumber,Size,Index",
+                "WHERE SerialNumber = " + diskStore.getSerial(),DRIVE_TYPES);
+        // TODO: null checks
+        diskStore.setName((String) vals.get("Name").get(0));
+        diskStore.setModel(String.format("%s %s", vals.get("Model").get(0), vals.get("Manufacturer").get(0)).trim());
+        String index = vals.get("Index").get(0).toString();
+        diskStore.setReads(MapUtil.getOrDefault(readMap, index, 0L));
```
dbwiddis: These are the values that change and I don't see where you update the maps.  I think you can delete this WMI call and instead call `populateReadWriteMaps()`.  (Don't think you need `populatePartitionMaps()` ... can any of that change while OSHI is running?)

oshi-core/src/main/java/oshi/hardware/platform/mac/MacDisks.java
```diff
@@ -211,161 +382,7 @@
 
                 // Now look up the device using the BSD Name to get its
```
dbwiddis: this comment belongs with the moved code :)

cjbrowne: True!

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsDisks.java
cjbrowne: That's a good point, I think my mind was somewhere else when I was doing the Windows bit.  I'll have another go in a bit.

cjbrowne:
Yeah the windows stuff is wrong for sure, just a brainfart there.  I'll have another crack at it.

dbwiddis:
Tested the macOS implementation. Works great.  However, you should make the update method with the `DASessionRef` argument private.

dbwiddis:
Solaris works as well.

One other thing needed -- you need to be able to deal with the cases where the disk no longer exists.  I'd suggest making the update method a boolean returning false if the disk isn't found.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsDisks.java
cjbrowne: Looked again and realised they're keyed on the disk name (diskStore.getName() gives me the index into the map, therefore)

dbwiddis: Actually it's not that simple. (I had the first thought at first.)

The Name in the disk is `\\.\PHYSICALDRIVE0` while the name in the stats is `0 C:` or similar.

One possibile fix is to repeat the query but only fetch the name and index.

Alternately, my research indicates that the Names are always consistently `\\.\PHYSICALDRIVEx` where x is the index.  

Alternately again, you will have the partition map attached and can parse the partition name `Disk #0, Partition #0` to get the index.  (I think `getMajor()` might also work.)

cjbrowne: Hmm.  I'm a little hesitant to trust either parsing the name or using the partition map as they are both subject to potential changes in the Win32 API (although that is historically rather stable).  I will repeat the query but fetch only the index, using the SerialNumber as the key since that is probably stable enough.  I can also use this to detect if the drive disappears.

dbwiddis:
Looks good and tests well on mac and SOlaris.  Two final requests:   can you use your IDE to remove unused imports (and explicitly fill out imported classes rather than using *)?  If this is easy, great. If not, I'll just merge and tweak later.   

Also, add a changelog entry!

PR #527, opened 2018-06-21
Correct process information caching and command line retrieval under Windows

dustin-johnson:
For the most part command line retrieval appears to be broken under Windows. When investigating the issue, I found that process caching also appears a bit suspect, so I tweaked that logic a bit as well. Since the changes are concise, I'll let the diff speak for the details.


dbwiddis:
D’oh!

PR #542, opened 2018-07-05
Disabled Windows performance collection leads to empty Process cache

MarcMil:
… to an empty cache

The empty cache caused an empty process list.


dbwiddis:
Thanks for this fix!  A question -- if the performance data is disabled is there any other backup source for the same information (e.g., WMI) or does WMI draw from the same source eventually (I think it does)?

dbwiddis:
Also is this a one-time fix of an empty cache, or does the disabled performance data cause the cache to be cleared every time?

dbwiddis:
The fix does solve one problem, but it creates some inefficiencies.  The cache is emptied at the end of `updateProcessMapFromRegistry()` in a one-by-one manner (vs. `clear()`) and then entirely rebuilt.

It may be preferable to try to preserve some of the objects in the cache in your case, although admittedly it may not be worth the effort for this edge case.

PR #559, opened 2018-07-19
Prevent skipped interfaces in the event of a SocketException.

cilki:
Moving the try/catch into the loop body ensures every interface is given
a chance.

Also this change makes it possible to alter
NetworkIF.setNetworkInterface to throw SocketException rather than
consuming it. Since that method is technically part of the API (although
it shouldn't be), I didn't mess with it.


cilki:
I hope you guys don't mind the PRs without issues first. It's just small things I've noticed while working on the redesign.

dbwiddis:
I'd rather have a PR with no issue than an Issue with no PR!  😄 

PR-only is perfectly fine when the diff is self-explanatory.

PR #600, opened 2018-08-30
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these [code quality badges](https://lgtm.com/projects/g/oshi/oshi/ci/#badges) to your project. They will indicate a very high code quality to potential users and contributors.
You can also check the alerts discovered by [LGTM](https://lgtm.com/projects/g/oshi/oshi).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,9 @@
 [![SonarQube Maintainability](https://sonarcloud.io/api/project_badges/measure?project=com.github.oshi%3Aoshi-parent&metric=sqale_rating)](https://sonarcloud.io/dashboard?id=com.github.oshi%3Aoshi-parent)
 [![SonarQube Reliability](https://sonarcloud.io/api/project_badges/measure?project=com.github.oshi%3Aoshi-parent&metric=reliability_rating)](https://sonarcloud.io/dashboard?id=com.github.oshi%3Aoshi-parent)
 [![SonarQube Security](https://sonarcloud.io/api/project_badges/measure?project=com.github.oshi%3Aoshi-parent&metric=security_rating)](https://sonarcloud.io/dashboard?id=com.github.oshi%3Aoshi-parent)
-[![Say Thanks!](https://img.shields.io/badge/Say%20Thanks-!-1EAEDB.svg)](https://saythanks.io/to/dbwiddis)
 [![Code Quality: Java](https://img.shields.io/lgtm/grade/java/g/oshi/oshi.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/oshi/oshi/context:java)
 [![Total Alerts](https://img.shields.io/lgtm/alerts/g/oshi/oshi.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/oshi/oshi/alerts)
+[![Say Thanks!](https://img.shields.io/badge/Say%20Thanks-!-1EAEDB.svg)](https://saythanks.io/to/dbwiddis)
 
 OSHI is a free JNA-based (native) Operating System and Hardware Information library for Java.
 It doesn't require the installation of any additional native libraries and aims to provide a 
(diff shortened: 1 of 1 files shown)


dbwiddis:
Yay, buttons!

Added, and appreciate another perspective on code quality.  Sonar is pretty thorough but you picked up a few things they missed.

One annoyance is that I have to litter my source code with `// lgtm` to suppress alerts that I'm not interested in.   And I'm not really sure passing an `int` to a method that takes a `long` arg really should be flagged as an error, or result in me getting a `B` rather than an `A`.

xcorail:
Thanks for the feedback. Really appreciate your taking the time.

You are not obliged to put `//lgtm` all over the place ... It's only if you want to suppress alerts on a specific part of the code ... but it you want to exclude one alert that doesn't matter to you, then do it in the [configuration](https://lgtm.com/help/lgtm/showing-hiding-query-results) of your project. It will affect all the code, and is not intrusive.

Another interesting thing concerning your feedback is that our rating of Error / Warning ... and of the severity of an alert is based on the analysis of the behaviour of the hundreds of thousands of developers on open source projects: basically, if an alert get always or often fixed, it will raise up the list, if an alert seems to be ignored by developers, it will be ranked down

Cheers

sjvs:
Hi @xcorail and @dbwiddis,

I hope you don't mind me chipping in here — I'm one of @xcorail's colleagues working on LGTM.com. Glad to hear that you found some of the results useful :slightly_smiling_face:!

@dbwiddis wrote:
> And I'm not really sure passing an int to a method that takes a long arg really should be flagged as an error [...]

I assume you're referring to [those 10 alerts](https://lgtm.com/projects/g/oshi/oshi/alerts/?mode=tree&ruleFocus=7900075). The alert message is indeed perhaps a bit confusing, but the alerts are genuine. For example, consider [this alert](https://lgtm.com/projects/g/oshi/oshi/snapshot/23660c30ca37a8178e69c1835c20c7c49c4300c3/files/oshi-core/src/main/java/oshi/jna/platform/windows/Cfgmgr32Util.java?#x9e841321f31c7df7:1) that flagged up the following code:

```
new Memory((pulLen.getValue() + 1) * charToBytes);
```

Your `Memory` constructor takes a `long`; both `getValue()` and `charToBytes` are integers. The multiplication of the two `int` values results in an integer, and that operation is performed *before* the implicit conversion to `long`. Therefore, if the multiplication overflows, the *already overflown* value will be cast to long, and passed to the `Memory` constructor. Of course there may be semantic reasons why the operands to the multiplication will never cause an overflow; that's hard for me (or LGTM) to tell without further knowledge of your code. However, the following fix will make sure such an overflow can never happen:

```
new Memory((pulLen.getValue() + 1) * (long)charToBytes);
```

The [documentation for that alert](https://lgtm.com/rules/7900075/) explains this in more detail. I'll make sure the alert message gets updated to make sure it's more informative :slightly_smiling_face:.

Give me a shout if you have any further questions/comments/suggestions!

dbwiddis:
Thanks... hadn't found the config area.   But a common use case for me is to want to ignore specific cases in specific files like the above (by the way changing the type of `charToBytes` would be a better option since that alert occurs 3 times :) ) but allow the system to alert me on new occurrences so I could make an individual decision on each line... again, something that can be done with comments but I'd prefer to do on the web, like an issue that I can close.    I'll still play with it to see what's preferred, of course.  Or I can just fix the alerts.

I like the self-learning nature you describe.

There is one alert I'm not quite understanding, the "always returns same value" on [FreeBsdCentralProcessor line 222](https://lgtm.com/projects/g/oshi/oshi/snapshot/23660c30ca37a8178e69c1835c20c7c49c4300c3/files/oshi-core/src/main/java/oshi/hardware/platform/unix/freebsd/FreeBsdCentralProcessor.java?sort=name&dir=ASC&mode=heatmap&showExcluded=false#L222).  Is it alerting on the `Libc.CP_USER * Libc.UINT64_SIZE` multiplication (which I prefer for readability, and the compiler will pre-multiply anyway)?  If so why aren't the next 5 lines also alerting?

xcorail:
Hey @dbwiddis 
I saw that you got an answer on the LGTM forum.
Ping me if you still have questions

PR #605, opened 2018-09-03
OSGi imports missing

swimmesberger:
Due to the change in https://github.com/oshi/oshi/pull/464 by @lprimak some OSGi imports where missing. I have readded them and deliberately made com.sun.management optional because in some OSGi environments this API is not exposed on default. oshi can be used without this package being imported as we can see in https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/hardware/common/AbstractCentralProcessor.java there is a graceful fallback when "com.sun.management.OperatingSystemMXBean" can't be used.

### DIFF
--- a/oshi-core-shaded/pom.xml
+++ b/oshi-core-shaded/pom.xml
@@ -101,6 +101,19 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
+				<groupId>biz.aQute.bnd</groupId>
+				<artifactId>bnd-maven-plugin</artifactId>
+				<version>3.5.0</version>
+				<configuration>
+				<bnd><![CDATA[
+					Export-Package: oshi.*;-noimport:=true
+					Import-Package: com.sun.management.*;optional:=true, !org.slf4j.*, *
+					Bundle-SymbolicName: ${project.groupId}.${project.artifactId}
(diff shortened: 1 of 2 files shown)


swimmesberger:
Considering the comment of @lprimak in https://github.com/oshi/oshi/pull/465:
The latest sl4jf-api artifact on maven (1.7.25) defines OSGi metadata:
`Export-Package: org.slf4j;version=1.7.25, org.slf4j.spi;version=1.7.25
 , org.slf4j.helpers;version=1.7.25, org.slf4j.event;version=1.7.25`
Therefore if the slf4j-api artifact exports the org.slf4j.* packages correctly we should correctly import these packages in oshi and not simply ignore them in the base oshi-core bundle (not talking about the shaded artifact). In the oshi-shaded bundle (though I'm not sure why it's still needed considering that 1.7.25 slf4j-api bundle artifact seems osgi compatible to me) we have to switch to a different Import-Package statement which does not import the org.sl4j.* package because it's bundled.

swimmesberger:
I made some changes to the oshi-shaded pom so that the org.slf4j package (which is included) is not imported for this artifact.

lprimak:
Thank you!

PR #629, opened 2018-09-25
Implement QueueLength in HWDiskStore and populate data in Linux, Free…

Space2Man:
…BSD, Solaris and Windows.

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/HWDiskStore.java
+++ b/oshi-core/src/main/java/oshi/hardware/HWDiskStore.java
@@ -54,6 +54,7 @@ public class HWDiskStore implements Serializable, Comparable<HWDiskStore> {
     private long readBytes;
     private long writes;
     private long writeBytes;
+    private long queueLength;
     private long transferTime;
     private HWPartition[] partitions;
     private long timeStamp;
@@ -84,6 +85,8 @@ public class HWDiskStore implements Serializable, Comparable<HWDiskStore> {
      *            Number of writes to the disk
      * @param writeBytes
      *            Number of bytes written to the disk
+     * @param queueLength
+     *            Number of I/O operations currently in progress
      * @param transferTime
      *            milliseconds spent reading or writing to the disk
      * @param partitions
@@ -92,7 +95,7 @@ public class HWDiskStore implements Serializable, Comparable<HWDiskStore> {
      *            milliseconds since the epoch
      */
     public HWDiskStore(String name, String model, String serial, long size, long reads, long readBytes, long writes,
-            long writeBytes, long transferTime, HWPartition[] partitions, long timeStamp) {
+                       long writeBytes, long queueLength, long transferTime, HWPartition[] partitions, long timeStamp) {
         setName(name);
         setModel(model);
         setSerial(serial);
@@ -101,6 +104,7 @@ public class HWDiskStore implements Serializable, Comparable<HWDiskStore> {
         setReadBytes(readBytes);
         setWrites(writes);
         setWriteBytes(writeBytes);
+        setQueueLength(queueLength);
         setTransferTime(transferTime);
         setPartitions(partitions);
         setTimeStamp(timeStamp);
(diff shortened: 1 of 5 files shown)


oshi-core/src/main/java/oshi/hardware/HWDiskStore.java
```diff
@@ -197,6 +201,13 @@ public long getWriteBytes() {
         return this.writeBytes;
     }
 
+    /**
+     * @return the lenght of the disk queue (#I/O's in progress)
```
dbwiddis: typo :)

dbwiddis: Is this instantaneous or average?

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisDisks.java
```diff
@@ -52,6 +52,7 @@ public static boolean updateDiskStats(HWDiskStore diskStore) {
             diskStore.setWrites(data.writes);
             diskStore.setReadBytes(data.nread);
             diskStore.setWriteBytes(data.nwritten);
+            diskStore.setQueueLength(data.rlentime / 1000000L);
```
dbwiddis: Reading the docs this is cumulative run time, which is not the same as queue length.   Should you not use `wcnt` (count of elements in wait state) or `rcnt` (count of elements in run state)?  I'm not sure this stat is consistent with the values you're getting from other OS's

oshi-core/src/main/java/oshi/hardware/HWDiskStore.java
```diff
@@ -119,24 +123,24 @@ public HWDiskStore(String name, String model, String serial, long size, long rea
...
-        case SOLARIS:
-            diskFound = SolarisDisks.updateDiskStats(this);
-            break;
-        case FREEBSD:
-            diskFound = FreeBsdDisks.updateDiskStats(this);
-            break;
-        default:
-            LOG.error("Unsupported platform. No update performed.");
-            break;
+            case WINDOWS:
```
dbwiddis: any chance you can change your formatter to not indent here?

Space2Man:
Hi, I am not sure about Solaris. For Linux (see [here](https://www.kernel.org/doc/Documentation/iostats.txt) it's 

```
Field  9 -- # of I/Os currently in progress
    The only field that should go to zero. Incremented as requests are
    given to appropriate struct request_queue and decremented as they finish.
```

For FreeBSD you are parsing iostat anyway so it's qlen (see [here](https://www.freebsd.org/cgi/man.cgi?iostat))

`qlen	   transactions	queue length`

And for Windows it's described [here](https://blogs.technet.microsoft.com/askcore/2012/03/16/windows-performance-monitor-disk-counters-explained/):

```
Current Disk Queue Length
Current Disk Queue Length is a direct measurement of the disk queue present at the time of the sampling. 
```

But for Solaris I am not sure, if it's wcnt or rcnt or wcnt+rcnt. According to [here](https://docs.oracle.com/cd/E23824_01/html/821-1451/spmonitor-4.html) it looks like wcnt is translated as queue length. My feeling says it might be wcnt+rcnt.

What's your view on this?

dbwiddis:
I expect rcnt to be either 0 or 1, while wcnt is only a positive number if rcnt is 1.  It's like going to a supermarket and asking how long the checkout line is.  The rcnt value is whether someone is actively checking out, and wcnt is those people waiting.   Essentially, "everyone being processed" is rcnt+wcnt and "those waiting" is just wcnt.

All that to say... match it up with what the other stats report!

Space2Man:
This is how OpenIndiana is doing it ([kr_iostat.c](https://github.com/OpenIndiana/oi-userland/blob/1ed47a2077bfb7e671c15a56afc3d9451fba23d7/components/sysutils/net-snmp/sun/agent/modules/healthMonitor/kr_iostat.c)):

```
	avw	= (double)DISK_DELTA(wlentime) / hr_etime;
		/* average number of transactions waiting */

	avr	= (double)DISK_DELTA(rlentime) / hr_etime;
		/* average number of transactions running */
```

Space2Man:
The problem is, for Linux and FreeBSD  those are average values (I would guess) but Solaris we only have wcnt (which is only the current value) or the wlentime (which is accumulative since boot if I read that correctly). So we  would need a delta calculation / time diff ...

dbwiddis:
Welcome to the world of cross-platform coding.   :)       What does the user really need?

Space2Man:
Or we need to parse iostat on Solaris, which does the average handling ...

dbwiddis:
The bigger picture is to ask what's valuable from a monitoring perspective.  Cumulative stats are generally good (and you can get the difference to calculate rate).  Instantaneous stats (queue length) are less useful.

Space2Man:
pfff ... just read that stuff again ... on Windows "Current Disk Queue Length" also seems to just be current value. And on Linux as well:

```
Field  9 -- # of I/Os currently in progress
    The only field that should go to zero. Incremented as requests are
    given to appropriate struct request_queue and decremented as they finish.
```

For the cumulative value I think Field 11 makes more sense:

```
Field 11 -- weighted # of milliseconds spent doing I/Os
    This field is incremented at each I/O start, I/O completion, I/O
    merge, or read of these stats by the number of I/Os in progress
    (field 9) times the number of milliseconds spent doing I/O since the
    last update of this field.  This can provide an easy measure of both
    I/O completion time and the backlog that may be accumulating.
```

Need to check if this kind of stat is also available on the other OS.

Space2Man:
pfff ... difficult if you don't have any test systems. But I think the instantaneous values deliver no real value. Either it has to be cumulative or averaged (for the last x seconds) ...

Issue #625, opened 2018-09-20
Sound card(s) information


BilalAM:
Information regarding audio card would be a useful addition.
`/proc/asound/cardX` gives information about the sound cards , X is the card number , it depends how many cards do you have . 
`cat /proc/asound/cards` gives me this 
`0 [HDMI           ]: HDA-Intel - HDA Intel HDMI
                      HDA Intel HDMI at 0xd1510000 irq 53`
` 1 [PCH            ]: HDA-Intel - HDA Intel PCH
                      HDA Intel PCH at 0xd1514000 irq 52`

Further information about these can be found in the `codec#0` file.
Interestingly `alsamixer` utility is also using `/proc/asound` to list audio cards and information.

Can this be a useful addition ?

cilki:
Yes, but it will have to wait until we decide on the new design of the next version. There's no sense in adding something new to an old architecture.

BilalAM:
We can still add it though , the new design will take some time .

dbwiddis:
There's no sense waiting until we decide if the decision is going to stretch years. ;)   

This seems a relatively simple implementation (Linux listed above, Windows has Win32_SoundDevice, mac has `system_profiler SPAudioDataType` or possibly registry, and I'm sure we could parse dmidecode for other OS's) and I'd welcome a PR if someone else wants to do it. 

If they want me to do it, then it will probably wait until the new version, and then some. :)

The bigger question is: what (cross-platform) information do we want to expose?  Do we want a specific SoundCard class or perhaps we should roll up miscellaneous hardware like Sound cards, GPUs (another old request hanging out there), etc. in a "Peripherals" or "IO" or similar.  

Again... I'd welcome a PR on this if you'd really like this capability in the current version.

BilalAM:
I can try to make this , the information I managed to discover related to the sound cards were not much in detail, only the number of cards , name , vendor and the type (Like the output shown above) .
GPUs are a bit difficult as it is vendor dependent and non standard , the closest I can get is by running this command I found posted at stackoverflow
`lspci  -v -s  $(lspci | grep ' VGA ' | cut -d" " -f 1)` 
BUT it is also written that it doesn't really show the actual memory of the GPU , for that we need to have vendor dependent libraries installed ...

dbwiddis:
For Mac OS x, parse this output (you can probably load a test file to implement, then replace with command line for final commit):

Command: `system_profiler SPAudioDataType`

Output:
```
Audio:

    Devices:

[...]

PR #630, opened 2018-09-26
SoundCard feature 

BilalAM:
Have made the SoundCard feature to get the information related to the sound cards.
Currently I have made this for **Linux** only , will make it for Windows once this is approved.
I have tested this on **Fedora 27** and **Manjaro 17.1.12.**
It retrieves the information from the **_asound_** folder in **_proc_**.
I also tried to follow the architecture of oshi (a SoundCard interface with an abstract class with specific OS dependent classes in the relevant packages).
The **SystemInfotest** main method can be run to see the results .
Again..this is for linux only (currently until this is approved).


BilalAM:
I don't know why the Travis CI build is failing  , I have removed the lambda expressions also but Travis CI build is failing on the SoundCardTest class at line 40.
The Travis CI log shows this :- 
**[INFO] Running oshi.hardware.SoundCardTest
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.008 s <<< FAILURE! - in oshi.hardware.SoundCardTest
[ERROR] testSoundCards(oshi.hardware.SoundCardTest)  Time elapsed: 0.007 s  <<< ERROR!
java.lang.NullPointerException
	at oshi.hardware.SoundCardTest.testSoundCards(SoundCardTest.java:40)**
Even though I have ran it several times on my machine and it is perfectly running fine on Fedora and Manjaro too without giving any Null pointer exceptions on the SoundCardTest class..

dbwiddis:
There's a plugin for IntelliJ called [Eclipse Code Formatter](https://plugins.jetbrains.com/plugin/6546-eclipse-code-formatter) that you can use OSHI's format file.  If you can't get it to work, no worries, I'll just do a format commit after merging.

The license headers are written using the [License Maven Plugin](https://www.mojohaus.org/license-maven-plugin/).  You can probably run [`mvn license:update-file-header`](https://www.mojohaus.org/license-maven-plugin/update-file-header-mojo.html) to fix them yourself.  Or, leave them blank and that's something else I can fix in a single commit.   Bottom line: either copy another file's header exactly, or leave it blank.

The NPE is in the iteration line.  Haven't yet dug deep enough to figure out where it is but I'd suspect something with listing sound card files and handling null there.

BilalAM:
`mvn test` is showing no build errors on my side.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxSoundCard.java
```diff
@@ -0,0 +1,144 @@
...
+        /**
+         * Method to find all the card folders contained in the <b>asound</b> folder denoting the cards
+         * currently contained in our machine.
+         *
+         * @return : A list of files starting with 'card'
+         */
+        private static List<File> getCardFolders() {
+                File cardsDirectory = new File(SC_PATH);
+                List<File> cardFolders = new ArrayList<>();
+                for (File card : Objects.requireNonNull(cardsDirectory.listFiles())) {
```
dbwiddis: I think this is where your NPE is being generated.  `requireNonNull()` throws NPE if the argument is null.  You would have to catch this exception and skip the iteration (or better yet have a temporary variable for the file list and null check it!).

BilalAM: And also I will fix the rest of the things first thing tomorrow morning at the earliest (some unused imports and debugging code and also the format)

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxSoundCard.java
```diff
@@ -57,10 +56,14 @@ public LinuxSoundCard(String kernelVersion, String name, String codec) {
...
-                for (File card : Objects.requireNonNull(cardsDirectory.listFiles())) {
-                        if (card.getName().startsWith(CARD_FOLDER) && card.isDirectory()) {
-                                cardFolders.add(card);
+                try {
+                        for (File card : cardsDirectory.listFiles()) {
+                                if (card.getName().startsWith(CARD_FOLDER) && card.isDirectory()) {
+                                        cardFolders.add(card);
+                                }
                         }
+                }catch (Exception e){
```
dbwiddis: Travis builds now!  I hope this was just a debugging step to confirm, because catching all exceptions here is a bad idea and even catching NPE's that you could avoid by explicitly checking null seems like a code smell.  Can we just do:
```
foo = cardsDirectory.listFiles();
if (foo != null) { for (File card : foo) { ... } }
```

BilalAM: Yes I did this for testing travis ci .
In response to your code suggestion on the else statement I just simply log it like "error: no files found! " or something like that and return the empty list ?
Sorry for this as it is a bit tedious to debug considering the only way to see if it works is by committing and waiting for travis's results.

dbwiddis: Thought it might be a test. No worries.  

A log message (warm level or lower) seems fine for the case where there are no audio cards.

BilalAM:
Build complete , also changed the SystemInfoTest class to include all features like it was before.

BilalAM:
Thank you :)
And yes I am planning to do other OS's , I have researched a bit for windows , it seems win32_sounddevice doesnt really give you info about drivers but win32_pnpdrivers do so I'll have to do tests first and then I will make a pull request !

dbwiddis:
Sounds good!  #635 will fix the formatting/headers and add the JSON implementation, I'll merge it soon and you should be good to go for next OS's.  Continuing discussion on #630.

dbwiddis:
Er, continuing on #625...

Issue #632, opened 2018-09-27
NullPointerException on WindowsHardwareAbstractionLayer.getProcessor


Kald0:
Hi,

I have a NullPointerException when calling the following code on Windows Server 2016, using the latest release of oshi:
```
SystemInfo si = new SystemInfo();
OperatingSystem os = si.getOperatingSystem();
CentralProcessor processor = si.getHardware().getProcessor();
```

The stacktrace:

> WARN oshi.util.platform.windows.PerfDataUtil - Failed to add PDH Counter : \System\Context Switches/sec, Error code: 0xC0000BB8
> WARN oshi.hardware.platform.windows.WindowsCentralProcessor - Unable to enumerate performance counter instances for Processor.
> Exception in thread "Thread-0" java.lang.NullPointerException
>     at oshi.util.platform.windows.PerfDataUtil.removeAllCounters(PerfDataUtil.java:227)
>     at oshi.hardware.platform.windows.WindowsCentralProcessor.initPdhCounters(WindowsCentralProcessor.java:246)
>    at oshi.hardware.platform.windows.WindowsCentralProcessor.<init>(WindowsCentralProcessor.java:125)
>    at oshi.hardware.platform.windows.WindowsHardwareAbstractionLayer.getProcessor(WindowsHardwareAbstractionLayer:64)

oshi can't retrive PDH counters and try to fall back to Wmi. To do that, it calls
`PerfDataUtil.removeAllCounters(PROCESSOR)` at _WindowsCentralProcessor:246_
The problematic lines:
```
HANDLEByReference query = queryMap.get(queryKey);
PDH.PdhCloseQuery(query.getValue());
```
In my case I think that the first call to `PerfDataUtil.createCounter(PROCESSOR, ...` failed: the _PROCESSOR_ key was never added to the _queryMap_ map and we end up calling `null.getValue()`.

This should do the trick:
```
HANDLEByReference query = queryMap.get(queryKey);
if (query != null) {
    PDH.PdhCloseQuery(query.getValue());
}
```

Should I submit a PR or am I missing something ?

dbwiddis:
Yep, looks like I missed that.  I recall finding a similar mistake elsewhere, so there were probably two similar bugs and I only caught one of them.  Please submit a PR!

dbwiddis:
Fixed in #634

PR #634, opened 2018-09-28
#632 removeAllCounters() - Null query check

Kald0:
Make sure that the **PdhQuery** is not null before closing it in the _removeAllCounters()_ method.
This behavior can happen when the **PdhCounter** retrieval fails on the first counter: the **PdhQuery** was never opened/added to the _queryMap_ in the first place.


Kald0:
This solves the #632 issue.

PR #639, opened 2018-10-03
Implement QueueLength metric in HWDiskStore (macOS does not provide t…

Space2Man:
…hat value -> always 0)

The following values are returned. They only show the current/instantaneous queue length:

- Linux (see linux/Documentation/iostat.txt):

> Field  9 -- # of I/Os currently in progress
>     The only field that should go to zero. Incremented as requests are
>     given to appropriate struct request_queue and decremented as they finish.

- Solaris (see kstat in JNA or man(3) kstat) - sum of wcnt + rcnt

> uint_t	  wcnt;		    /* count of	elements in wait state */
> uint_t	  rcnt;		    /* count of	elements in run	state */
 
- FreeBSD (see man iostat -Ix):

> qlen	   transactions	queue length

- Windows (see [https://blogs.technet.microsoft.com/askcore/2012/03/16/windows-performance-monitor-disk-counters-explained/](https://blogs.technet.microsoft.com/askcore/2012/03/16/windows-performance-monitor-disk-counters-explained/):

> Current Disk Queue Length
> Current Disk Queue Length is a direct measurement of the disk queue present at the time of the sampling. 

- on MacOS this metric is not defined or at least I could not find it

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/HWDiskStore.java
+++ b/oshi-core/src/main/java/oshi/hardware/HWDiskStore.java
@@ -54,6 +54,7 @@ public class HWDiskStore implements Serializable, Comparable<HWDiskStore> {
     private long readBytes;
     private long writes;
     private long writeBytes;
+    private long queueLength;
     private long transferTime;
     private HWPartition[] partitions;
     private long timeStamp;
@@ -62,7 +63,7 @@ public class HWDiskStore implements Serializable, Comparable<HWDiskStore> {
      * Create an object with empty/default values
      */
     public HWDiskStore() {
-        this("", "", "", 0L, 0L, 0L, 0L, 0L, 0L, new HWPartition[0], 0L);
+        this("", "", "", 0L, 0L, 0L, 0L, 0L, 0L, 0L, new HWPartition[0], 0L);
     }
 
     /**
@@ -84,6 +85,8 @@ public class HWDiskStore implements Serializable, Comparable<HWDiskStore> {
      *            Number of writes to the disk
      * @param writeBytes
      *            Number of bytes written to the disk
+     * @param queueLength
+     *            Number of I/O operations currently in progress
      * @param transferTime
      *            milliseconds spent reading or writing to the disk
      * @param partitions
@@ -92,7 +95,7 @@ public class HWDiskStore implements Serializable, Comparable<HWDiskStore> {
      *            milliseconds since the epoch
      */
     public HWDiskStore(String name, String model, String serial, long size, long reads, long readBytes, long writes,
-            long writeBytes, long transferTime, HWPartition[] partitions, long timeStamp) {
+            long writeBytes, long queueLength, long transferTime, HWPartition[] partitions, long timeStamp) {
         setName(name);
         setModel(model);
         setSerial(serial);
@@ -101,6 +104,7 @@ public class HWDiskStore implements Serializable, Comparable<HWDiskStore> {
         setReadBytes(readBytes);
         setWrites(writes);
         setWriteBytes(writeBytes);
+        setQueueLength(queueLength);
         setTransferTime(transferTime);
         setPartitions(partitions);
         setTimeStamp(timeStamp);
@@ -198,6 +202,13 @@ public class HWDiskStore implements Serializable, Comparable<HWDiskStore> {
     }
 
     /**
+     * @return the length of the disk queue (#I/O's in progress)
+     */
+    public long getQueueLength() {
+        return this.queueLength;
+    }
+
+    /**
      * @return the milliseconds spent reading or writing
      */
     public long getTransferTime() {
@@ -281,6 +292,12 @@ public class HWDiskStore implements Serializable, Comparable<HWDiskStore> {
     public void setWriteBytes(long writeBytes) {
         this.writeBytes = writeBytes;
     }
+
+    /**
+     * @param queueLength
+     *            the length of the disk queue (#I/O's in progress) to set
+     */
+    public void setQueueLength(long queueLength) { this.queueLength = queueLength; }
 
     /**
      * @param transferTime
--- a/oshi-core/src/main/java/oshi/hardware/platform/unix/freebsd/FreeBsdDisks.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/unix/freebsd/FreeBsdDisks.java
@@ -66,6 +66,7 @@ public class FreeBsdDisks implements Disks {
             // In KB
             diskStore.setReadBytes((long) (ParseUtil.parseDoubleOrDefault(split[3], 0d) * 1024));
             diskStore.setWriteBytes((long) (ParseUtil.parseDoubleOrDefault(split[4], 0d) * 1024));
+            diskStore.setQueueLength((long) (ParseUtil.parseDoubleOrDefault(split[5], 0d)));
             // In seconds, multiply for ms
             diskStore.setTransferTime((long) (ParseUtil.parseDoubleOrDefault(split[6], 0d) * 1000));
             diskStore.setTimeStamp(timeStamp);
@@ -106,6 +107,7 @@ public class FreeBsdDisks implements Disks {
             // In KB
             store.setReadBytes((long) (ParseUtil.parseDoubleOrDefault(split[3], 0d) * 1024));
             store.setWriteBytes((long) (ParseUtil.parseDoubleOrDefault(split[4], 0d) * 1024));
+            store.setQueueLength((long) (ParseUtil.parseDoubleOrDefault(split[5], 0d)));
             // In seconds, multiply for ms
             store.setTransferTime((long) (ParseUtil.parseDoubleOrDefault(split[6], 0d) * 1000));
             store.setTimeStamp(timeStamp);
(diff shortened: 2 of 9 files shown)


oshi-json/src/main/java/oshi/json/hardware/HWDiskStore.java
```diff
@@ -73,6 +73,8 @@
      *            Number of writes to the disk
      * @param writeBytes
      *            Number of bytes written to the disk
+     * @param queueLength
+     *            Number of I/O outstanding operations (waiting + running)
```
dbwiddis: More a question than a comment. Have you confirmed this interpretation (wait + run) is valid on all OS's?

oshi-core/src/main/java/oshi/hardware/HWDiskStore.java
```diff
@@ -197,6 +201,13 @@ public long getWriteBytes() {
         return this.writeBytes;
     }
 
+    /**
+     * @return the length of the disk queue (#I/O's in progress)
```
dbwiddis: Given the multiple interpretations (e.g. average, current) this should probably be named `currentQueueLength`

oshi-core/src/main/java/oshi/hardware/HWDiskStore.java
```diff
@@ -84,6 +85,8 @@ public HWDiskStore() {
      *            Number of writes to the disk
      * @param writeBytes
      *            Number of bytes written to the disk
+     * @param queueLength
+     *            Number of I/O operations currently in progress
```
dbwiddis: I'd like this to be a bit more verbose. The linux doc has this longer version, something in between what you have and this would be good:

> the number of I/O requests that have been issued to the device driver but have not yet completed.

Also be clear that it's both read and write.

oshi-core/src/main/java/oshi/hardware/platform/unix/freebsd/FreeBsdDisks.java
```diff
@@ -66,6 +66,7 @@ public static boolean updateDiskStats(HWDiskStore diskStore) {
             // In KB
             diskStore.setReadBytes((long) (ParseUtil.parseDoubleOrDefault(split[3], 0d) * 1024));
             diskStore.setWriteBytes((long) (ParseUtil.parseDoubleOrDefault(split[4], 0d) * 1024));
+            diskStore.setQueueLength((long) (ParseUtil.parseDoubleOrDefault(split[5], 0d)));
```
dbwiddis: Nitpick. For documentation you stuck these under the "In KB" header but aren't multiplying by 1024.  So leave a space or add another comment to indicate the units of this.

Also I'm wondering why we parse double and cast to long (same thing in some previous lines that you didn't change) rather than parsing directly to long...

oshi-core/src/main/java/oshi/hardware/platform/unix/freebsd/FreeBsdDisks.java
```diff
@@ -106,6 +107,7 @@ public static boolean updateDiskStats(HWDiskStore diskStore) {
             // In KB
             store.setReadBytes((long) (ParseUtil.parseDoubleOrDefault(split[3], 0d) * 1024));
             store.setWriteBytes((long) (ParseUtil.parseDoubleOrDefault(split[4], 0d) * 1024));
+            store.setQueueLength((long) (ParseUtil.parseDoubleOrDefault(split[5], 0d)));
```
dbwiddis: same as above, comment differently and if we can parse straight to Long do so

oshi-json/src/main/java/oshi/json/hardware/HWDiskStore.java
Space2Man: Well, for Windows and FreeBSD the definition does not define anything. But I found some hints on the net that it's only the Waiting I/Os ... but this Linux statement makes me wonder about how it's defined on Linux:

> Incremented as requests are given to appropriate struct request_queue and decremented as they finish.

When do they finish? If the are completed to / from disk or if they are dispatched to disk ...

oshi-core/src/main/java/oshi/hardware/platform/unix/freebsd/FreeBsdDisks.java
Space2Man: Not sure how to validate if we can parse to long directly ...

Space2Man:
Updated

Space2Man:
Updated again, not sure though if this should include the requests which are currently active (on Solaris) ... it's the only OS where we have counters for both waiting and active ...

Space2Man:
Updated once more (don't use setters inside constructor) :)

oshi-core/src/main/java/oshi/hardware/HWDiskStore.java
```diff
@@ -62,7 +63,18 @@
      * Create an object with empty/default values
      */
     public HWDiskStore() {
-        this("", "", "", 0L, 0L, 0L, 0L, 0L, 0L, new HWPartition[0], 0L);
+        name = "";
+        model = "";
+        serial = "";
+        size = 0L;
```
dbwiddis: this is probably unnecessary as it's the default initialization.

oshi-core/src/main/java/oshi/hardware/HWDiskStore.java
```diff
@@ -62,7 +63,18 @@
      * Create an object with empty/default values
      */
     public HWDiskStore() {
-        this("", "", "", 0L, 0L, 0L, 0L, 0L, 0L, new HWPartition[0], 0L);
+        name = "";
+        model = "";
+        serial = "";
```
dbwiddis: you can probably just put this initialization above where the variable is defined.

oshi-core/src/main/java/oshi/hardware/HWDiskStore.java
```diff
@@ -92,18 +106,19 @@ public HWDiskStore() {
      *            milliseconds since the epoch
      */
     public HWDiskStore(String name, String model, String serial, long size, long reads, long readBytes, long writes,
```
dbwiddis: just delete this constructor

oshi-json/src/main/java/oshi/json/hardware/HWDiskStore.java
dbwiddis: I think you've got it mostly correct, then...

oshi-core/src/main/java/oshi/hardware/platform/unix/freebsd/FreeBsdDisks.java
dbwiddis: Dug into the source code. Here's the format headers for FreeBSD:
```
"device     r/s   w/s    kr/s    kw/s qlen svc_t  %%b  "
"%-8.8s %5.1Lf %5.1Lf %7.1Lf %7.1Lf %4" PRIu64 " %5.1Lf %3.0Lf "
```
So all the outputs are floating point except:
- qlen is `%4" PRIu64 "` which is an unsigned 64 bit long...
- %b (busy) is truncated to 0 decimals so could be parsed as an int

oshi-json/src/main/java/oshi/json/hardware/HWDiskStore.java
```diff
@@ -81,22 +83,33 @@
      *            milliseconds since the epoch
      */
     public HWDiskStore(String name, String model, String serial, long size, long reads, long readBytes, long writes,
-            long writeBytes, long transferTime, HWPartition[] partitions, long timeStamp) {
+            long writeBytes, long currentQueueLength, long transferTime, HWPartition[] partitions, long timeStamp) {
```
dbwiddis: Replace this constructor with one that takes an `oshi.hardware.HWDiskStore` argument and sets values based on the argument's getters.

Space2Man: Hm, not sure what you mean? Do you want to have a copy constructor?

dbwiddis: See `HardwareAbstractionLayerImpl.java` line 136 where you added another param to the constructor.  Consistent with the goal to remove multi-arg constructors you still need a way for that Impl class to wrap the `oshi-core` object inside the `oshi-json` object.  So you will create a new `oshi.json.hardware.HWDiskStore` using an `oshi.hardware.HWDiskStore` argument.  Then for each of the elements in the `oshi-core` version, use its getter inside the setter for the `oshi-json` version.

Space2Man: pushed changes

oshi-core/src/main/java/oshi/hardware/platform/unix/freebsd/FreeBsdDisks.java
```diff
@@ -66,6 +66,8 @@ public static boolean updateDiskStats(HWDiskStore diskStore) {
             // In KB
             diskStore.setReadBytes((long) (ParseUtil.parseDoubleOrDefault(split[3], 0d) * 1024));
             diskStore.setWriteBytes((long) (ParseUtil.parseDoubleOrDefault(split[4], 0d) * 1024));
+            // # transactions
+            diskStore.setCurrentQueueLength((long) (ParseUtil.parseDoubleOrDefault(split[5], 0d)));
```
dbwiddis: This all looks good... I'll merge in the morning. One last fix if you get to it before I do, please use `parseLongOrDefault()` here and in line 112.  I confirmed from the source code this will be a uint64 value.

Space2Man: done

PR #640, opened 2018-10-04
Windows sound card implementation 

BilalAM:
Have done the windows version.
Slight issue with the codec info , mainly because sometimes windows shows the codec chip along with the card device (e.g Realtek ALC887 HD Audio Device)  and sometimes it doesn't (e.g Intel HD Audio Device) so returning the name of the card as codec too seems to be the viable solution ,but linux displays the codec by default..

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
@@ -121,7 +121,11 @@ public class WindowsSoundCard extends AbstractSoundCard {
 
     /**
      * Returns the sound cards..
-     *
+     * <br>
+     * NOTE : The reason why the codec name is same as the card name is because windows does not provide
+     * the name of the codec chip but sometimes the name of the card returned is infact the name of the codec
+     * chip also.
+     * Example : Realtek ALC887 HD Audio Device
      * @return List of sound cards
      */
     public static List<WindowsSoundCard> getSoundCards() {
--- a/oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
@@ -18,14 +18,14 @@
  */
 package oshi.hardware.platform.windows;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import oshi.hardware.common.AbstractSoundCard;
 import oshi.jna.platform.windows.WbemcliUtil;
 import oshi.util.platform.windows.WmiUtil;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Sound Card data obtained by Wmi Queries for Windows.
@@ -33,6 +33,9 @@ import java.util.List;
  * @author : BilalAM
  */
 public class WindowsSoundCard extends AbstractSoundCard {
+    public WindowsSoundCard(String kernelVersion, String name, String codec) {
+        super(kernelVersion, name, codec);
+    }
 
     enum SoundCardName {
         MANUFACTURER, NAME;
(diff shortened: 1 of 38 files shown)


oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    enum AudioCardKernel {
+        DRIVERPROVIDERNAME, DRIVERNAME, DRIVERVERSION, DEVICENAME
+    }
+
+    private static final Logger LOG = LoggerFactory.getLogger(FileUtil.class);
+    private static WbemcliUtil.WmiQuery<AudioCardName> AUDIO_CARD_QUERY;
+    private static final String AUDIO_CARD = "Win32_SoundDevice";
+    private static WbemcliUtil.WmiQuery<AudioCardKernel> AUDIO_CARD_KERNAL_QUERY;
```
dbwiddis: typo

BilalAM:
I have tried to correct the revisions according to your comments , kindly check.
p.s The logger line was a bit awkward for me :D

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
```diff
@@ -0,0 +1,127 @@
...
+     *     <li>The DeviceName property</li>
+     *     <li>The DriverProvider property</li>
+     * </ul>
+     * Note: The query of AUDIO_CARD_KERNEL_QUERY is like Get-WmiObject Win32_PnPSignedDriver |
+     * where {$_.devicename -like "*(<b>the name of the card returned by Win32_SoundDevice Name Property</b>)*"}
+     * @param index
+     *        The row to fetch from
+     * @return  The 'complete' name of the driver.
+     */
+    private static String getAudioCardKernelVersion(int index) {
```
dbwiddis: This seems inefficient: you run one query to get the card names, and then run this query for each card using a filter.

I think it would be better to run this query first without a filter; iterate over the results and populate a Map using the DeviceName as the key and calculated string below as the Value.   Then iterate over the `Win32_SoundDevice` results and lookup the kernel version from the map using the same name key.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
```diff
@@ -0,0 +1,127 @@
...
+     * Returns the sound cards..
+     * <br>
+     * NOTE : The reason why the codec name is same as the card name is because windows does not provide
+     * the name of the codec chip but sometimes the name of the card returned is infact the name of the codec
+     * chip also.
+     * Example : Realtek ALC887 HD Audio Device
+     * @return List of sound cards
+     */
+    public static List<WindowsSoundCard> getSoundCards() {
+        List<WindowsSoundCard> cards = new ArrayList<>();
```
dbwiddis: Suggestion:  populate the map described @ line 94 comment here.   Then query the audio card query and iterate over the results.   You'd pull the version from the map, and calculate the name and complete name from the query result.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
BilalAM: The reason I run this with a filter is because Win32_PnPSignedDriver returns a very long list of information of all the drivers , instead of getting all the results and then populating a map , I made a filter query that 
- returns only the driver info about a particular device(in our case it is the sound card returned by `AUDIO_CARD_QUERY_RESULT` i.e Win32_SoundDevice) 
- Then builds a string consisting of the attributes returned by our query

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
```diff
@@ -0,0 +1,127 @@
...
+        MANUFACTURER, NAME;
+    }
+
+    enum SoundCardKernel {
+        DRIVERPROVIDERNAME, DRIVERNAME, DRIVERVERSION, DEVICENAME;
+    }
+
+    private static final Logger LOG = LoggerFactory.getLogger(WindowsSoundCard.class);
+    private static WbemcliUtil.WmiQuery<SoundCardName> AUDIO_CARD_QUERY = new WbemcliUtil.WmiQuery<>("Win32_SoundDevice",SoundCardName.class);
+    private static WbemcliUtil.WmiResult<SoundCardName> AUDIO_CARD_QUERY_RESULT = WmiUtil.queryWMI(AUDIO_CARD_QUERY);
```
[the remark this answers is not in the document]
BilalAM: Wouldn't it better to just make the AUDIO_CARD_QUERY static and final since it is not going to change throughout along with the AUDIO_CARD_QUERY_RESULT instead of using static blocks ?

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
BilalAM: As far as my understanding the number of times Wmi gets queried 

- Once for Win32_SoundDevice.
- And depending on the number of cards returned by SoundDevice ,'n' times for Win32_PnPSignedDriver where n are the number of cards returned by Win32_SoundDevice.
So according to my understanding Wmi gets queried on a maximum of 2-3 times (if we consider a machine having 2 sound cards or more than that (rare) ) .
I may be wrong on this but this what I understand !

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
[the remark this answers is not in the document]
dbwiddis: That's what I'm trying to point out -- you don't need both.  Pick one or the other.  If you persist the result as final, you don't need the query any more.   So you could declare the result as final and put a static block after it to define/run the query, then the query can go away.

Or if you want to query and generate the result on-the-fly then persist the query but not the result -- generate the result down in the getter methods when you need it.  This is how we do it in most other OSHI classes: we persist the query as static and final, and only hold on to the results temporarily.    

However, I understand that SoundCards are not hot-swappable and all this information can probably be generated one time and stored in String lists or maps, and then you can dispose of all the WMI stuff and just fetch from the maps when the getters are called.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
dbwiddis: You still end up iterating the SoundDevice query and doing multiple individual queries.

Take a look at `WindowsUsbDevice.java` and see how `PNPENTITY_BASE_CLASS` is defined and used.  We collect the list of PnPDeviceIDs and then do a single query to get just those device IDs.

So my suggestion:
(One time on initialization) 
1. Run the audio_card_query and store results in a MANUFACTURER_MAP with the name as the key and manufacturer as the value. 
2. Iterate the manufacturer map keys and build the kernel query where clause similar to the USB Device example I gave and run that query, storing the results in a KERNEL_VERSION_MAP with name as the key, and constructing the string as you currently do.
(When the getter is called)
3. Iterate the manufacturer map keys and build the sound card objects using the values from the two different maps.

BilalAM: In the `WindowsUsbDevice.java` example , what you are trying to say is that I should make up my Win32_PnPSignedDriver query only once and then run it ?

- Like the query should be something like this 
`Win32_PnPSignedDriver WHERE DeviceName = "{name value from MANUFACTURER map}" OR "{second name value from MANUFACTURER map}"` and so on..till the size of the map , just like you have built it in the UsbClass.**And that I should do this only once in a static block.**
- Then when I call the `getSoundCards()` method , it is going to get the kernels by running the newly built query **ONLY ONCE** , then iterate over the results and build the sound cards accordingly ?

I hope I have understood this.

dbwiddis: Yes!

1. The list of sound cards doesn't ever change so you can pull that once on initialization, store in the map, and construct a single query like your first bullet.
2. When you call `getSoundCards()` you will run the saved query once (it will retrieve multiple rows) and iterate over the resultcounts, matching the name to the name/manufacturer map and building the sound card objects.
3. Repeat 2 as needed if information changes.

BilalAM: Okay now I am getting it , will start working on this ASAP today.
Sorry for taking a bit long though..

BilalAM:
Changes#2 .
Check and comment.
If its approved then I will correct the small formatting issues like removing the <p> tags in comments.

cilki:
@BilalAM We decided to let the user worry about this kind of thing.

dbwiddis:
As @cilki said, if we're only reporting on a situation that's already obvious to the user (e.g., empty array) then there's no need to add to the spam.  Feel free to add a log at DEBUG level, though, if you like.  But I don't think it's needed.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * Creates our Win32_PnPSignedDevice query with the WHERE clause taking the attributes
+     * from our map.
+     * @param map
+     *       The map whose keys will be used inside the Where Clause.
+     */
+    private static void createClause(Map<String, String> map) {
+        boolean isEnd = true;
+        for (String key : map.keySet()) {
```
dbwiddis: Any reason you're avoiding using a `StringBuilder` here?  Seems the classic case for it rather than using `+=` with Strings.  Do a `new Stringbuilder(DRIVER_QUERY_BASE)` and then append `" WHERE (foo)"` (1st iteration) and `" OR (foo)"` (2nd+ iteration) inside the loop.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
```diff
@@ -0,0 +1,140 @@
...
+    enum SoundCardName {
+        MANUFACTURER, NAME;
+    }
+
+    enum SoundCardKernel {
+        DRIVERPROVIDERNAME, DRIVERNAME, DRIVERVERSION, DEVICENAME;
+    }
+
+    private static Map<String, String> NAME_MAP = new HashMap<>();
+    private static final String driverQuery = "Win32_PnPSignedDriver";
```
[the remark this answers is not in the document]
dbwiddis: No reason to have a separate `driverQuery` and `whereClause` and concatenate when you use them.  If you want to keep them separate, declare this as `DRIVER_QUERY_BASE` and then use it to build a (non-final) `driverQuery` later that includes both this class and the WHERE clause.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
```diff
@@ -0,0 +1,140 @@
...
+     * @return List of sound cards
+     */
+    public static List<WindowsSoundCard> getSoundCards() {
+        WbemcliUtil.WmiQuery<SoundCardKernel> cardKernelQuery = new WbemcliUtil.WmiQuery<>(driverQuery +" "+ whereClause, SoundCardKernel.class);
+        WbemcliUtil.WmiResult<SoundCardKernel> cardKernelQueryResult = WmiUtil.queryWMI(cardKernelQuery);
+
+        List<WindowsSoundCard> soundCards = new ArrayList<>();
+
+        for (Map.Entry<String, String> nameSet : NAME_MAP.entrySet()) {
+            for (int i = 0; i < cardKernelQueryResult.getResultCount(); i++) {
```
dbwiddis: This works but it's O(n^2).  It seems less convoluted to make the `WmiResult` iteration the outer loop and then just get from the map (careful with nulls, see `MapUtil if you need some defaults) so the whole loop is just O(n).

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
BilalAM: Well the main reason of using += is that it is really not a large string to be using a StringBuilder (and I feel a bit comfortable using +=s for small strings like this one lol )

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
BilalAM: `...driverQuery = "Win32PnPSignedDriver WHERE DeviceName LIKE"  `, like this right ? This looks better instead of using two different variables.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
dbwiddis: I have never used `+=` for string concatenation as a matter of taste. :)   For short strings always adding fixed elements the `+` is just fine (as you've done elsewhere) but almost always if I end up adding to a string inside a loop I prefer a `StringBuilder`.  Again this is mostly preference/performance/style.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsSoundCard.java
dbwiddis: You can stick with one single variable, yes... just initialize empty and build it all in one place so someone editing this code later doesn't have to hunt down for various initializations in various places.

BilalAM: Well I managed to remove the outer loop , kept the inner loop as the main loop and used NAME_MAP.containsKey() inside the loop instead of iterating the keys one by one and checking. Pushed the changes just now..

BilalAM: I guess this is fine ?

BilalAM:
Sure thing ! I'll take care of the StringBuilder usage when I am doing the rest of the operating systems (if you want me to continue with the rest).

dbwiddis:
I already did it.... [see this commit](https://github.com/oshi/oshi/commit/981ccfc332f7c5f921d065af22f728e008126647).

dbwiddis:
And do please see if you can do the two unixes.  Mac driver version looks like something I can pull from a driver file somewhere, still researching it... I think on the unix machines you're going to have to parse `dmidecode`

BilalAM:
I was doing a bit of research for solaris and freeBsd (both of these are **completely new** to me and looks fun to tweak around with them :)  ) , can you give me some time because I have a somewhat busy schedule next week .

PR #652, opened 2018-10-12
Freebsd sound card

BilalAM:
FreeBSD sound card implementation.
Also have added another method inside ParseUtil.

### DIFF
--- a/oshi-core/src/main/java/oshi/util/ParseUtil.java
+++ b/oshi-core/src/main/java/oshi/util/ParseUtil.java
@@ -18,15 +18,15 @@
  */
 package oshi.util;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.math.BigInteger;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * String parsing utility.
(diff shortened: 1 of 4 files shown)


oshi-core/src/main/java/oshi/util/ParseUtil.java
```diff
@@ -498,6 +498,21 @@ public static String getSingleQuoteStringValue(String line) {
         return split[1];
     }
 
+    /**
+     * Gets a value between two quotes having multiple quotes between them.
+     * See ParseUtilTest.java for more clarification.
```
dbwiddis: I'd rather the javadoc be self explanatory.  Give an example rather than referring the API reader to the source code (which isn't exactly obvious, either).

Also while you've written this for a single quote it may be useful to make this more general for any delimiter, kind of like `trim()` does.  (Or would `trim()` work here if you just trimmed based on single quotes?)

BilalAM: Meaning the method should have another parameter where you have to pass the character so that it gets the value between the leading and trailing character right ? Also to remove the trailing and leading whitespaces 
Kinda like 
```
 public static String getStringBetween(String line , char c){
        int firstOcc = line.indexOf(c);
        if(firstOcc < 0){
            return "";
        }
        return (line.substring(firstOcc + 1 , line.lastIndexOf(c))).trim();
    }
```

BilalAM: Umm I am a bit confused on what to name this method :) `getStringBetween()` looks a bit vague

PR #653, opened 2018-10-13
Fix transferTime by using 1-%Idle

Space2Man:
or in this case: timestamp (in 100ns) - %idle (in 100ns)


oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsDisks.java
```diff
@@ -215,7 +215,7 @@ public static boolean updateDiskStats(HWDiskStore diskStore) {
             diskStore.setWrites(MapUtil.getOrDefault(writeMap, index, 0L));
             diskStore.setWriteBytes(MapUtil.getOrDefault(writeByteMap, index, 0L));
             diskStore.setCurrentQueueLength(MapUtil.getOrDefault(queueLengthMap, index, 0L));
-            diskStore.setTransferTime(MapUtil.getOrDefault(xferTimeMap, index, 0L));
+            diskStore.setTransferTime(MapUtil.getOrDefault(timeStampMap, index, 0L)-MapUtil.getOrDefault(xferTimeMap, index, 0L));
```
dbwiddis: Very minor performance tweak. If you do the next line (set timestamp) first, then you can just use `diskStore.getTimeStamp()` here.  Also swap lines 249 and 250 for the same reason.

Space2Man:
You mean in Changelog.md?

Issue #434, opened 2017-12-13
User information


bildechinger:
Feature could be added to get user information:

- username
- userid
- home-dir
- real name

On Linux it is simple (getent passwd).

henryx:
On Windows, is possible to use the `Win32_UserAccount` WMI class, but I'm not sure if it returns Active Directory users too

BilalAM:
`getent passwd` also works for Unix systems ( I ran it on GhostBSD) , but it returns a long list of all entries from the database , you have to find your entry.
 I was thinking first getting your usrname by `whoami ` then passing that value to the list of the entries returned by `getent` and simply `.contains()`-ing the usrname to return our entry , then parse it. 
(Or maybe there is a filter for getent where you can pass a value to filter result of which I do not know about it :) )

dbwiddis:
I'm closing this issue as it's been up for a year and the direction is vague enough that I don't see it actually going anywhere.   Feel free to re-open or submit a new issue to clearly specify missing functionality in OSHI.

PR #654, opened 2018-10-13
Solaris Sound Card (issue#625) and User Feature (issue#434)

BilalAM:
Couple of different additions.
1) `SolarisSoundCard `added.
2) User feature mentioned in #434 implemented for **Linux , Solaris and FreeBSD.**
3) Test classes of User feature.

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
+++ b/oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
@@ -122,4 +122,11 @@ public interface HardwareAbstractionLayer extends Serializable {
      *         present.
      */
     SoundCard[] getSoundCards();
+
+    /**
+     * Returns information about current user
+     *
+     * @return A User object
+     */
+    User getUser();
 }
--- a/oshi-core/src/main/java/oshi/hardware/User.java
+++ b/oshi-core/src/main/java/oshi/hardware/User.java
@@ -1,0 +1,49 @@
+/**
+ * Oshi (https://github.com/oshi/oshi)
+ *
+ * Copyright (c) 2010 - 2018 The Oshi Project Team
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Maintainers:
+ * dblock[at]dblock[dot]org
+ * widdis[at]gmail[dot]com
+ * enrico.bianchi[at]gmail[dot]com
+ *
+ * Contributors:
+ * https://github.com/oshi/oshi/graphs/contributors
+ */
+package oshi.hardware;
+
+/**
+ * @author: BilalAM
+ */
+public interface User {
+
+        /**
+         * Gets the user name currently logged in
+         * @return : The user name.
+         */
+        String getUserName();
+
+        /**
+         * Gets the current user's Id.
+         * @return : The user id
+         */
+        int getUserId();
+
+        /**
+         * Gets the user home directory path
+         * @return : The user's home directory path
+         */
+        String getHomeDir();
+
+        /**
+         * Gets the logged-in user's real name (results maybe equal to getUserName() )
+         * @return : The 'real' name of the user.
+         */
+        String getRealName();
+}
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractUser.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractUser.java
@@ -1,0 +1,95 @@
+/**
+ * Oshi (https://github.com/oshi/oshi)
+ *
+ * Copyright (c) 2010 - 2018 The Oshi Project Team
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Maintainers:
+ * dblock[at]dblock[dot]org
+ * widdis[at]gmail[dot]com
+ * enrico.bianchi[at]gmail[dot]com
+ *
+ * Contributors:
+ * https://github.com/oshi/oshi/graphs/contributors
+ */
+package oshi.hardware.common;
+
+import oshi.hardware.User;
+
+/**
+ * An abstract user class having information about the current user.
+ *
+ * @author : BilalAM
+ */
+public abstract class AbstractUser implements User {
+        private String userName;
+        private int userId;
+        private String homeDir;
+
+        public AbstractUser(String userName, int userId, String homeDir, String realName) {
+                this.userName = userName;
+                this.userId = userId;
+                this.homeDir = homeDir;
+                this.realName = realName;
+        }
+
+        private String realName;
+
+        @Override
+        public String getUserName() {
+                return userName;
+        }
+
+        public void setUserName(String userName) {
+                this.userName = userName;
+        }
+
+        @Override
+        public int getUserId() {
+                return userId;
+        }
+
+        public void setUserId(int userId) {
+                this.userId = userId;
+        }
+
+        @Override
+        public String getHomeDir() {
+                return homeDir;
+        }
+
+        public void setHomeDir(String homeDir) {
+                this.homeDir = homeDir;
+        }
+
+        @Override
+        public String getRealName() {
+                return realName;
+        }
+
+        public void setRealName(String realName) {
+                this.realName = realName;
+        }
+
+        @Override
+        public String toString(){
+                StringBuilder builder = new StringBuilder();
+                builder.append("User@");
+                builder.append(Integer.toHexString(hashCode()));
+                builder.append(" [name=");
+                builder.append(userName);
+                builder.append(", id=");
+                builder.append(userId);
+                builder.append(", home-dir=");
+                builder.append(homeDir);
+                builder.append(", real-name=");
+                builder.append(realName);
+                builder.append(']');
+                return builder.toString();
+        }
+
+}
--- a/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHardwareAbstractionLayer.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHardwareAbstractionLayer.java
@@ -18,16 +18,7 @@
  */
 package oshi.hardware.platform.linux;
 
-import oshi.hardware.CentralProcessor;
-import oshi.hardware.ComputerSystem;
-import oshi.hardware.Display;
-import oshi.hardware.GlobalMemory;
-import oshi.hardware.HWDiskStore;
-import oshi.hardware.NetworkIF;
-import oshi.hardware.PowerSource;
-import oshi.hardware.Sensors;
-import oshi.hardware.SoundCard;
-import oshi.hardware.UsbDevice;
+import oshi.hardware.*;
 import oshi.hardware.common.AbstractHardwareAbstractionLayer;
 
 public class LinuxHardwareAbstractionLayer extends AbstractHardwareAbstractionLayer {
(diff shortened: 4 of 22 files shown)


BilalAM:
Travis failing probably because travis log shows the platform as mac os , and mac os functionality is currently not built . Maybe thats why ?

dbwiddis:
You need to return at least something not null if you want to test for non-null :)

BilalAM:
Is it okay to return an empty object (with empty fields) just till the implementation is done ? 
Like in MacOS abstraction layer 
`public User getUser() { return new MacUser("","",""); }`

dbwiddis:
Yes, you can certainly return reasonable defaults.  Empty or "unknown".   I'm curious, though. What does a user have to do with a sound card?  Isn't one software and the other hardware?  Are sound cards restricted to only some users?

BilalAM:
Actually I have done two things on the same branch (the Solaris sound card implementation of #625  and the User feature mentioned in issue #434 ). 
The User feature has nothing to do with the SoundCard obviously :)

dbwiddis:
Ah.  In general you should use a different branch.  Next time

BilalAM:
Yes , sorry for that !

oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
```diff
@@ -122,4 +122,11 @@
      *         present.
      */
     SoundCard[] getSoundCards();
+
+    /**
+     * Returns information about current user
+     *
+     * @return A User object
+     */
+    User getUser();
```
dbwiddis: I don't think this should be on the HAL, User is a software/OS concept.  Move to the `OperatingSystem`.

oshi-core/src/main/java/oshi/hardware/User.java
```diff
@@ -0,0 +1,49 @@
...
+        int getUserId();
+
+        /**
+         * Gets the user home directory path
+         * @return : The user's home directory path
+         */
+        String getHomeDir();
+
+        /**
+         * Gets the logged-in user's real name (results maybe equal to getUserName() )
```
dbwiddis: Be more clear with language like "attempts to get" or "makes a best effort".

oshi-core/src/main/java/oshi/hardware/User.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Contributors:
+ * https://github.com/oshi/oshi/graphs/contributors
+ */
+package oshi.hardware;
+
+/**
+ * @author: BilalAM
+ */
+public interface User {
```
dbwiddis: I'm wondering if this should be focused on the currently logged in user or if we can make this more general to apply to any computer user; there could be an "isCurrentUser" boolean.  But this somewhat reminds me of the OSHI process as it's running. It may have its own PID but that's just a small picture of the rest of the system.   The current user is of interest (as is the current process) but it's also useful to know other users with accounts on the computer.

oshi-core/src/main/java/oshi/hardware/common/AbstractUser.java
```diff
@@ -0,0 +1,95 @@
...
+ * An abstract user class having information about the current user.
+ *
+ * @author : BilalAM
+ */
+public abstract class AbstractUser implements User {
+        private String userName;
+        private int userId;
+        private String homeDir;
+
+        public AbstractUser(String userName, int userId, String homeDir, String realName) {
```
dbwiddis: I know you're copying other existing constructors in OSH but I'm really trying to lean towards new versions not having constructors with arguments like this. Just construct the object with default values for the class variables, and have setters for them (you can keep the setters `protected` to limit API abilities.)

oshi-core/src/main/java/oshi/hardware/common/AbstractUser.java
```diff
@@ -0,0 +1,95 @@
...
+import oshi.hardware.User;
+
+/**
+ * An abstract user class having information about the current user.
+ *
+ * @author : BilalAM
+ */
+public abstract class AbstractUser implements User {
+        private String userName;
+        private int userId;
```
dbwiddis: The `int` type won't work with windows SIDs.   Won't approve this PR unless you explain your plan for Windows user Id's with this type. :)

oshi-core/src/main/java/oshi/hardware/common/AbstractUser.java
```diff
@@ -0,0 +1,95 @@
...
+        public String getRealName() {
+                return realName;
+        }
+
+        public void setRealName(String realName) {
+                this.realName = realName;
+        }
+
+        @Override
+        public String toString(){
```
dbwiddis: Glad to see you adopting what seems to be the new OSHI standard toString. :)

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHardwareAbstractionLayer.java
```diff
@@ -18,16 +18,7 @@
...
-import oshi.hardware.ComputerSystem;
-import oshi.hardware.Display;
-import oshi.hardware.GlobalMemory;
-import oshi.hardware.HWDiskStore;
-import oshi.hardware.NetworkIF;
-import oshi.hardware.PowerSource;
-import oshi.hardware.Sensors;
-import oshi.hardware.SoundCard;
-import oshi.hardware.UsbDevice;
+import oshi.hardware.*;
```
dbwiddis: FYI, my formatter will expand this again.  While this compiles if you can set up your formatter to list each import individually it'll save later spurious changes on commits.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxUser.java
```diff
@@ -0,0 +1,64 @@
...
+        static {
+                userName = ExecutingCommand.getFirstAnswer(WHO_AM_I);
+                for(String line : ExecutingCommand.runNative(GETENT_PASSWD)){
+                        if(line.contains(userName)){
+                                values = line.split(":");
+                        }
+                }
+        }
+        public static LinuxUser getUser(){
+              return new LinuxUser(values[0],Integer.parseInt(values[2]),values[5],values[4]);
```
dbwiddis: Not fond of setting `values` as a variable on initialization but then generating a new object for the getter each time here.   Either the getter should go all the way to the source, or you should do the new object creation once and then just return the object here (if certain the info won't change).

Also, need to do length checking on the `split()` here to avoid array index out of bounds.

oshi-core/src/main/java/oshi/hardware/platform/mac/MacHardwareAbstractionLayer.java
```diff
@@ -18,16 +18,7 @@
...
-import oshi.hardware.ComputerSystem;
-import oshi.hardware.Display;
-import oshi.hardware.GlobalMemory;
-import oshi.hardware.HWDiskStore;
-import oshi.hardware.NetworkIF;
-import oshi.hardware.PowerSource;
-import oshi.hardware.Sensors;
-import oshi.hardware.SoundCard;
-import oshi.hardware.UsbDevice;
+import oshi.hardware.*;
```
dbwiddis: repeat of earlier comment.  please set your IDE to expand import wildcards.

oshi-core/src/main/java/oshi/hardware/platform/mac/MacHardwareAbstractionLayer.java
```diff
@@ -129,4 +120,9 @@ public Sensors getSensors() {
     public SoundCard[] getSoundCards() {
         return MacSoundCard.getSoundCards().toArray(new SoundCard[0]);
     }
+
+    @Override
+    public User getUser() {
+        return new MacUser("",0,"","");
```
dbwiddis: for Mac, parse the `id` command.

> <danielwiddis> [10-13 21:46] ~/git/oshi $ id
> uid=501(danielwiddis) gid=20(staff) groups=20(staff),405(com.apple.sharepoint.group.4)  ... <snip>

> <danielwiddis> [10-13 21:46] ~/git/oshi $ id -F
> Daniel Widdis

> <danielwiddis> [10-13 21:47] ~/git/oshi $ id -p
> uid	danielwiddis
> groups	staff com.apple.sharepoint.group.4  ... <snip>
>

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsHardwareAbstractionLayer.java
```diff
@@ -120,4 +123,9 @@ public Sensors getSensors() {
     public SoundCard[] getSoundCards() {
         return WindowsSoundCard.getSoundCards().toArray(new SoundCard[0]);
     }
+
+    @Override
+    public User getUser() {
+        return new WindowsUser("",0,"","");
```
dbwiddis: I don't think I'm going to approve this PR for Users until it's implemented on Windows.   You may want to separate out the `User` portion from the other portions as this has a ways to go.

oshi-core/src/test/java/oshi/SystemInfoTest.java
```diff
@@ -18,39 +18,21 @@
...
-import oshi.hardware.PowerSource;
-import oshi.hardware.Sensors;
-import oshi.hardware.UsbDevice;
-import oshi.hardware.SoundCard;
-import oshi.software.os.FileSystem;
-import oshi.software.os.NetworkParams;
-import oshi.software.os.OSFileStore;
-import oshi.software.os.OSProcess;
-import oshi.software.os.OperatingSystem;
+import oshi.software.os.*;
```
dbwiddis: Wildcards are fun but not OSHi standard. :)

oshi-core/src/test/java/oshi/SystemInfoTest.java
```diff
@@ -347,4 +333,13 @@ private static void printSoundCards(SoundCard[] cards) {
             System.out.println(card.toString());
         }
     }
+
+    private static void printUser(User user){
+        System.out.println("User Information : ");
+        System.out.println("User ID        : " + user.getUserId());
```
dbwiddis: let's be a bit more creative with this format.   Don't need 5 lines for this!

dbwiddis:
Take a look at the "User" portions of the OSProcess object.  (user and group names and IDs/SIDs/GIDs).  If we need a new User Object it should replace the individual attributes of that object.

BilalAM:
I am closing this , I will make two separate PRs , one for FreeBsDSoundCard and the other for User . Its getting a bit complicated  . On the User PR I will make the changes as requested by your review comments !

BilalAM:
@dbwiddis 
> Take a look at the "User" portions of the OSProcess object. (user and group names and IDs/SIDs/GIDs). If we need a new User Object it should replace the individual attributes of that object.

There is already an `OSUser` class inside `oshi.software.os` , but that class is only implemented for Linux (usages are only inside `LinuxOperatingSystem` and` LinuxUserGroupInfo `) , why not for rest of the OSs ? 

Also I was thinking of a `User` class inside `oshi.software.os` and the implementation of getting the current User info be made inside the specific ` *OperatingSystem ` classes like for example `LinuxOperatingSystem `has a `getCurrentUser() `method and so on.. ? Just like you have made OSProcess's implementation in all `*OperatingSystem` classes. 

But I am still a bit confused on that lonely `OSUser` class..

dbwiddis:
There have been (and still are) a few one-off classes in Linux for ease of parsing.  I think there's some Udev-related class somewhere too.    This should all eventually be fixed in a future version of OSHI when we implement a "driver" model.

In the meantime, perhaps that lonely class should be promoted!

dbwiddis:
Check out JNA which has some of this User/Group stuff implemented on Windows: https://github.com/java-native-access/jna/blob/4d805210c8890b9597f3326debdf41fd092e04eb/contrib/platform/src/com/sun/jna/platform/win32/Netapi32Util.java#L71

And mac (this came from OSHI): https://github.com/java-native-access/jna/blob/45c2bb86e7d118fa27db3697b928eb1d21f6eda1/contrib/platform/src/com/sun/jna/platform/mac/SystemB.java#L519

PR #674, opened 2018-11-03
Use lombok to make your code more short and easy to learn

AlisherGM:


BilalAM:
I believe lombok doesnt work well with jdk 10 (till yet). 
See these issues : 
https://github.com/rzwitserloot/lombok/issues/1651

BilalAM:
Thats why the build is failing , upgrade the version to the latest maven i.e 1.18.4 , you are using a pretty old version of it in your pom

dbwiddis:
I'm looking into this.  Some of the items in [the "controversy" section](http://jnb.ociweb.com/jnb/jnbJan2010.html#controversy) of the description concern me.  Open to listening, but a few reasons I think I'd say no:

1. I try hard not to introduce additional dependencies to OSHI.  This adds a dependency.
2. The version is 0.x.  I'm very hesitant to rely on an early release version of any code.
3. We don't save much time "writing boilerplate code".  The primary use I'd see for this in our code is in the new "data" or "state" classes I envision for OSHI 4, which will be POJOs with getters and setters.  But it takes about 3 seconds to generate that code in my IDE so it's not saving me any work... probably adding more work to keep up with this developmental code.
4. I didn't read far enough into the docs to see how I can have more verbose descriptions of class attributes in javadocs, like I currently do with some getters.

I do think some of the items are great ideas and would fully support their inclusion in future versions of Java, if the Java maintainers see their value.  However, unless/until these annotations become java standard (or at least widely adopted pseudo-standards) I'd shy away from simplifying the actual java code with hacked annotations and move instead to automatically generating java code from framework documents.  For example, @cilki has proposed [automatically generating Java files](https://github.com/square/javapoet) with JavaPoet in the upcoming major rewrite.

dbwiddis:
I'm going to close this PR as I'm strongly leaning against it because of the dependency.   I'm all for auto-generated code but would prefer that's done up front before compilation, not as a run-time thing.   Please feel free to open a new Issue to discuss this.

hazendaz:
Lombok when used for @data and other associated items is only used at compilation and left as a provided scope.  It's never a good idea to use lomboks more tricky magic but always a good idea to use it to get rid of pojo boiler plate.  Not sure on asked context here as I haven't followed closely but wanted to point that out.  I've been using lombok this way for years.

Get Outlook for Android<https://aka.ms/ghei36>

hazendaz:
Looked at PR.  Makes sense to close it.  It's not being used just added.. needs code...

Get Outlook for Android<https://aka.ms/ghei36>

PR #694, opened 2018-11-25
Uci util method

Aashishthakur10:
Changes in order to have a Unique Computer Identifier in the Util method and test the same in the UtilTest method.

### DIFF
--- a/oshi-core/src/main/java/oshi/util/Util.java
+++ b/oshi-core/src/main/java/oshi/util/Util.java
@@ -92,13 +92,12 @@ public class Util {
         String processorIdentifier = centralProcessor.getIdentifier();
         int processors = centralProcessor.getLogicalProcessorCount();
 
-        //For the processor count.
         String delimiter = "#";
 
         return vendor.hashCode() +
-                delimiter.hashCode() +
+                delimiter +
                 processorSerialNumber.hashCode() +
-                delimiter.hashCode() +
+                delimiter +
                 processorIdentifier.hashCode() +
                 delimiter +
                 processors;
--- a/oshi-core/src/main/java/oshi/util/Util.java
+++ b/oshi-core/src/main/java/oshi/util/Util.java
@@ -20,12 +20,12 @@ package oshi.util;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 import oshi.SystemInfo;
 import oshi.hardware.CentralProcessor;
 import oshi.hardware.ComputerSystem;
 import oshi.hardware.HardwareAbstractionLayer;
 import oshi.software.os.OperatingSystem;
-
 
 /**
  * General utility methods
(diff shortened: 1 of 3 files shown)


oshi-core/src/test/java/oshi/util/UtilTest.java
```diff
@@ -42,5 +42,11 @@ public void testSleep() {
         then = now - 550;
         Util.sleepAfter(then, 500);
         assertTrue(System.currentTimeMillis() - now < 500);
+
+        String LicenseKeytest  = Util.generateLicenseKey();
+        System.out.println(LicenseKeytest);
```
dbwiddis: Rather than printing it out, use an `assert*()` test method.  Simplest might be asserting the string is not null; you could possibly `String.split()` on the delimiter and assert that the resulting array has a length of at least 4.

oshi-core/src/main/java/oshi/util/Util.java
```diff
@@ -65,4 +71,35 @@ public static void sleepAfter(long startTime, long ms) {
...
+        HardwareAbstractionLayer hardwareAbstractionLayer = systemInfo.getHardware();
+        CentralProcessor centralProcessor = hardwareAbstractionLayer.getProcessor();
+        ComputerSystem computerSystem = hardwareAbstractionLayer.getComputerSystem();
+
+        String vendor = operatingSystem.getManufacturer();
+        String processorSerialNumber = computerSystem.getSerialNumber();
+        String processorIdentifier = centralProcessor.getIdentifier();
+        int processors = centralProcessor.getLogicalProcessorCount();
+
+        String delimiter = "#";
```
dbwiddis: This is what the StackOverflow method used. Personally I would prefer hyphen-delimited hex strings using `String.format("%08x", foo.hashcode())` but that's optional.

oshi-core/src/main/java/oshi/util/Util.java
```diff
@@ -65,4 +71,35 @@ public static void sleepAfter(long startTime, long ms) {
...
     }
+
+    /**
+     * Generates a licence key based upon the processor serial number, vendor,
+     * processor identifier, and total processor count. These are obtained using
+     * the interfaces like OperatingSystem, HardwareAbstractionLayer, CentralProcessor
+     * ComputerSystem and class SystemInfo.
+     *
+     */
+    static String generateLicenseKey()
```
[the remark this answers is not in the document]
dbwiddis: As this bot indicates, use `public` for the method.

oshi-core/src/main/java/oshi/util/Util.java
```diff
@@ -65,4 +71,35 @@ public static void sleepAfter(long startTime, long ms) {
             sleep(until - now);
         }
     }
+
+    /**
+     * Generates a licence key based upon the processor serial number, vendor,
```
dbwiddis: Can you add a cautionary note to this javadoc that it may not be completely unique and can easily be defeated, or some other caveats from the StackOverflow thread?

Aashishthakur10: I was wondering though, if we share the actual value of our serial number instead of hash value will it be better? As Serial numbers are unique whereas hash value of the same may not be.

dbwiddis: Valid question.  (Same question could apply for "number of processors" which could just be left as an int.  I suspect the hash is unique enough for fingerprinting (and may not be accessible on many systems without root access anyway.)  My thought was to make the format easily parseable for comparison purposes, however, which would include transforming the string in such a manner that it's guaranteed not to contain the delimiter.   Which would you rather use as a user?  I'm not hard set one way or the other.

dblock:
Throwing my wrench into this as I had to deal with something similar before. Originally this was proposed as a unique identifier, but turned out not to be unique? So it now is named as a "license key", but that's something specific to the application calling it, not to the library. It's not "oshi's license key". 

So I guess my question is what is the original purpose of this?

oshi-core/src/main/java/oshi/util/Util.java
Aashishthakur10: True, I will mention the general warning of our key may not be unique.

Aashishthakur10:
Thanks, @dbwiddis for this opportunity and for being patient, appreciate it. It was a good learning experience for me.

Issue #719, opened 2018-12-10
NPE on HardwareAbstractionLayer::getComputerSystem for windows 7 and 10 when java as a service (admin account)


iXo:
Hi, I have a problem since some times now :

I am using oshi in a java application that is started as a service on windows 10 LTSB with an admin account.

The problem is for this code : 

`		final SystemInfo si = new SystemInfo();
		final HardwareAbstractionLayer hal = si.getHardware();
		final ComputerSystem computerSystem = hal.getComputerSystem();`

The last line result in a NPE : 

`java.lang.NullPointerException: null
	at oshi.util.platform.windows.WmiUtil.queryWMI(WmiUtil.java:140)
	at oshi.hardware.platform.windows.WindowsComputerSystem.init(WindowsComputerSystem.java:61)
	at oshi.hardware.platform.windows.WindowsComputerSystem.<init>(WindowsComputerSystem.java:55)
	at oshi.hardware.platform.windows.WindowsHardwareAbstractionLayer.getComputerSystem(WindowsHardwareAbstractionLayer.java:43)
	at infos.SystemInfos.getSystemInfos(SystemInfos.java:53)`

It doesn't seems to be a problem with WMI not ready as if I stop my service, and start it manually (from the service application) I still have this problem.

dbwiddis:
The NPE is on a line processing a COMException from JNA, for which the `HRESULT` is allowed to be `null`.  We need to check for that null.  Can you submit a PR?  If not, i'll toss this one to the first-timers-only list.

iXo:
Thanks for your lightning fast reply ! I don't really know what I could send for the PR, as if we check for the HRESULT is null, this will prevent the exception, but I don't know how to resolve the fact the com is not initialized. 
I would like to help more.
I can try any modification you want me to test.

dbwiddis:
Well, it's catching the exception and generating a log message that's creating the NPE.  At least you'll have a bit more clarity in the log message on what's failing?

iXo:
Hum... I tried to create a test unit just for the NPE, to be able to reproduce it I have to make the initCOM method from WmiUtil to always throw an exception. As the test environment doesn't use any mockup test framework, this will be quite difficult without a quite big modification of the code I want to test. And because the WmiUtil class is well protected I can't extend it.
Maybe I am wrong, sorry I can't really be of any help.

iXo:
(sorry for the right shift of the case entries for the switch)

dbwiddis:
[...]

PR #721, opened 2018-12-10
Better handling of null hresult in COMException

iXo:
I have done the minimum to solved the NPE described by the issue #719 

I wasn't able to create a unit test.

And because the underlying problem is not systematic, I haven't still see any changes.

### DIFF
--- a/oshi-core/src/main/java/oshi/util/platform/windows/WmiUtil.java
+++ b/oshi-core/src/main/java/oshi/util/platform/windows/WmiUtil.java
@@ -137,20 +137,24 @@ public class WmiUtil {
         } catch (COMException e) {
             // Ignore any exceptions with OpenHardwareMonitor
             if (!OHM_NAMESPACE.equals(query.getNameSpace())) {
-                switch (e.getHresult().intValue()) {
-                case Wbemcli.WBEM_E_INVALID_NAMESPACE:
-                    LOG.warn("COM exception: Invalid Namespace {}", query.getNameSpace());
-                    break;
-                case Wbemcli.WBEM_E_INVALID_CLASS:
-                    LOG.warn("COM exception: Invalid Class {}", query.getWmiClassName());
-                    break;
-                case Wbemcli.WBEM_E_INVALID_QUERY:
-                    LOG.warn("COM exception: Invalid Query: {}", queryToString(query));
-                    break;
-                default:
-                    LOG.warn(
-                            "COM exception querying {}, which might not be on your system. Will not attempt to query it again. Error was: {}:",
-                            query.getWmiClassName(), e.getMessage());
+                if (e.getHresult() == null) {
+                    LOG.warn("COM exception (HRESULT == null): {}", e.getMessage());
+                } else {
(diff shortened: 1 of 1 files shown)


oshi-core/src/main/java/oshi/util/platform/windows/WmiUtil.java
```diff
@@ -137,20 +137,24 @@ public static boolean hasNamespace(String namespace) {
...
-                    LOG.warn("COM exception: Invalid Class {}", query.getWmiClassName());
-                    break;
-                case Wbemcli.WBEM_E_INVALID_QUERY:
-                    LOG.warn("COM exception: Invalid Query: {}", queryToString(query));
-                    break;
-                default:
-                    LOG.warn(
-                            "COM exception querying {}, which might not be on your system. Will not attempt to query it again. Error was: {}:",
-                            query.getWmiClassName(), e.getMessage());
+                if (e.getHresult() == null) {
```
dbwiddis: I think it might be better to create a temp int variable and assign 0 to it if `getHresult()` returns null.  Then just use this integer in the switch, and you'll get the existing longer/verbose default log message.

iXo: Could we consider to use this code `throw new COMException("Failed to initialize COM library.", new HRESULT(0));` instead ? If so, the correction in only 2 lines in WmiUtil and what I have submited is not needed anymore.

oshi-core/src/main/java/oshi/util/platform/windows/WmiUtil.java
```diff
@@ -443,7 +439,7 @@ public static void initCOM() {
             // This can be safely ignored
             if (COMUtils.FAILED(hres) && hres.intValue() != WinError.RPC_E_TOO_LATE) {
                 Ole32.INSTANCE.CoUninitialize();
-                throw new COMException("Failed to initialize security.");
+                throw new COMException("Failed to initialize security.", new HRESULT(-1));
```
dbwiddis: Better solution for OSHI, definitely, although I still think we should trap the null in the other code.

But why use -1 here when we have a hres variable (already a `HRESULT` type available to us?  The correct fix here (and 12 lines earlier) is just to include `hres` as the second argument. :)

oshi-core/src/main/java/oshi/util/platform/windows/WmiUtil.java
dbwiddis: Yes, except use the existing hres variable (not sure why I missed that before.  I think I had to exclude it with the 4.x version of JNA and missed adding it when moving to 5.x (although 5.x created the default null that bit us here.)

oshi-core/src/main/java/oshi/util/platform/windows/WmiUtil.java
iXo: Oh ! sorry I didn't see the href variable, I will make the modifications. And at a case if href is null too.

PR #766, opened 2019-01-17
Use query key to update counters in groups

### LINKED ISSUE #759: Redesign PDH handle caching/WMI fallback
Most OSHI methods for fetching data come from resources that can be/are closed following the fetch.  The main exception to this rule is PDH counters, which are created, and for which OSHI needs to hold on to the Windows `HANDLE`.

The current code to cache these handles has a few flaws:
- It is not thread safe
- For convenience, I grouped counters on query by object.  (A single query is refreshed/updated at once but may contain multiple individual counters.)  This is not always optimal:
  - Multiple queries must be called to refresh one OSHI object, such as `WindowsGlobalMemory` having two counters on the `Memory` object and one on the `PagingFile` instance.  (Although perhaps this is an argument for a separate `PagingFile` class?)
  - Multiple OSHI objects might have a need to reference the same counter so we'd over-refresh.
- Not all counters are available on all systems, so a fallback system is required.  In all counter fallback cases, we're using WMI calls to the `PerfData_PerfRawData_*` set of classes.  This is resulting in:
  - Many unused static WMI objects/enums floating around individual classes
  - Use of a `null` `WmiQuery` object as a conditional to determine whether to fetch the data from WMI or PDH.
  - Complex methods in many different classes that all implement the same paradigm of "try to get this information from a counter; if it doesn't work, fetch that same data from WMI."

What is needed:
 - Create a new thread-safe `PerfDataQueryHandler` class and provide methods to:
   - Create new counters and arbitrarily group them by a key (`Enum` is ok.)
   - Handle exceptions when the counter creation fails, and establish a WMI fallback (transparent to the user)
      - This could probably involve another small class for each of the groups (matching the `Enum`), with the class housing the static enums and query String constants.
      - Keep track of a boolean variable indicating whether this fallback is occurring
   - Refresh (query) a group of counters
   - Fetch the data from the counters and return it to the user
   - Remove counters upon request by users (to release the resources) or automatically upon exception/WMI fallback (see above)

dbwiddis:
Fixes #759

### DIFF
--- a/oshi-core/src/main/java/oshi/data/windows/PerfCounterQuery.java
+++ b/oshi-core/src/main/java/oshi/data/windows/PerfCounterQuery.java
@@ -48,6 +48,7 @@ public class PerfCounterQuery<T extends Enum<T>> {
     protected final Class<T> propertyEnum;
     protected final String perfObject;
     protected final String perfWmiClass;
+    protected final String queryKey;
     protected CounterDataSource source;
     protected PerfCounterQueryHandler pdhQueryHandler;
     protected WmiQueryHandler wmiQueryHandler;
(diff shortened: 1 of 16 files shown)


dbwiddis:
This pull request **introduces 1 alert** and **fixes 4** when merging 08643b631a3409b70ab44e8ebdd33c8dde2822d6 into 34ff1c3e925294b0d4b03cc93a7a9d5ba60fc6ef - [view on LGTM.com](https://lgtm.com/projects/g/oshi/oshi/rev/pr-d328fe8b035ba221b3dc28b1d872305f6cfb8ab9)

**new alerts:**

* 1 for Spurious Javadoc @param tags

**fixed alerts:**

* 2 for Result of multiplication cast to wider type
* 2 for Dereferenced variable may be null

---

*Comment posted by [LGTM.com](https://lgtm.com)*

dbwiddis:
This pull request **fixes 4 alerts** when merging 7bddf8d0d743e6478d95fd4aecd9e365b85d0377 into 34ff1c3e925294b0d4b03cc93a7a9d5ba60fc6ef - [view on LGTM.com](https://lgtm.com/projects/g/oshi/oshi/rev/pr-0698dbefe9e9733a1bc56f38d5f9163aaf23a31e)

**fixed alerts:**

* 2 for Result of multiplication cast to wider type
* 2 for Dereferenced variable may be null

---

*Comment posted by [LGTM.com](https://lgtm.com)*

PR #811, opened 2019-03-02
ExecutingCommand fixed for windows

BilalAM:


BilalAM:
PR for issue #812

oshi-core/src/main/java/oshi/util/ExecutingCommand.java
```diff
@@ -50,7 +51,13 @@ private ExecutingCommand() {
      *         empty string if the command failed
      */
     public static List<String> runNative(String cmdToRun) {
-        String[] cmd = cmdToRun.split(" ");
+        String[] cmd;
+        if(Platform.isWindows() || Platform.isWindowsCE()){
+                // we have windows
+            cmd = new String[]{"cmd.exe","/c",cmdToRun};
```
dbwiddis: If `cmdToRun` has spaces, this doesn't split them out to an array as intended.  I believe there are some edge case argument lists where this would fail.

oshi-core/src/main/java/oshi/util/ExecutingCommand.java
```diff
@@ -50,7 +51,13 @@ private ExecutingCommand() {
      *         empty string if the command failed
      */
     public static List<String> runNative(String cmdToRun) {
-        String[] cmd = cmdToRun.split(" ");
+        String[] cmd;
+        if(Platform.isWindows() || Platform.isWindowsCE()){
```
dbwiddis: Should check based on `SystemInfo.getCurrentPlatformEnum()`.

oshi-core/src/main/java/oshi/util/ExecutingCommand.java
dbwiddis: Better yet, if we do include this we should include it in the array version, not this version.  And check that the string already being passed in doesn't already start with `cmd.exe /c` like the test case does, and this incorrectly adds on to.

BilalAM:
Closing this after discussion with @dbwiddis on #812

Issue #813, opened 2019-03-02
Return type checking for PerfCounterQuery is overly broad


dbwiddis:
I'm marking this issue for [first-timers-only](https://www.firsttimersonly.com/).  That means that I will only accept a PR for this one from someone who's never contributed to open source before. This one is easy (but don't make that statement make you feel bad if you have a hard time with it, there's more to contributing to open source than changing lines of code, especially if it's your first time). I'll hold your hand through this if you need me to. :-) Here are the steps to get a PR merged here.

- [ ] Read the [CONTRIBUTING](https://github.com/oshi/oshi/blob/master/CONTRIBUTING.md) file for reference.  I'll walk you through anything you don't understand.
- [ ] Find the [PerfCounterWildcardQuery.java](https://github.com/oshi/oshi/blob/3086b010758404cc4dcda3e0105c38b1877c8ac4/oshi-core/src/main/java/oshi/data/windows/PerfCounterWildcardQuery.java#L250) class.
- [ ] Line 250 parses a returned WMI result to a `Long` based on the variable's CIM Type (CIM_TYPE). This is a narrow check that assures correct parsing. This is good!
- [ ] Now find the [PerfCounterQuery.java](https://github.com/oshi/oshi/blob/1ea3575239c05ce75a71bbd8739fa691a3b672d2/oshi-core/src/main/java/oshi/data/windows/PerfCounterQuery.java#L235) class.
- [ ] Line 235 parses a returned WMI result to a `Long` based on the variable's Variant Type (VT_TYPE). This is OK for the I2 and I4 (2-bit and 4-bit integer) types but is overly broad on the BSTR type used for the UINT64 and would permit a developer to incorrectly map a string result to our Long-typed counters and get a useless "0" result without receiving any warnings.
- [ ] Alter the VT Type switch and cases starting at line 235 to match the preferred CIM Type check instead of VT Type.
- [ ] Bonus points: if you can figure out a nice way to refactor and reduce duplication, go for it!
- [ ] Fork the project, commit your changes to your fork, create a PR, and get it merged
- [ ] Celebrate

PR #814, opened 2019-03-04
changing the result from VT_type to CIM_Type in method queryWmi of Pe…

anwesh408:
…rfCounterQuery.java


dbwiddis:
Looks great! Thanks for contributing to open source!  Fixes #813

PR #830, opened 2019-04-03
[mvn] Update maven wrapper to 0.5.3

hazendaz:


mvnw.cmd
```diff
@@ -1,161 +1,172 @@
...
-if exist "%HOME%\mavenrc_post.cmd" call "%HOME%\mavenrc_post.cmd"
-:skipRcPost
-
-@REM pause the script if MAVEN_BATCH_PAUSE is set to 'on'
-if "%MAVEN_BATCH_PAUSE%" == "on" pause
-
-if "%MAVEN_TERMINATE_CMD%" == "on" exit %ERROR_CODE%
-
-exit /B %ERROR_CODE%
+@REM ----------------------------------------------------------------------------
```
dbwiddis: Lots of whitespace changes, probably unix vs. windows line endings. Can you confirm the new file uses unix line endings?

hazendaz: These files come straight from [maven team](https://github.com/takari/maven-wrapper).  In windows they will have windows line endings due to git attributes set to auto and in *nix they will have unix styled endings.  Since this is used for travis only, it's not something to be too overly concerned with when the CI system pulls it out as the build already showed.  For windows you would in fact want the cmd file to be windows line endings.  If using git bash in windows with maven wrapper which some might have a use case for, one would need to run dos2unix against a number of files to make it work but generally there isn't such a need.

hazendaz: Also should have noted, this latest bump in releases from maven team, they did a lot of work to clean things up, fixing license headers to match other projects, etc.

hazendaz: And just realized - no gitattributes on this project...  So at the moment, you are accurate in that this file is coming in wrong.  Do you recall why there is no gitattribute file?  I tend to have those on all projects.  Not sure if that was just missed here...

dbwiddis: Ah, that makes sense to have a .cmd file have windows line endings!  And I have no idea about gitattribute files...

dbwiddis:
I'm okay with you merging this unless you think it needs something!

hazendaz:
Possibly need to add git attributes at some point and run some normalization but that is generally hit or miss.  For now, this is good enough and may be better given no git attributes file.

Issue #843, opened 2019-04-26
Wrong time unit used in LinuxCentralProcessor


drdwo:
There is a problem with 
https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java

In the documentation of 
long[][] getProcessorCpuLoadTicks();

we have:

"Note that while tick counters are in units of milliseconds, they may advance ..."

But the actual time unit is different on Linux, implementation in 
https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
is

List<String> procStat = FileUtil.readFile("/proc/stat");
       for (String stat : procStat) {
           if (stat.startsWith("cpu") && !stat.startsWith("cpu ")) {
               // Split the line. Note the first (0) element is "cpu" so
               // remaining
               // elements are offset by 1 from the enum index
               String[] tickArr = ParseUtil.whitespaces.split(stat);

And from 
http://www.linuxhowtos.org/System/procstat.htm we know:
"Time units are in USER_HZ or Jiffies (typically hundredths of a second)."

So either LinuxCentralProcessor should use USER_HZ to convert the data in ms, or
the documentation needs to be adapted. I would prefer an OS-independent fixed time unit.

dbwiddis:
This is a case where I wrote the documentation before I changed the code, and then forget to change the code. :grin:

dbwiddis:
I'm marking this issue for [first-timers-only](https://www.firsttimersonly.com/).  That means that I will only accept a PR for this one from someone who's never contributed to open source before. This one is easy (but don't make that statement make you feel bad if you have a hard time with it, there's more to contributing to open source than changing lines of code, especially if it's your first time). I'll hold your hand through this if you need me to. :-) Here are the steps to get a PR merged here.

Linux tick counters return values in "Jiffies".  In response to #782, we decided to standardize all tick return values in milliseconds.  Your mission, should you choose to accept it, is to convert the Linux Jiffies to Milliseconds.

- [ ] Read the [CONTRIBUTING](https://github.com/oshi/oshi/blob/master/CONTRIBUTING.md) file for reference.  I'll walk you through anything you don't understand.
[...]

PR #844, opened 2019-04-27
#843: Convert Jiffies to Milliseconds

yudyack:
Solves #843 Wrong time unit used in LinuxCentralProcessor


dbwiddis:
Looks great, thanks for contributing to open source!

Issue #854, opened 2019-05-09
getCpuTemperature() at CentOS 7 seems to give always 0.0


rlouwerens:
I'm trying to get the current CPU temperature but always get zero's back from the getCpuTemperature() method.

We use version 3.12.2 but the same occurs with 3.12.0 (no other versions tried).

Files temp1_input until temp6_input exists at /sys/class/hwon/hwon0/ exists and holds values like 43000.

From logging I can see the following from FileUtil:
"Reading file temp3_input" 
"Reading file temp4_input" 
"Reading file temp5_input" 
"Reading file temp1_input" 
"Reading file temp2_input" 
Not sure, but it looks like the FileUtil is not getting the pull path and only the sensor?

Are any issues known with this setup?

Environment:
<artifactId>oshi-core</artifactId>
<groupId>com.github.oshi</groupId>
<version>3.13.2</version>

Centos 7.6.1810

dbwiddis:
OSHI's logic is to use the value in `temp1_input` if it exists; if not, average `temp2_input` thru `temp6_input`.  Can you tell me what the value in `temp1_input` is?

dbwiddis:
> Not sure, but it looks like the FileUtil is not getting the pull path and only the sensor

Yep, `LinuxSensors.java` line 66 does `FileUtil.getIntFromFile(f.getName())` but it needs to be the full path/name of the file on that line.  Can you test this fix and possibly submit a PR?

rlouwerens:
> > Not sure, but it looks like the FileUtil is not getting the pull path and only the sensor
> 
> Yep, `LinuxSensors.java` line 66 does `FileUtil.getIntFromFile(f.getName())` but it needs to be the full path/name of the file on that line. Can you test this fix and possibly submit a PR?

I can verify (with unit tests) that changing the check `&& FileUtil.getIntFromFile(f.getName()) > 0;` at line 66 to `&& FileUtil.getIntFromFile(f.getAbsolutePath()) > 0;`  fixes the issue. 

I changed to unit test SensorsTest to verify the value is greater then 0 in stead of >=:
`assertTrue(s.getCpuTemperature() > 0d && s.getCpuTemperature() <= 100d);`

dbwiddis:
Thanks for the test verification... I see you closed the PR, were you going to rework it?  A few comments:
[...]

PR #857, opened 2019-05-13
Fixed reading of sensor data

rlouwerens:
Changed filename for sensor to full path. #854
Changed unit test for sensors (based on availability of sys/class path)

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxSensors.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxSensors.java
@@ -62,8 +62,13 @@ public class LinuxSensors implements Sensors {
                 // Find any *_input files in that path
                 @Override
                 public boolean accept(File f) {
-                    return f.getName().startsWith(sensorPrefix) && f.getName().endsWith("_input")
-                            && FileUtil.getIntFromFile(f.getName()) > 0;
+                    try {
+                        return f.getName().startsWith(sensorPrefix) && f.getName().endsWith("_input")
+                                && FileUtil.getIntFromFile(f.getCanonicalPath()) > 0;
+                    }
+                    catch(Exception e) {
+                        return false;
+                    }
                 }
             });
         }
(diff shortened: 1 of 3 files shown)


oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxSensors.java
```diff
@@ -62,8 +62,13 @@ public LinuxSensors() {
                 // Find any *_input files in that path
                 @Override
                 public boolean accept(File f) {
-                    return f.getName().startsWith(sensorPrefix) && f.getName().endsWith("_input")
-                            && FileUtil.getIntFromFile(f.getName()) > 0;
+                    try {
+                        return f.getName().startsWith(sensorPrefix) && f.getName().endsWith("_input")
+                                && FileUtil.getIntFromFile(f.getCanonicalPath()) > 0;
+                    }
+                    catch(Exception e) {
```
dbwiddis: I don't like catching all exceptions here. Can we narrow this to the actual exception thrown?  And log a message other than just returning?

oshi-core/src/test/java/oshi/hardware/SensorsTest.java
```diff
@@ -1,9 +1,9 @@
 /**
  * OSHI (https://github.com/oshi/oshi)
- *
+ * <p>
```
dbwiddis: Good idea to include these, but this header will be overwritten by the maven license plugin when snapshot is deployed.   OK to leave these, for now.

oshi-core/src/test/java/oshi/hardware/SensorsTest.java
```diff
@@ -23,28 +23,40 @@
...
-import static org.junit.Assert.assertTrue;
-
 import org.junit.Test;
-
 import oshi.SystemInfo;
 
+import java.io.File;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeFalse;
```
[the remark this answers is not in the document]
dbwiddis: Codacy says this is an unused import.

dbwiddis:
@rlouwerens Can you address the comments in this PR?  All should be very quick fixes, then I can merge this.

Issue #869, opened 2019-05-20
Thread safety


Nadahar:
We've been using OSHI for a long time for logging system information, but we now wanted to take it a step further and start using the information gathered by OSHI for decisions/evaluations in the code.

I took a quick look in the code and can see that there is not handling of thread-safety in OSHI at all as far as I can see. Unfortunately our application is multi-threaded, and considering how slow it is to gather all the information needed, it's not really viable to do that every time an evaluation using this information is due.

Has this at all been considered? It shouldn't be that hard to use immutable classes where possible and use synchronization on the rest (all classes with setters). It will require changes to most classes though, so it's a comprehensive in that sense.

I could consider making a PR for this, but unfortunately I'm still stuck at 3.x (since we still support Java 7) and I doubt you'd be interested in merging something like that in 3.x? Also, I don't know if you'd be interested in having thread-safety in the first place, although I think it would make a lot of sense because of the time it takes to gather the information.

dbwiddis:
I'd love to make OSHI more thread safe, and would consider updating the 3.X branch.  Unfortunately I personally don't have the expertise nor bandwidth to make that happen.  See the discussion on #797.  If you are able to make the necessary changes I'd be happy to merge them.

Nadahar:
Good to hear. I won't be right now, but then I know that I have a path to do this when I get there :+1:

dbwiddis:
Closing this as we will implement thread safety in 4.1.0.

I understand you're still on JDK7 and the 3.x branch, but I think the work involved to update 3.x exceeds the work that it would take to backport 4.x to Java 7.  And if you think that's a useful endeavor, I invite you to open a new issue about it!

PR #870, opened 2019-05-22
Thread safety examples

Nadahar:
I wrote in #869 that I would consider making a PR implementing thread-safety. When looking at it closer, it seems that this is too much work for me. The reason for my "miscalculation" as to how much work it would be, is that I always create classes "defensively" when it comes to concurrency, regardless of if they are intended to be thread-safe or not. I forgot that not everybody does this :wink: 

The problem is the way the classes are initialized. Since there are performance penalties with synchronization (even though they are over-emphasized in many situations), I prefer immutability when possible. Immutable classes and/or fields means that you don't have to think about concurrency when using them, and there's no performance penalty. There might be a small penalty when constructing them, but often there's not or it's negligible. The essential challenge is that every final field must be set in the constructor. Java has some unfortunate limitations when it comes to how constructors work, which is why "factories" are often used instead to create immutable instances.

Thread-safety can be solved in many different ways. I've made implementations for `Baseboard` and `CentralProcessor` as examples of how I would solve this without making it too complicated.
I think the system of "initializers" I made up here would work fine for most of the classes in this project. They are basically very simplified builders that sub-classes can initialize, which is then used to set the final fields in the constructor. 

For `Baseboard` the implementation is completely immutable and all is handled using a such "initializer". `CentralProcessor` can't be immutable because some real-time data can be updated. I've implemented every field that can be immutable using an "initializer" like in `Baseboards`. The rest I've simply protected with a simple synchronization object. There were some `synchronized` statements in some of the methods, which I removed. The reason for this is two-fold: First, I prefer not to expose the synchronization object publicly unless there's a good reason to do so. This way we know that "we" (including sub-classes) control the synchronization object. Second, fields protected by a synchronization object *must* be protected everywhere it's used, or the synchronization is pointless.

For `Baseboard` I think this implementation is uncontroversial. The only change (seen from the outside), is that the setters are gone from `AbstractBaseboard`. I find it very unlikely that client code would call these setters, unless somebody has made their own sub-class implementation. While the same can be argued for `CentralProcessor`, the setters were in the interface here, which means I had to remove them from the interface itself. I consider that a "more serious" breach of the API, although I don't know if it really matters in practical use of the library.

In any case, I think doing this for all the classes will be too time-consuming for me, so I chose to simply create this PR to show how I would go about doing it. I don't intend for this to be merged, as it would add little value making these two classes alone thread-safe.

When it comes to our use-case and the need for a thread-safe instance that could be used for evaluations, I think the only practical solution is for me to make a thread-safe class that only contains the information we actually need, and then use OSHI to populate it from a separate thread during startup.

### DIFF
--- a/oshi-core/src/main/java/oshi/SystemInfo.java
+++ b/oshi-core/src/main/java/oshi/SystemInfo.java
@@ -52,6 +52,8 @@ public class SystemInfo implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
+    public static final String UNKNOWN = "unknown";
+
     private OperatingSystem os = null;
 
     private HardwareAbstractionLayer hardware = null;
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -101,27 +101,11 @@ public interface CentralProcessor extends Serializable {
     String getVendor();
 
     /**
-     * Set processor vendor.
-     *
-     * @param vendor
-     *            Vendor.
-     */
-    void setVendor(String vendor);
-
-    /**
      * Name, eg. Intel(R) Core(TM)2 Duo CPU T7300 @ 2.00GHz
      *
      * @return Processor name.
      */
     String getName();
-
-    /**
-     * Set processor name.
-     *
-     * @param name
-     *            Name.
-     */
-    void setName(String name);
 
     /**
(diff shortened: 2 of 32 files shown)


Nadahar:
By the way, I didn't update the tests, because I wasn't sure exactly why the setters were tested. It's obviously not possible to test the setters when they're no longer there. Thus, the tests will fail as this PR is currently.

dbwiddis:
Thanks for these examples!  I've implemented your suggestions or an alternative in #959 so I'll close this PR.

PR #872, opened 2019-05-23
Add vm mac address for Docker container.

KyongSik-Yoon:
I refer to genMAC method on https://github.com/docker/libnetwork/blob/master/netutils/utils.go

### DIFF
--- a/oshi-demo/src/main/java/oshi/demo/DetectVM.java
+++ b/oshi-demo/src/main/java/oshi/demo/DetectVM.java
@@ -25,6 +25,7 @@ package oshi.demo;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import oshi.SystemInfo;
 import oshi.hardware.HardwareAbstractionLayer;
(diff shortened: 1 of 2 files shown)


KyongSik-Yoon:
@dbwiddis ping

oshi-demo/src/main/java/oshi/demo/DetectVM.java
```diff
@@ -100,4 +104,11 @@ public static String identifyVM() {
         // Couldn't find VM, return empty string
         return "";
     }
+
+    static String findOuiByMacAddressIfPossible(String mac) {
```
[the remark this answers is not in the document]
dbwiddis: Can you make this public or private?

oshi-demo/src/main/java/oshi/demo/DetectVM.java
```diff
@@ -47,11 +49,12 @@
         vmMacAddressOUI.put("00:0F:4B", "Virtual Iron 4");
         vmMacAddressOUI.put("00:16:3E", "Xen or Oracle VM");
         vmMacAddressOUI.put("08:00:27", "VirtualBox");
+        vmMacAddressOUI.put("02:42", "Docker Container");
```
dbwiddis: An OUI is 24 bits, so this should be `02:42:AC`.  Your source only uses the first two bytes because the remaining 4 bytes are drawn from the IP address; however, the Docker IP range is `172.17.0.0/16` meaning the MAC always starts `02:42:AC:11` and only the last 16 bits vary.

KyongSik-Yoon: @dbwiddis Thanks for your feedback.
I push some modification.

PR #880, opened 2019-06-02
s/dblock/oshi/g

dblock:
Out with this dblock guy ;)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,16 +1,16 @@
 Contributing to Oshi
 =====================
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -21,7 +21,7 @@ env:
 addons:
   coverity_scan:
     project:
-      name: "dblock/oshi"
+      name: "oshi/oshi"
```
dbwiddis: Actually, our account on Coverity is [tied to the older project name](https://scan.coverity.com/projects/dblock-oshi?tab=overview).  I'd have to create a new Coverity account and lose project history to change this. 
image
While there are settings to change the project coordinates (which I've done) the name can apparently not be changed.

dblock: I undid this.

PR #887, opened 2019-06-06
WTSEnumerateProcessesEx need minimum window 7

kthaya:
WTSEnumerateProcessesEx need minimum window 7


dbwiddis:
Oops, merged this too fast.   Need to also change the checks later on in the method that depend on the initial setting.  I'll add that directly.

Issue #865, opened 2019-05-19
Map Intel Processor Identifier to CPU name


Sami32:
https://github.com/hirschmann/nbfc/blob/master/Core/Plugins/OpenHardwareMonitor/Hardware/CPU/IntelCPU.cs

As it is my wondering day ;-)
That could be only me, but having a name like "Haswell" for an Intel CPU, for example, wouldn't be more easily readable?

https://github.com/hirschmann/nbfc/tree/master/Core/Plugins/OpenHardwareMonitor/Hardware/CPU

Sami32:
It came to my mind because of this documentation for hardware acceleration:

https://en.m.wikipedia.org/wiki/Intel_Quick_Sync_Video#Hardware_decoding_and_encoding

dbwiddis:
This looks like a good add-in to toss into the `oshi-demo` project, as it appears the naming can be done by parsing the processor identifier.  It's similar to the "is this a VM" method.

Note you can't just port the code directly over, as it's MPL.  So we'd have to independently write the source code based on [other documentation](https://software.intel.com/en-us/articles/intel-architecture-and-processor-identification-with-cpuid-model-and-family-numbers).

Feel free to contribute!

tbradellis:
I can work this one.  Looking at the links above I'm not seeing where microarchitecture codename is actually stored as part of CPU info. Does this imply that we'd want to maintain a lookup based on  the full model string?  Or, am I overlooking  something.  I think I can run with this if I get pointed in the right direction.

dbwiddis:
The way I thought to do it was to map the appropriate bits of the Processor ID to the processor using the tables at [this link](https://software.intel.com/en-us/articles/intel-architecture-and-processor-identification-with-cpuid-model-and-family-numbers) (you'd need to hard code the lookup).

Actually, we're not guaranteed the endianness of the Processor ID so it'd be better to use the `getFamily()` and `getModel()` fields.

dbwiddis:
Actually the Intel link in the previous comment looks a bit dated.  [Here's another option](https://en.wikichip.org/wiki/intel/cpuid).

tbradellis:
I should have at least a first pass at a PR this week. Had some family in from out of town this weekend so playing tour guide until tomorrow morning.

Sami32:
For olders CPUs that are missing i found some more informations into these links:

http://www.paradicesoftware.com/specs/cpuid/  (7)
[...]

PR #894, opened 2019-06-13
#865 Lookup microarchitecture from processor identifier

tbradellis:
This is the direction I'm thinking for this implementation.  Just getting the initial idea posted before going further. I appreciate any feedback or pointers if some other direction seems better.

Ultimately, all codenames would be maintained in the ArchitectureCodename enum with the corresponding family and model as a string.  I considered using an interface and breaking out an enum for each chip type.


oshi-core/src/main/java/oshi/util/ArchitectureCodename.java
```diff
@@ -0,0 +1,36 @@
...
+     */
+
+
+public enum ArchitectureCodename {
+
+    /**
+     * Intel
+     */
+
+    SkyLake("6 94");
```
dbwiddis: This won't work as an enum.  Skylake (Client) has family 6 but models 78 and 94, Skylake (Server) is model 85.

You also may not want to hardcode these values -- that requires releasing OSHI every time a new processor comes out.  I'd suggest a key-value map based on the family/model, using a config file that can be updated by the user if desired.

oshi-core/src/main/java/oshi/util/ArchitectureCodename.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private String familyModel;
+
+    ArchitectureCodename(String familyModel){
+        this.familyModel = familyModel;
```
dbwiddis: depending on how you implement this you may want to separate family and model to two variables, as that's what data you'll have available from OSHI. (If you use as a lookup key you'll still end up combining them, but you could do that internally.)

dbwiddis:
I'm going to close this PR as it's been 8 months, it's not really mergeable as-is, and I think I've provided enough feedback for a submission the next time around!

dbwiddis:
I'm going to reopen this and work on it... major rewrite, though! :)

Issue #888, opened 2019-06-06
Global Config File


dbwiddis:
Create a global configuration file with the following specifications:
 - store the file in `src/main/resources`
 - easily read the file into a Java `Properties` object
 - have a method in the `SystemInfo` class to get the current `Properties` object
   - <s>Object is mutable, can get/set properties on the object.</s>
 - as an initial implementation
   - set a global default WMI timeout property and make it the default in the `WmiUtil` class.
     - If property not set or negative, defaults to infinite
   - set a global path for `/proc` filesystem and make it a constant in the `ProcUtil` class.
     - If property not set, defaults to `/proc`.

cilki:
The method that yields the `Properties` object needs to be `static` so users can modify the default configuration before the `SystemInfo` is created. OSHI classes that use configuration values should consume them on initialization rather than repeatedly querying the `Properties` object. This prevents the user from dynamically updating the values (which I think is ideal).

Also, naming conventions like `oshi.example.value` should be used throughout the config.

Otherwise this sounds good and I can make a PR when I have time.

dbwiddis:
One of the paradigms I was trying to get at with 4.x is sort of a "poor man's thread safety" where if you create a new `SystemInfo` object, everything you get from that is a unique object.  Static properties would end up being shared across threads and require jumping through some more hoops to be thread safe, unless we made the object immutable.

But what if I want to change these global defaults (e.g., WMI default timeout) at runtime?  Recall some of our users will be using OSHI in long-running systems that don't reboot the JVM.

dbwiddis:
The `.property` extension is redundant. The file itself will be `oshi.default.properties`.  I do like a dotted convention that corresponds to the class name and attribute/getter name where it's read on initialization, though, e.g., `oshi.util.platform.os.Windows.WmiTimeout = 500`.

dbwiddis:
Another (probably better) option to SystemInfo is a separate `OshiProperties` class.

cilki:
> The .property extension is redundant

I should have given a real example. I mean that each property name in the config will be in dotted notation exactly like you said. So by `oshi.example.property` I meant something like `oshi.wmi.timeout`. Sorry for the confusion.

cilki:
> But what if I want to change these global defaults (e.g., WMI default timeout) at runtime?

[...]

PR #896, opened 2019-06-16
Add global configuration utility

cilki:
Here's a simple attempt at #888. I didn't implement anything like `reloadUtilPropertiesFromFile` because the user won't be modifying the `default.properties` on the classpath.


oshi-core/src/main/java/oshi/util/ConfigUtil.java
```diff
@@ -0,0 +1,164 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(ConfigUtil.class);
+
+    private static final Properties configuration = new Properties();
+
+    static {
+        try (InputStream in = ConfigUtil.class.getResourceAsStream("/default.properties")) {
+            if (in != null)
+                configuration.load(in);
+            else
+                LOG.warn("Default configuration not found");
```
dbwiddis: Not sure we need a log warning on no default configuration.  Info at the highest, or perhaps trace.

oshi-core/src/main/java/oshi/util/ConfigUtil.java
```diff
@@ -0,0 +1,164 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The global configuration utility.
+ * 
+ * @author cilki
+ * @since 4.0.0
+ */
+public final class ConfigUtil {
```
dbwiddis: I know I probably suggested this name earlier, but I think given the methods you've developed a name like `GlobalConfig` would be more meaningful.  `GlobalConfig.set("oshi.util.wmi.timeout", 500);` is self-documenting.

oshi-core/src/main/java/oshi/util/ConfigUtil.java
```diff
@@ -0,0 +1,164 @@
...
+        return value == null ? def : Boolean.parseBoolean(value);
+    }
+
+    /**
+     * Set the given property, overwriting any existing value.
+     * 
+     * @param key The property key
+     * @param val The new value
+     */
+    public static void set(String key, Object val) {
```
dbwiddis: I like the `toString()` approach here.  Will it throw an exception on `val = null`?   Do we want to allow null and invoke `remove()` in that case?

oshi-core/src/main/java/oshi/util/ConfigUtil.java
```diff
@@ -0,0 +1,164 @@
...
+        configuration.setProperty(key, val.toString());
+    }
+
+    /**
+     * Load the given {@link Properties} into the global configuration.
+     * 
+     * @param properties The new properties
+     */
+    public static void load(Properties properties) {
+        configuration.putAll(properties);
```
dbwiddis: I see a way to add more properties but not to `remove()` them.  Or `removeAll()`...

oshi-core/src/main/java/oshi/util/ConfigUtil.java
```diff
@@ -0,0 +1,164 @@
...
+    /**
+     * Get the {@code Double} property associated with the given key.
+     * 
+     * @param key The property key
+     * @param def The default value
+     * @return The property value or the default if not found
+     */
+    public static Double getDouble(String key, double def) {
+        String value = configuration.getProperty(key);
+        return value == null ? def : Double.parseDouble(value);
```
dbwiddis: Should be able to use `ParseUtil.parseDoubleOrDefault(value, def)` here.

oshi-core/src/main/java/oshi/util/ConfigUtil.java
```diff
@@ -0,0 +1,164 @@
...
+    /**
+     * Get the {@code Integer} property associated with the given key.
+     * 
+     * @param key The property key
+     * @param def The default value
+     * @return The property value or the default if not found
+     */
+    public static int getInteger(String key, int def) {
+        String value = configuration.getProperty(key);
+        return value == null ? def : Integer.parseInt(value);
```
dbwiddis: Should be able to use `ParseUtil.parseIntOrDefault(value, def)` here.

oshi-core/src/main/java/oshi/util/ConfigUtil.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * Get the {@code String} property associated with the given key.
+     * 
+     * @param key The property key
+     * @return The property value or {@code null} if not found
+     */
+    public static String get(String key) {
+        return configuration.getProperty(key);
```
dbwiddis: This seems to be the only place to check whether a property even exists (`null` return if not).  I wonder if we really need the other convenience getter methods (particularly the ones without user-supplied defaults), since they  essentially do the same thing a user could do (but without giving the user any indication of whether the value was actually the default).

Java 8+ adds the ability for a default property list so we could actually guarantee never to get null....

oshi-core/src/main/java/oshi/util/ConfigUtil.java
```diff
@@ -0,0 +1,164 @@
...
+        return configuration.getProperty(key, def);
+    }
+
+    /**
+     * Get the {@code Integer} property associated with the given key.
+     * 
+     * @param key The property key
+     * @return The property value or {@code 0} if not found
+     */
+    public static int getInteger(String key) {
```
dbwiddis: Not sure we need this.

oshi-core/src/main/java/oshi/util/ConfigUtil.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * Get the {@code Integer} property associated with the given key.
+     * 
+     * @param key The property key
+     * @param def The default value
+     * @return The property value or the default if not found
+     */
+    public static int getInteger(String key, int def) {
```
dbwiddis: Since the `String` getter doesn't include the type, why not just use `get()` here as well?  The `int` default argument will ensure this one is called.

oshi-core/src/main/java/oshi/util/ConfigUtil.java
```diff
@@ -0,0 +1,164 @@
...
+        return value == null ? def : Integer.parseInt(value);
+    }
+
+    /**
+     * Get the {@code Double} property associated with the given key.
+     * 
+     * @param key The property key
+     * @return The property value or {@code 0.0} if not found
+     */
+    public static Double getDouble(String key) {
```
dbwiddis: Not sure we need this.

oshi-core/src/main/java/oshi/util/ConfigUtil.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * Get the {@code Double} property associated with the given key.
+     * 
+     * @param key The property key
+     * @param def The default value
+     * @return The property value or the default if not found
+     */
+    public static Double getDouble(String key, double def) {
```
dbwiddis: Since the `String` getter doesn't include the type, why not just use `get()` here as well?

oshi-core/src/main/java/oshi/util/ConfigUtil.java
```diff
@@ -0,0 +1,164 @@
...
+        return value == null ? def : Double.parseDouble(value);
+    }
+
+    /**
+     * Get the {@code Boolean} property associated with the given key.
+     * 
+     * @param key The property key
+     * @return The property value or {@code false} if not found
+     */
+    public static boolean getBoolean(String key) {
```
dbwiddis: Not sure we need this.

oshi-core/src/main/java/oshi/util/ConfigUtil.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * Get the {@code Boolean} property associated with the given key.
+     * 
+     * @param key The property key
+     * @param def The default value
+     * @return The property value or the default if not found
+     */
+    public static boolean getBoolean(String key, boolean def) {
```
dbwiddis: Since the `String` getter doesn't include the type, why not just use `get()` here as well?

oshi-core/src/main/java/oshi/util/platform/linux/ProcUtil.java
```diff
@@ -42,14 +43,27 @@
...
     private ProcUtil() {
     }
+    
+    /**
+     * The /proc filesystem location.
+     */
+    private static String proc = ConfigUtil.get("oshi.util.proc.path");
+
+    static {
+        // Ensure prefix begins with path separator, but doesn't end with one too
```
dbwiddis: Do we want to do any sanity checking of whether this path exists?

oshi-core/src/main/java/oshi/util/platform/windows/WmiQueryHandler.java
```diff
@@ -39,12 +39,14 @@
...
 
+import oshi.util.ConfigUtil;
+
 public class WmiQueryHandler {
 
     private static final Logger LOG = LoggerFactory.getLogger(WmiQueryHandler.class);
 
     // Timeout for WMI queries
-    private int wmiTimeout = Wbemcli.WBEM_INFINITE;
+    private int wmiTimeout = ConfigUtil.getInteger("oshi.util.wmi.timeout");
```
dbwiddis: I'd like to set the default in this class, either through getting the string and parsing here (with default) or passing the default to the getter.   Do we want to sanity check (positive value)?

oshi-core/src/main/resources/default.properties
```diff
@@ -0,0 +1,26 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+
+oshi.util.proc.path=/proc
```
dbwiddis: Let's add some comments above these to describe their use cases (and that -1 = infinite for WMI)

oshi-core/src/main/java/oshi/util/ConfigUtil.java
cilki: The only way the `default.properties` resource could not be found is if it's not included in the artifact (or the classloader is broken). However, if we require a default to be supplied to every `get`, then not finding `default.properties` is not so serious. I'll remove the condition altogether.

oshi-core/src/main/java/oshi/util/ConfigUtil.java
cilki: I can't think of a reason to undefine a property, but calling `remove()` is probably more useful than throwing `NullPointerException`.

dbwiddis: My use case was that by clearing custom user behavior, we would restore default behavior.  Like if I set a custom WMI timeout and then wanted to remove my custom application, it would automatically revert to the infinite, without me explicitly having to *set* infinite.

I think a common use case for property files is for users to comment-out properties that aren't used/overridden, implying those would be the defaults... so in the "normal" situation there actually wouldn't be any properties on the object, and allowing removal helps restore it to its original state.

cilki: Calling `remove()` won't restore the global default though.

dbwiddis: Remove will remove the property; so the behavior when `get()` is called will go back to "returning `null`" and restoring whatever the original "default" behavior was.

In the WMI case, with the "factory" implementation, the new global will be set on every new WMI initialization, and actually change behavior.  In the proc case, since that value is never read again, it won't change anything.

oshi-core/src/main/java/oshi/util/platform/windows/WmiQueryHandler.java
cilki: Should `getWmiTimeout()` and `setWmiTimeout()` be removed now?

dbwiddis: > Should `getWmiTimeout()` and `setWmiTimeout()` be removed now?

No!  The setter is necessary to override the default on an individual query basis.

oshi-core/src/main/java/oshi/util/GlobalConfig.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * Get the {@code Double} property associated with the given key.
+     * 
+     * @param key The property key
+     * @param def The default value
+     * @return The property value or the given default if not found
+     */
+    public static Double get(String key, double def) {
```
dbwiddis: Why are we returning the boxed Double?

oshi-core/src/main/java/oshi/util/GlobalConfig.java
```diff
@@ -0,0 +1,166 @@
...
+     * @param key The property key
+     * @param def The default value
+     * @return The property value or the given default if not found
+     */
+    public static String get(String key, String def) {
+        return configuration.getProperty(key, def);
+    }
+
+    /**
+     * Get the {@code Integer} property associated with the given key.
```
dbwiddis: You indicate the boxed type here but return value is primitive (which I like).  Same comment on the `Boolean`.

oshi-core/src/main/java/oshi/util/GlobalConfig.java
```diff
@@ -0,0 +1,166 @@
...
+     */
+    public static class PropertyException extends RuntimeException {
+
+        private static final long serialVersionUID = -7482581936621748005L;
+
+        /**
+         * @param property The property name
+         */
+        public PropertyException(String property) {
+            super("Invalid property: \"" + property + "\" = " + GlobalConfig.get(property, null));
```
dbwiddis: with the overloaded methods, which one does the `null` call match?   Do we really need a `null` default, as opposed to the `get()` method with no default (which returns a String or null)?

oshi-core/src/main/java/oshi/util/GlobalConfig.java
```diff
@@ -0,0 +1,166 @@
...
+ */
+public final class GlobalConfig {
+
+    private static final Logger LOG = LoggerFactory.getLogger(GlobalConfig.class);
+
+    private static final Properties configuration = new Properties();
+
+    static {
+        try (InputStream in = GlobalConfig.class.getResourceAsStream("/default.properties")) {
+            if (in != null)
```
dbwiddis: General comment throughout -- my IDE will reformat all conditionals using curly brackets.   I won't hold up merging for this, but will probably follow up with a "reformat" commit to standardize the code style.

oshi-core/src/main/java/oshi/util/GlobalConfig.java
cilki: It will match the `String` version. On this particular line, I think `get(property, null)` is acceptable, but I can bring back the `get()` without a default if you want it.

oshi-core/src/main/java/oshi/util/GlobalConfig.java
cilki: Are you using a custom formatting profile? I'm just using the "Java Conventions" profile in Eclipse.

dbwiddis: Huh.  So, actually it doesn't fix it.  Maybe I got tired of seeing CheckStyle warnings on it or something.  Or it was drilled into me in some programming class somewhere that it's optional in the language but a good idea (except when playing code golf.)

cilki: I was just wondering if there's a style guide or a special xml formatting profile that I should be using for OSHI. Some projects keep the profile in the repo if they are using custom conventions ([for example](https://github.com/spring-projects/spring-batch/blob/master/spring-eclipse-code-conventions.xml)).

This is something that we should fix because I can't run my formatter on most existing source files without it changing a bunch of things.

dbwiddis: I uploaded my Eclipse format file here:
https://github.com/oshi/oshi/tree/master/oshi-dist/src/main/resources/eclipse
It's the "Sun Java Conventions" with one alteration:  spaces instead of tabs.

dbwiddis: And possibly 120 character line width.

cilki: Exactly what I was looking for. Thanks.

oshi-core/src/main/java/oshi/util/GlobalConfig.java
dbwiddis: OK, this makes sense and I'm ok with always requiring a default.

dbwiddis:
This looks good to me.  I'll give @spyhunter99 a chance to look it over before merging.

dblock:
Before someone hits merge, and I am so sorry to be late to this game, but has anyone considered using the widely adopted [cfg4j](http://www.cfg4j.org/) instead of rolling oshi's own?

dbwiddis:
> developers want to configure it along with other configuration properties

Thanks for the suggestion, @dblock ... definitely something to consider.  On the one hand I want to avoid another dependency, and we're really just manipulating a `Properties` object that can be set using external code.   On the other hand, a common use case for OSHI is monitoring remote/distributed systems and having a URL-based config is something I know firsthand the benefits of.

cilki:
@dblock We can search the classpath to find a file called `oshi.properties` (or similar) that developers can include in their applications to override the values in `default.properties`. This is roughly how [Logback](https://github.com/qos-ch/logback) (and probably many other applications) work.

That doesn't do much for remote situations, though. We'll have to determine whether it's worth another dependency.

dblock:
> @dblock We can search the classpath to find a file called `oshi.properties` (or similar) that developers can include in their applications to override the values in `default.properties`. This is roughly how [Logback](https://github.com/qos-ch/logback) (and probably many other applications) work.
> 
> That doesn't do much for remote situations, though. We'll have to determine whether it's worth another dependency.

Yes, we can implement anything, but it's a huge can of worms. I don't like dependencies, but for something so common and so riddled with corner cases I would recommend taking one. Configuration is definitely not business logic for oshi.

dblock:
Just to be clear, I only wanted to raise my concern, @dbwiddis looking forward to seeing what you decide!

dbwiddis:
It's a good concern to raise!

My concern with adopting a particular framework is whether it conflicts with another. If we choose `cfg4j` will it conflict with an application using `commons-config`, etc?   One of the design features of this simple config class is the ability to add/update/replace using a simple `Properties` object which should be compatible with any config source out there, including just creating a class programmatically.  

Other than JNA, OSHI's only dependency is `slf4j`, a logging facade. I might love `logback` for example, but OSHI uses `slf4j` facade to make sure anyone can use any logging framework.  So what we need is a config facade...  oh, hey, here's one: https://github.com/agentgt/configfacade

dblock:
If there's already a slf4j dependency I'd go with cfg4j :)

dbwiddis:
So I've been thinking about this a while and think I want to merge as is.  We can revisit before the release if this ends up being more complicated.

Issue #908, opened 2019-07-17
DetectVM - Wrong comparison of result


jgueth:
In the class **DetectVM** (package _oshi.demo_), the method _findOuiByMacAddressIfPossible_ returns an empty string if no entry was found in the _vmMacAddressOUI_ map. However, the result is checked to null (not empty String), which means that the method always ends after the MAC address check.

dbwiddis:
Not only the MAC check, but the first MAC check.  Looks like the error was introduced in #872.  Can you submit a PR to fix, or should I toss it out to first-timers?

dbwiddis:
I'm marking this issue for [first-timers-only](https://www.firsttimersonly.com/).  That means that I will only accept a PR for this one from someone who's never contributed to open source before. This one is easy (but don't make that statement make you feel bad if you have a hard time with it, there's more to contributing to open source than changing lines of code, especially if it's your first time). I'll hold your hand through this if you need me to. :-) Here are the steps to get a PR merged here.

- [ ] Read the [CONTRIBUTING](https://github.com/oshi/oshi/blob/master/CONTRIBUTING.md) file for reference.  I'll walk you through anything you don't understand.
- [ ] Find the [DetectVM.java](https://github.com/oshi/oshi/blob/fa069ec975c6c06014b895c390ffb111bbf74c9b/oshi-demo/src/main/java/oshi/demo/DetectVM.java#L87) class.  At line 87, the code only checks for a non null string. This is a bug, because the code should also check for an empty string.
- [ ] Alter the code to additionally check for an empty String.
- [ ] Fork the project, commit your changes to your fork, create a PR, and get it merged
- [ ] Celebrate

shannondavid:
Hi @dbwiddis. I haven't contributed to open source before and would like to give it a go. Do you mind if I contribute?

yzhang2907:
It has been done, this is my first time contributing to OS besides adding my name in the first-contribution github tutorial. It is PR #909. Thank you!

dbwiddis:
Hi, sorry @shannondavid  it looks like @yzhang2907 beat you to the punch here!  I can suggest a slightly more challenging issue for you to work on if you'd still like to help!

shannondavid:
@dbwiddis I'd love to!

dbwiddis:
[...]

PR #909, opened 2019-07-17
Fixed Issue #908, now empty string also checked.

yzhang2907:
Thank you for giving a first timer a chance.

### DIFF
--- a/oshi-demo/src/main/java/oshi/demo/DetectVM.java
+++ b/oshi-demo/src/main/java/oshi/demo/DetectVM.java
@@ -84,7 +84,7 @@ public class DetectVM {
         for (NetworkIF nif : nifs) {
             String mac = nif.getMacaddr().toUpperCase();
             String oui = findOuiByMacAddressIfPossible(mac);
-            if (oui != null) {
+            if (oui != null && oui != "") {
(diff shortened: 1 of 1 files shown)


oshi-demo/src/main/java/oshi/demo/DetectVM.java
```diff
@@ -84,7 +84,7 @@ public static String identifyVM() {
         for (NetworkIF nif : nifs) {
             String mac = nif.getMacaddr().toUpperCase();
             String oui = findOuiByMacAddressIfPossible(mac);
-            if (oui != null) {
+            if (oui != null && oui != "") {
```
dbwiddis: You shouldn't use `!=` as a comparison for Strings, as they are objects.  The correct equality comparison would be `"".equals(oui)`.  However, it is slightly more efficient to do `!oui.isEmpty()`.

yzhang2907:
@dbwiddis Thank you for pointing that out! My Java is a bit rusty, I feel dumb now...
Anyhow, I think I did it correctly this time. Thank you so much!

dbwiddis:
Looks great!  I'll merge later tonight!

PR #910, opened 2019-07-18
Added method getBootTime() for system start date

shannondavid:
Method returns number of seconds since last boot. Changes occur in CentralProcessor interface and respective <os>CentralProcessor classes.

### DIFF
--- a/oshi-demo/pom.xml
+++ b/oshi-demo/pom.xml
@@ -49,7 +49,7 @@
 
 	<properties>
 		<main.basedir>${project.parent.basedir}</main.basedir>
-		<jackson.version>2.9.9</jackson.version>
+		<jackson.version>2.9.9.1</jackson.version>
 	</properties>
 
 	<dependencies>


oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
```diff
@@ -257,13 +257,21 @@
...
+     */
+    long getBootTime();
 
     /**
      * Get the number of logical CPUs available for processing. This value may
      * be higher than physical CPUs if hyperthreading is enabled.
      *
      * @return The number of logical CPUs available.
      */
+
```
dbwiddis: (Extremely minor point) Lots of extra whitespace lines here.  Do you use a formatter in your IDE?  For the project we use the "Sun Java Conventions" except we do spaces instead of tabs.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
```diff
@@ -57,6 +57,21 @@
...
+    	// in /proc/stat.
+        List<String> procStat = FileUtil.readFile("/proc/stat");
+        for (String stat : procStat) {
+            if (stat.startsWith("btime")) {
+                String[] bTime = ParseUtil.whitespaces.split(stat);
+                BOOTTIME = ParseUtil.parseLongOrDefault(bTime[1], 0L);
+                break;
+            }
+        }
+
```
dbwiddis: You need to handle the case where the final variable isn't initialized in the loop.

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisCentralProcessor.java
```diff
@@ -52,6 +52,15 @@
...
 
     private static final Logger LOG = LoggerFactory.getLogger(SolarisCentralProcessor.class);
+    
+    private static final long BOOTTIME;
+    static {
+    	Kstat ksp = KstatUtil.kstatLookup("unix", 0, "system_misc");
+        if (ksp != null && KstatUtil.kstatRead(ksp)) {
+            BOOTTIME = KstatUtil.kstatDataLookupLong(ksp, "boot_time");
+            break;
+        }
```
dbwiddis: You need to handle the case where the final variable isn't initialized in the conditional.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsCentralProcessor.java
```diff
@@ -79,6 +79,28 @@
...
+                }
+                // First 6005; tentatively assign and look for EventID 12
+                BOOTTIME = record.getRecord().TimeGenerated.longValue();
+                found = true;
+            } else if (found && record.getRecord().EventID.getLow().intValue() == 12) {
+                // First 12 after 6005, this is boot time
+                BOOTTIME = record.getRecord().TimeGenerated.longValue();
+                break;
+            }
+        }
```
dbwiddis: You need to handle the case where the final variable isn't initialized in the loop.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsCentralProcessor.java
```diff
@@ -79,6 +79,28 @@
...
+                if (found) {
+                    // Didn't find EventID 12, return first 6005
+                    break;
+                }
+                // First 6005; tentatively assign and look for EventID 12
+                BOOTTIME = record.getRecord().TimeGenerated.longValue();
+                found = true;
+            } else if (found && record.getRecord().EventID.getLow().intValue() == 12) {
+                // First 12 after 6005, this is boot time
+                BOOTTIME = record.getRecord().TimeGenerated.longValue();
```
dbwiddis: This line will fail if you ever get to it because you've made `BOOTTIME` a `final` variable and if you ever get here you've already assigned it on line 95.   (Note the line 94 comment "tentative" which doesn't work with "final".)  You may want to use a temporary variable here and then make the assignment at the end of this `static` block.

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisCentralProcessor.java
```diff
@@ -52,6 +52,15 @@
     private static final long serialVersionUID = 1L;
 
     private static final Logger LOG = LoggerFactory.getLogger(SolarisCentralProcessor.class);
+    
+    private static final long BOOTTIME;
+    static {
+    	Kstat ksp = KstatUtil.kstatLookup("unix", 0, "system_misc");
+        if (ksp != null && KstatUtil.kstatRead(ksp)) {
+            BOOTTIME = KstatUtil.kstatDataLookupLong(ksp, "boot_time");
+            break;
```
dbwiddis: This break is outside a loop.  Compile error.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsCentralProcessor.java
```diff
@@ -79,6 +79,28 @@
     private static final String PROCESSOR = "Processor";
 
     private Map<String, Integer> numaNodeProcToLogicalProcMap;
+    
+    private static final long BOOTTIME;
+    static {
+    	boolean found = false;
+        EventLogIterator iter = new EventLogIterator(null, "System", WinNT.EVENTLOG_BACKWARDS_READ);
```
dbwiddis: You need to add the include for the `EventLogIterator` and `EventLogRecord` and possibly other classes.  Your IDE should help you with these... you are using an IDE, right? :)

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
shannondavid: Hi @dbwiddis. How do you suggest I handle the case where BOOTTIME isn't initialized in the loop (here and in the solaris and windows cases)? If the record isn't found, in this instance, in the /proc/stat file, is there somewhere else I can find the value? Would you recommend I set the BOOTTIME to 0?

dbwiddis: I think the current time in seconds (`System.currentTimeMillis/1000`) minus uptime in seconds is a reasonable default.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsCentralProcessor.java
shannondavid: @dbwiddis I made changes to address all of the concerns above. I am using Eclipse to edit my code but for some reason that I don't understand fully the IDE is not picking up on the missing imports... from what I've been reading online it seems like Eclipse not suggesting including missing imports could be related to the fact the project is built with maven and the IDE isn't recognizing the project as a Java project. But I'm really unsure.

Either way, I added the imports and I have pushed those changes to this PR (although doing so appears to have closed this PR). Should I open another PR or can you access the updated code from your end? Thank you

dbwiddis:
Hmm, it seems the option to reopen the pull request is not available so I'm not sure what's happening here. :(    It looks like the only option is to submit a new PR.  If you look at my first github commit you'll see I ended up closing and reopening a new one... it's all a learning experience!   

I'm not sure what happened here... the way I set up my Eclipse is that i used the command line to clone the project to a local directory and then used the "import" function in Eclipse to import from an existing maven repository.   Not sure how that corresponds to your setup but happy to try to answer any specific questions you have.

Can you try to open a new pull request with your changes?

PR #914, opened 2019-07-22
Add system start date

shannondavid:
2nd pull request after 1st one closed. Fixed style (java docs, use of white spaces) and bugs for the getBootTime functions in the respective <os>CentralProcessor classes.

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -257,6 +257,13 @@ public interface CentralProcessor extends Serializable {
      * @return Number of seconds since boot.
      */
     long getSystemUptime();
+    
+    /**
+     * Get the System boot time.
+     * 
+     * @return Number of seconds since system last booted.
+     */
+    long getBootTime();
 
     /**
      * Get the number of logical CPUs available for processing. This value may
@@ -264,6 +271,7 @@ public interface CentralProcessor extends Serializable {
      *
      * @return The number of logical CPUs available.
      */
+
     int getLogicalProcessorCount();
 
     /**
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -254,14 +254,14 @@ public interface CentralProcessor extends Serializable {
     /**
      * Get the System uptime (time since boot).
      *
-     * @return Number of seconds since boot.
+     * @return Number of seconds since boot. 
      */
     long getSystemUptime();
     
     /**
-     * Get the System boot time.
+     * Get Unix time of boot.
      * 
-     * @return Number of seconds since system last booted.
+     * @return The approximate time at which the system booted, in seconds since the Unix epoch.
      */
     long getBootTime();
 
@@ -271,7 +271,6 @@ public interface CentralProcessor extends Serializable {
      *
      * @return The number of logical CPUs available.
      */
-
     int getLogicalProcessorCount();
 
     /**
--- a/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
@@ -57,10 +57,10 @@ public class LinuxCentralProcessor extends AbstractCentralProcessor {
 
     // See https://www.kernel.org/doc/Documentation/cpu-freq/user-guide.txt
     private static final String CPUFREQ_PATH = "/sys/devices/system/cpu/cpu";
-    
+
     private static final long BOOTTIME;
     static {
-    	// Boot time given by btime variable in /proc/stat.
+        // Boot time given by btime variable in /proc/stat.
         List<String> procStat = FileUtil.readFile("/proc/stat");
         long tempBT = 0;
         for (String stat : procStat) {
@@ -72,7 +72,7 @@ public class LinuxCentralProcessor extends AbstractCentralProcessor {
         }
         // If above fails, current time minus uptime.
         if (tempBT == 0) {
-        	tempBT = System.currentTimeMillis() / 1000L - (long) ProcUtil.getSystemUptimeSeconds();
+            tempBT = System.currentTimeMillis() / 1000L - (long) ProcUtil.getSystemUptimeSeconds();
         }
         BOOTTIME = tempBT;
     }
@@ -338,13 +338,13 @@ public class LinuxCentralProcessor extends AbstractCentralProcessor {
     public long getSystemUptime() {
         return (long) ProcUtil.getSystemUptimeSeconds();
     }
-    
+
     /**
      * {@inheritDoc}
      */
     @Override
     public long getBootTime() {
-    	return BOOTTIME;
+        return BOOTTIME;
     }
 
     /**
--- a/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
@@ -57,6 +57,21 @@ public class LinuxCentralProcessor extends AbstractCentralProcessor {
 
     // See https://www.kernel.org/doc/Documentation/cpu-freq/user-guide.txt
     private static final String CPUFREQ_PATH = "/sys/devices/system/cpu/cpu";
+    
+    private static final long BOOTTIME;
+    static {
+    	// Boot time given by btime variable
+    	// in /proc/stat.
+        List<String> procStat = FileUtil.readFile("/proc/stat");
+        for (String stat : procStat) {
+            if (stat.startsWith("btime")) {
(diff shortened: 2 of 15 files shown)


oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
```diff
@@ -252,11 +252,18 @@
     long[][] getProcessorCpuLoadTicks();
 
     /**
-     * Get the System uptime (time since boot).
+     * Get the Unix time of boot.
      *
-     * @return Number of seconds since boot.
+     * @return Number of seconds since the 1970 epoch.
```
dbwiddis: I think you switched comments here. :)   The Uptime needs the "number of seconds since last booted" while the BootTime is the Unix time that the system was last booted.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
```diff
@@ -57,6 +57,25 @@
...
+        for (String stat : procStat) {
+            if (stat.startsWith("btime")) {
+                String[] bTime = ParseUtil.whitespaces.split(stat);
+                tempBT = ParseUtil.parseLongOrDefault(bTime[1], 0L);
+                break;
+            }
+        }
+        // If above fails, current time minus uptime.
+        if (tempBT == 0) {
+        	tempBT = (long) ProcUtil.getSystemUptimeSeconds();
```
dbwiddis: you aren't subtracting from current time here.

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisCentralProcessor.java
```diff
@@ -52,6 +52,16 @@
...
 
     private static final Logger LOG = LoggerFactory.getLogger(SolarisCentralProcessor.class);
+    
+    private static final long BOOTTIME;
+    static {
+    	Kstat ksp = KstatUtil.kstatLookup("unix", 0, "system_misc");
+        if (ksp != null && KstatUtil.kstatRead(ksp)) {
+            BOOTTIME = KstatUtil.kstatDataLookupLong(ksp, "boot_time");
+        } else {
+        	BOOTTIME = System.currentTimeMillis / 1000L -  ksp.ks_snaptime / 1000000000L;
```
dbwiddis: This would work, but better to not rely on `ksp` being valid, and just use the existing `getSystemUptime()` method which will call a different kstat.

Also, I note different indentation here... did you use "Source > Format" after making these changes?

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsCentralProcessor.java
```diff
@@ -79,6 +81,40 @@
...
+                found = true;
+            } else if (found && record.getRecord().EventID.getLow().intValue() == 12) {
+                // First 12 after 6005, this is boot time
+                tempBT = record.getRecord().TimeGenerated.longValue();
+                break;
+            }
+        }
+        if (tempBT != 0) {
+        	BOOTTIME = tempBT;
+        } else if (IS_VISTA_OR_GREATER) {
```
dbwiddis: Similar to above, if you use the existing `getSystemUptime()` method you can avoid repeating code here.

dbwiddis:
In addition to the other code review comments, the code isn't compiling:
- You need `()` after `System.currentTimeMillis` in multiple locations
- You need to import JNA's EventLogIterator and EventLogRecord classes

shannondavid:
Do I need to do more than add the import statements at the top of the file to include JNA's EventLogIterator and EventLogRecord? When running the mvn test on my machine I'm showing it isn't found, even though I've included the import statements...

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
shannondavid: whoops! Fixed this

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisCentralProcessor.java
shannondavid: I was concerned the getSystemUptime() method wouldn't be available without an instance of the class but on second thought that doesn't seem right. I'm refactored to use the getSystemUptime() method

shannondavid: @dbwiddis I apologize for spamming you. I just have a few comments/questions. 

- First off -- It seems installing maven locally has triggered something in Eclipse because now I'm seeing helpful error markers in Eclipse where I wasn't before (which is good!).

- Second off -- These changes include now seeing the difference in indention (8 spaces instead of 4) that I couldn't see locally but were visible on your end (also good!). I will match my formatting to the source code (i.e. tab = 4 spaces).

- Third -- When trying to use getSystemUptime(), Eclipse is now complaining the way I thought it would above, namely that I "cannot make a static reference to the non-static method getSystemUptime() from the type Solaris Central Processor" . . . the two options I see to resolve this error are to make getSystemUptime() a static method or to create an instance of SolarisCentralProcessor. The second option doesn't seem right to me; not sure about the first.

- Fourth -- I did install pmd and checkstyle on eclipse but didn't realize they would find their way into the repo (wasn't my intent to include them), so I deleted .pmd and .checkstyle. That said, I submitted a PR request earlier (which I have since deleted) and deleting those files caused an issue. The project was unable to compile almost right away because the compiler was searching for the .pmd and .checkstyle files. . . which leads me to believe that by installing pmd and checkstyle on Eclipse a call to .pmd and .checkstyle was injected somewhere else in the project code -- do you know where I could find that code so I can possibly remove it?

Finally, thank you thank you for your assistance. I'm truly learning here as I go, and I really appreciate your time and patience

dbwiddis: > I was concerned the getSystemUptime() method wouldn't be available without an instance of the class 

Actually you are correct here!  So we can't call it from a static method.  We could make a private static method, and have the static boot time calculation and instance uptime method call it.  Alternately we could allow the boot time to be an instance-level constant, or even a normal class field calculated with a lazy getter.

dbwiddis: Regarding the spacing, don't use any tabs. OSHI's formatting is the "Sun Java Conventions" with that one change (tabs = 4 spaces) but there should be no tabs in submitted code.

I think you're on the right track installing maven locally to integrate with Eclipse... glad you're learning as you go, and thanks for all the effort you're putting into this!

It's okay to have those .pmd and .checkstyle files locally, just mark them in eclipse to "remove from version control". (You'll need to delete, commit, then re-add/install and then remove from version control before the next commit.)

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsCentralProcessor.java
```diff
@@ -79,6 +83,84 @@
...
+                tempBT = record.getRecord().TimeGenerated.longValue();
+                break;
+            }
+        }
+        if (tempBT != 0) {
+        	BOOTTIME = tempBT;
+        } else 
+            BOOTTIME = (long) System.currentTimeMillis() / 1000L - getSystemUptime();
+        }
+	@Override
```
dbwiddis: Your code has inserted a lot of method stubs here... they should be deleted, as they're handled in the abstract parent.

dbwiddis:
Hey @shannondavid I just committed to your branch, deleting the stray code and resolving the "calling instance method from static" problem by creating a new private static method that both the boot time static block, and the up time instance method can call.  

Tasks you still need to do to finish this PR:
 - Add a new test case to test for boot time > 0 and boot time < current time in millis
 - Add a CHANGELOG entry

dbwiddis:
Also I've realized that both the uptime and boot time methods probably belong in the `OperatingSystem` class rather than `CentralProcessor` and I expect I'll move them both over later... but let's get you through this PR first!

shannondavid:
Hi @dbwiddis apologies on delay. Thank you for the assist. I am pulling your changes into my local repo now. Do you have any suggestions of what to include in my CHANGELOG entry? I'm not clear if something like "Add query for System Start Time" is reasonable or not

shannondavid:
One last question -- am I adding thee test case to test for boot time > 0 and boot time < current time in millis in the static initializer or in the getBootTime() method? If in the former, I'm unclear what BOOTTIME should be initialized too; if the latter, do I not return a value?

dbwiddis:
Changelog entry sounds good... I might say "boot time" rather than start time ince that's what we've named the method.  The test case should be in the CentralProcessorTest class under `src\test\java`.

shannondavid:
Got it. Changed the wording and added the test cases!

dbwiddis:
Thanks!  Looks good.  Cleaned up one stray bit of text from a merge conflict. I'll merge once the tests finish.

dbwiddis:
Actually, seems some code has now gone missing from the Mac, Linux, and FreeBsd files. They no longer implement (override) the `getBootTime()` method....

PR #939, opened 2019-08-10
Rework SystemInfoTest for Better Usability (Fixes #829)

### LINKED ISSUE #829: SystemInfoTest - reconsider using an arrayList with logger rather than system out
This is a rehash of an old request.  I rewrote SystemInfoTests to stop using system out long ago. The data is collected in an array list and written to the logger at the end.  In turn I use this as a foundation class in psi probe to output the system information as this class defines on a JSP page.  It would be extremely helpful to raise this again as a pull request so I can avoid unnecessary merge conflicts while keeping up-to-date.

A lot has changed since the early days with this project thus currently a simple change to the simple logging setup to not write out to a log file will still provide the system out usability.  An additional change to send output from tests to redirected log file also solve that issue.  The resulting behaviour as best I can tell is the same.  I can send a PR for review if this sounds acceptable.

hazendaz:
Fixes #829 

The overall class is great for testing with oshi.  This class has become a baseline for [psi-probe](https://github.com/psi-probe/psi-probe/blob/master/core/src/main/java/psiprobe/controllers/oshi/OshiController.java).  In order to allow for easier porting of changes over to psi-probe, the structure needs to better adhere to accepted java standards such as not using system out printing.  Further, in this case, psi probe wants to show the results on a webpage rather than logging.  While Oshi did a little bit of both using a logger and system out printing, some slight adjustments and use of an array list result in similar outcome.  Essentially, outside of psi probe usage of setting up a controller, the port will mostly be intack same code without the current merge conflict concerns.

No performance hit was noticed during the test run.

Configuration now redirects logs via surefire rather than via slf4j simple.  While logs are spit per test which is a variance, the output to screen remains the same.  The slf4j configuration has simple been commented out here.  It could use a note as to why if the config line stays otherwise it can be removed.  If memory serves me, I believe this adjustment item was the original concern with the design change due to output that would have otherwise went to the console during a build.

For readibility, the powerSource has been remaned from pSource which also followed generally accepted coding practice.  This was not entirely necessary but helpful and simlar change was being made in psi probe port.

### DIFF
--- a/oshi-core/src/test/java/oshi/SystemInfoTest.java
+++ b/oshi-core/src/test/java/oshi/SystemInfoTest.java
@@ -26,6 +26,7 @@ package oshi;
 import static org.junit.Assert.assertFalse;
 
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
@@ -63,6 +64,11 @@ import oshi.util.Util;
  */
 public class SystemInfoTest {
 
+    /** The Constant logger. */
+    private static final Logger logger = LoggerFactory.getLogger(SystemInfoTest.class);
+
+    static List<String> oshi = new ArrayList<>();
+
     /**
      * Test system info.
      */
@@ -78,12 +84,7 @@ public class SystemInfoTest {
      *            the arguments
      */
     public static void main(String[] args) {
-        // Options: ERROR > WARN > INFO > DEBUG > TRACE
-        System.setProperty(org.slf4j.impl.SimpleLogger.DEFAULT_LOG_LEVEL_KEY, "INFO");
-        System.setProperty(org.slf4j.impl.SimpleLogger.LOG_FILE_KEY, "System.err");
-        Logger LOG = LoggerFactory.getLogger(SystemInfoTest.class);
-
-        LOG.info("Initializing System...");
+        logger.info("Initializing System...");
         SystemInfo si = new SystemInfo();
 
         HardwareAbstractionLayer hal = si.getHardware();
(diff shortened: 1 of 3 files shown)


dbwiddis:
I wonder if it should be moved to oshi-demo. Do we need two loggers? I only put in the one to get debug logging from the core artifact.

hazendaz:
I replaced the original logger noted as "LOG" to the more accepted "logger" and moved it in the class.  So this simply does as it was doing before in the log output with all the activities showing first then it outputs the work at the end.  Possibly that is a slight improvement as it takes away subjective configuration of that log and keeps things nicely separated.

Should this be in oshi demo?  I'm not 100% there yet. I'd like to see if this could be made even more user friendly so I don't have to copy any code into psi probe.  In that regard, it's likely something that belongs in the core but not sure how that might look at the moment.

hazendaz:
My usage in psi probe to be fair is a copy / paste job that ultimately isn't well formatted at least in relation to how psi probe generally outputs data.  I think I would need to look at that more to see how this might eventually look.  For the time being, its probably fine here or in the demo.  Longer term it might be something for the core that includes more advanced formatting than here today.  Either way for now, I would still be primarily copy / pasting and adjustment for the controller needs / output needs in psi probe.

dbwiddis:
> In that regard, it's likely something that belongs in the core but not sure how that might look at the moment.

Perhaps `toString()`s?

dbwiddis:
I'm going to go ahead and merge this as-is, and then do some more tweaks.  I like the idea of making lots of `toString()` methods on the objects that do the equivalent of this test (and replacing the test output with just logging the object, which will invoke the `toString()`).  Obviously that doesn't work for the CPU load "test" which has the 1-second delay so it can't completely replace this class, but I do think that in many cases, what we show in this test class is what should appear in a `toString()`.

hazendaz:
@dbwiddis Thanks!  I reported the code over to psi probe and updated to oshi 4.0.0.  Much easier now :)  Thanks again.  Also like the toString() suggestion and the write up you did for newer committers.  Very nice!

PR #941, opened 2019-08-10
Retool the travis ci jdk selection process

hazendaz:
All jdks will be auto pulled via normal procedure now.  The install-jdk is part of travis by default.  For the xcode 9.3, that one will remain with jdk not set and thus will use what is on the vm (oracle jdk 8).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ matrix:
   include:  
     - os: osx
       env: 
-      - JDK = "OpenJDK 8 - Mac"
+      - JDK = "OracleJDK 8 - Mac"
       - MATRIX_ID="first"
       jdk: openjdk8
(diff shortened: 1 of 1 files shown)


hazendaz:
This can be squashed.  The important thing was figuring out how to get here which the commit history shows.  But long term that isn't necessary ;)

dbwiddis:
Thanks for this, I'll go ahead and merge it.  I set up the external script back when Travis was way behind in getting their own script working.  I also was choosing versions based on 8 (OSHI minimum), 11 (LTS), 12 (current), and the EA build which I recently switched from 13 to 14.  I don't think we need 12 since we have 13 at release candidate phase, but I'll wait until next month to nix the 12 tests.

Issue #942, opened 2019-08-10
Add toString methods to API interface implementations


dbwiddis:
I'm marking this issue for [first-timers-only](https://www.firsttimersonly.com/).  That means that I will only accept a PR for this one from someone who's never contributed to open source before. This one is easy (but don't make that statement make you feel bad if you have a hard time with it, there's more to contributing to open source than changing lines of code, especially if it's your first time). I'll hold your hand through this if you need me to. :-) 

We'd like to add `toString()` methods to all of the API classes, with the implementation in the associated Abstract class for interfaces.  This has been done for [AbstractOperatingSystem](https://github.com/oshi/oshi/blob/05b143dd0dd2485cabc1bebc14d8b710ceef1da2/oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java#L151), [AbstractDisplay](https://github.com/oshi/oshi/blob/05b143dd0dd2485cabc1bebc14d8b710ceef1da2/oshi-core/src/main/java/oshi/hardware/common/AbstractDisplay.java#L58), [AbstractOSVersionInfoEx](https://github.com/oshi/oshi/blob/05b143dd0dd2485cabc1bebc14d8b710ceef1da2/oshi-core/src/main/java/oshi/software/common/AbstractOSVersionInfoEx.java#L81), [AbstractSoundCard](https://github.com/oshi/oshi/blob/05b143dd0dd2485cabc1bebc14d8b710ceef1da2/oshi-core/src/main/java/oshi/hardware/common/AbstractSoundCard.java#L109), [AbstractUsbDevice](https://github.com/oshi/oshi/blob/05b143dd0dd2485cabc1bebc14d8b710ceef1da2/oshi-core/src/main/java/oshi/hardware/common/AbstractUsbDevice.java#L131), [AbstractCentralProcessor](https://github.com/oshi/oshi/blob/fa4c75ff6a04ee36f5bb37a57fa2881e7aedba80/oshi-core/src/main/java/oshi/hardware/common/AbstractCentralProcessor.java#L400), and the class [OSProcess](https://github.com/oshi/oshi/blob/05b143dd0dd2485cabc1bebc14d8b710ceef1da2/oshi-core/src/main/java/oshi/software/os/OSProcess.java#L647)

Similar implementations should be made for:
- [ ] ComputerSystem
- [ ] Baseboard
- [ ] Firmware
- [ ] GlobalMemory
- [ ] VirtualMemory
- [ ] HWDiskStore
[...]

PR #948, opened 2019-08-12
Added toString implementation to PowerSource

colinbobolin:
PowerSource toString method as specified in issue #942

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
@@ -75,6 +75,32 @@ public abstract class AbstractPowerSource implements PowerSource {
     /** {@inheritDoc} */
     @Override
     public String toString() {
-        return String.format("%n %s @ %.1f%%", this.getName(), this.getRemainingCapacity() * 100d);
+        StringBuilder sb = new StringBuilder();
+        sb.append("Name: ").append(getName()).append(", ");
+        sb.append("Remaining Capacity: %").append(getRemainingCapacity() * 100d).append(", ");
+        sb.append("Time Remaining: ").append(getFormattedTimeRemaining());
+        return sb.toString();
+    }
+
+    /**
+     * Estimated time remaining on power source, formatted as HH:mm
+     *
+     * @return formatted String of time remaining
+     */
+    private String getFormattedTimeRemaining() {
+        double timeInSeconds = getTimeRemaining();
+        String formattedTimeRemaining;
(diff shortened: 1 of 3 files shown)


oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
```diff
@@ -71,4 +71,10 @@ public double getRemainingCapacity() {
     public double getTimeRemaining() {
         return this.timeRemaining;
     }
+
+    /** {@inheritDoc} */
+    @Override
+    public String toString() {
+        return String.format("%n %s @ %.1f%%", this.getName(), this.getRemainingCapacity() * 100d);
```
dbwiddis: Let's not include the `%n ` header here.

oshi-core/src/test/java/oshi/SystemInfoTest.java
```diff
@@ -278,8 +278,7 @@ private static void printPowerSources(PowerSource[] powerSources) {
             }
         }
         for (PowerSource powerSource : powerSources) {
-            sb.append(
-                    String.format("%n %s @ %.1f%%", powerSource.getName(), powerSource.getRemainingCapacity() * 100d));
+            sb.append(powerSource.toString());
```
dbwiddis: Append the `"\n "` here.

oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
dbwiddis: Also while it wasn't in the original `SystemInfoTest` we should add `getTimeRemaining()`.  If positive, seconds remaining (would be awesome to format as something like `1h 20m 15s` or even omit seconds and do `1:20`). If negative, -1.0 (calculating) or -2.0 (unlimited)... careful with floating point equality here too, you probably want to do "<1.5" and "<0" cases.

dbwiddis:
Thanks!  I'll do some formatting tweaks and changelog, and then merge this.  Also the modulo with the double worries me just enough that I'll do the int cast before the modulo.

colinbobolin:
Thanks for the review, I'll keep a closer eye on my type casting in the future. Happy I could contribute.

dbwiddis:
Heh, actually it looks like I did it exactly that way in SystemInfoTest but only for the first element of the array. I'm nixing that and leaving the casting. It works, I'm always iffy about floating point math!  I did change the "Unlimited" to "Charging" though... thanks again for your submission.

dbwiddis:
Merged! Thanks for contributing to open source!

PR #952, opened 2019-08-16
Add toString to OSFileStore class

GandalfTheJava:
… I am struggling to call it within the SystemInfoTest.java. Any guidance would be appreciated :)

### DIFF
--- a/.github/issue_label_bot.yaml
+++ b/.github/issue_label_bot.yaml
@@ -1,0 +1,4 @@
+label-alias:
+  bug: 'kind/bug?'
+  feature_request: 'new feature'
+  question: 'question'
--- a/.github/issue_label_bot.yaml
+++ b/.github/issue_label_bot.yaml
@@ -1,28 +1,4 @@
-#
-# OSHI (https://github.com/oshi/oshi)
-#
-# Copyright (c) 2010 - 2019 The OSHI Project Team:
-# https://github.com/oshi/oshi/graphs/contributors
-#
-# Permission is hereby granted, free of charge, to any person obtaining a copy
-# of this software and associated documentation files (the "Software"), to deal
-# in the Software without restriction, including without limitation the rights
-# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-# copies of the Software, and to permit persons to whom the Software is
-# furnished to do so, subject to the following conditions:
-# The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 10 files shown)


oshi-core/src/test/java/oshi/SystemInfoTest.java
```diff
@@ -118,6 +119,10 @@ public static void main(String[] args) {
         logger.info("Checking File System...");
         printFileSystem(os.getFileSystem());
 
+        logger.info("Checking OSFileStore...");
```
dbwiddis: You don't need to add this here.  The call above to print the File System will iterate over the OSFileStores -- the same location where you copied the information to put in the toString.  You'll end up with something like:
```
OSFileStore[] fsArray = fileSystem.getFileStores()
for (OSFileStore fs : fsArray) {
    oshi.add(fs.toString());
}
```

oshi-core/src/test/java/oshi/SystemInfoTest.java
```diff
@@ -171,6 +176,10 @@ private static void printMemory(GlobalMemory memory) {
         oshi.add("Swap: \n " + vm.toString());
     }
 
+    private static void printOSFileStore(OSFileStore OSfileStore)
```
dbwiddis: Don't need to add this either.

oshi-core/src/main/java/oshi/software/os/OSFileStore.java
```diff
@@ -349,4 +349,22 @@ public boolean updateAtrributes() {
         }
         return false;
     }
+
+    /** {@inheritDoc} */
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
```
dbwiddis: You're not including any punctuation between the entries here.   Also I'd prefer to keep the formatted output (GB rather than the entire integer) and the percentage from the [existing output](https://github.com/oshi/oshi/blob/fe457f3e4e5b1e4bc37b4401aa600ddea042da9c/oshi-core/src/test/java/oshi/SystemInfoTest.java#L296) here.

.github/issue_label_bot.yaml
```diff
@@ -1,3 +1,27 @@
+#
```
dbwiddis: Don't commit this file... looks like I need to add another exception to the license plugin. :)

Issue #797, opened 2019-02-05
[version 3.13.0] SystemInfo (and many classes available via it) is not thread-safe


stIncMale:
Instances of `SystemInfo` initialize its fields lazily, but without any synchronization, which leads to executions with data races and the possibilities of having either `NullPointerException` or observing default values for primitive fields.

After taking a first look, I thought "maybe users are not supposed to use `SystemInfo` concurrently", but then I saw the question [Are oshi SystemInfo instances thread safe?](https://github.com/oshi/oshi/issues/660), and an answer 
> I've made reasonable efforts to try to avoid issues with thread safety, and many classes will work without a problem, I certainly can't guarantee it

and realized that `SystemInfo` is actually supposed to be used concurrently... I am going to present a only a few issues, but I would expect that there are more problems.

**Problem 1: lazy initialization without synchronization**
Consider `SystemInfo.getHardware().getMemory().getTotal()`:
* a plain field `SystemInfo.hardware` is initialized lazily to a new instance of let's say `LinuxHardwareAbstractionLayer` when `SystemInfo.getHardware` is called.
* a plain field `LinuxHardwareAbstractionLayer.memory` is initialized lazily to a new instance of let's say `LinuxGlobalMemory` when `LinuxHardwareAbstractionLayer.getMemory` is called.
* a plain field `LinuxGlobalMemory.memTotal` is initialized lazily to a `long` value when `LinuxGlobalMemory.getTotal` is called.

All those fields are initialized without any memory synchronization. As a result writes to `SystemInfo.hardware` field produced by `SystemInfo.getHardware` method and reads from `SystemInfo.hardware` field produced by the same `SystemInfo.getHardware` method are not ordered with happens-before (HB) relation when `SystemInfo.getHardware` is called concurrently (unless a user of `SystemInfo` takes care of that...), and this means that executions contain data races. The mentioned reads of `SystemInfo.hardware` is allowed by the Java memory model to observe both writes to this field produced by `SystemInfo.getHardware` and the implicit write of the default value (this write exists in all executions regardless of whether the original code has it explicitly or not). The same is true for other mentioned fields, and thus there is no guarantee that we will not observe `null` in `SystemInfo.hardware`/`LinuxHardwareAbstractionLayer.memory` and 0 in `LinuxGlobalMemory.memTotal`.

[...]

PR #959, opened 2019-08-19
Thread Safety implementations

dbwiddis:
Building on suggestions from #797, #869, #870, attempting to make the API thread safe.

Work in progress, open for review/comment.

### DIFF
--- a/oshi-core/src/main/java/oshi/SystemInfo.java
+++ b/oshi-core/src/main/java/oshi/SystemInfo.java
@@ -23,7 +23,7 @@
  */
 package oshi;
 
-import java.io.Serializable;
+import java.util.function.Supplier;
 
 import com.sun.jna.Platform; // NOSONAR squid:S1191
 
@@ -39,6 +39,7 @@ import oshi.software.os.mac.MacOperatingSystem;
 import oshi.software.os.unix.freebsd.FreeBsdOperatingSystem;
 import oshi.software.os.unix.solaris.SolarisOperatingSystem;
 import oshi.software.os.windows.WindowsOperatingSystem;
+import oshi.util.Memoizer;
 
 /**
  * System information. This is the main entry point to Oshi.
@@ -47,9 +48,7 @@ import oshi.software.os.windows.WindowsOperatingSystem;
  * platform-specific implementations of {@link oshi.software.os.OperatingSystem}
  * (software) and {@link oshi.hardware.HardwareAbstractionLayer} (hardware).
  */
-public class SystemInfo implements Serializable {
-
-    private static final long serialVersionUID = 1L;
+public class SystemInfo {
 
     // The platform isn't going to change, and making this static enables easy
     // access from outside this class
@@ -71,9 +70,10 @@ public class SystemInfo implements Serializable {
         }
     }
 
-    private volatile OperatingSystem os = null;
+    private final Supplier<OperatingSystem> os = Memoizer.memoize(this::createOperatingSystem);
 
-    private volatile HardwareAbstractionLayer hardware = null;
+    private final Supplier<HardwareAbstractionLayer> hardware = Memoizer.memoize(this::createHardware);
+
 
     /**
      * <p>
@@ -93,16 +93,7 @@ public class SystemInfo implements Serializable {
      * @return A new instance of {@link oshi.software.os.OperatingSystem}.
      */
     public OperatingSystem getOperatingSystem() {
-        OperatingSystem localRef = this.os;
-        if (localRef == null) {
-            synchronized (this) {
-                localRef = this.os;
-                if (localRef == null) {
-                    this.os = localRef = createOperatingSystem();
-                }
-            }
-        }
-        return localRef;
+        return os.get();
     }
 
     private OperatingSystem createOperatingSystem() {
@@ -130,16 +121,7 @@ public class SystemInfo implements Serializable {
      * @return A new instance of {@link oshi.hardware.HardwareAbstractionLayer}.
      */
     public HardwareAbstractionLayer getHardware() {
-        HardwareAbstractionLayer localRef = this.hardware;
-        if (localRef == null) {
-            synchronized (this) {
-                localRef = this.hardware;
-                if (localRef == null) {
-                    this.hardware = localRef = createHardware();
-                }
-            }
-        }
-        return localRef;
+        return hardware.get();
     }
 
     private HardwareAbstractionLayer createHardware() {
--- a/oshi-core/src/main/java/oshi/SystemInfo.java
+++ b/oshi-core/src/main/java/oshi/SystemInfo.java
@@ -39,7 +39,7 @@ import oshi.software.os.mac.MacOperatingSystem;
 import oshi.software.os.unix.freebsd.FreeBsdOperatingSystem;
 import oshi.software.os.unix.solaris.SolarisOperatingSystem;
 import oshi.software.os.windows.WindowsOperatingSystem;
-import oshi.util.Memoizer;
+import static oshi.util.Memoizer.memoize;
 
 /**
  * System information. This is the main entry point to Oshi.
@@ -70,9 +70,9 @@ public class SystemInfo {
         }
     }
 
-    private final Supplier<OperatingSystem> os = Memoizer.memoize(this::createOperatingSystem);
+    private final Supplier<OperatingSystem> os = memoize(this::createOperatingSystem);
 
-    private final Supplier<HardwareAbstractionLayer> hardware = Memoizer.memoize(this::createHardware);
+    private final Supplier<HardwareAbstractionLayer> hardware = memoize(this::createHardware);
 
 
     /**
--- a/oshi-core/src/main/java/oshi/SystemInfo.java
+++ b/oshi-core/src/main/java/oshi/SystemInfo.java
@@ -51,10 +51,6 @@ public class SystemInfo implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
-    private OperatingSystem os = null;
-
-    private HardwareAbstractionLayer hardware = null;
-
     // The platform isn't going to change, and making this static enables easy
     // access from outside this class
     private static final PlatformEnum currentPlatformEnum;
@@ -75,6 +71,10 @@ public class SystemInfo implements Serializable {
         }
     }
 
+    private volatile OperatingSystem os = null;
+
+    private volatile HardwareAbstractionLayer hardware = null;
+
     /**
      * <p>
      * Getter for the field <code>currentPlatformEnum</code>.
@@ -93,29 +93,34 @@ public class SystemInfo implements Serializable {
      * @return A new instance of {@link oshi.software.os.OperatingSystem}.
      */
     public OperatingSystem getOperatingSystem() {
-        if (this.os == null) {
-            switch (currentPlatformEnum) {
-
-            case WINDOWS:
-                this.os = new WindowsOperatingSystem();
-                break;
-            case LINUX:
-                this.os = new LinuxOperatingSystem();
-                break;
-            case MACOSX:
-                this.os = new MacOperatingSystem();
-                break;
-            case SOLARIS:
-                this.os = new SolarisOperatingSystem();
-                break;
-            case FREEBSD:
-                this.os = new FreeBsdOperatingSystem();
-                break;
-            default:
-                throw new UnsupportedOperationException("Operating system not supported: " + Platform.getOSType());
+        OperatingSystem localRef = this.os;
+        if (localRef == null) {
+            synchronized (this) {
+                localRef = this.os;
+                if (localRef == null) {
+                    this.os = localRef = createOperatingSystem();
+                }
             }
         }
-        return this.os;
+        return localRef;
+    }
+
+    private OperatingSystem createOperatingSystem() {
+        switch (currentPlatformEnum) {
+
+        case WINDOWS:
+            return new WindowsOperatingSystem();
+        case LINUX:
+            return new LinuxOperatingSystem();
+        case MACOSX:
+            return new MacOperatingSystem();
+        case SOLARIS:
+            return new SolarisOperatingSystem();
+        case FREEBSD:
+            return new FreeBsdOperatingSystem();
+        default:
+            throw new UnsupportedOperationException("Operating system not supported: " + Platform.getOSType());
+        }
     }
 
     /**
@@ -125,28 +130,33 @@ public class SystemInfo implements Serializable {
      * @return A new instance of {@link oshi.hardware.HardwareAbstractionLayer}.
      */
     public HardwareAbstractionLayer getHardware() {
-        if (this.hardware == null) {
-            switch (currentPlatformEnum) {
-
-            case WINDOWS:
-                this.hardware = new WindowsHardwareAbstractionLayer();
-                break;
-            case LINUX:
-                this.hardware = new LinuxHardwareAbstractionLayer();
-                break;
-            case MACOSX:
-                this.hardware = new MacHardwareAbstractionLayer();
-                break;
-            case SOLARIS:
-                this.hardware = new SolarisHardwareAbstractionLayer();
-                break;
-            case FREEBSD:
-                this.hardware = new FreeBsdHardwareAbstractionLayer();
-                break;
-            default:
-                throw new UnsupportedOperationException("Operating system not supported: " + Platform.getOSType());
+        HardwareAbstractionLayer localRef = this.hardware;
+        if (localRef == null) {
+            synchronized (this) {
+                localRef = this.hardware;
+                if (localRef == null) {
+                    this.hardware = localRef = createHardware();
+                }
             }
         }
-        return this.hardware;
+        return localRef;
+    }
+
+    private HardwareAbstractionLayer createHardware() {
+        switch (currentPlatformEnum) {
+
+        case WINDOWS:
+            return new WindowsHardwareAbstractionLayer();
+        case LINUX:
+            return new LinuxHardwareAbstractionLayer();
+        case MACOSX:
+            return new MacHardwareAbstractionLayer();
+        case SOLARIS:
+            return new SolarisHardwareAbstractionLayer();
+        case FREEBSD:
+            return new FreeBsdHardwareAbstractionLayer();
+        default:
+            throw new UnsupportedOperationException("Operating system not supported: " + Platform.getOSType());
+        }
     }
 }
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractComputerSystem.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractComputerSystem.java
@@ -23,12 +23,13 @@
  */
 package oshi.hardware.common;
 
+import static oshi.util.Memoizer.memoize;
+
 import java.util.function.Supplier;
 
 import oshi.hardware.Baseboard;
 import oshi.hardware.ComputerSystem;
 import oshi.hardware.Firmware;
-import static oshi.util.Memoizer.memoize;
 
 /**
  * Computer System data.
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractComputerSystem.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractComputerSystem.java
@@ -28,16 +28,16 @@ import java.util.function.Supplier;
 import oshi.hardware.Baseboard;
 import oshi.hardware.ComputerSystem;
 import oshi.hardware.Firmware;
-import oshi.util.Memoizer;
+import static oshi.util.Memoizer.memoize;
 
 /**
  * Computer System data.
  */
 public abstract class AbstractComputerSystem implements ComputerSystem {
 
-    private final Supplier<Firmware> firmware = Memoizer.memoize(this::createFirmware);
+    private final Supplier<Firmware> firmware = memoize(this::createFirmware);
 
-    private final Supplier<Baseboard> baseboard = Memoizer.memoize(this::createBaseboard);
+    private final Supplier<Baseboard> baseboard = memoize(this::createBaseboard);
 
     @Override
     public Firmware getFirmware() {
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractComputerSystem.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractComputerSystem.java
@@ -26,7 +26,6 @@ package oshi.hardware.common;
 import oshi.hardware.Baseboard;
 import oshi.hardware.ComputerSystem;
 import oshi.hardware.Firmware;
-import oshi.util.Constants;
 
 /**
  * Computer System data.
@@ -35,38 +34,59 @@ public abstract class AbstractComputerSystem implements ComputerSystem {
 
     private static final long serialVersionUID = 1L;
 
-    protected String manufacturer;
-    protected String model;
-    protected String serialNumber;
-    protected Firmware firmware;
-    protected Baseboard baseboard;
+    protected volatile String manufacturer;
+
+    protected volatile String model;
+
+    protected volatile String serialNumber;
+
+    private volatile Firmware firmware;
+
+    private volatile Baseboard baseboard;
 
     /** {@inheritDoc} */
     @Override
-    public String getManufacturer() {
-        if (this.manufacturer == null) {
-            this.manufacturer = Constants.UNKNOWN;
+    public Firmware getFirmware() {
+        Firmware localRef = this.firmware;
+        if (localRef == null) {
+            synchronized (this) {
+                localRef = this.firmware;
+                if (localRef == null) {
+                    this.firmware = localRef = createFirmware();
+                }
+            }
         }
-        return this.manufacturer;
+        return localRef;
     }
+
+    /**
+     * Instantiates the platform-specific {@link Firmware} object
+     * 
+     * @return platform-specific {@link Firmware} object
+     */
+    protected abstract Firmware createFirmware();
 
     /** {@inheritDoc} */
     @Override
-    public String getModel() {
-        if (this.model == null) {
-            this.model = Constants.UNKNOWN;
+    public Baseboard getBaseboard() {
+        Baseboard localRef = this.baseboard;
+        if (localRef == null) {
+            synchronized (this) {
+                localRef = this.baseboard;
+                if (localRef == null) {
+                    this.baseboard = localRef = createBaseboard();
+                }
+            }
         }
-        return this.model;
+        return localRef;
     }
 
-    /** {@inheritDoc} */
-    @Override
-    public String getSerialNumber() {
-        if (this.serialNumber == null) {
-            this.serialNumber = Constants.UNKNOWN;
-        }
-        return this.serialNumber;
-    }
+    /**
+     * Instantiates the platform-specific {@link Baseboard} object
+     * 
+     * @return platform-specific {@link Baseboard} object
+     */
+    protected abstract Baseboard createBaseboard();
 
     /** {@inheritDoc} */
     @Override
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractComputerSystem.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractComputerSystem.java
@@ -23,40 +23,25 @@
  */
 package oshi.hardware.common;
 
+import java.util.function.Supplier;
+
 import oshi.hardware.Baseboard;
 import oshi.hardware.ComputerSystem;
 import oshi.hardware.Firmware;
+import oshi.util.Memoizer;
 
 /**
  * Computer System data.
  */
 public abstract class AbstractComputerSystem implements ComputerSystem {
 
-    private static final long serialVersionUID = 1L;
+    private final Supplier<Firmware> firmware = Memoizer.memoize(this::createFirmware);
 
-    protected volatile String manufacturer;
+    private final Supplier<Baseboard> baseboard = Memoizer.memoize(this::createBaseboard);
 
-    protected volatile String model;
-
-    protected volatile String serialNumber;
-
-    private volatile Firmware firmware;
-
-    private volatile Baseboard baseboard;
-
-    /** {@inheritDoc} */
     @Override
     public Firmware getFirmware() {
-        Firmware localRef = this.firmware;
-        if (localRef == null) {
-            synchronized (this) {
-                localRef = this.firmware;
-                if (localRef == null) {
-                    this.firmware = localRef = createFirmware();
-                }
-            }
-        }
-        return localRef;
+        return firmware.get();
     }
 
     /**
@@ -66,19 +51,9 @@ public abstract class AbstractComputerSystem implements ComputerSystem {
      */
     protected abstract Firmware createFirmware();
 
-    /** {@inheritDoc} */
     @Override
     public Baseboard getBaseboard() {
-        Baseboard localRef = this.baseboard;
-        if (localRef == null) {
-            synchronized (this) {
-                localRef = this.baseboard;
-                if (localRef == null) {
-                    this.baseboard = localRef = createBaseboard();
-                }
-            }
-        }
-        return localRef;
+        return baseboard.get();
     }
 
     /**
@@ -88,7 +63,6 @@ public abstract class AbstractComputerSystem implements ComputerSystem {
      */
     protected abstract Baseboard createBaseboard();
 
-    /** {@inheritDoc} */
     @Override
     public String toString() {
         StringBuilder sb = new StringBuilder();
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractHardwareAbstractionLayer.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractHardwareAbstractionLayer.java
@@ -38,8 +38,6 @@ import oshi.util.Memoizer;
  */
 public abstract class AbstractHardwareAbstractionLayer implements HardwareAbstractionLayer {
 
-    private static final long serialVersionUID = 1L;
-
     private final Supplier<ComputerSystem> computerSystem = Memoizer.memoize(this::createComputerSystem);
 
     private final Supplier<CentralProcessor> processor = Memoizer.memoize(this::createProcessor);
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractHardwareAbstractionLayer.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractHardwareAbstractionLayer.java
@@ -30,7 +30,7 @@ import oshi.hardware.ComputerSystem;
 import oshi.hardware.GlobalMemory;
 import oshi.hardware.HardwareAbstractionLayer;
 import oshi.hardware.Sensors;
-import oshi.util.Memoizer;
+import static oshi.util.Memoizer.memoize;
 
 /**
  * Common fields or methods used by platform-specific implementations of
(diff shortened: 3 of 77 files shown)


dbwiddis:
I'd appreciate comments/criticisms from @stIncMale and @Nadahar as I make progress here.

oshi-core/src/main/java/oshi/hardware/common/AbstractHardwareAbstractionLayer.java
```diff
@@ -37,12 +37,99 @@
...
-    protected CentralProcessor processor;
+    private volatile CentralProcessor processor;
 
-    protected GlobalMemory memory;
+    private volatile GlobalMemory memory;
 
-    protected Sensors sensors;
+    private volatile Sensors sensors;
 
+    /** {@inheritDoc} */
```
stIncMale: All javadocs are inherited by default, so it is not necessary to specify the `@inheritDoc` tag unless we want to extend the javadoc. The below snippet explains some usages of the tag:
```java
interface A {
    /**
     * [A x]
     */
    void x();

    /**
     * [A y]
     * 
     * @param foo [A foo]
     * @param bar [A bar]
     * @throws IllegalArgumentException [A IllegalArgumentException]
     * @throws NullPointerException [A NullPointerException]
     */
    void y(int foo, String bar) throws IllegalArgumentException, NullPointerException;
}

interface B extends A {
    //inherits javadoc from A as is
    void x();

    /**
     * {@inheritDoc}
     * <p>
     * [B y]
     *
     * @param bar {@inheritDoc}[B bar]
     * @throws IllegalArgumentException [B IllegalArgumentException]{@inheritDoc}
     * @throws NullPointerException [B NullPointerException]
     */
    //extends method description of A
    //inherits foo param description from A as is
    //extends bar param description of A
    //extends IllegalArgumentException description of A
    //overrides NullPointerException description of A
    void y(int foo, String bar) throws IllegalArgumentException, NullPointerException;
}
```

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxComputerSystem.java
```diff
@@ -44,48 +44,68 @@
...
     @Override
     public String getManufacturer() {
-        if (this.manufacturer == null && !queryManufacturerFromSysfs() && !queryManufacturerFromProcCpu()) {
-            this.manufacturer = Constants.UNKNOWN;
+        String localRef = this.manufacturer;
+        if (localRef == null) {
+            synchronized (this) {
+                localRef = this.manufacturer;
+                if (localRef == null) {
+                    if (!queryManufacturerFromSysfs() && !queryManufacturerFromProcCpu()) {
```
stIncMale: The `manufacturer` field is declared in the super class `AbstractComputerSystem` but is written in subclasses. This violates encapsulation for no good reason. I am suggesting the following:

- Declare `AbstractComputerSystem.manufacturer` as `private`.
- Remove the `getManufacturer` method from all subclasses of `AbstractComputerSystem` and move it to `AbstractComputerSystem`.
- Introduce the `abstract protected String AbstractComputerSystem.manufacturer()` method specifying the logic of getting the value of `manufacturer`.
- Use the `manufacturer()` method in `AbstractComputerSystem` in order to calculate the value of `manufacturer`.
- Implement the `manufacturer()` method in all subclasses of `AbstractComputerSystem`. E.g. for `LinuxComputerSystem` the implementation would be
```java
protected final String manufacturer() {
  String result;
  if ((result = queryManufacturerFromSysfs()) == null &&
      (result = queryManufacturerFromProcCpu()) == null) {
    return Constants.UNKNOWN;
  }
  return result;
}
```
This would of course require to change `queryManufacturerFromSysfs()` and `queryManufacturerFromProcCpu()` methods so that they return manufacturer instead of writing to the `manufacturer` field. But such a change again would make the code much less error-prone and more readable, because we would have a single place where the field `manufacturer` is written, instead of having multiple places as it is now.

Nadahar:
There are so many ways to solve this, and so much is about style/what one is used to rather than what is "correct". As such, I think it's hard to give "general" feedback - it all depends on so many other things. The only thing I can say is use `final` for all fields where that's possible. A class where all fields are `final` is immutable if all the field types (classes) are also immutable, and is as such thread-safe without the need for any synchronization.

I always start out with making everything i can `final`, and then figures out a synchronization strategy for what's left. In OSHI, a lot of things could be immutable (everything that can't change while an application is running), which should significantly simplify the task.

oshi-core/src/main/java/oshi/SystemInfo.java
```diff
@@ -93,29 +93,34 @@ public static PlatformEnum getCurrentPlatformEnum() {
...
-                break;
-            case SOLARIS:
-                this.os = new SolarisOperatingSystem();
-                break;
-            case FREEBSD:
-                this.os = new FreeBsdOperatingSystem();
-                break;
-            default:
-                throw new UnsupportedOperationException("Operating system not supported: " + Platform.getOSType());
+        OperatingSystem localRef = this.os;
```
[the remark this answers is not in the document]
dbwiddis: Thanks for this suggestion!  I'll look into writing my own Memoizer.

oshi-core/src/main/java/oshi/hardware/common/AbstractHardwareAbstractionLayer.java
dbwiddis: The vast majority of the "InheritDoc" statements were written by the maven javadoc plugin... I guess I can remove them but I don't see them as harmful and they serve as a visual reference to me... then again, so does the `@Override` annotation.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxComputerSystem.java
dbwiddis: Thanks for this suggestion, and I did attempt to do something similar to that.   However, the pattern breaks down (and possibly the earlier proposed memoizer pattern) because there is not a 1-to-1 relationship of value-fetching logic to value assignment.  For the `ComputerSystem` object, as an example, while Linux fetches the three values separately:
- Windows uses a WMI query that returns both Manufacturer and Model
- Mac uses the `system_profiler` command line to get both Model and Serial Number
- FreeBSD uses `dmidecode` command line to get Manufacturer, Model, and Serial Number
- Solaris uses the `smbios` command line that returns both Manufacturer and Model

The pattern above, fetching `manufacturer()` does not allow us to use the result to populate the `model` field, unless I mess around with arrays or other encapsulation.

Certainly I could fetch things individually, and in these cases the cost of repeating the commands is minimal, but there are many similar situations where a single command fetches values for multiple fields in a more expensive manner (usually WMI queries) and I'd like a (thread safe) way of populating all the fields at once.

stIncMale: I see. Yeah, that may affect both the idea mentioned here (not very important) and the idea of using memoizer (important). I understand and agree with the reason you don't want to initialize the fields individually. However, there is at least one another approach that can be taken: move the fields initialized in groups out of `AbstractComputerSystem ` into subclasses.

You mentioned that `manufacturer` and `model` are initialized at once in `WindowsComputerSystem` via `queryManufacturerAndModel`. This means we may remove these fields from `AbstractComputerSystem` and have the following:
```java
final class WindowsComputerSystem extends AbstractComputerSystem {
  private final Supplier<ManufacturerAndModel> manufacturerAndModel = memoized(this:: queryManufacturerAndModel);

  public String getManufacturer() {
    return manufacturerAndModel.get().manufacturer;
  }

  public String getModel() {
    return manufacturerAndModel.get().model;
  }

  private ManufacturerAndModel queryManufacturerAndModel() {
    ...
    return new ManufacturerAndModel(...);
  }

  private static final class ManufacturerAndModel {
    private final String manufacturer;
    private final String model;

    private ManufacturerAndModel() {
      this.manufacturer = manufacturer;
      this.model = model;
    }
  }
}
```

Other concrete implementations of `ComputerSystem ` may do a similar thing. If this approach is taken, then:
- only the `baseboard ` and `firmware ` fields would stay in `AbstractComputerSystem`;
- `LinuxComputerSystem` would have to declare all three fields `manufacturer`, `model`, `serialNumber` because `AbstractComputerSystem` would not be declaring them anymore.

While this approach misses the opportunity of declaring the state in a single place (`AbstractComputerSystem`), it leaves us with the benefit of using memoizers. In other words we are effectively choosing between
- duplicating field declarations
- duplicating multi-threaded lazy initialization code.

Despite both options are not good, I think the former is less bad than the latter. And unfortunately, I can not see a way of solving both of the problems in this case.

dbwiddis: @stIncMale Thanks so much for taking the time to educate me!

I've updated the code with your suggestions:
- Used a Memoizer instead of double checked locking, implemented in SystemInfo, AbstractHardwareAbstrationLayer, and AbstractComputerSystem.  Working through the platform-specific ones, have only done the WindowsComputerSystem per your suggestion
- Re-established encapsulation

It's getting a bit more complex now with SolarisComputerSystem as one method call only returns two of the fields and another returns three, so it will take a little bit of thought... but I think I'm on a roll here!

oshi-core/src/main/java/oshi/util/Memoizer.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * Store a supplier in a delegate function to be computed only once.
+     * 
+     * @param s
+     *            The {@link java.util.function.Supplier} to memoize
+     * @return A memoized version of the supplier
+     */
+    public static <T> Supplier<T> memoize(Supplier<T> s) {
```
stIncMale: I have to be honest, your implementation of `memoize` got me into thinking :) It looks like a clever trick, but it has a problem: there is a data race caused by the lack of happens-before relations between write to `delegate` in `getFirst()` and reads from `delegate` in `get()`. As a result, reads from `delegate` in `get()` are allowed to never observe the result of the write to `delegate` in `getFirst()`, i.e. the method `get()` is allowed to always observe the initial delegate calling `getFirst`, and thus the method `get()` is allowed to always acquire a monitor (requested by `synchronized`), which is a much more expensive operation than a volatile read that we have in DCL.

Interestingly, the same logic may not applicable when talking about accesses to `value` provided that `() -> value` captures `value` with the semantics of `final` fields. However, I did not find in JLS any statements saying this. Will ask the [concurrency-interest](http://altair.cs.oswego.edu/mailman/listinfo/concurrency-interest) list because this is interesting.

Anyway, because of the data race I described above, I would suggest to use a straightforward DCL implementation that acquires a monitor only on pessimistic code path, which happens a finite number of times (if the number of threads calling `get()` is final), so the absolute majority of calls is guaranteed to do only a single volatile read per `get()` invocation. Please refer to the comment for `memoize(Function)` for the example of the code.

oshi-core/src/main/java/oshi/util/Memoizer.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * Store the function in a thread-safe map to be computed only once.
+     * 
+     * @param f
+     *            The {@link java.util.function.Function} to memoize
+     * @return A memoized version of the function
+     */
+    public static <I, O> Function<I, O> memoize(Function<I, O> f) {
```
stIncMale: While implementing lazy-initialization with CHM is short, it is an overkill at the very least on terms of memory footprint. It's possible to implement `memoize(Function)` with DCL, and then implement `memoize(Supplier)` by simply reusing `memoize(Function)` (because `Supplier` is essentially a `Function` that ignores its argument):
```java
public static <I, O> Function<I, O> memoize(Function<I, O> f) {
  return new Function<I, O>() {
    private volatile O value;

    public O apply(I arg) {
      O value = this.value;
      if (value == null) {
        synchronized (this) {
          value = this.value;
          if (value == null) {
            this.value = value = f.apply(i);
            if (value == null) {
              throw new AssertionError("Internal error. Function must not return null");
            }
          }
        }
      }
      return value;
    }
  };
}

public static <T> Supplier<T> memoize(Supplier<T> s) {
  return () -> memoize(i -> s.get()).apply(null);
}
```

oshi-core/src/main/java/oshi/SystemInfo.java
```diff
@@ -47,9 +48,7 @@
  * platform-specific implementations of {@link oshi.software.os.OperatingSystem}
  * (software) and {@link oshi.hardware.HardwareAbstractionLayer} (hardware).
  */
-public class SystemInfo implements Serializable {
```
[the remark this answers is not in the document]
dbwiddis: Good call... I'll mention it but don't think it requires a major version bump.

oshi-core/src/main/java/oshi/SystemInfo.java
```diff
@@ -71,9 +70,10 @@
         }
     }
 
-    private volatile OperatingSystem os = null;
+    private final Supplier<OperatingSystem> os = Memoizer.memoize(this::createOperatingSystem);
```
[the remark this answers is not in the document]
dbwiddis: I learned something new!

oshi-core/src/main/java/oshi/util/Memoizer.java
dbwiddis: Thanks. I might try this later.  I haven't used the function version yet and may not see a need to.  The supplier code below seems rather efficient as well.

oshi-core/src/main/java/oshi/util/Memoizer.java
dbwiddis: I copied that code directly from [this StackOverflow answer](https://stackoverflow.com/questions/35331327/does-java-8-have-cached-support-for-suppliers/35335467#35335467) (and have added a comment with the source). The comments discuss the happens-before situation.  Note that `getFirst()` is synchronized so only one thread will enter that, updating the values of `initialized` and `delegate` before exiting.  Other threads which may have been waiting for the synchronized method will skip by the `initiialized` check, and any future threads will see the new delegate and never get to the synchronized method at all.

oshi-core/src/main/java/oshi/util/Memoizer.java
dbwiddis: Also I need to figure out how to implement expiration. Currently only one real use case in the code, reading swapfile stats.  I read used and total at the same time and if the getUsed and getTotal are called "close together" not require a second read.

oshi-core/src/main/java/oshi/util/Memoizer.java
stIncMale: I think the explanation of the code in that thread on SO is not correct for the reasons I mentioned earlier here. The implementation also relies on the assumption that captures of variables in lambda expressions have final fields semantics, which may be true (would be reasonable), but I do not see anything about this in JLS. If this assumption does not hold, then the code is completely incorrect; if the assumption holds, then the code has performance problems because the following claim about the synchronized block: "_may happen at most once for every thread, as afterwards, it has the most recent values_" is technically incorrect.

I expressed all my thoughts with more details in the email to concurrency-interest (you are in Bcc). You can follow the thread `another lazy initialization, now with method reference` here [http://cs.oswego.edu/pipermail/concurrency-interest/2019-August/thread.html](http://cs.oswego.edu/pipermail/concurrency-interest/2019-August/thread.html), or by subscribing to the list here [http://altair.cs.oswego.edu/mailman/listinfo/concurrency-interest](http://altair.cs.oswego.edu/mailman/listinfo/concurrency-interest). Let's see what the replies will be, if any (it's an interesting piece of code, and obviously I can be mistaken).

oshi-core/src/main/java/oshi/util/Memoizer.java
stIncMale: > Also I need to figure out how to implement expiration

As I mentioned, both Guava and my [`MemoizedObject`](https://gist.github.com/stIncMale/2fce5f39aa37ac4584e0635ca6f67fe5) implement expiration, and both contain concurrency tests. I wrote `MemoizedObject` before I discovered it's present in Guava (last time I checked, the implementation were similar).

There are though memory footprint optimizations one may add to `MemoizedObject`:
- use the monitor associated with `this` instead of having a dedicated `monitor ` object;
- use a more simple implementation for situations when there is no expiration to save space by not having fields related to TTL;
- a pair of fields `ttlNanos`, `lastRefreshedTNanos` may potentially be replaced with a single field, but I am not completely sure and this is may not be very straightforward.

oshi-core/src/main/java/oshi/hardware/common/AbstractComputerSystem.java
```diff
@@ -88,7 +63,6 @@ public Baseboard getBaseboard() {
      */
     protected abstract Baseboard createBaseboard();
 
-    /** {@inheritDoc} */
     @Override
     public String toString() {
         StringBuilder sb = new StringBuilder();
```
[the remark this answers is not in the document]
dbwiddis: Fair comment, and I don't overuse them when simple concatenation (e.g., a single line) is used.  In the case of these toStrings, though, I find the toString version more readable.  I'll leave these as-is.

oshi-core/src/main/java/oshi/SystemInfo.java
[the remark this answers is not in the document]
Nadahar: I happen to disagree with this, I simply can't see how implementing this "marker interface" can harm anybody. It's like if the whole world was only about web services, and the only thing that "matters" is the risk that follows by deserializing untrusted or otherwise "manipulatable" sources. It is however entirely up to the application developer if he/she will use serialization, and what sources that will be used for deserialization. 

I don't mean to say that serialization is "perfect" in any way, I wish it had more capabilities, especially that it was possible to initialize final fields from a custom deserializer. That said, it's certainly better than nothing, or having to include huge external libraries to achieve serialization if the needs are simple. I don't see the big problem with using it for things like caching, either in a local database or in memory, for many applications.

Why take away these possibilities just because some web developers abuse the functionality..?

oshi-core/src/main/java/oshi/util/Memoizer.java
dbwiddis: @stIncMale -- I just tried your DCL version of the memoizer.. it doesn't work (run the main method in the latest commit) .... yes I know you have suggested to use the MemoizedObject, I'm trying to do as much "roll my own" as I can here...

oshi-core/src/main/java/oshi/util/Memoizer.java
```diff
@@ -0,0 +1,167 @@
...
+                        }
+                    }
+                }
+                return value;
+            }
+        };
+    }
+
+    public static <T> Supplier<T> memoize1(Supplier<T> s) {
+        return () -> memoize1(i -> s.get()).apply(null);
```
stIncMale: > I just tried your DCL version of the memoizer... it doesn't work (run the main method in the latest commit) ...

Ahah :) Yeah, it's not DCL, I was just too quick to convert `Supplier` into a `Function`. Obviously the correct code is:
```java
public static <T> Supplier<T> memoize1(Supplier<T> s) {
  Function<Void, T> f = memoize1(i -> s.get());
  return () -> f.apply(null);
}
```
Notice how this time a function is created only once, while in my first attempt I was creating a new function (by calling `memoize1(Function<I, O> f)`) each time `Supplier.get()` was called.

The manual way you are using to check wether an implementation of a memoizer is correct or not is not automatic and can only find obvious logical problems (as the one addressed above), but it can't detect memory synchronization issues.

It's very easy to modify [`MemoizedObjectConcurrencyTest.java`](https://gist.github.com/stIncMale/2fce5f39aa37ac4584e0635ca6f67fe5) so that it can be used for the methods in `Memoizer`. I did this and checked `memoize(Supplier<T> original, long ttlNanos)` implementation you adopted from Guava, and I also checked the `memoize1(Supplier<T> s)` with the fix specified above. Both passed the test, which is good news! :)

dbwiddis: Nice!   I just did a bunch of microbenchmarks on the Guava version vs. the one from the SO post.  The Guava version takes half the time as the SO version for the first read, and only slightly longer for later reads... so I'll ditch the current version and go with the Guava-based version ... and yes, use your test class eventually.  That's tonight's project.  I've got the first set of classes complete, so I'll probably merge this PR and start on another group.

dbwiddis: Once again, @stIncMale, I want to thank you for your patience as I learn this. I wish I'd have taken the time to figure it out before, it actually makes so many things easier that I struggled with for a while!

I attempted to implement your test class, and had to make a few changes:
- Removed `@Nullable` annotation
- Replaced `var` with explicit typing (testing on JDK8)
- In the calculation of `actualNumberOfRefreshes` the code was asking for `s.value` but since I'm testing a method rather than an object, my only way to get that was to call `s.get()` which pre-increments value before returning it, so I decremented for the test.  (Not sure if that's correct or thread-safe to do!)

When I run the class as a test by itself, it passes.  However, if I run tests on my entire project, that test consistently fails:

> java.lang.AssertionError: ttlNanos=100000, minExpectedNumberOfRefreshes=2.0, maxExpectedNumberOfRefreshes=110.0, actualNumberOfRefreshes=1

I suspect it's my decrementing hack at fault here, but I'm not sure, and not sure of the appropriate workaround to get the test working.

Related, this test clearly takes 60 seconds of CPU time, and I'd like to avoid running it with every build.  Is there a standard way of triggering the test only when required?

stIncMale: > I suspect it's my decrementing hack at fault here

No, `actualNumberOfRefreshes = s.get() - 1` is fine because the thread doing this is the only thread accessing the `s` object (we know that all the other threads started in the executor `ex` are dormant because we waited for them to complete).

I think here is what happens:
- when you run this test from a full build, something terminates the test because all tests together (or the whole build) takes too much time (I did not look at how the build process is done in OSHI).
- when the test is terminated, executer's threads are interrupted
- interrupt is handled incorrectly in the test, it affects the `while` loop logic, when it should just throw an exception instead.

I was even able to reproduce the problem you are observing by simply adding a single line `Thread.currentThread().interrupt();` right before the `while` loop, which confirms my hypothesis.

The fix is simple, move the check of the interrupted flag inside the loop and throw an `InterruptedException`:
```
while (System.nanoTime() - beginNanos < iterationDurationNanos
    || (guaranteedLastIteration = !guaranteedLastIteration)) {
  if (Thread.currentThread().isInterrupted()) {
    throw new InterruptedException();
  }
  ...
}
```
I am confident this will solve the problem. It's also a correct way of handling interruptions in this test. I also fixed this test in the gist with the test that I linked earlier.

> Is there a standard way of triggering the test only when required?

I believe this is the standard way: https://github.com/junit-team/junit4/wiki/categories

oshi-core/src/test/java/oshi/util/MemoizerTest.java
```diff
@@ -40,7 +39,7 @@ public void after() throws InterruptedException {
         ex.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
     }
 
-    @Test
+    // @Test
```
stIncMale: Does this mean the fix proposed here https://github.com/oshi/oshi/pull/959#discussion_r317253433 not helped?

dbwiddis: Haven't gotten to that.  I'm hunting down the other failing tests involving the Linux C library requiring Unix functions.

dbwiddis: And... got to it, but it still fails.

dbwiddis: I still think it's not hitting that last guaranteed iteration when it's interrupted, and is slow when running alongside other tests, so the interruption is more likely to prevent that.  Can we set the interrupted flag and read it later to know whether to decrement the minimum by one?

stIncMale: > I still think it's not hitting that last guaranteed iteration when it's interrupted

With the proposed fix, the interrupted flag does not affect the `while` condition. Which means a thread definitely enters the loop at least once, and then if it's interrupted, it throws an `InterruptedException` thus skipping the `m.get()` call. This would have caused observing only a single iteration when asserting at the end of the test, but the thrown exception is re-thrown by `result.get()` which leads to test termination.

Could you tell me how do you run the build to reproduce the failure? I'd like to observe what you see and debug it, but `mvn clean package` finishes successfully for me.

dbwiddis: The test is only failing on one machine, and only in one situation... using Eclipse, I highlight my project `oshi-core` and select the `src\test\java` package, then "Run as... Junit Test".  It executes all the tests in the project.  It doesn't fail every time, but does at least half the time I try. Might be a hardware limitation (2 physical/4 logical cores).

 Earlier when the tests were failing on Travis it was only on JDK 1.8 (Mac) so that's another possibility.

Capture

dbwiddis: Interestingly when it succeeds it takes longer than when it fails... successes are over 60 seconds, while failures are ~40 seconds.

Nadahar: Inconsistent failures like that, often only reproducible on certain systems/environments, are typical for concurrency issues. It often takes specific circumstances for the nanosecond timing to be in such a way that the bug is triggered.

stIncMale: @Nadahar
In this case I am not sure this is related to concurrency issues. Flaky behaviour of this test may (and likely is) be caused by unfortunate timings within a single thread.

@dbwiddis 
Are you sure you are running the test after fixing it? The version in the branch `dbwiddis:thread-safety` (commit d66d7ebbbbfaa6be541218f0c7b4412472459d6e) does not contain the fix? May I ask you to commit the version of the test which you are running?

Nadahar:
@stIncMale You would know this much better, I haven't studied the code - I just meant to say that I generally consider such behavior as a "red flag" for potential concurrency issues.

oshi-core/src/test/java/oshi/util/MemoizerTest.java
stIncMale: Oh, you introduced 98a03a74db08b719acdbe798e905cae327bc0d0e, and I missed it, sorry. So the current version is:
```
while ((System.nanoTime() - beginNanos < iterationDurationNanos
    && !Thread.currentThread().isInterrupted())
    || (guaranteedLastIteration = !guaranteedLastIteration)) {
  if (Thread.currentThread().isInterrupted()) {
    throw new InterruptedException();
  }
...
```
The check of the interrupted flag is still there, and is causing the problem. Could you try with the version I suggested originally?
```
while (System.nanoTime() - beginNanos < iterationDurationNanos
    || (guaranteedLastIteration = !guaranteedLastIteration)) {
  if (Thread.currentThread().isInterrupted()) {
    throw new InterruptedException();
  }
  ...
}
```

dbwiddis: Ah, sorry, I missed the removal of the middle clause in the surrounding loops and just added the new conditional.  I was too busy debugging silly copy/paste errors instead of actually learning what was happening.   Will make that change and test...

dbwiddis: Still the same error on my Win10-64 laptop in Eclipse (4 out of 5 times).  Confirmed latest commit:

> HEAD is now at ff51f0187d Fix MemoizerTest

The loop is:
```
while (System.nanoTime() - beginNanos < iterationDurationNanos
        || (guaranteedLastIteration = !guaranteedLastIteration)) {
    if (Thread.currentThread().isInterrupted()) {
        throw new InterruptedException();
    }
...
}
```

dbwiddis: Latest code also failed the Mac Java 8 build on Travis:
https://travis-ci.org/oshi/oshi/jobs/576107414

stIncMale: Uh, I am absolutely failing to see what may be causing this. We know, that either one of the following or both are incorrect:
- `MemoizerTest`;
- `Memoizer.memoize`.

I am suggesting to try playing around with them trying to exclude (to some extend) one or the other out of the equation:
1. set `numberOfThreads = 1` in `MemoizerTest`. Check if the problem reproduces;
2. _undo the changes_, mark `synchronized` the whole `get` method in the `Supplier` produced by `Memoizer.memoize`. Check if the problem reproduces.

If any of the aforementioned changes the reproducibility of the problem, we will at least discover the most plausible culprit and will continue thinking from there. But this is just what I think, maybe you have other ideas.

P.S. I tried reproducing it on JDK 1.8.0_162 (HotSpot), and it still never fails for me, which is not helping.

dbwiddis: Tried:
- `numberOfThreads = 1` ... problem remains
- synchronizing entire `get()` in the Memoizer... problem remains
- both of the above simultaneously ... problem remains

Other observations:
- I can reproduce the problem at will by causing disk activity from within Eclipse, such as editing and saving a file.  This is interesting because it introduces some latency.
- The comment on the `while()` loop indicates that it is guaranteed to run once -- either because we are under the ttl or because (when that fails) of the guaranteed iteration. (But why not a do...while instead of the boolean? But I digress.) . But what of the situation where we do not reach the `while()` loop until ttl has expired?  We will only get one iteration.

dbwiddis: To clarify:  the loop will run once by guarantee but is not guaranteed to run twice which the (failing) test assumes.  I have an idea on how to test this... will report back shortly.

dbwiddis: So there are multiple times the loop only executes once.   These do not all cause failures. But every time it does fail, it is immediately after a one-loop case.

stIncMale: > But what of the situation where we do not reach the while() loop until ttl has expired? We will only get one iteration.

Exactly (great catch)! If you introduce a pause (`Thread.sleep`) longer than `iterationDurationNanos` right at the beginning of the task submitted to an executor (i.e. right before `Long previousValue = m.get();`), the problem will be reproduced almost always. Unfortunately, I have some things to do right now, but I will get back to this as soon as I can.

P.S.
> why not a do...while instead of the boolean?
Yes, it very well may be possible. I so rarely use do-while (never), that I forgot it exists...

stIncMale: > So there are multiple times the loop only executes once. These do not all cause failures.

These situations don't cause failure when the memoizer is supposed to remember the calculated value forever (ttl is -1). And again, we are counting not the loop iterations, but the number of the calls to the supplier wrapped into a supplier that implements memoization.

dbwiddis: I think I've figured it out.  You are assuming that the `value` will reach 2:
- Initial call to `m.get()` sets it to 1 before reaching the loop
- In the loop, the second call will increment after ttl has expired.

However, in the case where (a) the loop only executes once, and (b) ttl has not expired, the second call to `m.get()` does not increment the value.

Armed with this information I can cause the failure every time on any system by calling `run()` with `ttlNanos = 1`, forcing the one-iteration case, and relying on catching an instance where code executes faster than `System.nanoTime()` can increment, and thus the `get()` returns the same value.

dbwiddis: Er... ignore that `ttlNanos = 1`, mental math mistake.  It's the other direction, a long `ttlNanos` equal to the first argument, that reliably fails.

stIncMale: I thought you were talking about this same idea initially) And yes, this is exactly the case. And the approach of reproducibility I proposed here (https://github.com/oshi/oshi/pull/959#discussion_r317366400) effectively leads to the same situation you have with `ttl = iterationDurationNanos`.

However, we still want the test to call the original supplier at least twice (this increases our chances of spotting a bug). This means that the condition of the `while` loop must take into account not only the time passed since the iteration started, but also the time passed since we called the original supplier before entering the loop. Something like this:
```
Long previousValue = m.get();
final long firstSupplierCallNanos = System.nanoTime();
assertNotNull(previousValue);
assertTrue(previousValue > 0);
boolean guaranteedLastIteration = false;
// this loop is guaranteed to be executed at least once regardless of the values
// returned by System.nanoTime
while (System.nanoTime() - beginNanos < iterationDurationNanos
    || System.nanoTime() - firstSupplierCallNanos < ttlNanos
    || (guaranteedLastIteration = !guaranteedLastIteration)) {
...
```
And if this works, we can further think about rewriting it with do-while, but it's important to not introduce any occasional additional iterations while doing so, because it in that case w may also need to re-estimate the exact Max value we are asserting.

dbwiddis: Hmmm remember I altered `s.value` to `s.get() - 1`.   But `s` is the non-threadsafe supplier.  Should I be testing `m.get() - 1` after an enforced delay of `ttlNanos`?

dbwiddis: The latest suggestion fixed the problem.   I also changed it to a do-while and removed the boolean.   Can't get this to fail....

```
                Long previousValue = m.get();
                final long firstSupplierCallNanos = System.nanoTime();
                assertNotNull(previousValue);
                assertTrue(previousValue > 0);
                // this loop is guaranteed to be executed at least once regardless of the values
                // returned by System.nanoTime
                long now;
                do {
                    if (Thread.currentThread().isInterrupted()) {
                        throw new InterruptedException();
                    }
                    final Long newValue = m.get();
                    assertNotNull(newValue);// check that we never get uninitialized value
                    assertTrue(String.format("newValue=%s, previousValue=%s", newValue, previousValue),
                            newValue >= previousValue);// check that the counter never goes down
                    previousValue = newValue;
                } while ((now = System.nanoTime()) - beginNanos < iterationDurationNanos
                        || now - firstSupplierCallNanos < ttlNanos);

```

stIncMale: As I mentioned previously, `s.get() - 1` is safe because the loop doing `Future.get()` provides both a happens-before relation which gives us visibility and guarantees that all other threads that had access to the unsafe supplier are now dormant. So the assertion part of the test has exclusive access to the unsafe supplier.

Doing the same via `m` is less trivia because you would have to figure if it is going to call the underlying supplier or not.

stIncMale: > The latest suggestion fixed the problem. I also changed it to a do-while and removed the boolean. Can't get this to fail....

@dbwiddis Great! And looks good also, thank you! I'm so glad this mystery is solved.

dbwiddis:
> * 1 for Useless comparison test

LGTM is telling me I'm useless, and Travis is reminding me that StackOverflow exists. Not my night. :D

dbwiddis:
I've gotten all the hardware that's not an array call done.   Probably will merge this version before I start digging into the array returns (debating removing setters and having gazillions of arguments in the constructors, or just documenting "setters are not thread safe"...

stIncMale:
@dbwiddis 
Interestingly, I was not crazy when introduced the variable `guaranteedLastIteration` and logic around it. It's easy to quite reliably reproduce a problem we have in the test right now by adding runs with the following settings:
```
iterationDurationNanos = TimeUnit.MICROSECONDS.toNanos(1300);
for (int r = 0; r < 4000; r++) {
  run(iterationDurationNanos, iterationDurationNanos + 1);
}
```

`guaranteedLastIteration` existed not to make sure the loop iterates at least once, but to make sure the loop iterates once more _after timing conditions will have stopped being met_. `do-while` is not help achieving this. So once we checked and observed that `now - firstSupplierCallNanos < ttlNanos` is not true anymore, we are right now assuming a refresh corresponding to this expiration happened and therefore expect min = 2. But we may only guarantee this only if we do one more iteration (which we are not doing right now).

So what I am saying, we need to return that `guaranteedLastIteration` back.

dbwiddis:
Busy now but will look at this later. I thought about this, and agree your testcase above will cause it to fail but it will also completely negate the purpose of the test to run multiple iterations.  I suppose it could fail in the rare instance where the first `get()` works and there's a long delay before the second.

In which case the variable name should be `guaranteedExtraIteration`. ;)

dbwiddis:
Take a look at #965

PR #960, opened 2019-08-19
Added OSProcess constructor with PID (#957)

### LINKED ISSUE #957: OSProcess constructor with PID
Add the capability to construct a new process object given a pid:  `new OSProcess(pid)`.

Steps to implement:
1. Write a new constructor that takes a pid as an argument, sets the `processID` attribute, and calls an `updateAttributes()` method. 
2. Write a new public `updateAttributes()` method which calls the platform-specific `getProcess(this.processId)` and then copies all the values from the return value OSProcess to the class attributes.  You'll have to switch based on `SystemInfo.getCurrentPlatformEnum()`.

Potat0x:
Fixes #957

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/OSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSProcess.java
@@ -23,6 +23,15 @@
  */
 package oshi.software.os;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import oshi.SystemInfo;
+import oshi.software.os.linux.LinuxOperatingSystem;
+import oshi.software.os.mac.MacOperatingSystem;
+import oshi.software.os.unix.freebsd.FreeBsdOperatingSystem;
+import oshi.software.os.unix.solaris.SolarisOperatingSystem;
+import oshi.software.os.windows.WindowsOperatingSystem;
+
 import java.io.Serializable;
 
 /**
@@ -34,6 +43,8 @@ import java.io.Serializable;
 public class OSProcess implements Serializable {
 
     private static final long serialVersionUID = 3L;
+
+    private static final Logger LOG = LoggerFactory.getLogger(OSProcess.class);
 
     private String name = "";
     private String path = "";
@@ -93,6 +104,44 @@ public class OSProcess implements Serializable {
          * Other or unknown states not defined
          */
         OTHER
+    }
+
+    /**
+     * <p>
+     * Constructor for OSProcess.
+     * </p>
+     */
+    public OSProcess() {
+    }
+
+    /**
+     * <p>
+     * Constructor for OSProcess.
+     * </p>
+     *
+     * @param processID
+     *                 process ID
+     */
+    public OSProcess(int processID) {
+        this.processID = processID;
+        updateAttributes();
+    }
+
+    /**
+     * Updates all process attributes.
+     */
+    public void updateAttributes() {
+        OperatingSystem operatingSystem = getCurrentOperatingSystem();
+        if (operatingSystem != null) {
+            OSProcess process = operatingSystem.getProcess(this.processID);
+            if (process != null) {
+                copyValuesToThisProcess(process);
+            } else {
+                LOG.error("No process found: {}", this.processID);
+            }
+        } else {
+            LOG.error("Unsupported platform. No update performed.");
+        }
     }
 
     /**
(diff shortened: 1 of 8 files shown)


oshi-core/src/main/java/oshi/software/os/OSProcess.java
```diff
@@ -95,6 +106,44 @@
...
+     * <p>
+     * Constructor for OSProcess.
+     * </p>
+     */
+    public OSProcess() {
+    }
+
+    /**
+     * <p>
+     * Constructor for OSProcess.
```
dbwiddis: Can we be a little more verbose with the javadoc here, stating that passing the PID will create an object populated with current stats?  You can even say it's equivalent to `{@link OperatingSystem.getProcess(processID)}`.

oshi-core/src/main/java/oshi/software/os/OSProcess.java
```diff
@@ -674,4 +723,48 @@ public int getBitness() {
...
+            case LINUX:
+                return new LinuxOperatingSystem();
+            case MACOSX:
+                return new MacOperatingSystem();
+            case SOLARIS:
+                return new SolarisOperatingSystem();
+            case FREEBSD:
+                return new FreeBsdOperatingSystem();
+            default:
+                return null;
```
dbwiddis: This should never happen unless we add a new OS in the platform enum and don't implement it here, so it's probably safe to throw an exception here, and then you don't have to handle the null return value.

oshi-core/src/main/java/oshi/software/os/OSProcess.java
```diff
@@ -95,6 +106,44 @@
...
+     * Updates all process attributes.
+     */
+    public void updateAttributes() {
+        OperatingSystem operatingSystem = getCurrentOperatingSystem();
+        if (operatingSystem != null) {
+            OSProcess process = operatingSystem.getProcess(this.processID);
+            if (process != null) {
+                copyValuesToThisProcess(process);
+            } else {
+                LOG.error("No process found: {}", this.processID);
```
dbwiddis: This may be a reasonably common return value so error logging isn't appropriate.  I think perhaps we should make this method return boolean, and just do debug level logging here.

oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
```diff
@@ -279,4 +279,17 @@ public void testGetCommandLine() {
 
         assertTrue(processesWithNonEmptyCmdLine >= 1);
     }
+
+    @Test
+    public void testConstructProcessWithGivenPid() {
+        SystemInfo si = new SystemInfo();
+        OperatingSystem os = si.getOperatingSystem();
+        OSProcess oldProcess = os.getProcess(os.getProcessId());
+        int givenPid = oldProcess.getProcessID();
```
dbwiddis: You get major bonus points for writing a testcase!

I'm curious why store the result of `getProcessID()` in a variable unless you want to test against it... why not just store `os.getProcessId()` two lines up and then use it twice?  (Or just not use a temp variable and only use `os.getProcessId()`.)

oshi-core/src/main/java/oshi/software/os/OSProcess.java
dbwiddis: Hmm. This also introduces the question of what happens when a user attempts to instantiate a process with an invalid pid -- a boolean return from "update" still has to be handled in the constructor. Clearly the stats fail, but since we are instantiating a new object it needs to be non-null and have some way for the user to test whether it was successful.

Potat0x: How about static factory method which returns Optional?
`Optional<OSProcess> byPid(int processID);`

oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
Potat0x: Thanks :)

Good point, I change code to use temp variable twice (I think temporary variable is more meaningful and easier to read in this case).

Potat0x:
In OperatingSystemTest.java I found always true conditions (inside assert):
`            if (onePid >= 0) {
                assertTrue(0 <= os.getChildProcesses(onePid, 0, null).length);
            }`
There is two more similar assertions. Should I fix it?

oshi-core/src/main/java/oshi/software/os/OSProcess.java
dbwiddis: That's not much different than the current `getProcess(int processID)` except with an optional rather than null.  (And the API for that is fixed so we can't change it.)  

That's not the problem I'm discussing... the problem is that someone writing `new OSProcess(1234)` will definitely get a Process object unless we throw an exception.  We could throw an `InstantiationError`, for example, and require them to handle it.  Or a `MissingResourceException`. Or return a class with the pid and essentially zero data.  Perhaps I should have thought about this dilemma before suggesting the feature!

dbwiddis:
OK, this looks great!  I think the fact that the process name, path, etc. will be empty is sufficient testing material for a return value.  I'll tweak the javadoc a bit for that and then merge.  Thanks for your contribution!

(And thanks for pointing out those testcases... that one has always been problematic with race conditions.... )

dbwiddis:
Actually, after reviewing, I decided the best thing to do for an invalid PID was to throw an `InstantiationException` so I added that code, updated the docs and the tests (and removed the always true ones you pointed out).

However, while running the testcase, it took a long time.  I tracked it down to creating a new (platform specific) OperatingSystem object every time the update method was called, and all the overhead we get in a new OS (in Windows, multiple WMI calls, reading the registry, reading the event log...).  

If this was only for creating the object I'd live with it, but the "update" method is intended to be one that a user will call regularly so we need to save that value somewhere, ideally passed as part of the constructor when the class is first created (required) rather than creating a new one ourselves.  I'm thinking:
- private no-arg constructor, so we can rely on this field
- `OSProcess(OperatingSystem os)`  constructor that sets the internal attribute, that we call from the other `getProcess()` code
- `OSProcess(OperatingSystem os, int ProcessID)`  constructor that sets the internal attribute and updates for that PID, thus requiring the user to provide us the OS as an argument.

Potat0x:
Why not all three? It should be good.

dbwiddis:
If you don't let anyone call with no-args you can be sure the operating system field is populated on construction.  If you leave it blank you have to create the object to store it, which is a (relatively) expensive operation and should be avoided.  There's really no good reason to offer it.

Potat0x:
You are right, of course (I misread your post, sorry). Changes pushed.

dbwiddis:
Looks great!  You no longer needed the `getCurrentOperatingSystem()` method so I deleted it -- but then it complained about not initializing `operatingSystem` in the non-used constructor so I deleted that too!   We don't need it -- code trying to invoke it doesn't even compile.

Letting the tests run and then I'll merge this!  Thanks for your efforts!

PR #964, opened 2019-08-25
Processor identifier to centralProcessor

Praveen101997:
Move the ProcessorIdentifier inner class to the CentralProcessor class

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -22,6 +22,9 @@
  * SOFTWARE.
  */
 package oshi.hardware;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import oshi.util.ParseUtil;
 
 /**
  * This class represents the entire Central Processing Unit (CPU) of a computer
@@ -362,7 +365,7 @@ public interface CentralProcessor {
      *		   and family information (also called the signature of a CPU)
      */
     ProcessorIdentifier getProcessorIdentifier();
-    
+
 
     /**
      * A class representing a Logical Processor and its replationship to physical
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -22,8 +22,10 @@
  * SOFTWARE.
  */
 package oshi.hardware;
+
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
 import oshi.util.ParseUtil;
 
 /**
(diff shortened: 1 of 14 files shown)


Praveen101997:
> Move the ProcessorIdentifier inner class to the CentralProcessor class
@dbwiddis 
I commit the changes as per requirement
But not sure as it is my first contribution
If anything is wrong, please guide and correct me

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
```diff
@@ -36,38 +36,38 @@
...
-    String getVendor();
+    @Deprecated String getVendor();
 
     /**
      * Name, eg. Intel(R) Core(TM)2 Duo CPU T7300 @ 2.00GHz
      *
      * @return Processor name.
      */
-    String getName();
+    @Deprecated String getName();
```
dbwiddis: While technically correct, for style purposes put in a newline after the `@Deprecated`.  Also add a `@deprecated` annotation to the javadoc (note the lower case) with a comment `replaced by {@link ProcessorIdentifier#getName()}`

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
```diff
@@ -453,4 +458,84 @@ public int getProcessorGroup() {
...
+            Matcher matcher = pattern.matcher(cpuName);
+            if (matcher.find()) {
+                String unit = matcher.group(1);
+                this.cpuVendorFreq = ParseUtil.parseHertz(unit);
+            } else {
+                this.cpuVendorFreq = -1L;
+            }
+        }
+
+        public String getVendor() {
```
dbwiddis: Good, but these need javadocs. You can copy the ones from above that you are deprecating.

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
```diff
@@ -339,6 +339,11 @@ public int getIndex() {
         }
     }
 
+
+    ProcessorIdentifier getProcessorIdentifier();
```
dbwiddis: You'll need to write a new javadoc for this.  Something like:

> Returns the CPU's identifier strings, including stepping, model, and family information (also called the signature of a CPU)

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
```diff
@@ -453,4 +458,84 @@ public int getProcessorGroup() {
...
+        }
+
+        public boolean isCpu64bit() {
+            return cpu64bit;
+        }
+
+        public long getVendorFreq() {
+            return cpuVendorFreq;
+        }
+
```
dbwiddis: for completeness, let's add a `toString()` that returns `getIdentifier()`.

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
dbwiddis: These are valid, but for style purposes add a new line between the `@Deprecated` annotation and the `get`.  You also need to add the (lowercase) `@deprecated` annotation in the Javadoc, pointing to the new location this information will be.  Such as `@deprecated Replaced by {@link ProcessorIdentifier#getName}`

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
dbwiddis: This is good here but you haven't implemented it in the AbstractCentralProcessor class, so you're failing compilation.  Also you need a Javadoc for this, something like 

> Returns the CPU's identifier strings such as the CPU's stepping, model, and family information (also called the signature of a CPU).

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
dbwiddis: Good but you need javadocs here.  Copy the ones from the deprecated getters.

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
dbwiddis: Let's add a toString() that returns `getIdentifier()`.

oshi-core/src/main/java/oshi/hardware/common/AbstractCentralProcessor.java
```diff
@@ -98,47 +98,56 @@ public AbstractCentralProcessor() {
 
     @Override
     public String getVendor() {
-        return cpuid.get().cpuVendor;
+        return getProcessorIdentifier().getVendor();
```
dbwiddis: These are all good, but don't forget to add a few lines above, a `getProcesserIdentifier()` implementation that returns `cpuid.get()`.

dbwiddis:
Also this needs a [CHANGELOG](https://github.com/oshi/oshi/blob/master/CHANGELOG.md) entry.

Praveen101997:
> Also this needs a [CHANGELOG](https://github.com/oshi/oshi/blob/master/CHANGELOG.md) entry.

Do I have to also import this library in CentralProcessor which is previously imported in abstractCentralProcessor
import java.util.regex.Matcher;
import java.util.regex.Pattern;
import oshi.util.ParseUtil;

Praveen101997:
@dbwiddis  I didn't add the issue name in the changelog.
So, I again pushed changes in it.
And the process is in way.

PR #965, opened 2019-08-25
Memoizer test updates and comments

dbwiddis:
Logic tweaks and ample comments for clarity.  @stIncMale please review!

### DIFF
--- a/oshi-core/src/test/java/oshi/util/MemoizerTest.java
+++ b/oshi-core/src/test/java/oshi/util/MemoizerTest.java
@@ -115,8 +115,7 @@ public final class MemoizerTest {
                 boolean guaranteedIteration = false;
                 long now;
                 while ((now = System.nanoTime()) < beginNanos + iterationDurationNanos
-                        || now < firstSupplierCallNanos + ttlNanos
-                        || (guaranteedIteration = !guaranteedIteration)) {
+                        || now < firstSupplierCallNanos + ttlNanos || (guaranteedIteration = !guaranteedIteration)) {
                     // guaranteedIteration will only be set true when the first two timing
                     // conditions are false, which will allow at least one iteration. After that
                     // final iteration the boolean will toggle false again to stop the loop.
@@ -148,13 +147,9 @@ public final class MemoizerTest {
 
     }
 
-    private void finishAllThreads(Collection<Future<Void>> results) throws Throwable {
+    private void finishAllThreads(Collection<Future<Void>> results) throws InterruptedException, ExecutionException {
         for (final Future<Void> result : results) {
-            try {
-                result.get();
-            } catch (final ExecutionException e) {
-                throw e.getCause();
-            }
+            result.get();
         }
     }
 
--- a/oshi-core/src/test/java/oshi/util/MemoizerTest.java
+++ b/oshi-core/src/test/java/oshi/util/MemoizerTest.java
@@ -114,8 +114,8 @@ public final class MemoizerTest {
                 // least once regardless of whether we have exceeded time delays
                 boolean guaranteedIteration = false;
                 long now;
-                while ((now = System.nanoTime()) < beginNanos + iterationDurationNanos
-                        || now < firstSupplierCallNanos + ttlNanos || (guaranteedIteration = !guaranteedIteration)) {
+                while ((now = System.nanoTime()) - beginNanos < iterationDurationNanos
+                        || now - firstSupplierCallNanos < ttlNanos || (guaranteedIteration = !guaranteedIteration)) {
                     // guaranteedIteration will only be set true when the first two timing
                     // conditions are false, which will allow at least one iteration. After that
                     // final iteration the boolean will toggle false again to stop the loop.
@@ -147,13 +147,14 @@ public final class MemoizerTest {
 
     }
 
-    private void finishAllThreads(Collection<Future<Void>> results) throws InterruptedException, ExecutionException {
+    private static void finishAllThreads(Collection<Future<Void>> results)
+            throws InterruptedException, ExecutionException {
         for (final Future<Void> result : results) {
             result.get();
         }
     }
 
-    private void testIncrementCounts(long actualNumberOfIncrements, long iterationDurationNanos, long ttlNanos) {
+    private static void testIncrementCounts(long actualNumberOfIncrements, long iterationDurationNanos, long ttlNanos) {
         if (ttlNanos < 0) {
             assertEquals(String.format("ttlNanos=%d, expectedNumberOfIncrements=%d, actualNumberOfIncrements=%s",
                     ttlNanos, 1, actualNumberOfIncrements), 1, actualNumberOfIncrements);
--- a/oshi-core/src/test/java/oshi/util/MemoizerTest.java
+++ b/oshi-core/src/test/java/oshi/util/MemoizerTest.java
@@ -23,10 +23,11 @@
  */
 package oshi.util;
 
-
-
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static oshi.util.Memoizer.memoize;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -43,7 +44,8 @@ import org.junit.Before;
 import org.junit.Test;
 
 public final class MemoizerTest {
-    private static final int numberOfThreads = Math.max(5, Runtime.getRuntime().availableProcessors());
+    // We want enough threads that some of them are forced to wait
+    private static final int numberOfThreads = Math.max(5, Runtime.getRuntime().availableProcessors() + 2);
 
     private ExecutorService ex;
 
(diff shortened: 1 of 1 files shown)


oshi-core/src/test/java/oshi/util/MemoizerTest.java
```diff
@@ -65,39 +67,59 @@ public void after() throws InterruptedException {
...
-        final Supplier<Long> m = Memoizer.memoize(s, ttlNanos);
+        // The memoizer we are testing
+        final Supplier<Long> m = memoize(s, ttlNanos);
+        // Hold the results until all threads terminate
         final Collection<Future<Void>> results = new ArrayList<>();
+        // Mark the start time, end after iterationDuration
         final long beginNanos = System.nanoTime();
         for (int tid = 0; tid < numberOfThreads; tid++) {
             results.add(ex.submit(() -> {
+                // First read from the memoizer. Only one thread will win this race but all
```
stIncMale: Actually, `m.get()` calls here from each of the threads may lead to a refresh (a call to `s.get()` ) depending on the timings (though this does not change the max number of refreshes).

dbwiddis: Right. The first one is guaranteed to increment the value, the rest should at least see an incremented value (if not increment again if they are delayed until after ttl)

dbwiddis: The point is to note the test 2 lines lower that guarantees the >0 condition

oshi-core/src/test/java/oshi/util/MemoizerTest.java
```diff
@@ -65,39 +67,59 @@ public void after() throws InterruptedException {
...
+        ttlNanos = iterationDurationNanos / 100;
+        assertNotEquals(0, ttlNanos); // avoid div/0 later
         for (int r = 0; r < 4000; r++) {
-            assertTrue(run(iterationDurationNanos, iterationDurationNanos / 100));
+            run(iterationDurationNanos, ttlNanos);
         }
     }
 
-    private boolean run(final long iterationDurationNanos, final long ttlNanos) throws Throwable {
+    private void run(final long iterationDurationNanos, final long ttlNanos) throws Throwable {
```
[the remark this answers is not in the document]
stIncMale: How is this method long? :) Looks like the checker does not ignore lines containing only comments.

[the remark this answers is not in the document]
dbwiddis: [Cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) calculation.  15 is the threshold.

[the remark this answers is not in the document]
stIncMale: Are you sure it assesses cyclomatic complexity? This is not obvious from the description of the problem: _When methods are excessively long this usually indicates that the method is doing more than its name/signature might suggest. They also become challenging for others to digest since excessive scrolling causes readers to lose focus. Try to reduce the method length by creating helper methods and removing any copy/pasted code._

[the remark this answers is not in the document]
dbwiddis: Not really sure... but can't imagine that it includes comments.  In any case I refactored some bits to make the problem go away ;)

oshi-core/src/test/java/oshi/util/MemoizerTest.java
```diff
@@ -150,6 +144,21 @@ public Long get() {
...
         final long actualNumberOfIncrements = s.get() - 1;
+        testIncrementCounts(actualNumberOfIncrements, iterationDurationNanos, ttlNanos);
+
+    }
+
+    private void finishAllThreads(Collection<Future<Void>> results) throws Throwable {
+        for (final Future<Void> result : results) {
+            try {
+                result.get();
+            } catch (final ExecutionException e) {
```
stIncMale: Now that this is not a part of the original method anymore, it makes even more sense to no unwrap the original exception in order to not lose the whole trace leading not only to the executor thread but also to this `finishAllThreads` method. I think my original attempt to make the stack trace shorter has the potential of reducing clarity rather than improving it (in other words, I did a stupid thing originally).

dbwiddis: So I should not catch exceptions here?

oshi-core/src/test/java/oshi/util/MemoizerTest.java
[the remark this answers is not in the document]
dbwiddis: Digging into Codacy it's using the legacy PMD 5.x for this rule. I switched to 6.x which clearly states NCSS.

oshi-core/src/test/java/oshi/util/MemoizerTest.java
stIncMale: Yep, just let it be thrown without unwrapping:
```java
private static void finishAllThreads(Collection<Future<Void>> results) throws ExecutionException, InterruptedException {
    for (final Future<Void> result : results) {
        result.get();
    }
}
```
Also, if a method is not designed to use instance variables (fields) and or be overridden, it always makes sense to make it static.

oshi-core/src/test/java/oshi/util/MemoizerTest.java
```diff
@@ -148,13 +147,9 @@ public Long get() {
 
     }
 
-    private void finishAllThreads(Collection<Future<Void>> results) throws Throwable {
+    private void finishAllThreads(Collection<Future<Void>> results) throws InterruptedException, ExecutionException {
```
stIncMale: 👍

oshi-core/src/test/java/oshi/util/MemoizerTest.java
```diff
@@ -115,8 +115,7 @@ public Long get() {
                 boolean guaranteedIteration = false;
                 long now;
                 while ((now = System.nanoTime()) < beginNanos + iterationDurationNanos
-                        || now < firstSupplierCallNanos + ttlNanos
-                        || (guaranteedIteration = !guaranteedIteration)) {
+                        || now < firstSupplierCallNanos + ttlNanos || (guaranteedIteration = !guaranteedIteration)) {
```
stIncMale: Oh! How did we get from `now - firstSupplierCallNanos < ttlNanos` to `now < firstSupplierCallNanos + ttlNanos`?
It's not the same thing when we are talking about `System.nanoTime`. The results of this method are allowed to overflow, therefore we can only compare deltas, but we can't compare absolute values (see the javadoc of the method).

So `now < firstSupplierCallNanos + ttlNanos` is not the same as `now - firstSupplierCallNanos < ttlNanos. And we want the latter. The same applies to `(now = System.nanoTime()) < beginNanos + iterationDurationNanos`. Somehow I missed this previously.

dbwiddis: I changed that for readability.  Didn't think about the potential for overflow.

oshi-core/src/test/java/oshi/util/MemoizerTest.java
stIncMale: Notice also how `expirationNanos ` is compared with `now ` in `Memoizer.memoize`:
```
long nanos = expirationNanos;
now - nanos >= 0
```
A delta is compared with 0 rather than just comparing `now >= nanos`. That's not just for fun :)

oshi-core/src/test/java/oshi/util/MemoizerTest.java
stIncMale: Notice also how expirationNanos is compared with now in Memoizer.memoize:
```
long nanos = expirationNanos;
now - nanos >= 0;
```
A delta is compared with 0 rather than just comparing `now >= nanos`. That's not just for fun :)

oshi-core/src/test/java/oshi/util/MemoizerTest.java
```diff
@@ -114,8 +114,8 @@ public Long get() {
                 // least once regardless of whether we have exceeded time delays
                 boolean guaranteedIteration = false;
                 long now;
-                while ((now = System.nanoTime()) < beginNanos + iterationDurationNanos
-                        || now < firstSupplierCallNanos + ttlNanos || (guaranteedIteration = !guaranteedIteration)) {
+                while ((now = System.nanoTime()) - beginNanos < iterationDurationNanos
```
stIncMale: 👍

oshi-core/src/test/java/oshi/util/MemoizerTest.java
dbwiddis: Well, the only concievable case where A - B < C gives a different result than A < B + C is if "B+C" overflows.  But given that we are assured that A will exceed the value of B+C during this program execution (which would invalidate the interval method) I maintain the original calculation is still valid.   Unless expirationNanos is on the order of hundreds of years and the test runs forever. :)

stIncMale: > the only conceivable case where A - B > C gives a different result than A > B + C is if "B+C" overflows

No, consider the following:
```java
long a = Long.MIN_VALUE;
long b = (Long.MAX_VALUE - 1);
long c = 1;
System.out.println("(a - b > c) is " + (a - b > c));//true
System.out.println("(a > b + c) is " + (a > b + c));//false
```
`b + c` does not overflow, and yet the results are different. The thing is that `Long.MIN_VALUE < Long.MAX_VALUE`. But if you received these values from `System.nanoTime` in the order `Long.MAX_VALUE`, `Long.MIN_VALUE` it means that you did two invocations of this method with an interval of about 1ns between them, i.e. `Long.MIN_VALUE`ns is greater than `Long.MAX_VALUE`ns when we are talking about `System.nanoTime`, and `Long.MIN_VALUE - Long.MAX_VALUE > 0` is `true`.

stIncMale: And when you start a JVM, `System.nanoTime` does not return values starting from 0: _The value returned represents nanoseconds since some fixed but arbitrary origin time (perhaps in the future, so values may be negative)_; it can easily start from `Long.MAX_VALUE` or from any other valid long.

dbwiddis: OK... :)

PR #977, opened 2019-09-01
Rename default configuration

### LINKED ISSUE #974: default.properties resource file would conflict with other libraries
In 4.0.0, a new resource file "default.properties" was introduced.
The problem is that the same file would exist in other libs, so we do not know which one will be loaded when we try to get "default.properties" resource.
For example, if an application contains both Felix and Oshi libraries, it crashes on start sometimes (because a random "default.properties" resource file is loaded).
I reckon there could be other libraries with "default.properties" file inside. Renaming this file (for example, with "oshi.properties") will prevent possible conflicts with any other libraries.

cilki:


dbwiddis:
Can you add an appropriate comment in the changelog as well, possibly in the "API Change" section as well as this PR?

cilki:
Yes.. I always forget about the changelog.

oshi-core/src/main/java/oshi/util/GlobalConfig.java
```diff
@@ -41,9 +44,21 @@
     private static final Properties configuration = new Properties();
 
     static {
-        try (InputStream in = GlobalConfig.class.getResourceAsStream("/default.properties")) {
-            if (in != null) {
-                configuration.load(in);
+        // Load the configuration file from the classpath
+        try {
+            List<URL> resources = Collections.list(GlobalConfig.class.getClassLoader().getResources("oshi.properties"));
```
dbwiddis: Per the [`Class#getClassLoader`](https://docs.oracle.com/javase/6/docs/api/java/lang/Class.html#getClassLoader())  javadoc:
> Some implementations may use null to represent the bootstrap class loader. This method will return null in such implementations if this class was loaded by the bootstrap class loader.

A brief search suggests we prefer `Thread.currentThread().getContextClassLoader()` but I'm not sure if that allows a relative path?

cilki: Good catch. I think it will still work with the context classloader, but I'll check.

PR #981, opened 2019-09-03
Get services

### LINKED ISSUE #257: Get all services running on system
How to get all underlying services running on machine.
For example,
Windows Services.

agithyogendra:
Get Services Implementation.  Fixes #257


oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
```diff
@@ -89,6 +89,12 @@ public String getManufacturer() {
         return this.manufacturer;
     }
 
+    /** {@inheritDoc} */
+    @Override
+    public OSService[] getServices() {
```
dbwiddis: Missing the import statement for `OSService` so this is failing compilation.  Same thing for all the platform implementations of this.

oshi-core/src/main/java/oshi/software/os/OSService.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * <p>
+     * Getter for the field <code>pathName</code>.
+     * </p>
+     *
+     * @return Returns the pathName of the service.
+     */
+    public State getPathName() {
```
dbwiddis: Return type should be `String`?

oshi-core/src/main/java/oshi/software/os/OSService.java
```diff
@@ -0,0 +1,126 @@
...
+        this.processID = processID;
+    }
+
+    /**
+     * Set the status of the service.
+     *
+     * @param status
+     *            status
+     */
+    public void setState(Status state) {
```
dbwiddis: Argument type should be `State`

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -712,4 +715,26 @@ public static long getHz() {
         return USER_HZ;
     }
 
+    /** {@inheritDoc} */
+    @Override
+    public OSService[] getServices() {
+        //Sort by PID
+        OSProcess[] process =  getChildProcesses(1, 0, OperatingSystem.ProcessSort.PID);
+        File etc = new File("/etc/init");
+        File[] files = etc.listFiles();
```
dbwiddis: Need a null check on `etc` here (I know `/etc/init` should always exist, but that shouldn't crash OSHI if it doesn't.)  Same issue on the FreeBSD and Solaris implementations.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -712,4 +715,26 @@ public static long getHz() {
...
+        OSService[] svcArray = new OSService[files.size()];
```
dbwiddis: use `length` not `size()` for an array.  Same issue on FreeBSD and Solaris.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -712,4 +715,26 @@ public static long getHz() {
...
+        for (int i = 0; i < files.length; i++) {
+            svcArray[i].setName(files[i].getName());  
+            for (int j = 0; j < process.length; j++) {
+                if(process[j].getName().equals(files[i].getName())) {
+                    svcArray[i].setProcessId(process[j].getProcessID());
```
dbwiddis: setProcessID not Id ... same issue on other OS's.

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -393,4 +393,92 @@ public NetworkParams getNetworkParams() {
...
+                        // pbi_comm contains first 16 characters of name
+                        // null terminated
+                        for (int t = 0; t < taskAllInfo.pbsd.pbi_comm.length; t++) {
+                            if (taskAllInfo.pbsd.pbi_comm[t] == 0) {
+                                name = new String(taskAllInfo.pbsd.pbi_comm, 0, t);
+                                break;
+                            }
+                        }
+                    }
+                    svcArray.setPathName(path);
```
dbwiddis: Need a `[i]` after `svcArray`

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
```diff
@@ -770,4 +770,34 @@ private static void enableDebugPrivilege() {
         }
         Kernel32.INSTANCE.CloseHandle(hToken.getValue());
     }
+
+    /** {@inheritDoc} */
+    @Override
+    public OSService[] getServices() {
+        W32ServiceManager sm = new W32ServiceManager();
+        sm.open(Winsvc.SC_MANAGER_ALL_ACCESS);
```
dbwiddis: This fails without running as administrator.  Use `SC_MANAGER_ENUMERATE_SERVICE `.  I think we might want a try with resources block anyway, to catch `com.sun.jna.platform.win32.Win32Exception` here.

agithyogendra:
I have a Linux VM

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -393,4 +396,92 @@ public NetworkParams getNetworkParams() {
         return new MacNetworkParams();
     }
 
+    /** {@inheritDoc} */
+    @Override
+    public OSService[] getServices() {
+        //Sort by PID
+        OSProcess[] process =  getChildProcesses(1, 0, OperatingSystem.ProcessSort.PID);
+        //Get Directories
+        File ua = new File("~/Library/LaunchAgents");
```
dbwiddis: The `~` for this path is a relative path, that doesn't work from inside a Java application.  You need to replace the tilde with `System.getProperty("user.home")`

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -393,4 +396,92 @@ public NetworkParams getNetworkParams() {
...
+            for(File file : uaFiles) files.add(file);
+            for(File file : gaFiles) files.add(file);
+            for(File file : gdFiles) files.add(file);
+            for(File file : saFiles) files.add(file);
+            for(File file : sdFiles) files.add(file);
+
+            OSService[] svcArray = new OSService[files.size()];
+            for (int i = 0; i < svcArray.length; i++) {
+                for (int j = 0; j < process.length; j++) {
+                    svcArray[i].setName(files.get(i).getName());
```
dbwiddis: You haven't initialized `svcArray[i]` so this throws a NPE here.  Also I suspect it should be outside the iteration over j.

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -393,4 +396,92 @@ public NetworkParams getNetworkParams() {
...
+                    svcArray[i].setName(files.get(i).getName()); 
+                    svcArray[i].setState(OSService.State.STOPPED);
+                    ProcTaskAllInfo taskAllInfo = new ProcTaskAllInfo();
+                    if (0 > SystemB.INSTANCE.proc_pidinfo(process[j].getProcessID(), SystemB.PROC_PIDTASKALLINFO, 0, taskAllInfo, taskAllInfo.size())) {
+                        continue;
+                    }
+                    String name = null;
+                    String path = "";
+                    Pointer buf = new Memory(SystemB.PROC_PIDPATHINFO_MAXSIZE);
+                    if (0 < SystemB.INSTANCE.proc_pidpath(process[j].getProcessID(), buf, SystemB.PROC_PIDPATHINFO_MAXSIZE)) {
```
dbwiddis: I'm getting the identical path for every service, so something's wrong with this logic.

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -393,4 +396,92 @@ public NetworkParams getNetworkParams() {
...
+            for(File file : saFiles) files.add(file);
+            for(File file : sdFiles) files.add(file);
+
+            OSService[] svcArray = new OSService[files.size()];
+            for (int i = 0; i < svcArray.length; i++) {
+                for (int j = 0; j < process.length; j++) {
+                    svcArray[i].setName(files.get(i).getName()); 
+                    svcArray[i].setState(OSService.State.STOPPED);
+                    ProcTaskAllInfo taskAllInfo = new ProcTaskAllInfo();
+                    if (0 > SystemB.INSTANCE.proc_pidinfo(process[j].getProcessID(), SystemB.PROC_PIDTASKALLINFO, 0, taskAllInfo, taskAllInfo.size())) {
```
dbwiddis: It also takes a really long time to query services, which is bad.  I think the problem is here, your outer loop has 656 services in it, and you're then looping over this inner loop for every single process.  This ends up O(n^2) with multiple system calls.   Just call the existing `getProcesses()` method once and cache the results in a map with pid as the key and path as the value.... then look up that path here.   If the key exists it's running, otherwise it's stopped.

oshi-core/src/main/java/oshi/software/os/OSService.java
```diff
@@ -0,0 +1,126 @@
...
+ * SOFTWARE.
+ */
+
+package oshi.software.os;
+
+import java.io.Serializable;
+
+public class OSService implements Serializable {
+
+	private String name = "";
```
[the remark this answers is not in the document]
agithyogendra: I'm using a text editor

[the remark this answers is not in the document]
dbwiddis: You might consider downloading Eclipse (my preference) or IntelliJ.  It automates a lot of the things like formatting and the correct includes and lets you run the tests and identify issues.  Big time-saver!   I'll get to the rest of the review later today.

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -393,4 +401,77 @@ public NetworkParams getNetworkParams() {
...
+                pidPath[0] = String.valueOf(process[i].getProcessID());
+                pidPath[1] = process[i].getPath();
+                processMap.put(name, pidPath);
+            }
+
+            OSService[] svcArray = new OSService[files.size()];
+            for (int i = 0; i < svcArray.length; i++) {
+                svcArray[i].setName(files.get(i).getName()); 
+                svcArray[i].setState(OSService.State.STOPPED);
+                if(processMap.containsKey(svcArray[i].getName())) {
```
dbwiddis: OK, so this isn't always working. In fact it's rarely working.  There are a lot of services that only have the name in the process list, e.g.: `suggestd` even though we have `com.apple.suggestd`.  Most of these are in the `/System/Library` directory but doing a simple search for `plist` files doesn't work: this directory only has subdirectories... and the plist files are buried down a chain of directory structures.  I will need to take a look at how to do this better.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -712,4 +717,32 @@ public static long getHz() {
...
+    @Override
+    public OSService[] getServices() {
+        //Sort by PID
+        OSProcess[] process =  getChildProcesses(1, 0, OperatingSystem.ProcessSort.PID);
+        File etc = new File("/etc/init");
+        try {
+            File[] files = etc.listFiles();
+            OSService[] svcArray = new OSService[files.length];
+            for (int i = 0; i < files.length; i++) {
+                svcArray[i].setName(files[i].getName());
```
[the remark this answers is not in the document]
agithyogendra: I'm a little confused by this. Do you mean I should add, `svcArray[i] = new OSService();`

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
dbwiddis: Ah, I remember the suggestion I originally made -- any process with a ParentPid of 1 should count as a service.  You're only using the services from the file paths to fill out the list of any "stopped" services we missed.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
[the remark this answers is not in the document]
dbwiddis: Yes, sorry I missed this question. :)

Issue #757, opened 2019-01-11
API should return Lists, not Arrays


dbwiddis:
For 4.x, change all API methods that return `Foo[]` to return `List<Foo>` instead.

dbwiddis:
Generally return values should be `Collections.unmodifiableList(foo)`

PR #1005, opened 2019-10-02
Physical memory additional information implementation

### LINKED ISSUE #419: Memory speed
Hello, I noticed that I can only retrieve the used/total memory in system.
Is it possible to retrieve the memory clock speed?
Thank you

rohitkukreja1508:
Fixes #419

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
+++ b/oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
@@ -23,7 +23,6 @@
  */
 package oshi.hardware;
 
-
 /**
  * The GlobalMemory class tracks information about the use of a computer's
  * physical memory (RAM) as well as any available virtual memory.
@@ -56,7 +55,7 @@ public interface GlobalMemory {
      * @return A VirtualMemory object.
      */
     VirtualMemory getVirtualMemory();
-    
+
     /**
      * Physical memory, such as banks of memory.
      *
--- a/oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
+++ b/oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
@@ -58,6 +58,9 @@ public interface GlobalMemory {
 
     /**
      * Physical memory, such as banks of memory.
+     * <p>
+     * On Linux, requires elevated permissions. On FreeBSD and Solaris, requires
+     * installation of dmidecode.
      *
      * @return A list of PhysicalMemory objects.
      */
--- a/oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
+++ b/oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
@@ -23,6 +23,8 @@
  */
 package oshi.hardware;
 
+import java.util.List;
+
 /**
  * The GlobalMemory class tracks information about the use of a computer's
  * physical memory (RAM) as well as any available virtual memory.
@@ -55,4 +57,11 @@ public interface GlobalMemory {
      * @return A VirtualMemory object.
      */
     VirtualMemory getVirtualMemory();
+    
+    /**
+     * Physical memory, such as banks of memory.
+     *
+     * @return A list of PhysicalMemory objects.
+     */
+    List<PhysicalMemory> getPhysicalMemory();
 }
--- a/oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
+++ b/oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
@@ -23,7 +23,6 @@
  */
 package oshi.hardware;
 
-import java.util.List;
 
 /**
  * The GlobalMemory class tracks information about the use of a computer's
@@ -63,5 +62,5 @@ public interface GlobalMemory {
      *
      * @return A list of PhysicalMemory objects.
      */
-    List<PhysicalMemory> getPhysicalMemory();
+    PhysicalMemory[] getPhysicalMemory();
 }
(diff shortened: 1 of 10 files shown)


oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
```diff
@@ -1,58 +1,67 @@
...
-    long getPageSize();
-
-    /**
-     * Virtual memory, such as a swap file.
-     *
-     * @return A VirtualMemory object.
-     */
-    VirtualMemory getVirtualMemory();
-}
+/**
```
dbwiddis: Check your CR/LF settings. Unchanged lines are showing as changed.  The repository uses "Unix" line endings (`\n`) but I think you've committed with Windows line endings (`\r\n`).  This should be a one-time settings change in either your IDE or git client.

oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
```diff
@@ -1,58 +1,67 @@
...
+     * @return A VirtualMemory object.
+     */
+    VirtualMemory getVirtualMemory();
+    
+    /**
+     * Physical memory, such as banks of memory.
+     *
+     * @return A list of PhysicalMemory objects.
+     */
+    List<PhysicalMemory> getPhysicalMemory();
```
dbwiddis: See #757 where I initially thought to change all of OSHI's "collection" type API to Lists, and ended up not doing it due to serialization issues.  That said, I've since decided to get rid of serialization so that's not a barrier.  Do Lists work with Jackson's ObjectMapper?

Even then, for consistency, I think I'd prefer to return a fixed array here, and consider a List in the next major release.

oshi-core/src/main/java/oshi/hardware/PhysicalMemory.java
```diff
@@ -0,0 +1,106 @@
...
+
+    public String getMemoryType() {
+        return memoryType;
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append(" Bank label: " + getBankLabel());
+        sb.append("\n Capacity: " + FormatUtil.formatBytes(getCapacity()));
```
dbwiddis: Minor point, for output that will possibly have multiple entries, probably want to have this all on one line.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsGlobalMemory.java
```diff
@@ -1,102 +1,194 @@
...
+		int type = WmiUtil.getUint16(bankMap, PhysicalMemoryProperty.MEMORYTYPE, index);
+		String memoryType = convertMemoryTypeToString(type);
+		memory = new PhysicalMemory(bankLabel, capacity, speed, manufacturer, memoryType);
+		physicalMemoryArray.add(memory);
+    	   }
+	}
+   	return physicalMemoryArray;
+   }
+    
+    public static String convertMemoryTypeToString(int type) {
```
dbwiddis: this can probably be private

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsGlobalMemory.java
```diff
@@ -1,102 +1,194 @@
...
+	switch(type) {
+        	case 0:
```
dbwiddis: Interestingly, this seems to be "0" for most modern systems, and I'm wondering if this WMI field is even useful to retrieve any more.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsGlobalMemory.java
rohitkukreja1508: Done

oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
rohitkukreja1508: Changed from list to array and set size to number of banks

oshi-core/src/main/java/oshi/hardware/PhysicalMemory.java
rohitkukreja1508: Done

oshi-core/src/main/java/oshi/hardware/GlobalMemory.java
rohitkukreja1508: I have changed the setting.It will be reflected on the next commit

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsGlobalMemory.java
dbwiddis: OK, so I finally had the time to do some testing on my physical Win10 machine.  The `MemoryType` field has a limited enum set and is useless on modern hardware (my system returns 0).  I just committed a change that uses the `SMBIOSMemoryType` field instead, which is 26 (0x1A) on my system.  [This StackOverflow answer](https://stackoverflow.com/a/14227777/1161484) indicates this is DDR4.  I haven't managed to find the source document for this, but it's a better mapping to use here.

I made a few formatting tweaks with my commit as well.

dbwiddis: Also it appears that SMBIOS type 17 mapping is cross platform so you might want to toss this method into the abstract class (and put it back to `protected`, sorry!)

dbwiddis: [Source here](https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.2.0.pdf), page 98

dbwiddis:
I've done the mac and Unix implementations. Can you review/test my changes?

PR #1042, opened 2019-11-11
Fix broken "openHAB Systeminfo Binding" link

wborn:
The repository has been restructured so the link needs to be updated.


dbwiddis:
Thanks!

Issue #696, opened 2018-11-27
NetworkIF add support for additional fields


spyhunter99:
would be awesome if we could support the same fields that sigar supported

- [x] Rx dropped packets
- [ ] RX frames
- [ ] RX overruns
- [ ] TX Carrier
- [x] TX collisions
- [ ] TX overruns
- [ ] uptime (time since the nic was last administratively enabled or system boot time
- [ ] connected time, time elapsed since the nic was connected to a network or 0 if it is currently not connected
- [x] NIC subnet mask

dbwiddis:
Would be awesome if you could submit a PR for that!

dbwiddis:
Slightly more serious reply... good idea but before we get into the same trap we did with `OSProcess` data and using lots of CPU, this may be something that might wait until we have a better handle on "don't grab info until you need it" per #548

dbwiddis:
If someone's working on this, see #976 for another potential addition.

dbwiddis:
I took a look at doing this myself, and am not sure why the specific statistics shown were selected from a lot more alterantives (Linux) and the data is sparse on other OS's.   Some feedback:
 - Windows does not seem to provide any of this information.  The closest it gets is "Discards" which would probably be highly correlated to "Dropped Packets"
 - It does look like we could use our existing queries to add "rx_dropped" and "collisions" on most of the *nix-based OS's
 - The other fields (frames, overruns, carrier) appear to be subsets of the errors already reported.  There are many more error types on Linux, but no more detail on other OS's.  
 - I didn't see how to get uptime/connected time on any OS.  It's probably there somewhere, probably in a command line call.   Do you have links to the SIGAR source for those that I can look at to see if it's easily accessible?

Bottom line:  I'll probably work on adding dropped (using rx_discards for Windows) and collisions (using tx_discards for Windows) for the 4.4.0 release, and look more into how easy timing data is to get; but other than that, I don't think there's much value in adding only a select few of the many Linux error details, which are accessible via `getLongFromFile()` and similar methods in the util class, for those who want to automate them.

spyhunter99:
for windows, there is a wmi query for network connected time and i think there's another timestamp for last reset time

dbwiddis:
[...]

PR #1057, opened 2019-12-11
Subnet Mask & Prefix Length

Vesyrak:
First time PR.

Adds Subnet Mask and Prefix Length to the NetworkIF.
This was one of the problems addressed in #696 

If there is interest, I also have something which can read most of the other problems defined in the issue. This code is, however, Linux specific, as it uses the corresponding interface files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 4.3.0 (in progress)
 ================
-* Your contribution here.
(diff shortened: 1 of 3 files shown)


dbwiddis:
This looks great!  Thanks for your contribution.

I'd welcome your contribution for the Linux implementation of the other code, I can port the other OS's.

Issue #1061, opened 2019-12-16
TransferTime is wrong again


Space2Man:
Hi,

on Windows the transfertime seems to be wrong again. We fixed this in #653 but somehow on an completely idle disk (like CDROM) I see that transfertime == timestamp.

This is with OSHI 3.13.4.

Best regards, Space

Space2Man:
Issue seems to come from #763 ...

Space2Man:
I opened pull request #1063

Space2Man:
Ah, now I see why #763 broke it ... it applied the 1-x part, but it uses the diskTime, not the idleTime. I guess we should rename it for 3.13.x as well.

Although it might very well be that 4.2.x was working before that. I noticed that my test systems all are running against 3.13.4 at the moment. And there we have the 1-%diskTime which is of course wrong.

dbwiddis:
or remove the 1-x part? Heh.

Space2Man:
I think it's better to use 1-%idle, see http://support.demandtech.com/index.php?pg=kb.page&id=47
On laptops this probably does not matter (that much) but I think on pretty busy database servers the numbers get wrong otherwise. I don't expect a disk to be >100% busy ...

But 1-%disk is definetly wrong :)

dbwiddis:
I concur 1-%idle is probably the best.   Is the PR good to go? I haven't had time to look at it yet.

Space2Man:
Hi, I have not been able to test this as I switched my workstation to Linux and for the other systems there is currently a change freeze :-\

dbwiddis:
Fixed in #1063 and #1065

PR #1063, opened 2019-12-17
#1061 Fix transferTime in WindowsDisks by using 1-%Idle

Space2Man:
Hi,

first try at pull request. It compiles fine and uses 1 - %IdleTime to set the TransferTime.
Can you test this? I switched my system to Linux and don't have a test system at hand right now.

Thanks, Space

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsDisks.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsDisks.java
@@ -157,7 +157,7 @@ public class WindowsDisks implements Disks {
             diskStore.setWrites(stats.writeMap.getOrDefault(index, 0L));
             diskStore.setWriteBytes(stats.writeByteMap.getOrDefault(index, 0L));
             diskStore.setCurrentQueueLength(stats.queueLengthMap.getOrDefault(index, 0L));
-            diskStore.setTransferTime(stats.timeStamp-stats.idleTimeMap.getOrDefault(index, 0L));
+            diskStore.setTransferTime(stats.timeStamp-stats.idleTimeMap.getOrDefault(index, stats.timeStamp));
             diskStore.setTimeStamp(stats.timeStamp);
             return true;
         } else {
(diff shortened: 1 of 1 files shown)


dbwiddis:
It seems like it would work, except the default doesn't make sense (using 0 as default idle time means the disk is running 100%, it should be the opposite).  But in my testing I can't see why it's wrong the other way.  The idle and disk time always sum up to the same value, so the original method should work.

Is it WMI or performance counters giving you the odd readings in the old code?

Space2Man:
How can I find out which method is used? What logging do I need to enable?

dbwiddis:
Doesn't look like it's logged. You'd have to instrument the classes yourself. I'll try it tomorrow... manual WMI queries look sane, so it may be the performance counters don't read like I think they read

Space2Man:
I tried to do some further investigations but I failed. On the system where I see the transferTime == timeStamp in my app both Get-Counter and Get-CIMInstance return value of 0 for PercentDiskTime and 100 for PercentIdleTime in PowerShell ... how do you get the PERCENTDISKTIME? Is it maybe mapped incorrectly in JNA?

I am quite lost but I am not that much into Windows :)

Space2Man:
But I think the reason to use 1 - % Idle Time was something like this:

http://support.demandtech.com/index.php?pg=kb.page&id=47

dbwiddis:
The "fallback" for percent disk time comes from WMI in the [Win32_PerfRawData_PerfDisk_PhysicalDisk](https://docs.microsoft.com/en-us/previous-versions/aa394308(v%3Dvs.85)) class. You can try this on your commandline:
`wmic path Win32_PerfRawData_PerfDisk_PhysicalDisk get PercentDiskTime,PercentIdleTime `

You'll notice they add up to the same number consistently.  The primary choice of data, however, is from the performance counter `\\PhysicalDisk(0 C:)\% Disk Time` (or whichever drive letter is the instance) which you can look at the processed data with the `Perfmon.exe` program, and again see that idle and disk time add up to 100%.  However, you can't see the raw numbers that go into the counter... that's possibly what OSHI is returning to you, however.

Anyway, the 1-%Idle is probably a good thing to use and I'll accept this PR, but please look at your default as noted in my first comment.

dbwiddis:
I just tested the disk time vs. idle time on my hard drive, over 20 seconds, got:
```
9496222 94963206
9496263 94964174
...
9497756 94980229
9497787 94981199
```
The numbers sum up to elapsed milliseconds (e.g., each row sum is about 1000 more than the previous).   So this confirms subtracting idle time from elapsed time should give the correct result.

Space2Man:
Ok, I set default to stats.timeStamp. Then transferTime should be set to 0 if nothing is returned.

dbwiddis:
Looks good, merging.  If you want to submit the same PR against the 3.13.X branch I can update that one too.

Issue #273, opened 2016-11-22
Support getting Options of OSFileStore


CodingFabian:
In Sigar there exists FileSystem.getOptions() which returns stuff like "rw,local,rootfs,nfs" 
Would be great if OSFileStore has something similar.

dbwiddis:
OSHI already contains most of the information in this type of query, that is consistently cross-platform.

Some additional information can be retrieved using Java/Swing's FileSystemView, but we had problems with the swing framework and eventually got rid of it.   Given that we already display most/all of what's available in `mount` and similar commands, I'm not sure how we'd implement this cross platform.

CodingFabian:
Sigar gets them this way:
mac:
https://github.com/hyperic/sigar/blob/ad47dc3b494e9293d1f087aebb099bdba832de5e/src/os/darwin/darwin_sigar.c#L2079
win:
https://github.com/hyperic/sigar/blob/ad47dc3b494e9293d1f087aebb099bdba832de5e/go_bindings/Sigar/src/os/win32/win32_sigar.c#L1902
linux:
https://github.com/hyperic/sigar/blob/ad47dc3b494e9293d1f087aebb099bdba832de5e/src/os/linux/linux_sigar.c#L1168 (which is basically https://linux.die.net/man/3/getmntent_r # mnt_opts)
aix:
https://github.com/hyperic/sigar/blob/ad47dc3b494e9293d1f087aebb099bdba832de5e/src/os/aix/aix_sigar.c#L1171
solaris:
https://github.com/hyperic/sigar/blob/ad47dc3b494e9293d1f087aebb099bdba832de5e/src/os/solaris/solaris_sigar.c#L1398


so it seems that on *ix the string is taken from the os. on windows and mac the string is "emulated based on the type" I think thats a reasonable idea.

dbwiddis:
So, I'm wondering what the best approach is for all these sigar-isms.   Implement similarly performing (but different) functionality in oshi, or develop a new project that uses the same API as SIGAR to ease transition.

dblock:
I vote for not caring about Sigar at all.

CodingFabian:
@dblock thats a perfectly reasonable position.
I want to get away from sigar, and there is also a certain loss of capabilities I am willing to incur. But my feature requests are from my user perspective, which you certainly care about. I just used the links to the sigar implementation to document how its done in sigar.

[...]

PR #1098, opened 2020-02-09
Implement option to gather only local filestore stats.

Space2Man:
Initial version. Please check and verify ... did not even test it yet ...

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
@@ -40,7 +40,7 @@ import org.slf4j.LoggerFactory;
 import com.sun.jna.Native; // NOSONAR
 import com.sun.jna.platform.linux.LibC;
 
-import oshi.software.os.FileSystem;
+import oshi.software.common.AbstractFileSystem;
 import oshi.software.os.OSFileStore;
 import oshi.util.FileUtil;
 import oshi.util.ParseUtil;
@@ -51,7 +51,7 @@ import oshi.util.ParseUtil;
  * implementation specific means of file storage. In Linux, these are found in
  * the /proc/mount filesystem, excluding temporary and kernel mounts.
  */
-public class LinuxFileSystem implements FileSystem {
+public class LinuxFileSystem extends AbstractFileSystem {
 
     private static final Logger LOG = LoggerFactory.getLogger(LinuxFileSystem.class);
 
@@ -107,11 +107,6 @@ public class LinuxFileSystem implements FileSystem {
         return false;
     }
 
-    /**
-     * {@inheritDoc}
-     *
-     * Gets File System Information.
-     */
     @Override
     public OSFileStore[] getFileStores() {
         return getFileStores(false);
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
@@ -136,15 +136,12 @@ public class LinuxFileSystem extends AbstractFileSystem {
             // Exclude pseudo file systems
             String path = split[1].replaceAll("\\\\040", " ");
             String type = split[2];
-            if (PSEUDO_FS.contains(type) // exclude non-fs types
+            if ((localOnly && NETWORK_FS_TYPES.contains(type)) // Skip non-local drives if requested
+                    || PSEUDO_FS.contains(type) // exclude non-fs types
                     || path.equals("/dev") // exclude plain dev directory
                     || ParseUtil.filePathStartsWith(TMP_FS_PATHS, path) // well known prefixes
                     || path.endsWith("/shm") // exclude shared memory
             ) {
-                continue;
-            }
-            // Skip non-local drives if requested
-            if (localOnly && isNetworkFsType(type)) {
                 continue;
             }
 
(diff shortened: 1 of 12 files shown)


oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -163,6 +175,10 @@ private boolean listElementStartsWith(List<String> aList, String charSeq) {
                 continue;
             }
 
+            if (localOnly && (type.startsWith("nfs") || type.equals("cifs"))) {
```
dbwiddis: Probably should add `smbfs` and `sshfs` for completeness (here and other places these are used).  Actually found this list...

`afs, cifs, smbfs, sshfs, ncpfs, ncp, nfs, nfs4, gfs, gds2, glusterfs`

Plus `afp` on mac, although looks like we handle that separately.

Might want a protected static `isNetworkFsType(String type)` method in the abstract parent.

oshi-core/src/main/java/oshi/software/os/mac/MacFileSystem.java
```diff
@@ -107,6 +106,9 @@
                     if (volume.equals("devfs") || volume.startsWith("map ")) {
                         continue;
                     }
+                    if (localOnly && (volume.startsWith("localhost:") || volume.startsWith("//"))) {
```
dbwiddis: I'll do some testing later but I'm iffy about this filter.  We do have a `LOCAL_DISK` pattern matcher for the `/dev` path that might be appropriate here.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisFileSystem.java
```diff
@@ -105,7 +104,11 @@ private boolean listElementStartsWith(List<String> aList, String charSeq) {
         String key = null;
         String total = null;
         String free = null;
-        for (String line : ExecutingCommand.runNative("df -g")) {
+        String command = "df -g";
```
dbwiddis: These 4 lines might read more simply with a ternary operator.

oshi-core/src/main/java/oshi/software/os/mac/MacFileSystem.java
dbwiddis: Yep, this is incomplete.  There's at least `smb://192.168.foo.bar/disk` for example and a `foobar@oxfuse`.   And the `/dev` search might not screen out some non-local systems.

Some options are to parse `df -l` similar to the *nixes, parse `mount` looking for the 'local' flag, or use the `f_fstypename` field to grab the string for the same test as the other *nix systems.

I'll dig into the source code for `mount` or `df` later to see if I can figure out what it uses internally.

dbwiddis: Aha, we have an `f_flags` field already, which we can bitmask with `MNT_LOCAL` (`0x00001000`).  Which makes me realize that perhaps we should reopen #273.

dbwiddis: Here's the actual code implementing the `l` flag in `df -l`

```
if (lflag && (mntbuf[i].f_flags & MNT_LOCAL) == 0)
			continue;
```

oshi-core/src/main/java/oshi/software/os/windows/WindowsFileSystem.java
```diff
@@ -141,7 +136,9 @@ public WindowsFileSystem() {
                 volumeMap.get(wmiVolume.getMount()).setName(wmiVolume.getName());
             } else {
                 // Otherwise add the new volume in its entirety
-                result.add(wmiVolume);
+                if (!localOnly) {
```
dbwiddis: This can be moved up as an `else if`.

dbwiddis:
I went ahead and added the macOS code and more thorough fs type checking for the other *nix systems.  Still TODO is [filter the WMI query to only ask for the local drives](https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-logicaldisk).   We could query on the `DriveType`: (Unknown (0), No Root Directory (1), Removable Disk (2), Local Disk (3), Network Drive (4), Compact Disc (5), RAM Disk (6)) but should decide whether to include only 3, or just exclude 4 (my vote).  The `MediaType` field is another option (non-zero?) but I think isn't as good as `DriveType`.

dbwiddis:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisFileSystem.java  4
- oshi-core/src/main/java/oshi/software/common/AbstractFileSystem.java  3
- oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdFileSystem.java  2
- oshi-core/src/main/java/oshi/software/os/mac/MacFileSystem.java  3
- oshi-core/src/main/java/oshi/software/os/windows/WindowsFileSystem.java  3
- oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/widdis/oshi/pullRequest?prid=4976929&bid=16439462)

dbwiddis:
I think I'm done tweaking your PR.  I'll merge tonight unless you have anything else you'd like to comment on it.

Space2Man:
Man, you are fast! I hope to be able to test it on one of our systems in the next few days.

Thanks a lot!

Issue #1104, opened 2020-02-13
autofs filesystems not detected as pseudo or network FS


Space2Man:
Hi,

I just tested OSHI 4.4.0 and found out that NFS filesystems which are mounted from automounter via /etc/auto.direct are still detected as local filesystems. Also the info seems strange:

`13/Feb/2020 11:17:58,433- DEBUG FilesystemStats: Found FileStore: /ixos/devices/backup/ixos6_18, {"volume":"/etc/auto.direct","type":"autofs","mount":"/ixos/devices/backup/ixos6_18","freeSpace":62603329536,"usableSpace":62603329536,"freeInodes":31876197,"options":"rw,relatime,fd=12,pgrp=9690,timeout=600,minproto=5,maxproto=5,direct","UUID":"","totalSpace":1932735283200,"description":"Mount Point","logicalVolume":"","totalInodes":31876689,"name":"/etc/auto.direct"}
`

I just checked /proc mounts and there are two entries for that:

```
# cat /proc/mounts | grep /ixos/devices/backup/ixos6_18
/etc/auto.direct /ixos/devices/backup/ixos6_18 autofs rw,relatime,fd=12,pgrp=9690,timeout=600,minproto=5,maxproto=5,direct 0 0
xd00387a:/vol/ix20872_18/ixos6_18 /ixos/devices/backup/ixos6_18 nfs rw,sync,relatime,vers=3,rsize=32768,wsize=32768,namlen=255,acregmin=0,acregmax=0,acdirmin=0,acdirmax=0,hard,noac,proto=tcp,timeo=600,retrans=2,sec=sys,mountaddr=10.253.19.169,mountvers=3,mountport=4046,mountproto=tcp,local_lock=none,addr=10.253.19.169 0 0

```
The nfs filesystem is filtered correctly but I guess autofs needs to be added to the pseudo filesystems. Otherwise the autofs mounts get triggered when getFileStores() is called.

Space2Man:
What do you think? Is autofs a pseudo filesystem or a network filesystem? I think it's more a pseudo filesystem...

Space2Man:
Wouldn't all the filesystems which are "nodev" filesystems be either pseudo or network?

```
# cat /proc/filesystems 
nodev	sysfs
nodev	rootfs
nodev	bdev
nodev	proc
nodev	cgroup
nodev	cpuset
nodev	tmpfs
nodev	devtmpfs
nodev	debugfs
nodev	securityfs
nodev	sockfs
[...]

PR #1105, opened 2020-02-13
Added additional pseudo filesystems

Space2Man:


oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -58,32 +58,47 @@
 
     // Linux defines a set of virtual file systems
     private static final List<String> PSEUDO_FS = Arrays.asList(//
```
dbwiddis: Can you please move this to the AbstractFileSystem object (and make it `protected`)?  The SolarisFileSystem and FreeBSDFileSystem need something similar.

Space2Man: Ok, and you want those from Solaris and FreeBSD added into the same list?

dbwiddis:
Looks great!  Merging!  Fixes #1104

Space2Man:
Hi, any plans to do a release in the near future?

dbwiddis:
I was planning on one later this week after completing some WMI Refactoring (#1108) but can push one out earlier if needed.

Space2Man:
An early release would be great ... need to push the fix regarding the local mounts to systems soon. And I would prefer to use a real release so I can just include OSHI via maven ...

dbwiddis:
It is possible to directly link to snapshot releases via Maven, but I get it.  Pushing out 4.4.1 now, expect it on Maven within a few hours.

Space2Man:
Thanks a lot ... already building :)

PR #1170, opened 2020-04-15
Add final annotation to class

Bob160:
This pull request adds the `final` annotation to the class.


dbwiddis:
This looks perfect, thanks!   This is the "hello world" equivalent in GitHub, figuring out how to contribute... you've got the hard part out of the way.  Now go forth and contribute here and elsewhere!

PR #1184, opened 2020-04-29
minor correction for catching future fault

ehgus:
- set initial 'state' of OSProcess instance as 'null' to detect OSProcess instance that does not reflect real process information
- Change testing of 'os.getParentProcessID()' to see whether it points real process id
- Change testing of 'proc.getBitness()' to rigorously check its condition (32 or 64 bit)

### DIFF
--- a/oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
+++ b/oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
@@ -103,7 +103,7 @@ public class OperatingSystemTest {
         if (os.getBitness() == 32) {
             assertEquals("Process on 32-bit OS must have bitness 0 or 32", 0, proc.getBitness() & ~32);
         } else {
-            assertEquals("Process on 64-bit OS must have bitness 0, 32, or 64", 0, proc.getBitness() & ~(32 + 64));
+            assertTrue("Process on 64-bit OS must have bitness 0, 32, or 64", (0==(proc.getBitness() & ~(32 + 64)))&&(poc.getBitness!=(32+64)));
         }
         assertTrue(proc.getOpenFiles() >= -1);
     }
--- a/oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
+++ b/oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
@@ -80,7 +80,7 @@ public class OperatingSystemTest {
         assertNotNull(proc.getGroupID());
         assertNotNull(proc.getState());
         assertEquals(proc.getProcessID(), os.getProcessId());
-        assertTrue(proc.getParentProcessID() >= 0);
+        os.getProcess(proc.getParentProcessID());
         assertTrue(proc.getThreadCount() > 0);
         assertTrue(proc.getPriority() >= -20 && proc.getPriority() <= 128);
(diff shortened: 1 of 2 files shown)


ehgus:
Yes. Coverage should not be changed because I didn't add additional test suites.
What I do is to change codes to kill the parts of survived mutants. pitest deliberately insert/delete/modify codes to generate faulty mutants and it checks whether our test suites kill the cases. Mutation coverage is number of mutants killed. 
coverage_before
<original coverage>
coverage_after
<current coverage>
It shows that mutant coverage increases.

oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
```diff
@@ -80,7 +80,7 @@ public void testOperatingSystem() {
         assertNotNull(proc.getGroupID());
         assertNotNull(proc.getState());
         assertEquals(proc.getProcessID(), os.getProcessId());
-        assertTrue(proc.getParentProcessID() >= 0);
+        os.getProcess(proc.getParentProcessID());
```
dbwiddis, line 83: This test as you've written it introduces a race condition, as the parent may have terminated by the time you get to this test, and I'm actually surprised it didn't fail the CI tests. 

Given that this is a snapshot of the process state at the time you ran the query, you should be able to test the membership of the parent in the set of process IDs returned.

oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
```diff
@@ -100,8 +100,7 @@ public void testOperatingSystem() {
         assertTrue(proc.getStartTime() >= 0);
         assertTrue(proc.getBytesRead() >= 0);
         assertTrue(proc.getBytesWritten() >= 0);
-        assertTrue(proc.getBitness() >= 0);
-        assertTrue(proc.getBitness() <= 64);
+        assertTrue(proc.getBitness() == 64||proc.getBitness() == 32||proc.getBitness() == 0);
```
dbwiddis, line 103: I like this improvement.  You could also test <= against OS bitness (You can't have a 64-bit process on a 32-bit OS).

dbwiddis: This works:
```
        if (os.getBitness() == 32) {
            assertEquals("Process on 32-bit OS must have bitness 0 or 32", 0, proc.getBitness() & ~32);
        } else {
            assertEquals("Process on 64-bit OS must have bitness 0, 32, or 64", 0, proc.getBitness() & ~(32 + 64));
        }
```

ehgus:
Thank you for your feedback. Then I will change the above two parts back.

dbwiddis:
Great, thanks!  Good that you pointed them out and I'll consider them when I refactor the OSProcess object!

dbwiddis:
After merging I realized (based on your commit you retracted) that you'd identified the fact that bitmask testing 32 and 64 would in theory allow 96 as a value.   When I rewrite the OSProc in a few days I'll update this to two tests:  1 the bitmask (which would allow 0, 32, 64, and 96 only) and a second one that's simply proc bitness <= os bitness (which must be 32 or 64).  That should remove the need for the conditional.

Thanks again for pointing out how these tests can be improved!

PR #1203, opened 2020-05-08
Calculate size using string

Manas-Goel:


oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxGlobalMemory.java
```diff
@@ -104,23 +104,23 @@ private static long queryPageSize() {
             if (memorySplit.length > 1) {
                 switch (memorySplit[0]) {
                 case "MemTotal:":
-                    memTotal = parseMeminfo(memorySplit);
+                    memTotal = ParseUtil.parseDecimalMemorySizeToBinary(checkLine.substring(memorySplit[0].length()).trim());
```
dbwiddis, line 107: Good start, but this doesn't quite work.  The full `memorySplit` string includes the header text ("MemTotal" in this case), while the `ParseUtil` method is only expecting the part after the split.

The original string looks like "MemTotal:        1882064 kB".  Currently the split on line 103 turns this into an array `{ "MemTotal:", "1882064", "kB" }` and you want to send the string `"1882064 kB"` to `ParseUtil`.  If you add a limit parameter of 2 to the split on Line 103 (`ParseUtil.whitespaces.split(checkline, 2)`) that will split the string into only two parts... `split[0]` will be `"MemTotal:"` and `split[1]` will be `"1882064 kB"` and what you want to send.

dbwiddis: Oh, wait.... sorry I didn't look close enough, you did strip off that first part and trim the rest, so that does work exactly as you have it.  I still think I'd prefer using the limit on the split, though!

dbwiddis:
Thanks for contributing to open source!  I hope you stick around!

PR #1226, opened 2020-05-16
Enhancing unit test case message, first time contribution for open so…

dPramod:
Updated the unit test case messages. This is my first pull request on github.

### DIFF
--- a/oshi-core/src/test/java/oshi/util/TupleTest.java
+++ b/oshi-core/src/test/java/oshi/util/TupleTest.java
@@ -63,6 +63,6 @@ public class TupleTest {
         assertEquals("quartet.getD().charValue() should be c",'c', quartet.getD().charValue());
         assertEquals("quintet.getD().charValue() should be d",'d', quintet.getD().charValue());
 
-        assertEquals("quintet.getE() should be "+BigInteger.ZERO,BigInteger.ZERO, quintet.getE());
+        assertEquals("quintet.getE() should be BigInteger.ZERO",BigInteger.ZERO, quintet.getE());
     }
 }
--- a/oshi-core/src/test/java/oshi/util/TupleTest.java
+++ b/oshi-core/src/test/java/oshi/util/TupleTest.java
@@ -47,22 +47,23 @@ public class TupleTest {
         Quintet<String, Integer, Long, Character, BigInteger> quintet = new Quintet<>("D", 4, Long.valueOf("0"), 'd',
                 BigInteger.ZERO);
 
-        assertEquals("pair.getA() should be A","A", pair.getA());
-        assertEquals("triplet.getA() should be B" ,"B", triplet.getA());
-        assertEquals("quartet.getA() should be C","C", quartet.getA());
-        assertEquals("quintet.getA() should be D","D", quintet.getA());
+        assertEquals("pair.getA() should be A", "A", pair.getA());
+        assertEquals("triplet.getA() should be B", "B", triplet.getA());
+        assertEquals("quartet.getA() should be C", "C", quartet.getA());
+        assertEquals("quintet.getA() should be D", "D", quintet.getA());
 
-        assertEquals("pair.getB().intValue() should be 1",1, pair.getB().intValue());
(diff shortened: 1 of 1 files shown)


dPramod:
The only update was made on unit test file. Is there a rule set up to increase the coverage for every other build? Is there a test case I can add to fix the issue? Any help will be highly appreciated.

oshi-core/src/test/java/oshi/util/TupleTest.java
```diff
@@ -47,23 +47,22 @@ public void testTuples() {
         Quintet<String, Integer, Long, Character, BigInteger> quintet = new Quintet<>("D", 4, Long.valueOf("0"), 'd',
                 BigInteger.ZERO);
 
-        assertEquals("A", pair.getA());
-        assertEquals("B", triplet.getA());
-        assertEquals("C", quartet.getA());
-        assertEquals("D", quintet.getA());
+        assertEquals(pair.getA()+" should be equal to A","A", pair.getA());
```
dbwiddis, line 50: JUnit actually does include the "expected" and "actual" values for an equality test so your output on a failure would be redundant with that, and not help.  On failure, I might get the message "B should be equal to A" and it wouldn't let me find the line.  Better to have the string "Pair.getA() should be A".

oshi-core/src/test/java/oshi/util/TupleTest.java
```diff
@@ -47,23 +47,22 @@ public void testTuples() {
...
 
-        assertEquals(1, pair.getB().intValue());
-        assertEquals(2, triplet.getB().intValue());
-        assertEquals(3, quartet.getB().intValue());
-        assertEquals(4, quintet.getB().intValue());
+        assertEquals(pair.getB().intValue()+" should be equal to 1",1, pair.getB().intValue());
+        assertEquals(triplet.getB().intValue()+" should be equal to 2",2, triplet.getB().intValue());
+        assertEquals(quartet.getB().intValue()+" should be equal to 3",3, quartet.getB().intValue());
+        assertEquals(quintet.getB().intValue()+" should be equal to 4",4, quintet.getB().intValue());
+        assertEquals(triplet.getC().longValue()+" should be equal to "+Long.MAX_VALUE,Long.MAX_VALUE, triplet.getC().longValue());
```
dbwiddis, line 59: Rather than outputting the large number that max value actually is, just say "Long.MAX_VALUE" in plain words in the string.

oshi-core/src/test/java/oshi/util/TupleTest.java
dPramod: Thank you for the feedback, I will work on the review.

oshi-core/src/test/java/oshi/util/TupleTest.java
dPramod: sure, I will work on the given comments.

dPramod:
I have updated the test as per review comments. Thank you !

oshi-core/src/test/java/oshi/util/TupleTest.java
```diff
@@ -47,23 +47,22 @@ public void testTuples() {
         Quintet<String, Integer, Long, Character, BigInteger> quintet = new Quintet<>("D", 4, Long.valueOf("0"), 'd',
                 BigInteger.ZERO);
 
-        assertEquals("A", pair.getA());
-        assertEquals("B", triplet.getA());
-        assertEquals("C", quartet.getA());
-        assertEquals("D", quintet.getA());
+        assertEquals(pair.getA()+" should be A","A", pair.getA());
```
dbwiddis, line 50: This should be `assertEquals("pair.getA() should be A", "A", pair.getA());` and similar for the others.

dbwiddis:
Made another comment with an example of what it should be!

oshi-core/src/test/java/oshi/util/TupleTest.java
dPramod: Sorry for my wrong understanding. I have updated as suggested .

dbwiddis:
This looks great!  I'll do a formatting commit to standardize whitespace and then merge!

PR #1230, opened 2020-05-16
Change linux swapin/swapout counters

Roeezz:
-Change Linux virtual memory to use pswpin and pswap out instead of pgpgin and pgpgout.
-Change Solaris virtual memory to use pgswpin and pgswap out instead of pgpgin and pgpgout.

Cherry-picked from roeezz-linux-swp


dbwiddis:
Perfect, thanks!

Issue #1201, opened 2020-05-07
OSHI 10th anniversary virtual happy hour 🍾🙌🎉


dbwiddis:
[OSHI was started](https://code.dblock.org/2010/06/23/introducing-oshi-operating-system-and-hardware-information-java.html) on June 23, 2010, and we're coming up on the project's 10 year anniversary! 

In celebration of this milestone, I'd like to host an online happy hour were we can all bring our favorite beverage 🍷 🍺 🥃 (or the official "OSHI Martini" 🍸 in the comment below) and toast the project's success!  

I'm setting the date for this meetup at Tuesday, 23 June 2020 at 1pm PDT / 4pm EDT / 2000 GMT /  9pm BST / 10pm CEST.  [Here's a countdown to the event](https://www.timeanddate.com/countdown/generic?iso=20200623T13&p0=234&msg=OSHI+10th+Anniversary+online+meetup&font=cursive).

Here's the meeting info:
[Join Zoom Meeting](https://us04web.zoom.us/j/76760427867?pwd=cE9PNUhSMmNwR04rTjdlTDZWMVRBQT09)
Meeting ID: 767 6042 7867
Password: 9Yb1k0

[Add to Google Calendar](https://us04web.zoom.us/meeting/upIqcemsrDwiHNNaev16THMd1z-ZeHPvTZ69/calendar/google/add)
[Add to Outlook Calendar](https://us04web.zoom.us/meeting/upIqcemsrDwiHNNaev16THMd1z-ZeHPvTZ69/ics)

dbwiddis:
As promised, a bartender has concocted a custom celebratory cocktail for us!  Introducing...

### The OSHI Martini

#### Ingredients:
 - 3 oz. Sobieski Vodka or espresso flavored vodka
 - 3/4 oz. Drambuie Liquor
 - 3/4 oz. white grape juice
 - Orange peel

#### Directions:
In a martini shaker or tall glass filled with ice, combine the vodka, Drambuie, and white grape juice. Shake well and pour into a martini glass. Using a match or lighter, carefully and lightly burn and smoke the orange peel to create a toasted orange essence for garnish. Twist the peel over the cocktail and garnish drink.

#### Tasting notes:
[...]

PR #1236, opened 2020-05-21
Add a Sponsor button.

dblock:
I believe this will add a proper sponsor button to this repo. Please double check the package name for me. Inspired by [ruby-grape](https://github.com/ruby-grape/grape/blob/master/.github/FUNDING.yml).

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+tidelift: "pkg/maven-com-github-oshi-oshi-core"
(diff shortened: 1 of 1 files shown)


dbwiddis:
Thanks for this! Button's active with the correct link after a few trial and error and accidentally editing the wrong branch and other silly errors.  I actually have an application pending for a contributor sponsor button that I'll add once it's approved. Happy to add you as well if you want!

dbwiddis:
By the way, did you see the official OSHI Martini recipe in #1201 ? :)

dblock:
Thanks @dbwiddis!

PR #1241, opened 2020-05-23
Implement code-assert, fix javadocs

### LINKED ISSUE #1239: Make checking javadocs easier to find errors
Currently, the only way we notice javadoc problems (misnamed `@param` tags or failed `@link`s) is when the appveyor build fails.  Usually it is at least easier to find the problem, but yesterday it killed the build too fast to log the problems and I spent probably 3 hours hunting down two misnamed parameters.

Earlier I found [code-assert](https://github.com/nidi3/code-assert), that covers this exact case (and many more!), and it's probably worth setting up, even if just for this single case.

dbwiddis:
Fixes #1239


dbwiddis:
@hazendaz you may be interested in doing something similar for your other projects.

dbwiddis:
For the record, I'm really glad I added this. I was able to pre-test a recent PR and fix `@param` issues, and add private constructors and make classes final long before the chain of events which would lead to failed appveyor builds or CI tools informing me of defects. :)

Issue #1027, opened 2019-10-28
Can OSHI show process threads similar to Windows Task Manager?


dawzhang:
such as use chrome,  windows task manager show 10 chrome process, but oshi getProcesses() only show 1 .

dbwiddis:
There's actually only one Chrome process.  The rest of the Chrome "processes" you see on Task Manager are actually threads.

You may notice on Task manager that the statistics for the parent process CPU is the sum of the CPU of the subordinate threads.

OSHI doesn't (yet) expose all the thread details, but it may be a useful future addition.

tausiflife:
Hi, Can i work on this?

dbwiddis:
Absolutely! What plan do you have for implementing it?

tausiflife:
Thanks. My plan is to implement this in a step by step manner for each OS starting with Linux. 
I think that i can get the thread details of a process for Linux systems from /proc/{processid}/task/ directory. I want to add a method in `OSProcess.java` named `getThreadDetails` to expose the details. I haven't thought much about displaying the results in OshiGUI yet. 

This is the first time i am trying to contribute a functionality to opensource. Please bear with me if i have not answered your question correctly. Also i would like to listen to your ideas about this.

dbwiddis:
OK, great, that sounds like a good direction to start.  I like to start by defining the top level API which you have with `getThreadDetails` as the method.  Now to identify what details are available!

It would be good to look at WIndows, at least, to try to identify the direction you're going to go as you design the API.  It looks like the CreateToolhelp32Snapshot along with the Thread32First and Thread32Next functions can iterate that list similar to how we do processes, you would just have to filter for process ID.  Good to see what stats that API exposes that match the same in Linux.

tausiflife:
sure. I will have a look at it as well.

dbwiddis:
Great, looking forward to it!  

One other implementation detail: please try to put as much of the actual implementation code in the `oshi.driver` package structure.  The main implementation methods will be short, calling out to these other packages.

I haven't gone through all existing code to move things there yet (it's more of a move-while-fixing) I do want to try to put new code there to start.   This will enable more modularity in the future (e.g., JDK9+ has some methods such as iterating `ProcessHandles` that may be more efficient at some of the things OSHI does (directory list) and we could substitute one of those drivers in a multi-release JAR).

tausiflife:
[...]

PR #1258, opened 2020-06-08
Add threads details of a process #1027

tausiflife:


oshi-core/src/main/java/oshi/driver/linux/proc/LinuxOSThread.java
```diff
@@ -0,0 +1,140 @@
...
+    private String exePath;
+    private String commandLine;
+    private String userID;
+    private String user;
+    private String groupID;
+    private String group;
+
+
+    public LinuxOSThread(int pID, int tID) {
+        this.processID = pID;
```
dbwiddis, line 57: Personal preference, for short names like this should be all lowercase (`pid`, `tid`). Above I think I'd prefer `processId` to `processID` although I've done both elsewhere in the code...

oshi-core/src/main/java/oshi/software/common/AbstractOSProcess.java
```diff
@@ -72,6 +74,10 @@ public double getProcessCpuLoadBetweenTicks(OSProcess priorSnapshot) {
         return getProcessCpuLoadCumulative();
     }
 
+    @Override
+    public List<OSThread> getThreadDetails() {
+        return null;
```
dbwiddis, line 79: Better to return `Collections.emptyList()` here. I try to be very deliberate with my nulls!

oshi-core/src/main/java/oshi/driver/linux/proc/LinuxOSThread.java
tausiflife: Ok. I think it is fine if we use either format, but consistency matter. Hence i will use `processId` and similar variable names elsewhere.

oshi-core/src/main/java/oshi/software/common/AbstractOSProcess.java
tausiflife: Sure.

oshi-core/src/main/java/oshi/software/common/AbstractOSThread.java
```diff
@@ -0,0 +1,59 @@
...
+        this.parentProcess = process;
+    }
+
+    @Override
+    public OSProcess getParentProcess() {
+        return this.parentProcess;
+    }
+
+    @Override
+    public int getProcessID() {
```
dbwiddis, line 46: Hmm... this is a key design question.  Should a thread return its parent process ID or have its own numeric thread ID?  Should we have a separate `getThreadID()` or just return the thread id in the `processID` field?  Interested in your thoughts here, something doesn't sit right with returning the parent ID here.

oshi-core/src/main/java/oshi/software/common/AbstractOSThread.java
```diff
@@ -0,0 +1,59 @@
...
+        return getParentProcess().getProcessID();
+    }
+
+    @Override
+    public int getParentProcessID() {
+        return getParentProcess().getParentProcessID();
```
dbwiddis, line 52: Continuation of above discussion. It seems odd to worry about a parent of a process when we are in a thread of that process. It seems to me the "parent" of the thread should be the process it is a thread of, so this should return the "owning" process ID, and the getProcessID should return the thread ID (or 0 if such a thing is undefined).  Again, not sure what the right answer is but we should make a design decision here.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -312,25 +322,11 @@ public boolean updateAttributes() {
...
-            this.state = State.STOPPED;
-            break;
-        default:
-            this.state = State.OTHER;
-            break;
+        this.state = ProcessStat.getState(status.getOrDefault("State", "U").charAt(0));
+        List<Integer> threadIds = ProcessStat.getThreadIds(getProcessID());
+        this.threadDetails = new ArrayList<OSThread>();
+        for (Integer threadId : threadIds) {
+            this.threadDetails.add(new LinuxOSThread(this, threadId));
```
dbwiddis, line 329: I don't think you need to assign the `threadIDs` temporary variable, just include it in the foreach clause.  Although I think these 3 lines would read more clearly as a stream.  Something like this (untested):

this.threadDetails = ProcessStat.getThreadIds(getProcessID()).stream().map(id -> new LinuxOSThread(this, id)).collect(Collectors.toList()));

oshi-core/src/main/java/oshi/util/platform/linux/ProcPath.java
```diff
@@ -62,6 +62,10 @@
     public static final String UPTIME = PROC + "/uptime";
     public static final String VERSION = PROC + "/version";
     public static final String VMSTAT = PROC + "/vmstat";
+    public static final String THREADS_PATH = PROC + "/%d/task";
```
dbwiddis, line 65: I'm thinking `TASK` would be better than `THREAD` here.   Also the list does happen to be alphebetized, although that doesn't matter too much :)

oshi-core/src/main/java/oshi/software/os/OSThread.java
```diff
@@ -0,0 +1,29 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package oshi.software.os;
+
+public interface OSThread extends OSProcess {
+    int getThreadId();
```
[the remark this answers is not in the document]
tausiflife: Sure i will add JavaDocs everywhere i have touched.

oshi-core/src/main/java/oshi/driver/linux/proc/ProcessStat.java
```diff
@@ -453,4 +456,52 @@ private ProcessStat() {
...
+        }
+        return threadIDs;
+    }
+
+    /***
+     * Returns Enum STATE for the state value obtained from status file of any process/thread.
+     * @param stateValue state value from the status file
+     * @return OSProcess.State
+     */
+    public static OSProcess.State getState(char stateValue) {
```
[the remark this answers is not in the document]
tausiflife: Thanks. The thought behind this refactoring was just to reuse.

oshi-core/src/main/java/oshi/software/common/AbstractOSProcess.java
```diff
@@ -72,6 +75,10 @@ public double getProcessCpuLoadBetweenTicks(OSProcess priorSnapshot) {
         return getProcessCpuLoadCumulative();
     }
 
+    @Override
+    public List<OSThread> getThreadDetails() {
+        return Collections.emptyList();
+    }
     @Override
```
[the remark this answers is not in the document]
tausiflife: Sure, though i use IntelliJ i will find something and fix this.

oshi-core/src/main/java/oshi/util/platform/linux/ProcPath.java
tausiflife: Sorry, i didn't check the alphabetical order of the list. Will fix this.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
tausiflife: Thats a nice way to do :) Will do it.

oshi-core/src/main/java/oshi/software/common/AbstractOSProcess.java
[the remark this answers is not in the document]
dbwiddis: I think IntelliJ let’s you use an eclipse format file.  Alternately I think we are close to “google” defaults but with 4 space indent.

oshi-core/src/main/java/oshi/software/common/AbstractOSThread.java
tausiflife: I think i did a mistake naming the process of the thread to be `parentProcess`. It should be just `process`, and hence the method `getParentProcess` should be renamed to `getProcess` to avoid confusion. But you are right, this sounds more logical to return processId in `getParentProcessID()` and 0 in `getProcessID` . What do you think?

oshi-core/src/main/java/oshi/util/platform/linux/ProcPath.java
dbwiddis: It really doesn’t matter that much. :)

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -206,6 +211,11 @@ public long getBytesWritten() {
         return this.bytesWritten;
     }
 
+    @Override
+    public List<OSThread> getThreadDetails() {
+        return this.threadDetails;
```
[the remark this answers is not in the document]
tausiflife: ok.

oshi-core/src/main/java/oshi/driver/linux/proc/LinuxOSThread.java
```diff
@@ -0,0 +1,256 @@
...
+import oshi.util.ExecutingCommand;
+import oshi.util.FileUtil;
+import oshi.util.ParseUtil;
+import oshi.util.platform.linux.ProcPath;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class LinuxOSThread extends AbstractOSThread {
```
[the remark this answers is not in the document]
tausiflife: `TaskStat` sounds much better. Will rename both the class and the variables.

oshi-core/src/main/java/oshi/software/common/AbstractOSProcess.java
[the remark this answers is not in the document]
dbwiddis: https://imagej.net/Eclipse_code_style_profiles_and_IntelliJ

oshi-core/src/main/java/oshi/software/common/AbstractOSThread.java
dbwiddis: If you are inheriting you can’t really rename the method but you can redefine what it means.  

I think the “parent” of a thread is logically it’s process and the “processid” is the thread I’d, if it exists.

tausiflife: True, so i will add `getProcessID()` to return 0 as we already have a method `getThreadId` to return threadId. Also `getParentProcessID()` will now return processId, which i think makes sense after reading your comment.

dbwiddis: Or just have getProcessId return the thread id and don’t have a separate thread Id method.

tausiflife: ok but i still think that it should have a method to return threadId

dbwiddis: Are there any cases where the thread id is not a “lightweight” process id?  The problem with a separate “thread” (this object) and parent (owning process) is that the “process id” becomes ambiguous.  

If there are always only 2 numbers to ever return and we have sensibly named methods to return those 2 numbers, I don’t see added value in adding a new thread id method and also returning 0 for the process id.  I’m not strongly opposed to a new thread Id method that is just an alias to the process id, though.

dbwiddis: OK, after reading up on this for about an hour, I think I have this sorted.  Here's some background:
 - The Windows [THREADENTRY32](https://docs.microsoft.com/en-us/windows/win32/api/tlhelp32/ns-tlhelp32-threadentry32) includes both a thread ID and an "Owning Process ID".  So here it's not clear whether we want the pid or ppid as owner, but we are free to choose.
 - Linux thread info retains a reference to the "parent PID" so there's no reason not to return the same thing here, so let's keep parent process ID pointing to the parent process's parent, and not the "owning" process.  This means we need process ID as the owner.  
 - FreeBSD and Solaris have, in addition to a thread ID, a thread name. This would be useful info. Let's add a threadName field (default empty string).
 - Solaris thread ID is not unique (basically it's 1, 2, 3, etc. for each process). So keeping the process ID as part of a thread's identity (separate from the thread ID) is a must.
 - Mac threads don't have their own numbers easily obtainable so we can ignore them.  I'll keep digging around whatever macOS source I can find to see if there's anything useful.

So, basically:
 - keep the "new" thread ID field
 - also add a new threadName field to populate with info on FreeBSD/Solaris
 - have getProcessID return the "owning" process
 - have getParentProcessID return the owning process's parent

tausiflife: Thanks for the explanation and very good points indeed. Few questions -
1. For name, there is a method `getName` which is returning `getParentProcess().getName()` for Linux. Do you want to change this to return "" for linux and return certain value for FreeBSD/Solaris?

tausiflife: Also how did you test this for multiple OS. I work on ubuntu and hence could test Linux easily. Do you create VM and test it there?

dbwiddis: Yeah, we can probably reuse the getName field where appropriate. Leave blank except for where it's filled in.   

And for multiple OS's, yes, I mostly work on my Mac, with VMs for Win7, Ubuntu, Solaris, and FreeBSD. I also have a physical Win10 machine for work that I test windows code on.  I recently got a Synology Diskstation that supposedly has VM support and am trying to figure out how to set that up as well, but haven't gotten very far.

And I just committed the Windows mappings for `Thread32First` and `Thread32Next` to your fork/branch.

dbwiddis: FYI, dug in to macOS source code. `ps -M` is the closest it gets to listing threads, and all it really does is iterate from `j=0` to `j<numthreads` using `j` as the thread number and copying all the process info.  Thread 0 gets all the usage stats and threads 1-(numthreads-1) get 0's.  So it's basically useless, but I don't feel bad doing the same thing. :)
```
$ ps -aM           
USER           PID   TT   %CPU STAT PRI     STIME     UTIME COMMAND
root          1296 s004    0.0 S    31T   0:00.02   0:00.02 login -pfl danielwiddis /bin/bash -c exec -la zsh /bin/zsh
              1296         0.0 S    31T   0:00.00   0:00.00 
danielwiddis  1301 s004    0.0 S    31T   0:00.64   0:00.57 -zsh
root         60632 s004    0.0 R    31T   0:00.00   0:00.00 ps -aM
```

tausiflife: Thanks for the commit for windows. I don't both windows and macOS unfortunately, but i have created VM for both solaris and FreeBSD and can work on it.

dbwiddis: No problem.   Look for further Windows commits :)

oshi-core/src/main/java/oshi/driver/linux/proc/ProcessStat.java
```diff
@@ -40,8 +43,152 @@
...
 public final class ProcessStat {
 
+    /**
+     * Constant defining the number of integer values in {@code /proc/pid/stat}. 2.6
+     * Kernel has 44 elements, 3.3 has 47, and 3.5 has 52.
+     */
+    public static final int PROC_PID_STAT_LENGTH;
     private static final Pattern DIGITS = Pattern.compile("\\d+");
 
+    static {
```
dbwiddis, line 53: There's a lot of stuff moved here, where there any other changes?  Any reason to rearrange the order?   Generally like to keep big whitespace changes out of diffs to ease review.

oshi-core/src/main/java/oshi/driver/linux/proc/TaskStat.java
```diff
@@ -0,0 +1,259 @@
...
+import oshi.util.ParseUtil;
+import oshi.util.platform.linux.ProcPath;
+
+public class TaskStat extends AbstractOSThread {
+
+    private static final Logger LOG = LoggerFactory.getLogger(TaskStat.class);
+    private static final int PAGE_SIZE = ParseUtil
+            .parseIntOrDefault(ExecutingCommand.getFirstAnswer("getconf PAGESIZE"), 4096);
+
+    private static final int[] THREAD_PID_STAT_ORDERS = new int[TaskStat.ThreadPidStat.values().length];
```
dbwiddis, line 46: Minor comment on naming things (one of the two hard things).  The constant you're coping was intended to allude to `/proc/<pid>/stat` and you're still pulling stuff from the `/proc` filesystem.  So this should probably be `PROC_TASK_STAT_ORDERS` I think? `PROC_PID_TASK_STAT_ORDERS` is too verbose :)

oshi-core/src/main/java/oshi/driver/linux/proc/TaskStat.java
```diff
@@ -0,0 +1,259 @@
...
+import oshi.software.common.AbstractOSThread;
+import oshi.software.os.OSProcess;
+import oshi.software.os.linux.LinuxOSProcess;
+import oshi.software.os.linux.LinuxOperatingSystem;
+import oshi.util.ExecutingCommand;
+import oshi.util.FileUtil;
+import oshi.util.ParseUtil;
+import oshi.util.platform.linux.ProcPath;
+
+public class TaskStat extends AbstractOSThread {
```
dbwiddis, line 40: Why are we extending `AbstractOSThread`?  I think you renamed LinuxOSThread rather than the file closer to `ProcessStat`.  Which is what I suggested, but it's because I was confused!  Sorry.

This should be LinuxOSThread (like you had it) but you had it in the `driver` package hierarchy instead of the `oshi.software.os.linux` package.  You can either add to the `ProcessStat` driver package or create a new `TaskStat` package if there's not a lot of overlap.

oshi-core/src/main/java/oshi/software/common/AbstractOSThread.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public List<OSThread> getThreadDetails() {
+        return Collections.emptyList();
```
dbwiddis, line 57: Hmm, why are we returning ThreadDetails from the OSThread?   I guess we want to override the parent that returns the list of threads?  Not sure we need this overridden method, or if it should throw an exception.

oshi-core/src/main/java/oshi/software/os/OSProcess.java
```diff
@@ -362,6 +364,12 @@
      */
     boolean updateAttributes();
 
+    /**
+     * Retrieves the threads of the process and their details.
+     * @return a list of threads
```
dbwiddis, line 369: Let's get a little verbose here. Copy one of the others, should probably be "returns an unmodifiable list of OSThreads" or similar.

oshi-core/src/main/java/oshi/software/os/OSThread.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+package oshi.software.os;
+
+/**
+ * Represents a Thread/Task on the operating system.
+ */
+public interface OSThread extends OSProcess {
+
+    /**
+     * <p>
```
dbwiddis, line 32: Given this is an API level thing and "thread ID" is a variable thing, I would like a little more than the default "getter".  Something like "Returns the thread ID, whose interpretation is operating-system dependent."

oshi-core/src/main/java/oshi/software/os/OSThread.java
```diff
@@ -0,0 +1,47 @@
...
+     * Getter for the field <code>threadId</code>.
+     * </p>
+     *
+     * @return Returns the id of the thread.
+     */
+    int getThreadId();
+
+    /**
+     * <p>
+     * Getter for the field <code>parentProcess</code> which is the parent process of this thread.
```
dbwiddis, line 42: Let's be very clear here that this is the parent of the process to which the thread belongs. You might also suggest that the thread's *owning* process is accessible with getProcessId().

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -23,7 +23,18 @@
  */
 package oshi.software.os.linux;
 
-import static oshi.util.Memoizer.memoize;
+import org.slf4j.Logger;
```
dbwiddis, line 26: Uh oh. I bet our different IDEs have dueling import orderings... heh. I'll just do a mass formatting before merging this.

dbwiddis:
I'm going to go ahead and try to sort all this out tonight and make a commit.

dbwiddis:
OK, I've gone through and fixed most of the small "preference" things and some refactoring.  I think this is about how you wanted it.  I still have a few questions.
 
I'm not sure having OSThread extend OSProcess is a good thing.  It forces us to implement methods that make no sense (`getThreadDetails` from the thread) or which need thread equivalents/replacements (thread CPU load rather than process CPU load methods, for example).  
 - My initial thought to "keep everything" was based on Linux having the full "stat" output for tasks, but I'm not sure how much of that is relevant.
 - Mac and the other Unixes don't have much to offer here, and Windows has a very clearly defined subset of interesting information, including some not included (thread start address, stopped reason).
 - I wonder how much of the thread (task) `stat` data is actually useful. We may want to really shrink the `OSThread` API to be closer to the Windows output, plus anything on Linux that's interesting.

oshi-core/src/main/java/oshi/driver/linux/proc/ProcessStat.java
tausiflife: I guess it was the IDE which formatted it  :( . I applied the setting but seems it is not working.

oshi-core/src/main/java/oshi/driver/linux/proc/TaskStat.java
tausiflife: Sure i can do this.

oshi-core/src/main/java/oshi/software/common/AbstractOSThread.java
tausiflife: My thought was to return empty list as threads won't have anything to return. I am not quite sure about throwing an exception for accessing a method.

oshi-core/src/main/java/oshi/software/os/OSProcess.java
tausiflife: Will take it as a general review for all the java docs and apply where ever possible.

oshi-core/src/main/java/oshi/software/os/OSThread.java
tausiflife: Makes sense to add more information.

oshi-core/src/main/java/oshi/software/os/OSThread.java
tausiflife: Yes true, i was thinking about this but then i let it be like that. Will change this.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
tausiflife: Thanks i am not sure why my IDE is not formatting correctly. Sorry for the inconvenience caused.

oshi-core/src/main/java/oshi/software/common/AbstractOSThread.java
dbwiddis: It's reasonable to throw an `UnsupportedOperationException` for a method that shouldn't be called.  But see my other comments about perhaps not extending OSProcess overall.  This, combined with the need for different "usage" method naming, is a strong argument for a standalone class that we can name things like we want and not be forced to inherit/implement things that provide no added value.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
dbwiddis: It's not a big deal. I should probably figure out the equivalent formatter settings for IntelliJ and save a file for those.  The formatting and "ordering imports" are separate things, though.

oshi-core/src/main/java/oshi/driver/linux/proc/TaskStat.java
dbwiddis: Not sure if you saw my note, I already tried to put things in the right place. Sorry for going back and forth on suggestions here, but this is normal in the process of developing a new API.

tausiflife:
Thanks for all the commits. I haven't been as productive as you for this change :(

Nevertheless my initial thoughts were inline of what you just said as i was not sure what common information we have for all the OS. But after you suggested for `OSThread` to extend `OSProcess`, it made sense as thread is a lightweight process. 
I agree that we should limit this to only useful information specific to a thread, which should be a super set of the interesting thread information from different OS. I haven't yet gone through what you have done for windows to understand it correctly.

Also I am still working on setting up of solaris and freebsd VM to be able to test there.

dbwiddis:
A few more design comments:

I just did a detailed comparison of the `proc/pid/task/tid/stats` output for a few multi-threaded processes, and am even more convinced we should not inherit OSProcess.  Things that differ that we should model:
 - thread id, name, state
 - page faults (we don't use)
 - cpu usage ticks
 - start time 
 - memory address info

This is the same, no reason to inherit it
 - parent pid, group id
 - vsize and rss (all threads share the same memory)
 - thread count (in theory 1 for a thread but it reports the total # despite `man proc` saying it is 0)

This isn't in `/proc/pid/task/tid/stat`, if we want it we need to find it in another file
 - open files (not sure if this is a thread thing?)
 - bytes read/bytes written (not sure if this is a thread thing, it's in /proc/pid/io (sudo only) but is only nonzero for the primary task in all the places I checked)

Given this, there's a nice overlap with the available information on Windows (See the new `oshi.driver.windows.registry.ProcesPerformanceData` class).  In addition to what I mapped we could also get process start address, reason waiting, and context switches, which I'd consider including if we can get that easily from Linux.

dbwiddis:
>  I haven't been as productive as you for this change :(

On the contrary!  Don't measure productivity by lines of mostly boilerplate/copypaste stuff.  Figuring out the API is usually the hardest part.  There's a reason that feature request sat there for a year.

I think we're getting close to a really good feature, though!

dbwiddis:
One more thing to consider is performance: current linux version pre-populates all the thread details info, but would probably be a better implementation just reading per-process on request.  For Windows, registry is an all-at-once query but perfcounter/wmi fallback could be more efficient per-process.

So, summing up, let's split out tasks:

@tausiflife:
 - Revise API to not extend OSProcess, and only include the subset of per-thread information from [Linux](https://github.com/oshi/oshi/pull/1258#issuecomment-644844716) and [Windows](http://library.wmifun.net/cimv2/win32_perfrawdata_perfproc_thread.html).
  - I'll leave it to you to decide if you want to include "context switches" (I lean toward yes) and "process start address" / "reason for waiting" (I lean toward no).
 - Complete linux port per above
 - Do Unix ports per above, probably just parsing the appropriate `ps` output.

@dbwiddis:
 - Fully implement Windows port per the API. The drivers are there, just needs connecting to API
 - Implement a mac port that matches `ps -M` (which is mostly useless)

oshi-core/src/main/java/oshi/driver/linux/proc/TaskStat.java
tausiflife: I completely understand this and i believe great things evolve this way only. To tell you, this has been quite a bit a learning for me in terms of collaboration :). Thanks a lot for being such a supportive mentor.

tausiflife:
Now since we have per thread information, i think we don't need to have `owningProcess` inside `AbstractOSThread` class because the thread details are inside the process instance only. What do you think?

dbwiddis:
Generally since the only way to get to an `OSThread` objecct is from the `OSProcess` it's attached to, I agree it's not really that relevant to have the whole object.  However, I do think `OwningProcessID` is an important field.  Cross-platform, many threads are identified only by an index # and need to be combined with the process ID to be unique.

tausiflife:
Makes sense to keep the `owningProcessId`. Also i am bit confused what do you mean by memory address info of the thread. I see all the memory information from the status file, which can be parsed.

dbwiddis:
>  Also i am bit confused what do you mean by memory address info of the thread

Windows has the thread start address as one of the counter outputs. Linux has a lot more memory info, one of which is probably the start address. FreeBSD also has the START option on [`procstat`](https://www.unix.com/man-page/FreeBSD/1/procstat/). Don't see anything on `prstat` for Solaris but it might be there.  We could display it, but I'm not really sure how useful it is to anyone. :)

tausiflife:
I agree that the only way to get to an OSThread object should be from the OSProcess it's attached to, what i interpret is OSThread should only be accessible from inside the OSProcess. If we do so then we can have lazy initialization of the threads only on call from the process instance. This will improve the performance. But one question arises-  why should we allow independent initialization of OSThread class?

dbwiddis:
> But one question arises- why should we allow independent initialization of OSThread class?

Not sure there's any strong reason to disallow it.

If all the information in the `OSThread` can be generated without reference to the `OSProcess` then it's possible for methods which must iterate the whole thread list to create the `OSThread` objects up front.  Take a look at `oshi.driver.windows.registry.ThreadPerformanceData` for example.  The output of its query method is a `Map<Integer, PerfCounterBlock>` with thread ID as the key.  There's no reason we couldn't make that a `Map<Integer, WindowsOSThread>` instead.

(Actually there are probably some reasons to prefer not to, dealing with thread safety and the ability to modify members, and separating API from driver, so it's probably best to keep the existing implementation.)

Of note, in the case of Windows, the overhead of pulling data from the registry is incurred for _all threads_ even if I only want threads for one process.  So it might be better to return a `Map<Integer, Map<Integer, PerfCounterBlock>>`, with the outer map containing Process IDs and the inner map Thread IDs.  We would only have to query that once, and then just pass the map to the various constructors.  Take a look at `WindowsOperatingSystem` lines 119-122, which use a memoizer to return the same object when the registry/perfcounters are queried repeatedly in a short timespan.  That object is later passed to the `WindowOSProcess` constructor.

Which also brings up the thought that we might want to generate the thread list from `WindowsOperatingSystem` (with optional `int pid` or `Collection<Integer> pids` overloads just like we have for `getProcess()`).   Or, probably better, make sure we pass the memoized "thread map" variable as an argument to each `WindowsOSProcess` alongside the current process maps, so they would all have access to it.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSThread.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * Enum used to update attributes. The order field represents the 1-indexed
+     * numeric order of the stat in /proc/pid/task/tid/stat per the man file.
+     */
+    private enum ThreadPidStat {
+        // The parsing implementation in ParseUtil requires these to be declared
+        // in increasing order
+        PPID(4), MINOR_FAULTS(10), MAJOR_FAUT(12), USER_TIME(14), KERNEL_TIME(15), PRIORITY(18), THREAD_COUNT(20), START_TIME(22), VSZ(23), RSS(24), START_CODE(26);
```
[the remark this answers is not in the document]
dbwiddis: Also an aside. We don't display page faults at the process level, but perhaps we should.  But let's keep that on a separate issue/PR and just finish the threads first. :)

oshi-core/src/main/java/oshi/software/os/OSThread.java
```diff
@@ -0,0 +1,161 @@
...
+     * Since the context switches could be voluntary and non-voluntary, this gives the sum of both.
+     * </p>
+     * @return sum of both voluntary and involuntary context switches.
+     */
+    long getContextSwitches();
+
+    /**
+     * <p>
+     * Getter for the field <code>minorFaults</code>, which gives the number of minor faults the thread has made
+     * which have not required loading a memory page disk.
```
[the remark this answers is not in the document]
dbwiddis: Or maybe not. They are in posix ps output, so maybe we can get them in other unixes.   Not windows at thread level (and sadly it only gives one number at process level)

oshi-core/src/main/java/oshi/software/os/OSThread.java
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Gets cumulative CPU usage of this task/thread.
+     * <p>
+     * This calculation sums CPU ticks across all processors.
+     *
+     * @return The proportion of up time that the thread was executing in kernel or
+     *         user mode.
+     */
+    double getCpuLoadCumulative();
```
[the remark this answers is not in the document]
tausiflife: Ok i guess it will have the same logic as in OSProcess?

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSThread.java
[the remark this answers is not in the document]
tausiflife: Sorry for the typo. Will fix it.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSThread.java
```diff
@@ -0,0 +1,182 @@
...
+        // BOOT_TIME could be up to 500ms off and start time up to 5ms off. A process
+        // that has started within last 505ms could produce a future start time/negative
+        // up time, so insert a sanity check.
+        if (this.startTime >= now) {
+            this.startTime = now - 1;
+        }
+        this.minorFaults = statArray[ThreadPidStat.MINOR_FAULTS.ordinal()];
+        this.majorFaults = statArray[ThreadPidStat.MAJOR_FAUT.ordinal()];
+        this.startMemoryAddress = statArray[ThreadPidStat.START_CODE.ordinal()];
+        Long voluntaryContextSwitches = Long.valueOf(status.get("voluntary_ctxt_switches") != null ? status.get("voluntary_ctxt_switches") : "0");
```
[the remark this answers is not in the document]
tausiflife: Using Long was unintentional, will fix this.

oshi-core/src/main/java/oshi/util/platform/linux/ProcPath.java
```diff
@@ -59,6 +59,10 @@
     public static final String SELF_STAT = PROC + "/self/stat";
     public static final String STAT = PROC + "/stat";
     public static final String SYS_FS_FILE_NR = PROC + "/sys/fs/file-nr";
+    public static final String TASK_PATH = PROC + "/%d/task";
+    public static final String TASK_STATUS = TASK_PATH + "/%d/status";
+    public static final String TASK_STAT = TASK_PATH + "/%d/stat";
+    public static final String TASK_IO = TASK_PATH + "/%d/io";
```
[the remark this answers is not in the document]
tausiflife: Removed.

oshi-core/src/main/java/oshi/driver/linux/proc/ProcessStat.java
```diff
@@ -453,4 +456,58 @@ private ProcessStat() {
...
+     * directory with only numeric digit filenames, corresponding to the threads.
+     *
+     * @param pid
+     *            process id
+     * @return A list of thread id.
+     */
+    public static List<Integer> getThreadIds(int pid) {
+        File threadDir = new File(String.format(ProcPath.TASK_PATH, pid));
+        File[] threads = threadDir
+                .listFiles(file -> DIGITS.matcher(file.getName()).matches() && Integer.valueOf(file.getName()) != pid);
```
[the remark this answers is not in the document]
tausiflife: I have added the null check and have left the optimization for later.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -206,6 +208,11 @@ public long getBytesWritten() {
         return this.bytesWritten;
     }
 
+    @Override
+    public List<OSThread> getThreadDetails() {
+        return Collections.unmodifiableList(this.threadDetails);
```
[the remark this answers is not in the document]
tausiflife: Done. This will improve the performance a bit also :)

dbwiddis:
Looking good!  The Windows port is done.  I tweaked the OSThread docs a bit, renaming the CPU methods, and put "default" returns in the abstract class, so we don't have to write any code in the OS class for things we don't have (I skipped page faults for Windows, for example).  Also added a toString to help with debugging.

I've also written the driver that I hope works for macOS, but have to wait until I'm actually at my mac to test it :)

oshi-core/src/main/java/oshi/software/os/OSThread.java
```diff
@@ -0,0 +1,161 @@
...
+     * Getter for the field <code>startTime</code>.
+     * </p>
+     *
+     * @return Returns the start time of the task/thread in number of milliseconds since
+     *         January 1, 1970.
+     */
+    long getStartTime();
+
+    /**
+     * Attempts to updates process attributes. Returns false if the update fails,
```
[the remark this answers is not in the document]
tausiflife: One question regarding updateAttributes method. Java docs says that it will return false if something, shouldn't we have a try catch block to return false in case of any exception?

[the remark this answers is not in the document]
dbwiddis: We don't try to catch exceptions there either, so the program would probably halt on an exception.  We should handle exceptions at the lowest level so these update methods never see them.

dbwiddis:
Windows and mac are now both done, and I think you're done with Linux!  Just the Unixes to go!

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
```diff
@@ -0,0 +1,162 @@
...
+                break;
+        }
+        // Avoid divide by zero for processes up less than a second
+        long elapsedTime = ParseUtil.parseDHMSOrDefault(split[3], 0L); //etimes
+        this.upTime = elapsedTime < 1L ? 1L : elapsedTime;
+        long now = System.currentTimeMillis();
+        this.startTime = now - this.upTime;
+        this.kernelTime = ParseUtil.parseDHMSOrDefault(split[4], 0L); //systime
+        this.userTime = ParseUtil.parseDHMSOrDefault(split[5], 0L) - this.kernelTime; //time
+        //this.startMemoryAddress = split[6]; hexadecimal
```
tausiflife, line 155: Should we change the memory address data type to string, as freebsd is returning hexdecimal value?

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    @Override
+    public long getStartTime() {
+        return this.startTime;
+    }
+
+    @Override
+    public boolean updateAttributes() {
+        String psCommand = "ps -awwxo tdname,lwp,state,etimes,systime,time,tdaddr,nivcsw,nvcsw,majflt -H -p "
```
tausiflife, line 108: I couldn't find any filter for threads, hence had to us grep command. do you know about any other way?

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
dbwiddis: Parsing hex is easy using `Long.decode()`.  See `ParseUtil.parseLastLong()`.

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
dbwiddis: I'm not sure what you're asking here.  If I do `ps -awwxo pid,tdname,lwp,state,etimes,systime,time,tdaddr,nivcsw,nvcsw,majflt -H -p 11` I get only the threads for pid 11.

dbwiddis:
```
$> ps -awwxo pid,tdname,lwp,state,etimes,systime,time,tdaddr,nivcsw,nvcsw,majflt,minflt -H -p 11
PID TDNAME                 LWP STAT ELAPSED SYSTIME      TIME   TDADDR  NIVCSW     NVCSW MAJFLT MINFLT
 11 swi6: task queue    100004 WL    760562 0:00.00   0:00.03 c5f5a880      18        39      0      0
 11 swi6: Giant taskq   100005 WL    760562 0:00.00   0:00.02 c5f5a440       0        74      0      0
 11 swi5: fast taskq    100007 WL    760562 0:00.00   0:00.00 c5f59880       0         0      0      0
 11 swi1: netisr 0      100010 WL    760562 0:00.00  11:41.43 c5f95880  121124  14073708      0      0
 11 swi4: clock (0)     100011 WL    760562 0:00.00  14:52.95 c5f95440 4772170  14241267      0      0
 11 swi4: clock (1)     100012 WL    760562 0:00.00   0:00.00 c5f95000       0         0      0      0
 11 swi3: vm            100013 WL    760562 0:00.00   0:00.00 c5f94880       0         0      0      0
 11 gic0,s19: cesa0     100026 WL    760562 0:00.01   0:01.78 c5fd8440       0    289936      0      0
 11 gic0,s20: cesa1     100027 WL    760562 0:00.00   0:00.00 c5fd8000       0         0      0      0
 11 gic0,s1: spi0       100028 WL    760562 0:00.00   0:00.00 c61b9880       0         0      0      0
 11 swi0: uart uart     100029 WL    760562 0:00.00   0:00.00 c61b9440       0       776      0      0
 11 mpic0: mvneta0      100031 WL    760562 0:00.00   0:00.00 c61b8880       0         0      0      0
 11 mpic0: mvneta1      100034 WL    760562 0:00.00  81:33.78 c61b7880       0 116198736      0      0
 11 gic0,s18: ehci0     100035 WL    760562 0:00.00   0:00.00 c61b7440       0         0      0      0
 11 mpic0: mvneta2      100042 WL    760562 0:00.00 108:48.18 c6340000       0 147196077      0      0
 11 gic0,s26: ahci0     100043 WL    760562 0:00.00   0:22.66 c633b880    8944   3281193      0      0
 11 gic0,s25:-dhci_fdt0 100044 WL    760562 0:00.00   0:00.01 c633b440       0      1504      0      0
 11 gic0,s17: xhci0     100045 WL    760562 0:00.00   0:00.00 c633b000       0         0      0      0
 11 swi1: pf send       100053 WL    760562 0:00.00   0:00.00 c651d440       0        24      0      0
 11 swi1: pfsync        100054 WL    760562 0:00.00   0:00.00 c651d000       0         0      0      0
 11 swi1: netisr 1      100073 WL    760562 0:00.00   0:00.00 c6d08880       0         0      0      0
```

dbwiddis: Oh, maybe I should look at the source code first to see the context. We're in the thread object!  

No, I don't think there's a switch to filter PS.  But don't pipe `grep` to the OS. Just fetch the `List<String>` and parse the returned string using `contains()` as the Java equivalent.

dbwiddis: You might also consider using `procstat` for some info as well:
```
> procstat -t 11
  PID    TID COMM                TDNAME              CPU  PRI STATE   WCHAN
   11 100004 intr                swi6: task queue     -1   48 wait    -
   11 100005 intr                swi6: Giant taskq    -1   48 wait    -
   11 100007 intr                swi5: fast taskq     -1   44 wait    -
   11 100010 intr                swi1: netisr 0       -1   28 wait    -
   11 100011 intr                swi4: clock (0)      -1   40 wait    -
   11 100012 intr                swi4: clock (1)      -1   40 wait    -
   11 100013 intr                swi3: vm             -1   36 wait    -
   11 100026 intr                gic0,s19: cesa0      -1    8 wait    -
   11 100027 intr                gic0,s20: cesa1      -1    8 wait    -
   11 100028 intr                gic0,s1: spi0        -1   20 wait    -
   11 100029 intr                swi0: uart uart      -1   24 wait    -
   11 100031 intr                mpic0: mvneta0       -1    8 wait    -
   11 100034 intr                mpic0: mvneta1       -1    8 wait    -
   11 100035 intr                gic0,s18: ehci0      -1   12 wait    -
   11 100042 intr                mpic0: mvneta2       -1    8 wait    -
   11 100043 intr                gic0,s26: ahci0      -1   12 wait    -
   11 100044 intr                gic0,s25:-dhci_fdt0  -1   20 wait    -
   11 100045 intr                gic0,s17: xhci0      -1   12 wait    -
   11 100053 intr                swi1: pf send        -1   28 wait    -
   11 100054 intr                swi1: pfsync         -1   28 wait    -
   11 100073 intr                swi1: netisr 1       -1   28 wait    -
```

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
```diff
@@ -0,0 +1,162 @@
...
+        this.upTime = elapsedTime < 1L ? 1L : elapsedTime;
+        long now = System.currentTimeMillis();
+        this.startTime = now - this.upTime;
+        this.kernelTime = ParseUtil.parseDHMSOrDefault(split[4], 0L); //systime
+        this.userTime = ParseUtil.parseDHMSOrDefault(split[5], 0L) - this.kernelTime; //time
+        //this.startMemoryAddress = split[6]; hexadecimal
+        long nonVoluntaryContextSwitches = ParseUtil.parseLongOrDefault(split[7], 0L);
+        long voluntaryContextSwitches = ParseUtil.parseLongOrDefault(split[8], 0L);
+        this.contextSwitches = voluntaryContextSwitches + nonVoluntaryContextSwitches;
+        this.majorFaults = ParseUtil.parseLongOrDefault(split[9], 0L);
```
dbwiddis, line 159: Should be able to add minor faults also.

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
dbwiddis: (`procstat` has the priority, not sure if that's available on `ps`?)

dbwiddis: OK, you can add `pri` to the `ps` string as well but it's interesting, it shows up as a negative number offset by 100, e.g., TID 100004 has pri 48 in `procstat` and -52 on `ps`.

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
dbwiddis: Actually since the string doesn't have the `0x` it's probably better to just use `Long.parseLong(split[6], 16)`.  (Except don't do it here, put it in the `ParseUtil` class and catch the `NumberFormatException` and return a default if it fails.)

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
```diff
@@ -223,6 +226,30 @@ private int queryBitness() {
         return 0;
     }
 
+    @Override
+    public List<OSThread> getThreadDetails() {
+        List<OSThread> threads = new ArrayList<>();
+        String psCommand = "ps -awwxo tdname,lwp,state,etimes,systime,time,tdaddr,nivcsw,nvcsw,majflt -H";
```
dbwiddis, line 232: add `minflt` and `pri` to this.

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
```diff
@@ -223,6 +226,30 @@ private int queryBitness() {
...
+        if (getProcessID() >= 0) {
```
dbwiddis, line 233: I don't think this is nesseary. We should never have a process 0 on Linux and if we did, I'd want to get its threads!  Just add the `-p pid` to the above line.

tausiflife: Ok.

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
tausiflife: Man page of ps says `minflt	total page reclaims`. I am not sure if we should use that. Also we don't have any method for priority in the OSThread interface. Should we add priority for everything?

dbwiddis: I missed that we removed it.  Yes, we should have priority.  

Also minor page faults and reclaims are a synonym.  The very common paging with memory that’s not a big problem.  Major page faults (disk paging) are more serious.  Windows doesn’t distinguish between them,  it we should include both if possible.

dbwiddis: Have been on the road posting from phone. Summarizing:
1. Please add back in priority.  Driver for windows already has CURRENTPRIORITY field to include. Not available on mac. I assume linux has it in the proc/pid/task/tid/stat output. Hopefully you can include all that.
2. Please include minflt from unix, it has same meaning/different words as linux.   

Other than that it really looks good and hopefully Solaris with ps and/or prstat should be similar.

tausiflife: Relax i am working on it and probably finish it this weekend. Although Some of the test cases are failing in freebsd `SensorTest.java` `testSensors` and `GlobalMemoryTest : Available memory should be less than or equal to total memory`. This may be because of the VM thing but i am no of sure.

dbwiddis: I’m on the road all weekend so take your time. :)

dbwiddis: I’m on the road all weekend so take your time. :)

dbwiddis: Re: FreeBSD memory, check the relevant sysctl calls

dbwiddis: Re: FreeBSD memory, check the relevant sysctl calls

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
tausiflife: Added priority. `ps` has priority and a value of `actual priority - 100` . Checked with `procstat` and it matches.

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
tausiflife: ok.

oshi-core/src/main/java/oshi/software/common/AbstractOSProcess.java
```diff
@@ -72,6 +75,11 @@ public double getProcessCpuLoadBetweenTicks(OSProcess priorSnapshot) {
         return getProcessCpuLoadCumulative();
     }
 
+    @Override
+    public List<OSThread> getThreadDetails() {
+        return Collections.emptyList();
```
dbwiddis, line 80: Just a reminder, this was a placeholder while developing all the implementations. Once you do Solaris this can/should be removed.

oshi-core/src/main/java/oshi/util/ParseUtil.java
```diff
@@ -1120,4 +1120,21 @@ public static String parseUtAddrV6toIP(int[] utAddrV6) {
...
         }
     }
+
+    /**
+     * Parses a string of hex digits to long value.
+     * @param hexString A sequence of hex digits
+     * @param defaultValue default value to return if parsefails
+     * @return The corresponding long value
+     */
+    public static long hexStringToLong(String hexString, long defaultValue) {
```
dbwiddis, line 1130: This looks great.   As the failed "codecov" prompts say, you need to write a testcase  (one passing one failing) in `ParseUtilTest`.

tausiflife: I am unable to convert memory addres to long using this method as this throws `NumberFormatException`, however using `BigInteger`, i can get the long value, but this long value is overflowing. I still suggest that we should have the `address` as `String`.

oshi-core/src/main/java/oshi/software/common/AbstractOSProcess.java
tausiflife: yes sure.

oshi-core/src/main/java/oshi/util/ParseUtil.java
dbwiddis: Ah, the problem is that it is an unsigned long on the C side which is bigger than max long.  We can still use a negative value.   I think there is actually already another parsing method in ParseUtil does character by character parsing and bit-shifting to do this.  

I’ll rewrite that parse method, you can just have it return 0 for now and move in to Solaris.

dbwiddis: The existing method `ParseUtil.byteArrayToLong()` will do what you need.  Just convert the String to a byte array using `ParseUtil.hexStringToByteArray()`.

Although it would be nicer to parse the hex string directly to the `long` instead of using the array, using `hexStringToByteArray()` as a starting point and adding/bitshifting bytes instead of storing in an array.

tausiflife: I tried adding both the functions, and the result is same `
byte[] byteArray = hexStringToByteArray("ffffffff830f53a0"); 
 long result = byteArrayToLong(byteArray, 8);` **and** `new BigInteger("ffffffff830f53a0", 16).longValue()`both return the same value, which makes me think, why are we not using `String` for memory address. I think then i can easily picture a `memory address` just by looking at its value. Returning a long value is fine but when i look at it, it doesn't make any sense to make.

dbwiddis: Yes, the string value is more user friendly but we can leave that to the user to format how they want with %x on String.format().  The underlying data is a numeric value.  If not long, I’d consider a Pointer (which wraps a long as it’s peer value).

The value you get from big integer long value is the correct one that we want to return.

dbwiddis: Also the long only takes 8 bytes to store.  A string would take 8 bytes for the pointer to the memory location of the String object (a long!) plus all the overhead of an object plus the bytes for the String (2bytes per hex character turning 16 characters into 32 bytes).  :)

dbwiddis: You bring up a good point on readability and we should use the %x format in the toString().

tausiflife: done.

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
```diff
@@ -0,0 +1,177 @@
...
+        this.kernelTime = ParseUtil.parseDHMSOrDefault(split[4], 0L); // systime
+        this.userTime = ParseUtil.parseDHMSOrDefault(split[5], 0L) - this.kernelTime; // time
+        this.startMemoryAddress = ParseUtil.hexStringToLong(split[6], 0L);
+        long nonVoluntaryContextSwitches = ParseUtil.parseLongOrDefault(split[7], 0L);
+        long voluntaryContextSwitches = ParseUtil.parseLongOrDefault(split[8], 0L);
+        this.contextSwitches = voluntaryContextSwitches + nonVoluntaryContextSwitches;
+        this.majorFaults = ParseUtil.parseLongOrDefault(split[9], 0L);
+        this.minorFaults = ParseUtil.parseLongOrDefault(split[10], 0L);
+        // ps gives values offset by 100, hence adding 100
+        this.priority = 100 + ParseUtil.parseIntOrDefault(split[11], 0);
```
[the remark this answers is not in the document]
dbwiddis: OK I think this explains it:
> For processes with a real time priority of something other than 'normal' the
priority is not adjusted dynamically.
Processes with rtprio=idle always have lower priority than all other
processes.
Processes with an actual real time priority set will always have higher
priority than non-real time processes.

The way I see it, the "normal" ones have a priority that auto-adjusts more negative as needed to balance CPU among threads.   When real time priority is set it does the -100 offset so they are always at a higher priority (more negative) than everything else.  (Except for idle which is lowest priroity = highest positive value = 255 but adjusted -100 by the real time.)

So basically it's -100 for real time, and -(something between 0 an 100) for "normal" auto-adjusting.   I think we should probably just return whatever `pri` reports without modifying it.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOSProcess.java
```diff
@@ -281,4 +281,76 @@ private boolean updateAttributes(String[] split) {
...
+                IntStream.rangeClosed(0, 6).forEach(idx -> split[idx] = splitPs[idx]);
+                map.put(ParseUtil.parseIntOrDefault(splitPs[0], 0), split);
+            });
+            // 0-pid, 1-username, 2-usertime, 3-sys, 4-trp, 5-tfl, 6-dfl, 7-lck, 8-slp,
+            // 9-lat, 10-vcx, 11-icx, 12-scl, 13-sig, 14-processid/lwpid
+            if (threadListInfo2.size() > 1) { // first row is header
+                threadListInfo2.stream().skip(1).forEach(threadInfo -> {
+                    String[] splitPrstat = ParseUtil.whitespaces.split(threadInfo.trim(), 10);
+                    String[] processLwpid = ParseUtil.slash.split(splitPrstat[14].trim()); // format is process/lwpid
+                    String[] split1 = map.get(ParseUtil.parseIntOrDefault(processLwpid[1], 0));
```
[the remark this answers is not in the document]
tausiflife: I was thinking about this also and then i didn't do considering that it will always be present. Thanks for pointing out.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOSProcess.java
```diff
@@ -281,4 +281,76 @@ private boolean updateAttributes(String[] split) {
 
         return true;
     }
+
+    /**
+     * Merges results of a ps and prstat query, since Solaris thread details are not
+     * available in a single command. Package private to permit access by
+     * SolarisOSThread.
+     *
+     * @param threadListInfo1
```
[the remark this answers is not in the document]
tausiflife: ok

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOSProcess.java
```diff
@@ -281,4 +281,76 @@ private boolean updateAttributes(String[] split) {
...
+    static Map<Integer, String[]> parseAndMergeThreadInfo(List<String> threadListInfo1,
+            List<String> threadListInfo2) {
+        Map<Integer, String[]> map = new HashMap<>();
+        final String[] split = new String[9];
+        // 0-lwpid, 1-state,2-elapsedtime,3-kerneltime, 4-usertime, 5-address,
+        // 6-priority
+        if (threadListInfo1.size() > 1) { // first row is header
+            threadListInfo1.stream().skip(1).forEach(threadInfoPs -> {
+                String[] splitPs = ParseUtil.whitespaces.split(threadInfoPs.trim(), 6);
+                IntStream.rangeClosed(0, 6).forEach(idx -> split[idx] = splitPs[idx]);
```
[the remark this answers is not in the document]
tausiflife: done.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOSProcess.java
```diff
@@ -281,4 +281,76 @@ private boolean updateAttributes(String[] split) {
...
+     * @param threadListInfo1
+     *            output from ps command.
+     * @param threadListInfo2
+     *            output from the prstat command.
+     * @return a map with key as thread id and an array of command outputs as value
+     */
+    static Map<Integer, String[]> parseAndMergeThreadInfo(List<String> threadListInfo1,
+            List<String> threadListInfo2) {
+        Map<Integer, String[]> map = new HashMap<>();
+        final String[] split = new String[9];
```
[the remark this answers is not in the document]
tausiflife: renamed.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOSProcess.java
```diff
@@ -281,4 +281,76 @@ private boolean updateAttributes(String[] split) {
...
+        // 0-lwpid, 1-state,2-elapsedtime,3-kerneltime, 4-usertime, 5-address,
+        // 6-priority
+        if (threadListInfo1.size() > 1) { // first row is header
+            threadListInfo1.stream().skip(1).forEach(threadInfoPs -> {
+                String[] splitPs = ParseUtil.whitespaces.split(threadInfoPs.trim(), 6);
```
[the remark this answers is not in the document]
tausiflife: done.

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+        if (split != null && split.length == 12) {
+            return updateAttributes(split);
+        }
+        this.state = OSProcess.State.INVALID;
+        return false;
+    }
+
+    private boolean updateAttributes(String[] split) {
+        this.name = split[0];
```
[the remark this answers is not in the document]
tausiflife: ok.

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
```diff
@@ -0,0 +1,177 @@
...
+        String psCommand = "ps -awwxo tdname,lwp,state,etimes,systime,time,tdaddr,nivcsw,nvcsw,majflt,minflt,pri -H -p "
+                + getOwningProcessId();
+        // there is no switch for thread in ps command, hence filtering.
+        List<String> threadList = ExecutingCommand.runNative(psCommand);
+        String[] split = null;
+        for (String psOutput : threadList) {
+            split = ParseUtil.whitespaces.split(psOutput.trim(), 12);
+            int id = ParseUtil.parseIntOrDefault(split[1], 0);
+            if (id == this.getThreadId()) {
+                break;
```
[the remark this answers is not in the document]
tausiflife: done.

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
```diff
@@ -0,0 +1,177 @@
...
+
+    @Override
+    public boolean updateAttributes() {
+        String psCommand = "ps -awwxo tdname,lwp,state,etimes,systime,time,tdaddr,nivcsw,nvcsw,majflt,minflt,pri -H -p "
+                + getOwningProcessId();
+        // there is no switch for thread in ps command, hence filtering.
+        List<String> threadList = ExecutingCommand.runNative(psCommand);
+        String[] split = null;
+        for (String psOutput : threadList) {
+            split = ParseUtil.whitespaces.split(psOutput.trim(), 12);
```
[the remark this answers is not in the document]
tausiflife: ok.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOSProcess.java
```diff
@@ -281,4 +281,76 @@ private boolean updateAttributes(String[] split) {
...
+                IntStream.rangeClosed(0, 6).forEach(idx -> split[idx] = splitPs[idx]);
+                map.put(ParseUtil.parseIntOrDefault(splitPs[0], 0), split);
+            });
+            // 0-pid, 1-username, 2-usertime, 3-sys, 4-trp, 5-tfl, 6-dfl, 7-lck, 8-slp,
+            // 9-lat, 10-vcx, 11-icx, 12-scl, 13-sig, 14-processid/lwpid
+            if (threadListInfo2.size() > 1) { // first row is header
+                threadListInfo2.stream().skip(1).forEach(threadInfo -> {
+                    String[] splitPrstat = ParseUtil.whitespaces.split(threadInfo.trim(), 10);
```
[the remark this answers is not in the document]
tausiflife: done.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOSProcess.java
```diff
@@ -281,4 +281,76 @@ private boolean updateAttributes(String[] split) {
...
+                    String[] processLwpid = ParseUtil.slash.split(splitPrstat[14].trim()); // format is process/lwpid
```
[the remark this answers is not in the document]
tausiflife: ok

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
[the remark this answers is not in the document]
tausiflife: Returned as is.

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSThread.java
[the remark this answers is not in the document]
dbwiddis: I'm also updating this to set the state to INVALID.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOSProcess.java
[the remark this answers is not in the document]
dbwiddis: This looks good. One more thing I'm adding (not because it will ever happen but because I code very defensively!) is in the conditional on line 320, to also check that `idxForwardSlash` is less than length (to avoid using substring out of range) or less than length-1 (to ensure that we actually have something to parse later).

dbwiddis:
OK, this looks good!  I'm ready to merge this.  Anything else you think needs to be done?

tausiflife:
Looks fine to me. I just want to thank you for your patience with my not so great code. Its been a great leaning for me :)

dbwiddis:
Merged!  Do you want to take a crack at #1270 since you're already familiar with the data structures involved?

tausiflife:
Sure i will take it up.

PR #1261, opened 2020-06-11
Outputs for IPv4 & subnetMask and IPv6 & prefix lengths combined

### LINKED ISSUE #1260: macOS 10.16 support
Apple will announce the name of version 10.16 on [June 22](https://developer.apple.com/wwdc20/).  The mac version parsing should be updated to use a properties file rather than annually adding to the switch/case statements.

JapneetSingh5:
This fix will work for single element arrays, but not for multi-element arrays I think.
Am I supposed to only manipulate the String builder part? One solution I thought would be to combine the two arrays when they are being initially fed values, but that would mean only 2 arrays remain instead of current 4.

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
@@ -225,8 +225,7 @@ public abstract class AbstractNetworkIF implements NetworkIF {
         sb.append("Name: ").append(getName()).append(" ").append("(").append(getDisplayName()).append(")").append("\n");
         sb.append("  MAC Address: ").append(getMacaddr()).append("\n");
         sb.append("  MTU: ").append(getMTU()).append(", ").append("Speed: ").append(getSpeed()).append("\n");
-        sb.append("  IPv4: ").append(Arrays.toString(getIPv4addr())).append("\n");
-        sb.append("  Netmask:  ").append(Arrays.toString(getSubnetMasks())).append("\n");
+        sb.append("  IPv4: ").append(Arrays.toString(getIPv4addr())).append("/").append(Arrays.toString(getSubnetMasks())).append("\n");
         sb.append("  IPv6: ").append(Arrays.toString(getIPv6addr())).append("\n");
         sb.append("  Prefix Lengths:  ").append(Arrays.toString(getPrefixLengths())).append("\n");
         sb.append("  Traffic: received ").append(getPacketsRecv()).append(" packets/")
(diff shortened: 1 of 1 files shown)


dbwiddis:
Yeah it's not as simple as appending the arrays. You need to iterate the arrays (they should be the same size) and combine the terms, then put them back into an array.  Something like this should work for v4:

```
String[] ip4withSubnet = Arrays.copyOf(getIPv4addr(), getIPv4addr().length);
for (int i = 0; i < getSubnetMasks().length && i < ip4withSubnet.lenght; i++) {
    ip4withSubnet[i] += "/" + getSubnetMasks()[i];
}
```
Then just use `Arrays.toString()` on the `ip4withSubnet` variable.  Do the same for ipv6/prefixlength.

There may be other ways to do it, if you think you can improve on the above!

dbwiddis:
Actually, looking at the source code, the "getIPv4" already does the `copyOf()`. So my above suggestion includes a lot of extra copying.  No need for the initial copyOf() if using the getter for ipv4.   Also, repeatedly calling `getPrefixLengths()` invokes that copy every iteration, so better to use `this.subnetmasks`.  You could also just create an empty array and iterate both `this.ipv4` and `this.subnetMasks` simultaneously to fill it.

dbwiddis:
To summarize, this is probably the best way to generate the ipv4/netmask array:
```
        String[] ipv4 = getIPv4addr(); // does copyOf
        if (this.ipv4.length == this.subnetMasks.length) {
            for (int i = 0; i < this.subnetMasks.length; i++) {
                ipv4[i] += "/" + this.subnetMasks[i];
            }
        }
```

JapneetSingh5:
I think this should work now

oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
```diff
@@ -221,14 +222,27 @@ private static Properties queryVmMacAddrProps() {
...
+        String[] ipv4withmask = getIPv4addr(); 
+        if (this.ipv4.length == this.subnetMasks.length) {
+            for (int i = 0; i < this.subnetMasks.length; i++) {
+                ipv4withmask[i] += "/" + this.subnetMasks[i];
+            }
+        }
+
+        String[] ipv6withprefixlength = getIPv6addr(); 
+        if (this.ipv6.length == this.prefixLengths.length) {
+            for (int j = 0; j < this.prefixLengths.length; i++) {
```
dbwiddis, line 235: It's not compiling because you used `j` here (but still `i++` and `[i]` below).

JapneetSingh5:
Super sorry about that, fixed it.

dbwiddis:
Heh, no need to apologize, that's why we have testing! We all make those mistakes!

Speaking of testing, the tests are still failing, with a checkstyle issue. Trailing spaces on lines 226, 233, and 239. Not a big deal, but many of us use auto-formatters so it's nice to not have random whitespace line changes when we're making an unrelated change.  I'll go ahead and fix the formatting and then merge.  Thanks for your submission!

JapneetSingh5:
Thankyou sir for providing me this opportunity, I learnt a lot !

dbwiddis:
Glad to help! Now go forth and help another project! :)

PR #1262, opened 2020-06-12
Read macOS versions from properties file

hkbiet:
#  macOS 10.16 support #1260 

Changed code in parseCodeName() method of the MacOperatingSystem to read macOS version properties file and return the version name :)

Fixes # (1260)

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Ran mvn test and it ran with no errors.

- [x] mvn test


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 5.2.0 (in progress)
 ================
 * [#1247](https://github.com/oshi/oshi/pull/1247): Add Automatic-Module-Name to support JPMS - [@dbwiddis](https://github.com/dbwiddis).
-* Your contribution here 
+* [#1248](https://github.com/oshi/oshi/pull/1248): macOS versions from file - [@hkbiet](https://github.com/hkbiet).
```
dbwiddis, line 4: I like to keep the "Your contribution" part, so just insert your line here above it.

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -154,6 +157,18 @@ private String parseCodeName() {
                 return "Cheetah";
             default:
             }
+            */
+            final String OSHI_VERSIONS_PROPERTIES = "oshi.versions.properties";
```
dbwiddis, line 161: Declare this `public static` at the top of the file.  Also probably want `macos.versions`.

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -154,6 +157,18 @@ private String parseCodeName() {
...
+            Properties verProps = FileUtil.readPropertiesFromFilename(OSHI_VERSIONS_PROPERTIES);
+
+            String major = String.valueOf(this.major);
+            String minor = String.valueOf(this.minor);
+            String version_number = major+"."+minor;
```
dbwiddis, line 166: You don't need to do the `String.valueof()` before this, as string concatenation will convert to strings anyway.   
 Just use `this.major + "." + this.minor`.  (Also using the same name for a local variable as a class/instance variable isn't a good idea.).   Also, checkstyle is complaining that you didn't leave whitespace before and after the `+`.

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -74,6 +74,8 @@
     private final int major;
     private final int minor;
 
+    public static String MACOS_VERSIONS_PROPERTIES;
```
dbwiddis, line 77: Also need a `final` modifier here, and the full string declaration `= "oshi.macos.versions.properties";`

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -119,52 +121,11 @@ public FamilyVersionInfo queryFamilyVersionInfo() {
...
-            }
-            */
-            final String OSHI_VERSIONS_PROPERTIES = "oshi.versions.properties";
-            Properties verProps = FileUtil.readPropertiesFromFilename(OSHI_VERSIONS_PROPERTIES);
-
-            String major = String.valueOf(this.major);
-            String minor = String.valueOf(this.minor);
-            String version_number = major+"."+minor;
             
+            MACOS_VERSIONS_PROPERTIES = "macos.versions.properties";
```
dbwiddis, line 125: Sorry I wasn't more clear here.  Need to keep the `oshi` to prevent conflicting with any other possible properties files on the class path.   But define the string `final` up above instead of changing it here.

hkbiet: Thank you so much for being with at each step @dbwiddis , i am on it to make the changes.

dbwiddis:
This looks great!  The CI errors are from checkstyle whitespace issues, I'll go ahead and fix those and then merge it!  Thanks for your contribution.

hkbiet:
> This looks great! The CI errors are from checkstyle whitespace issues, I'll go ahead and fix those and then merge it! Thanks for your contribution.

I am really thankful to you @dbwiddis for this opportunity , i felt really happy and learnt how complex projects with many many lines of code can be managed and worked upon by different people... it was awesome and also learnt a few things about git too .... thanks a lot ! Have a great day !

PR #1282, opened 2020-06-27
AIX port

### LINKED ISSUE #1166: AIX UNIX ppc64 7 or higher support
is there a plan to support the aix unix ppc/ppc64 platform?

dbwiddis:
Fixes #1166

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
@@ -35,9 +35,7 @@ import static oshi.util.Memoizer.memoize;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.function.Supplier;
 
 import oshi.annotation.concurrent.ThreadSafe;
@@ -221,25 +219,24 @@ public class AixOSProcess extends AbstractOSProcess {
         // ps -m -o THREAD -p 12345
         // BND field for PID is either a dash (all processors) or the processor it's
         // bound to, do 1L << # to get mask
-        long affinityMask = 0L;
+        long mask = 0L;
         List<String> processAffinityInfoList = ExecutingCommand.runNative("ps -m -o THREAD -p " + getProcessID());
-        if (processAffinityInfoList.size() > 2) { //what happens when the process has not thread?
-            processAffinityInfoList.remove(0); //remove header row
-            processAffinityInfoList.remove(1); //remove process row
-            for (String processAffinityInfo : processAffinityInfoList) { //affinity information is in thread row
+        if (processAffinityInfoList.size() > 2) { // what happens when the process has not thread?
+            processAffinityInfoList.remove(0); // remove header row
+            processAffinityInfoList.remove(1); // remove process row
+            for (String processAffinityInfo : processAffinityInfoList) { // affinity information is in thread row
                 String[] threadInfoSplit = ParseUtil.whitespaces.split(processAffinityInfo.trim());
-                if (threadInfoSplit.length > 13 && threadInfoSplit[4].charAt(0) != 'Z') { //only non-zombie threads
-                    if (threadInfoSplit[11].trim().charAt(0) == '-') { //affinity to all processors
+                if (threadInfoSplit.length > 13 && threadInfoSplit[4].charAt(0) != 'Z') { // only non-zombie threads
+                    if (threadInfoSplit[11].trim().charAt(0) == '-') { // affinity to all processors
                         return this.affinityMask.get();
                     } else {
                         int affinity = ParseUtil.parseIntOrDefault(threadInfoSplit[11], 0);
-                        affinityMask |= 1L << affinity;
-                        return affinityMask;
+                        mask |= 1L << affinity;
                     }
                 }
             }
         }
-        return affinityMask;
+        return mask;
     }
 
     @Override
@@ -352,6 +349,7 @@ public class AixOSProcess extends AbstractOSProcess {
 
     /**
      * Returns affinity mask from the number of CPU in the OS.
+     *
      * @return affinity mask
      */
     private static long getAffinityMaskFromCpuCount() {
--- a/oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
@@ -40,7 +40,9 @@ import java.util.Map;
 import java.util.function.Supplier;
 
 import oshi.annotation.concurrent.ThreadSafe;
+import oshi.driver.unix.aix.perfstat.PerfstatCpu;
 import oshi.software.common.AbstractOSProcess;
+import oshi.software.os.OSProcess;
 import oshi.software.os.OSThread;
 import oshi.util.ExecutingCommand;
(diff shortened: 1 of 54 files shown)


tausiflife:
Added few things for process. Haven't been able to test it yet.
An observation - 
For process count, i think i will just query `ps` again to list process and count them. But if we do so as we have done for other OS, this might give inconsistent results. Lets says if in between the calls to `getProcesses` and `getProcessCount` or vice-versa, some process get killed, the output will be inconsistent.

oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
```diff
@@ -349,17 +356,19 @@ static State getStateFromOutput(char stateValue) {
...
             break;
-        case 'S':
-            state = SLEEPING;
+        case 'I':
+            state = NEW;
             break;
-        case 'R':
         case 'W':
-            state = WAITING;
+            state = WAITING; //need confirmation on this
```
dbwiddis, line 368: `man ps` has a few more letters here.  I'd bring back the `R` for running and `S` for sleeping.  I think "swapped" might be sleeping rather than waiting?

            For processes:
              O                   Nonexistent
              A                   Active
              W                   Swapped
              I                   Idle (waiting for startup)
              Z                   Canceled
              T                   Stopped
            For kernel threads:
              O                   Nonexistent
              R                   Running
              S                   Sleeping
              W                   Swapped
              Z                   Canceled
              T                   Stopped

oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
```diff
@@ -349,17 +356,19 @@ static State getStateFromOutput(char stateValue) {
...
             break;
         case 'Z':
-            state = ZOMBIE;
+            state = CANCELED;
```
dbwiddis, line 371: I think this might be the same thing as Zombie.  [From this page](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.2.0/com.ibm.zos.v2r2.bpxa500/ps.htm):
> Z
> Canceled and parent has not performed wait (Z for zombie).

oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
tausiflife: Yeah i saw those but haven't added states relevant to thread. Will add those when i work on thread details.

oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
```diff
@@ -349,17 +356,19 @@ static State getStateFromOutput(char stateValue) {
...
         case 'O':
+            state = INVALID;
+            break;
+        case 'A':
             state = RUNNING;
             break;
-        case 'S':
-            state = SLEEPING;
+        case 'I':
+            state = NEW;
```
dbwiddis, line 365: In other OS's, I've not separately mapped New.  I think this happens if `fork()` is in the middle of operation, so it's probably reasonable to toss this in with waiting.

dbwiddis: Or if you keep the I, call it Idle.  But I prefer Waiting :)

oshi-core/src/main/java/oshi/software/os/unix/aix/AixOperatingSystem.java
```diff
@@ -99,18 +99,16 @@ public InternetProtocolStats getInternetProtocolStats() {
 
     @Override
     public List<OSProcess> getProcesses(int limit, ProcessSort sort) {
-        // Needs updating for flags on AIX or to use /proc fs
         List<OSProcess> procs = getProcessListFromPS(
-                "ps -eo s,pid,ppid,user,uid,group,gid,nlwp,pri,vsz,rss,etime,time,comm,args", -1);
+                "ps -e -o st,pid,ppid,user,uid,group,gid,thcount,pri,vsize,rssize,etime,time,args,comm,pagein", -1);
```
dbwiddis, line 103: `args` needs to be the last thing on the list here.  This is the case where you have to bound the `split()` by the exact number so the `args` (which may contain spaces) doesn't get split up.

oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
tausiflife: ok i think i need to read both [man ps](https://www.ibm.com/support/knowledgecenter/ssw_aix_72/p_commands/ps.html) and [ps returns](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.2.0/com.ibm.zos.v2r2.bpxa500/ps.htm). Thanks for pointing out

oshi-core/src/main/java/oshi/software/os/unix/aix/AixOperatingSystem.java
tausiflife: hmm makes sense.

dbwiddis:
Looking good.  I don't see a problem with process count being different between queries, a user should expect that.  I think I prefer counting numeric filenames in proc (same code as in Linux, and in fact it's working as-is) to a command line ps call.

tausiflife:
ohh and btw you don't have to `source .bashrc` again. its fixed :)

dbwiddis:
> ohh and btw you don't have to `source .bashrc` again. its fixed :)

Thank you! I was just about to ask what you did and then wondered, did I ever chmod it to be executable? Was that the problem?

tausiflife:
> > ohh and btw you don't have to `source .bashrc` again. its fixed :)
> 
> Thank you! I was just about to ask what you did and then wondered, did I ever chmod it to be executable? Was that the problem?

i don't think executable rights was the problem. What i found was that it never ran the `.bashrc` file. I just created a `.bash_profile` and added the code to source `.bashrc`. You could remove the code from `.bashrc` and add it to `.bash_profile`

tausiflife:
As i now see the thread details in `ps` doesn't have values. I guess it just inherits them from the parent process. 
image

Do you know any other way of finding thread details?

oshi-core/src/main/java/oshi/software/os/unix/aix/AixOperatingSystem.java
```diff
@@ -0,0 +1,268 @@
...
+        }
+        // Get installed services from /etc/rc.d/init.d
+        File dir = new File("/etc/rc.d/init.d");
+        File[] listFiles;
+        if (dir.exists() && dir.isDirectory() && (listFiles = dir.listFiles()) != null) {
+            for (File file : listFiles) {
+                String installedService = ExecutingCommand.getFirstAnswer(file.getAbsolutePath() + " status");
+                // Apache httpd daemon is running with PID 3997858.
+                if (installedService.contains("running")) {
+                    Matcher m = ParseUtil.AIX_RUNNING_SERVICE_INFO.matcher(installedService);
```
dbwiddis, line 256: This can be greatly simplified with `ParseUtil.parseLastInt()` to grab the PID.

tausiflife: done.

dbwiddis: You might also want to delete the regexp constant from the `ParseUtil`.

Issue #1264, opened 2020-06-14
Cognitive Complexity Issues


dbwiddis:
@dbwiddis I may have looked into many issues but have not contributed any. But if you have anything else that I can work on that would be great.

_Originally posted by @berdesourabh in https://github.com/oshi/oshi/issues/1263#issuecomment-643795094_

dbwiddis:
For those who are looking to contribute, it would be helpful continuing a long-term effort to refactor native methods into the `oshi.driver` package structure.  Basically, we'd like to move all native calls into a separate utility method.  We can also move common parsing or lookup functions into an appropriate utility method as well.

The best place to start are methods with high cognitive complexity already.  These can be found in the [SonarQube issues](https://sonarcloud.io/project/issues?id=com.github.oshi%3Aoshi-parent&resolved=false) and also the [Scrutinizer low quality functions](https://scrutinizer-ci.com/g/oshi/oshi/code-structure/master?elementType=java-function).

Any existing `driver` class is a good template to use (we generally want `final` classes with a `private` constructor, and `public static` methods).

Feel free to open a new issue for any one of these tasks you want to take on, so we can keep the various conversation threads on-topic!

dbwiddis:
Here's a good specific list of improvements that can be made!
- [ ] [LinuxOperatingSystem.filenameToFamily()](https://scrutinizer-ci.com/g/oshi/oshi/code-structure/master/java-function/oshi.software.os.linux.LinuxOperatingSystem.filenameToFamily%28String%29) - move these cases to a properties file, similar to what was done in #1262 
- [ ] [WindowsOperatingSystem.parseVersion()](https://scrutinizer-ci.com/g/oshi/oshi/code-structure/master/java-function/oshi.software.os.windows.WindowsOperatingSystem.parseVersion%28WmiResult,int,String%29) - move this logic parsing major/minor to a properties file, similar to what was done in #1262 
[...]

PR #1284, opened 2020-07-02
Moved cases from function to properties

akdanwlf:
Hello @dbwiddis 

I am trying to address the issue you pointed here ([#1264](https://github.com/oshi/oshi/issues/1264#issuecomment-646229626))

moved cases from  LinuxOperatingSystem.filenameToFamily() to oshi.linux.filename.properties

I will update the changelog file if there is nothing more to change here.

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
@@ -167,7 +167,7 @@ public class LinuxOperatingSystem extends AbstractOperatingSystem {
 
         // now for each file (with digit name) get process info
         for (File pidFile : pids) {
-                int pid = ParseUtil.parseIntOrDefault(pidFile.getName(), 0);
+            int pid = ParseUtil.parseIntOrDefault(pidFile.getName(), 0);
             OSProcess proc = new LinuxOSProcess(pid);
             if (!proc.getState().equals(State.INVALID)) {
                 procs.add(proc);
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
@@ -538,7 +538,7 @@ public class LinuxOperatingSystem extends AbstractOperatingSystem {
 
         if (name.isEmpty()) {
             return "Solaris";
-        } else if ("issue".equals(name.toLowerCase())) {
+        } else if ("issue".equalsIgnoreCase(name)) {
             // /etc/issue will end up here
             return "Unknown";
         } else {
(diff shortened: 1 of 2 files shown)


oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -533,64 +535,12 @@ protected static String getReleaseFilename() {
      * @return Mixed case family
      */
     private static String filenameToFamily(String name) {
-        switch (name.toLowerCase()) {
-        // Handle known special cases
-        case "":
+        Properties filenameProps = FileUtil.readPropertiesFromFilename(FILENAME_PROPERTIES);
```
dbwiddis, line 538: If the name is empty we can skip reading the properties, so this can be moved down into the "else" block.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -533,64 +535,12 @@ protected static String getReleaseFilename() {
...
+        if (name.equals("")) {
```
dbwiddis, line 539: Better to use `name.isEmpty()` or even `Util.isBlank(name)` here.

oshi-core/src/main/resources/oshi.linux.filename.properties
```diff
@@ -0,0 +1,52 @@
...
+e-smith=SME Server
+eos=FreeEOS
+hlfs=HLFS
+lfs=Linux-From-Scratch
+linuxppc=Linux-PPC
+meego=MeeGo
+mandakelinux=Mandrake
+mklinux=MkLinux
+nld=Novell Linux Desktop
+#case "novell":
```
dbwiddis, line 38: In the switch/case, this fell through to the lower one, so this should probably read `novell=SUSE Linux`.

oshi-core/src/main/resources/oshi.linux.filename.properties
```diff
@@ -0,0 +1,52 @@
...
+sun=Sun JDS
+synoinfo=Synology
+tinysofa=Tiny Sofa
+turbolinux=TurboLinux
+ultrapenguin=UltraPenguin
+va=VA-Linux
+vmware=VMWareESX
+yellowdog=Yellow Dog
+issue=Unknown
+#/etc/issue will end up here:
```
dbwiddis, line 52: this comment can be removed. Or, maybe this "issue" case should be handled in the code.

akdanwlf:
@dbwiddis 

Thanks for the quick review. I appreciate you taking the time to explain the code changes. 

I will continue to work the rest of the things mentioned ([#1264](https://github.com/oshi/oshi/issues/1264#issuecomment-646229626)). This is my first open-source contribution, I'm glad started with oshi.

dbwiddis:
Looks great!  Merging this!  Thanks for your contribution!

PR #1287, opened 2020-07-06
Fallback to read properties file from classloader of the class

### LINKED ISSUE #1285: Warning log when oshi.properties not found
Hi, 

Thanks for the useful library
We have a warning log `No default configuration found` when loading oshi where the `Thread.currentThread().getContextClassLoader()` is not the classloader that loads the oshi jar 

https://github.com/oshi/oshi/blob/1e56bf1be27f44fdedcfb98dc65f95443cccb5da/oshi-core/src/main/java/oshi/util/FileUtil.java#L226

I have a proposal that if the resources is empty from the given classloader, we could try the ClassLoader of GlobalConfig.class.getClassLoader()

do you think it make sense?
I can create the pull request if it looks good to you

ngyukman:
resolve #1285

### DIFF
--- a/oshi-core/src/main/java/oshi/util/FileUtil.java
+++ b/oshi-core/src/main/java/oshi/util/FileUtil.java
@@ -37,9 +37,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
-import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


oshi-core/src/main/java/oshi/util/FileUtil.java
```diff
@@ -217,40 +221,56 @@ public static String getStringFromFile(String filename) {
...
-        // Load the configuration file from the classpath
-        try {
-            ClassLoader loader = Thread.currentThread().getContextClassLoader();
-            if (loader == null) {
-                loader = ClassLoader.getSystemClassLoader();
-                if (loader == null) {
-                    throw new IOException();
-                }
+        // Load the configuration file from the different classloaders
+        Set<ClassLoader> loaders = Stream.of(
```
dbwiddis, line 234: This whole sequence seems overly complicated for what's essentially three if/then checks.  

Why not something like:
```
    if (readPropertiesFromClassLoader(propsFilename, archProps,
            Thread.currentThread().getContextClassLoader())
    || readPropertiesFromClassLoader(propsFilename, archProps,
            ClassLoader.getSystemClassLoader())
    || readPropertiesFromClassLoader(propsFilename, archProps,
            FileUtil.class.getClassLoader())) {
        return archProps;
    }
```

ngyukman: when I was testing, Thread.currentThread().getContextClassLoader() == ClassLoader.getSystemClassLoader() but perhaps it's early optimization for that

dbwiddis: If you'd prefer to reorder them (FileUtil class 2nd and System 3rd?) that's fine with me.

ngyukman: In most of the application I think all 3 classloaders are the same, so it shouldn't matter too much
`FileUtil.class.getClassLoader()` should always gets the default properties as it is the classloader loading the oshi-core.jar unless the jar is being altered / shaded, keeping it as the last one seems better for others need to override the properties

dbwiddis: OK, great!  I'll go ahead and merge this!

dbwiddis:
Heh. I just realized you committed this to 4.X.  Pulling it over to the main branch too.

oshi-core/src/main/java/oshi/util/FileUtil.java
dbwiddis: Almost a year later, I see why you did this and wish I'd left your initial contribution alone. :)

PR #1288, opened 2020-07-06
Moved windows version parsing logic to a properties file

akdanwlf:
Hello Daniel(@dbwiddis),

I moved the windows version names from WindowsOperatingSystem.parseVersion to the properties file, oshi.windows.versions.properties. 

Please review my changes and let me know if anything needs to updated or even a changelog updation is required.

Thanks.

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
@@ -158,7 +158,6 @@ public class WindowsOperatingSystem extends AbstractOperatingSystem {
         // http://msdn.microsoft.com/en-us/library/windows/desktop/ms724833%28v=vs.85%29.aspx
         boolean ntWorkstation = WmiUtil.getUint32(versionInfo, OSVersionProperty.PRODUCTTYPE,
                 0) == WinNT.VER_NT_WORKSTATION;
-        boolean fivePtTwo = major == 5 && minor == 2;
 
         StringBuilder verLookup = new StringBuilder(major).append(".").append(minor);
 
@@ -166,7 +165,7 @@ public class WindowsOperatingSystem extends AbstractOperatingSystem {
             verLookup.append(".nt");
         } else if (major == 10 && ParseUtil.parseLongOrDefault(buildNumber, 0L) > 17762) {
             verLookup.append(".17763+");
-        } else if (fivePtTwo) {
+        } else if (major == 5 && minor == 2) {
             if (ntWorkstation && getBitness() == 64) {
                 verLookup.append(".nt.x64");
             } else if ((suiteMask & 0x00008000) != 0) { // VER_SUITE_WH_SERVER
--- a/oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
@@ -144,7 +144,7 @@ public class WindowsOperatingSystem extends AbstractOperatingSystem {
         return new FamilyVersionInfo("Windows", new OSVersionInfo(version, codeName, buildNumber));
     }
 
-    private static String parseVersion(WmiResult<OSVersionProperty> versionInfo, int suiteMask, String buildNumber) {
+    private String parseVersion(WmiResult<OSVersionProperty> versionInfo, int suiteMask, String buildNumber) {
         // Initialize a default, sane value
         String version = System.getProperty("os.version");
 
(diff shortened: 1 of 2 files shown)


oshi-core/src/main/resources/oshi.windows.versions.properties
```diff
@@ -0,0 +1,44 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+
+#Mapping Windows version names to major, minor and build values
+
+10.0.true=10
+10.0.false.true=Server 2019
```
dbwiddis, line 28: I'd prefer if these names were self-documenting here, rather than undocumented booleans.  Something like:
```
10.0.nt
10.0.17762
10.0
6.3.nt
6.3
6.2.nt
6.2
```
... and so on

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
```diff
@@ -155,47 +158,31 @@ private static String parseVersion(WmiResult<OSVersionProperty> versionInfo, int
...
-            if (minor == 3) {
-                version = ntWorkstation ? "8.1" : "Server 2012 R2";
-            } else if (minor == 2) {
-                version = ntWorkstation ? "8" : "Server 2012";
-            } else if (minor == 1) {
-                version = ntWorkstation ? "7" : "Server 2008 R2";
-            } else if (minor == 0) {
-                version = ntWorkstation ? "Vista" : "Server 2008";
+
+        if (major == 10 && minor == 0) {
```
dbwiddis, line 162: There's still too many if/then choices in this file.  You should be able to just use a StringBuilder to build major.minor, then conditioned on `VER_NT_WORKSTATION` add on `.nt` (or the home server conditoned on the suitemask).  Then you only have to call the `getProperty` once.

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
```diff
@@ -155,49 +158,22 @@ private static String parseVersion(WmiResult<OSVersionProperty> versionInfo, int
...
-            } else if (minor == 1) {
-                version = "XP"; // 32 bits
-            } else if (minor == 0) {
-                version = "2000";
-            }
-            break;
-        default:
-            break;
+
+        StringBuilder verLookup = new StringBuilder(major + "." + minor);
```
dbwiddis, line 162: While simple string appending is fine (it simplifies to stringbuilder when compiled) it seems odd to wrap it inside a stringbuilder initiation.  I'd suggest `new StringBuilder(major).append('.').append(minor);`

Except I see later that we need this version number for a `ntWorkstation` check, so assign the majorMinor as a temporary variable, and then re-use it later...

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
```diff
@@ -155,49 +158,22 @@ private static String parseVersion(WmiResult<OSVersionProperty> versionInfo, int
...
+
+        if ((suiteMask & 0x00008000) != 0) { // VER_SUITE_WH_SERVER
+            verLookup.append(".HS");
```
dbwiddis, line 165: Putting this first in an if/else means the other ones won't be checked. I am not sure if this will prevent you from testing `ntWorkstation` on 6.x or 10.x.  Probably best to specifically check for `5.2` on this,..

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
```diff
@@ -155,49 +158,22 @@ private static String parseVersion(WmiResult<OSVersionProperty> versionInfo, int
...
+        } else if (ntWorkstation && !"5.1".equals(verLookup.toString()) && !"5.0".equals(verLookup.toString())) {
```
dbwiddis, line 166: I think using `String.compareTo()` might be better here.   Something like:
`if (ntWorkstation && majorMinor.compareTo("5.2") >= 0)`
(I'm just guessing, verify that works!)

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
```diff
@@ -155,49 +158,22 @@ private static String parseVersion(WmiResult<OSVersionProperty> versionInfo, int
...
+            verLookup.append(".nt");
+        } else if (ParseUtil.parseLongOrDefault(buildNumber, 0L) > 17762) {
+            verLookup.append(".17762");
```
dbwiddis, line 169: So I know I suggested this value but I think a config of `.17763+` might be more clear that it's build number 17763 or later.

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
dbwiddis: Actually, that's not even needed. The JNA project (a dependency and accessible from our code) has a[ `VersionHelpers` class](http://java-native-access.github.io/jna/5.5.0/javadoc/com/sun/jna/platform/win32/VersionHelpers.html) (that I wrote) where you can just check `IsWindowsVersionOrGreater(int wMajorVersion, int wMinorVersion, int wServicePackMajor)`.  Check out [`the source code`](https://github.com/java-native-access/jna/blob/cc1acdac02e4d0dda93ba01bbe3a3435b8933dab/contrib/platform/src/com/sun/jna/platform/win32/VersionHelpers.java).  There's also a `IsWindowsServer()` method there.

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
akdanwlf: I added a check before checking for suiteMask

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
```diff
@@ -159,14 +159,14 @@ private static String parseVersion(WmiResult<OSVersionProperty> versionInfo, int
...
-        if ((suiteMask & 0x00008000) != 0) { // VER_SUITE_WH_SERVER
+        if (!IS_VISTA_OR_GREATER && (suiteMask & 0x00008000) != 0) { // VER_SUITE_WH_SERVER
             verLookup.append(".HS");
-        } else if (ntWorkstation && !"5.1".equals(verLookup.toString()) && !"5.0".equals(verLookup.toString())) {
+        } else if (ntWorkstation && VersionHelpers.IsWindowsVersionOrGreater(5,2,0)) {
             verLookup.append(".nt");
         } else if (ParseUtil.parseLongOrDefault(buildNumber, 0L) > 17762) {
-            verLookup.append(".17762");
+            verLookup.append(".17763+");
         } else if (User32.INSTANCE.GetSystemMetrics(WinUser.SM_SERVERR2) == 0) {
```
akdanwlf, line 170: updated the version name as suggested

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
```diff
@@ -159,14 +159,14 @@ private static String parseVersion(WmiResult<OSVersionProperty> versionInfo, int
...
 
-        StringBuilder verLookup = new StringBuilder(major + "." + minor);
+        StringBuilder verLookup = new StringBuilder(major).append(".").append(minor);
 
-        if ((suiteMask & 0x00008000) != 0) { // VER_SUITE_WH_SERVER
+        if (!IS_VISTA_OR_GREATER && (suiteMask & 0x00008000) != 0) { // VER_SUITE_WH_SERVER
             verLookup.append(".HS");
-        } else if (ntWorkstation && !"5.1".equals(verLookup.toString()) && !"5.0".equals(verLookup.toString())) {
+        } else if (ntWorkstation && VersionHelpers.IsWindowsVersionOrGreater(5,2,0)) {
             verLookup.append(".nt");
```
akdanwlf, line 167: I did not know if servicepackmajor will take 0 but we want to check only if it's greater than 5.2

dbwiddis:
OK, this looks great!  I just bounced off the [official list](https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-osversioninfoexa) and see we missed XP Pro x64 in the earlier logic, so I added that and tweaked your logic a bit.  I think this is good to merge!

PR #1296, opened 2020-07-14
Add checking if cpu has boost enabled (only for linux currently)

Szwendacz99:
Sorry i am new in contributing, but i saw in maven test that it shows failure if detect cpu frequency higher than maxFrequency and it happens cause cpu can have boost enabled which allow cpu to have bigger Frequency than detected max, so i tried to make boost detecting feature (partially implemented, dont have other systems to make it work on them)

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -81,12 +81,14 @@ public interface CentralProcessor {
     long[] getCurrentFreq();
 
     /**
+     * checks if cpu have boost mode enabled
+     * if cpu don't support or it is not enabled then returns false,
+     * otherwise returns true
      *
-     *
-     *
-     *
+     *@return state of boost mode
      */
-    Boolean boostEnabled();
+    boolean boostEnabled();
+
     /**
      * Returns an {@code UnmodifiableList} of the CPU's logical processors. The list
      * will be sorted in order of increasing NUMA node number, and then processor
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -81,6 +81,13 @@ public interface CentralProcessor {
     long[] getCurrentFreq();
 
     /**
+     *
+     *
+     *
+     *
+     */
+    Boolean boostEnabled();
+    /**
      * Returns an {@code UnmodifiableList} of the CPU's logical processors. The list
      * will be sorted in order of increasing NUMA node number, and then processor
      * number. This order is consistent with other methods providing per-processor
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractCentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractCentralProcessor.java
@@ -49,6 +49,7 @@ public abstract class AbstractCentralProcessor implements CentralProcessor {
 
     private final Supplier<ProcessorIdentifier> cpuid = memoize(this::queryProcessorId);
     private final Supplier<Long> maxFreq = memoize(this::queryMaxFreq);
+    private final Supplier<Boolean> boost = memoize(this::queryBoostEnabled);
     private final Supplier<long[]> currentFreq = memoize(this::queryCurrentFreq, defaultExpiration());
     private final Supplier<Long> contextSwitches = memoize(this::queryContextSwitches, defaultExpiration());
     private final Supplier<Long> interrupts = memoize(this::queryInterrupts, defaultExpiration());
@@ -114,6 +115,17 @@ public abstract class AbstractCentralProcessor implements CentralProcessor {
      * @return The max frequency.
      */
     protected abstract long queryMaxFreq();
+
+    public Boolean boostEnabled() {
+        return boost.get();
+    }
+
+    /**
+     * Check if processor have boost mode enabled.
+     *
+     * @return cpu boost state.
+     */
+    protected abstract Boolean queryBoostEnabled();
 
     @Override
     public long[] getCurrentFreq() {
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractCentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractCentralProcessor.java
@@ -49,7 +49,7 @@ public abstract class AbstractCentralProcessor implements CentralProcessor {
 
     private final Supplier<ProcessorIdentifier> cpuid = memoize(this::queryProcessorId);
     private final Supplier<Long> maxFreq = memoize(this::queryMaxFreq);
-    private final Supplier<Boolean> boost = memoize(this::queryBoostEnabled);
+    private final boolean boost = queryBoostEnabled();
     private final Supplier<long[]> currentFreq = memoize(this::queryCurrentFreq, defaultExpiration());
     private final Supplier<Long> contextSwitches = memoize(this::queryContextSwitches, defaultExpiration());
     private final Supplier<Long> interrupts = memoize(this::queryInterrupts, defaultExpiration());
@@ -116,8 +116,8 @@ public abstract class AbstractCentralProcessor implements CentralProcessor {
      */
     protected abstract long queryMaxFreq();
 
-    public Boolean boostEnabled() {
-        return boost.get();
+    public boolean boostEnabled() {
+        return boost;
     }
 
     /**
@@ -125,7 +125,7 @@ public abstract class AbstractCentralProcessor implements CentralProcessor {
      *
      * @return cpu boost state.
      */
-    protected abstract Boolean queryBoostEnabled();
+    protected abstract boolean queryBoostEnabled();
 
     @Override
     public long[] getCurrentFreq() {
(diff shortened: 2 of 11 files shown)


dbwiddis:
Thanks, this is a great suggestion!  Some quick notes:
 - I'm about to cut a new release today so if you want this in a new release you better be quick. Otherwise, we can take our time on this feature, which I think I'd prefer.
 - Return type should be primitive `boolean` not the boxed type.
 - Need the top level API Javadoc completed. It's okay to say "only on Linux" for now.
 - Rather than implement separately for each other OS, just put a stub in the `Abstract` class (and return `false` rather than null).
 - you can have the test case only apply if boost is false.

dbwiddis:
For windows, we already use `CallNTPowerInformation()` with the `ProcessorInformation` index to get a [PROCESSOR_POWER_INFORMATION](https://docs.microsoft.com/en-us/windows/win32/power/processor-power-information-str) structure.  We currently just use `MaxMhz` from that but I'm guessing the `MhzLimit` from that same structure might give us the information we need.

Szwendacz99:
It is really not necessary for me to add these changes in today's update, but i just commited changes based on what You said, hopefully all of them and properly :]

dbwiddis:
Just tested my own windows machine.  Processor is Intel Core i7-1065G7.  [Docs say it's got a base 1.30GHz, max 3.90Ghz](https://ark.intel.com/content/www/us/en/ark/products/196597/intel-core-i7-1065g7-processor-8m-cache-up-to-3-90-ghz.html
).

OSHI prints out vendor freq of 1.30GHz and max of 1498 MHz which matches the structure fetch in my previous comment.
```
PowrProf$ProcessorPowerInformation(allocated@0x5b0390 (24 bytes) (shared from allocated@0x5b0390 (192 bytes) (shared from allocated@0x5b0390 (192 bytes)))) {
  int number@0x0=0x0000
  int maxMhz@0x4=0x05DA
  int currentMhz@0x8=0x0512
  int mhzLimit@0xC=0x0508
  int maxIdleState@0x10=0x0002
  int currentIdleState@0x14=0x0002
}
```
The "max" is 0x05DA = 1498 MHz.  The mhzLimit 0x0508 is 1288 (1.3 GHz).  So it's a boost-capable, but the "max" vs. "vendor" from OSHI already answers that.

Seems a reasonable "default" for the "AbstractCentralProcessor" method would be a boolean test if Max > Vendor.  But then that test you noted would never fail on Windows.  So perhaps we're reading Max wrong on Linux?

dbwiddis:
What I meant to do is actually return something from AbstractCentralProcessor, which will apply unless you override the method in one of the platform classes.

But digging into this I'm wondering if this test is needed, as we already have a Max and Vendor frequency value and it seems that is supposed to answer the same question.

Which means if you're getting current > max on Linux, then there's a bug in Max on Linux and we should try to fix that...

Szwendacz99:
this is piece of ouput from SystemInfoTest:
------------------------------
Max Frequency: 3,4 GHz
Is boost enabled:true
Current Frequencies: 3,7 GHz, 3,1 GHz, 1,6 GHz, 1,6 GHz, 2,1 GHz, 2,5 GHz, 3,5 GHz, 2,8 GHz, 2,9 GHz, 3,4 GHz, 3,0 GHz, 3,0 GHz
------------------------------
my cpu is ryzen 5 2600 and its default max without boost is 3.4 as detected, but boost is actually enabled  and i got higher current frequencies

dbwiddis:
Does your "max" equal your "vendor"  If so, then the root of the problem is not detecting the max.  Currently OSHI is checking `/sys/devices/system/cpu/cpu/cpufreq/scaling_max_freq`.   If this isn't the right location on your system, we can/should fix this.

dbwiddis:
Hmm, [this post indicates](https://askubuntu.com/questions/353383/cpu-frequency-scaling-issue-scaling-max-freq-wrong) the "max" may change if running in an "on demand" mode.

So I think this is less a "new feature" as a "bug report" and figuring out how to correctly deduce a batter max value.

Szwendacz99:
in my system scaling_max_freq can be found in   /sys/devices/system/cpu/cpufreq/policy0/scaling_max_freq but it shows 3400000 
and there is no file in path /sys/devices/system/cpu/cpufreq/cpuinfo_max_freq

dbwiddis:
It looks [from this post](https://www.pantz.org/software/cpufreq/usingcpufreqonlinux.html) that we should be iterating `scaling_available_frequencies` and picking the max from that.

Szwendacz99:
in scaling_available_frequencies i found "3400000 2800000 1550000 "

dbwiddis:
But you can still get 3700000.  Very interesting.

So what is your actual max?  Anywhere else that data is?  Bios command line?

dbwiddis:
Here's a thought.  Iterate through all the possible policy subdirs... `/sys/devices/system/cpu/cpufreq/*/scaling_max_freq`. Do any of them have a higher value?

dbwiddis:
Reading [this documentation](https://www.kernel.org/doc/html/v4.14/admin-guide/pm/cpufreq.html) I am convinced we have a bug here, or at least an incompatibility with current kernels.  I think iterating over the `cpufreq/policy*/scaling_max_freq` files should work, based on those docs. Can you confirm at least one of your policy max's has 3700000?

dbwiddis:
(And I'll hold off on the release to try to fix this bug.)

Szwendacz99:
checked all of /sys/devices/system/cpu/cpufreq/policy*/scaling_max_freq , all of them have the same value (policy from 0 to 11)

dbwiddis:
OK one other place, please check `/sys/devices/system/cpu/cpu0/cpufreq/bios_limit` (one cpu is probably enough) or the same limit in all the policy* subdirs.

Szwendacz99:
in /sys/devices/system/cpu/cpu0/cpufreq/bios_limit also  3400000

dbwiddis:
For intel processors there is `/sys/devices/system/cpu/cpufreq/intel_pstate/`.  Is there a similar file for your AMD?

Szwendacz99:
in /sys/devices/system/cpu/
cpu0/            cpu11/           cpu4/            cpu7/            cpufreq/         isolated         modalias         possible         smt/
cpu1/            cpu2/            cpu5/            cpu8/            cpuidle/         kernel_max       offline          power/           uevent
cpu10/           cpu3/            cpu6/            cpu9/            hotplug/         microcode/       online           present          vulnerabilities/

and in /sys/devices/system/cpu/cpufreq/
boost      policy0/   policy1/   policy10/  policy11/  policy2/   policy3/   policy4/   policy5/   policy6/   policy7/   policy8/   policy9/   schedutil/

checket most or even all of them and did not find more than 3400000 or sth that shows bigger frequency

Szwendacz99:
even the kde plasma tool for system info shows 3,4 GHz max

dbwiddis:
One last command from googling: [cpupower frequency-info](https://community.chakralinux.org/t/amd-cpu-frequency-scheduler-issues-since-kernel-4-19/8244/2)

I'm at a loss here for how you can have a current exceeding what's obviously your system's "max".  Unless there is a set factor (e.g., 1.25x) for "boost mode" that we can always apply, but I'd need documentation for that.

Szwendacz99:
cpupower frequency-info shows:

analyzing CPU 0:
  driver: acpi-cpufreq
  CPUs which run at the same hardware frequency: 0
  CPUs which need to have their frequency coordinated by software: 0
  maximum transition latency:  Cannot determine or is not supported.
  hardware limits: 1.55 GHz - 3.40 GHz
  available frequency steps:  3.40 GHz, 2.80 GHz, 1.55 GHz
  available cpufreq governors: performance schedutil
  current policy: frequency should be within 1.55 GHz and 3.40 GHz.
                  The governor "schedutil" may decide which speed to use
                  within this range.
  current CPU frequency: 1.55 GHz (asserted by call to hardware)
  boost state support:
    Supported: yes
    Active: yes
    Boost States: 0
    Total States: 3
    Pstate-P0:  3400MHz
    Pstate-P1:  2800MHz
    Pstate-P2:  1550MHz

dbwiddis:
OK... so I think all the existing commands we are using assume intel processors, but you have a (superior) AMD processor.  The key part in your output there is the `acpi-cpufreq` driver.

Still researching but I think all the tools we are looking at aren't necessarily updated/patched to match acpi-cpufreq....

Szwendacz99:
lshw seems to do it right:
*-cpu
          description: CPU
          product: AMD Ryzen 5 2600 Six-Core Processor
          vendor: Advanced Micro Devices [AMD]
          physical id: 2c
          bus info: cpu@0
          version: AMD Ryzen 5 2600 Six-Core Processor
          serial: Unknown
          slot: AM4
          size: 2162MHz
          capacity: 3900MHz
          width: 64 bits
          clock: 100MHz

dbwiddis:
Aha!  Well, I suppose I can add a check for that.

dbwiddis:
Do you want me to take over this PR or just close it and add a new one that reads `lshw` (if installed) as a backup to raise the value if "capacity" > existing max?

Szwendacz99:
What suits you best

dbwiddis:
Probably easiest to close this and start fresh.  Thanks for the attempt at a PR and the bug report and detailed troubleshooting!  I'll ask you to test my fix.

BTW, any other commands besides `lshw` helpful?  Like `lscpu`?

Szwendacz99:
nope, lshw is the only one that worked for me

dbwiddis:
OK, can you check out #1298? Should do the trick.  I'm going to also tweak the existing code to iterate the `policy*` directories.

Szwendacz99:
Lshw.queryCpuCapacity() returns 3400000000 so sth still wrong

dbwiddis:
Yep.   I think under the hood there's a `cpuid` query that's impossible from within Java. :(

dbwiddis:
I guess I can always check and update max based on current.... then the test will never fail. :D

PR #1298, opened 2020-07-14
Use lshw as backup for max CPU frequency

### LINKED ISSUE #1297: Linux max cpu frequency assumes intel processors
The existing `cpufreq` implementation is intel `pstate` focused.  `lshw` can be used to check the max, as reported by @Szwendacz99

*-cpu
          description: CPU
          product: AMD Ryzen 5 2600 Six-Core Processor
          vendor: Advanced Micro Devices [AMD]
          physical id: 2c
          bus info: cpu@0
          version: AMD Ryzen 5 2600 Six-Core Processor
          serial: Unknown
          slot: AM4
          size: 2162MHz
          capacity: 3900MHz
          width: 64 bits
          clock: 100MHz

_Originally posted by @Szwendacz99 in https://github.com/oshi/oshi/pull/1296#issuecomment-658385205_

dbwiddis:
Fixes #1297


Szwendacz99:
Lshw.queryCpuCapacity() returns unproper so sth still wrong
probably because lshw returns good value only with root access

dbwiddis:
Well, I just set it to query current first, then max, and if any current is higher it'll use that.

Pretty sure that makes the test meaningless, but at least you should get a max equal to or greater than your max current.

Issue #1215, opened 2020-05-12
Add messages to unit test assertions


dbwiddis:
### Hacktoberfest

~I'm marking this issue for [first-timers-only](https://www.firsttimersonly.com/).  That means that I will only accept a PR for this one from someone who's never contributed to an open source project on GitHub before.~

I'm reopening this issue for #Hacktoberfest2020.  This is an easy PR for those of you who want T-shirts. :)

~The main goal of this task is to learn the process for submitting a PR -- the GitHub equivalent of writing a "Hello World!" program.  If you already know the PR workflow, please leave this issue for someone else!~

This one is easy (but don't make that statement make you feel bad if you have a hard time with it, there's more to contributing to open source than changing lines of code, especially if it's your first time). I'll hold your hand through this if you need me to. 😄 

### What you're changing

A key to making sure OSHI code is working properly is unit testing.  Every time we make a change to the code, these tests run to verify the output is as expected.  

While these tests point out failures, many times the log messages simply point to the line number of the error without providing more information.

The JUnit `Assert` class methods have an optional `message` parameter that can be included for more detailed logs.   OSHI has not used these very much in the past but would like to add them going forward.  While it's mostly an administrative chore fixing old ones, these are a perfect exercise for a first PR!

### What to do 

1. Find a test class under the `oshi-core` directory tree `src\test\java`.  There are many to choose from!
2. Look for methods like `assertEquals(foo, bar);`
3. Add a message parameter: `assertEquals("foo should equal bar", foo, bar);`

Repeat as much as you like.  Don't worry about trying to get them all (there are literally hundreds of them) but it would be nice to do a whole class at a time, or at least a whole method. I'll leave this Issue open for other first timers to revisit!

### How to submit your change to the repository

Here are the steps to get a PR merged here.
- Read the [CONTRIBUTING](https://github.com/oshi/oshi/blob/master/CONTRIBUTING.md) file for reference.  We'll follow those steps, and I'll walk you through anything you don't understand.
- Fork the project to your own GitHub account, make a local clone, and create a branch for your submission.
- Make the changes described above.
- Commit your changes, push to to your GitHub fork, and create a Pull Request
- Respond to the usually minor comments on code review
- Celebrate! :tada:

tomokos2:
Hi, I would like to work on this.

dbwiddis:
[...]

PR #1349, opened 2020-10-04
Added messages for unit test assertions

24apurv:
Added messages for unit test assertions for remaining classes in oshi-core/src/test/java/oshi/hardware/ 
as per #1215


24apurv:
I cant see why the CI check failed for Mac considering there were no actual changes to the code. Let me know if anything has to be changed. I cant find the issue here.

dbwiddis:
Thanks, this looks great! Merging.

Issue #1345, opened 2020-09-29
Resulting List<NetworkIF> does not include loopback and virtual interfaces


rnaval:
Is there a reason to ignore the loopback and virtual interfaces when calling `getNetworkIFs()`?

Based on https://github.com/oshi/oshi/blame/master/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java#L137 - this seems to be intentionally done. Why is that? Is there a way to configure oshi to include these?

dbwiddis:
As you can see ignoring them is hardcoded there is not "currently" a way to include them, but I'd welcome a PR to make that filtering configurable.   Probably could do the same thing for the types of filesystems which are ignored as well.

As for the "why"?  Probably trying to suppress noise, and because nobody has asked for it until now.

Issue #1339, opened 2020-09-21
does oshi support overlay filesystem types?


brettbandy22:
When I run oshi.SystemInfoTest I do not see my overlay filesystem in the list.

File System:
 File Descriptors: 195440/58657580
 /dev/mapper/3624a9370cb6d0b7215af4a2e00013c4d (Local Disk) [xfs] 26.2 GiB of 30.0 GiB free (87.5%), 15.7 M of 15.7 M files free (100.0%) is /dev/mapper/3624a9370cb6d0b7215af4a2e00013c4d ...
 /dev/sdb1 (Local Disk) [xfs] 354.5 GiB of 357.4 GiB free (99.2%), 187.5 M of 187.5 M files free (100.0%) is /dev/sdb1  ...
 /dev/sdb2 (Local Disk) [xfs] 662.4 GiB of 1.4 TiB free (46.3%), 143.5 M of 150.0 M files free (95.7%) is /dev/sdb2  ...
 /dev/sdb2 (Local Disk) [xfs] 662.4 GiB of 1.4 TiB free (46.3%), 143.5 M of 150.0 M files free (95.7%) is /dev/sdb2  ...
 /dev/sdb1 (Local Disk) [xfs] 354.5 GiB of 357.4 GiB free (99.2%), 187.5 M of 187.5 M files free (100.0%) is /dev/sdb1 ...

When I run mount I see the overlay mount point

overlay on / type overlay (rw,relatime,...)

This is being run docker container running Centos.

dbwiddis:
Issue #438 may be relevant.   That format for `mount` seems unusual.  Can you post the actual output?

brettbandy22:
mount output:

[...]

PR #1350, opened 2020-10-04
Feature/get local interfaces

zalintyre:
Hello,

here is my pull request for #1345. Feel free to leave feedback :)

Franz

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,29 @@
+#
+# MIT License
+#
+# Copyright (c) 2010 - 2020 The OSHI Project Contributors: https://github.com/oshi/oshi/graphs/contributors
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+
+root = true
+
+[*.java]
+ij_java_class_count_to_use_import_on_demand = 1000
+ij_java_names_count_to_use_import_on_demand = 1000
--- a/.editorconfig
+++ b/.editorconfig
@@ -22,8 +22,24 @@
 # SOFTWARE.
 #
 
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
 root = true
 
-[*.java]
-ij_java_class_count_to_use_import_on_demand = 1000
-ij_java_names_count_to_use_import_on_demand = 1000
+# Global editorconfig settings
+[*]
+end_of_line = lf
+insert_final_newline = true # Insert a newline at the end of a file
+charset = utf-8 # Set default charset
+indent_style = space # Use spaces for indents
+indent_size = 4
+trim_trailing_whitespace = true # delete whitespace at the end of each line
+
+# 2 space indentation for yaml
+[*.{yml, yaml}]
+indent_size = 2
+
+# IntelliJ specific settings
+ij_java_class_count_to_use_import_on_demand = 1000 # Do not group imports with an asterisk
+ij_java_names_count_to_use_import_on_demand = 1000 # Do not group imports with an asterisk
--- a/oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
+++ b/oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
@@ -69,7 +69,7 @@ public interface HardwareAbstractionLayer {
 
     /**
      * Instantiates an {@code UnmodifiableList} of {@link oshi.hardware.HWDiskStore}
-     * objects, representing physical hard disks or other similar storage devices
+     * objects, representing physical hard disks or other similar storage devices.
      *
      * @return An {@code UnmodifiableList} of HWDiskStore objects or an empty list
      *         if none are present.
@@ -77,8 +77,8 @@ public interface HardwareAbstractionLayer {
     List<HWDiskStore> getDiskStores();
 
     /**
-     * Gets a list of non-local {@link NetworkIF} objects, representing a network interface
-     * The list excludes local interfaces.
+     * Gets a list of non-local {@link NetworkIF} objects, representing a network
+     * interface. The list excludes local interfaces.
      *
      * @return An {@code UnmodifiableList} of {@link NetworkIF} objects representing
      *         the interfaces
@@ -86,9 +86,11 @@ public interface HardwareAbstractionLayer {
     List<NetworkIF> getNetworkIFs();
 
     /**
-     * Gets a list {@link NetworkIF} objects, representing a network interface
+     * Gets a list {@link NetworkIF} objects, representing a network interface.
      *
-     * @param includeLocalInterfaces include local interfaces in the result
+     * @param includeLocalInterfaces
+     *            whether to include local interfaces (loopback or no hardware
+     *            address) in the result
      * @return An {@code UnmodifiableList} of {@link NetworkIF} objects representing
      *         the interfaces
      */
@@ -105,7 +107,7 @@ public interface HardwareAbstractionLayer {
 
     /**
      * Instantiates a {@link oshi.hardware.Sensors} object, representing CPU
-     * temperature and fan speed
+     * temperature and fan speed.
      *
      * @return A Sensors object
      */
--- a/oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
+++ b/oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
@@ -77,7 +77,7 @@ public interface HardwareAbstractionLayer {
     List<HWDiskStore> getDiskStores();
 
     /**
-     * Gets a list of inet {@link NetworkIF} objects, representing a network interface
+     * Gets a list of non-local {@link NetworkIF} objects, representing a network interface
      * The list excludes local interfaces.
      *
      * @return An {@code UnmodifiableList} of {@link NetworkIF} objects representing
--- a/oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
+++ b/oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
@@ -77,12 +77,21 @@ public interface HardwareAbstractionLayer {
     List<HWDiskStore> getDiskStores();
 
     /**
-     * Gets a list of {@link NetworkIF} objects, representing a network interface
+     * Gets a list of inet {@link NetworkIF} objects, representing a network interface
+     * The list excludes local interfaces.
      *
      * @return An {@code UnmodifiableList} of {@link NetworkIF} objects representing
      *         the interfaces
      */
     List<NetworkIF> getNetworkIFs();
+
+    /**
+     * Gets a list of all {@link NetworkIF} objects, representing a network interface
+     *
+     * @return An {@code UnmodifiableList} of {@link NetworkIF} objects representing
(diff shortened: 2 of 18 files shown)


zalintyre:
It seems like the tests are quite system-dependent. Here are the current failures from AppVeyor:

```
[ERROR] oshi.hardware.NetworksTest.testAllNetworkInterfaces  Time elapsed: 0.572 s  <<< FAILURE!
java.lang.AssertionError: NetworkIF MTU should not be negative
	at oshi.hardware.NetworksTest.testAllNetworkInterfaces(NetworksTest.java:123)
[ERROR] oshi.hardware.NetworksTest.testInetNetworkInterfaces  Time elapsed: 10.74 s  <<< FAILURE!
java.lang.AssertionError: NetworkIF MTU should not be negative
	at oshi.hardware.NetworksTest.testInetNetworkInterfaces(NetworksTest.java:71)
```

On my Linux development machine, I get a test failure because `OperatingSystemTest.testGetServices` doesn't report stopped services.

.editorconfig
```diff
@@ -0,0 +1,29 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+
```
dbwiddis, line 24: Don't commit your IDE settings files like this.  There should be an option to "not track" these.  If you'd like to add `.editorconfig` to the `.gitignore` file that's fine.

oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
```diff
@@ -77,13 +77,22 @@
     List<HWDiskStore> getDiskStores();
 
     /**
-     * Gets a list of {@link NetworkIF} objects, representing a network interface
+     * Gets a list of inet {@link NetworkIF} objects, representing a network interface
```
dbwiddis, line 80: I'd prefer not to add `inet` here as it's not clearly defined, and the object returned is part of the OSHI API.

I see throughout the changes you've used this term, though, so perhaps you need to just clearly define what "inet" means.  Or use a more verbose term that's more clear.

oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
```diff
@@ -77,13 +77,22 @@
...
      */
     List<NetworkIF> getNetworkIFs();
 
+    /**
+     * Gets a list of all {@link NetworkIF} objects, representing a network interface
+     *
+     * @return An {@code UnmodifiableList} of {@link NetworkIF} objects representing
+     *         the interfaces
+     */
+    List<NetworkIF> getAllNetworkIFs();
```
dbwiddis, line 94: Instead of a new name, let's just overload the original one.  Make this second method `getNetworkIFs(boolean includeLocalAndVirtual);`.   Calling the first method will simply redirect to `getNetworkIFs(false)` and then you only have to really write the implementations for one method.

oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
```diff
@@ -68,8 +69,7 @@
      * Construct a {@link NetworkIF} object backed by the specified
      * {@link NetworkInterface}.
      *
-     * @param netint
-     *            The core java {@link NetworkInterface} backing this object.
+     * @param netint The core java {@link NetworkInterface} backing this object.
```
dbwiddis, line 72: Not that big a deal, but we do use consistent formatting to avoid whitespace/format noise in the diffs.  If possible, I'd prefer you set up your IDE to use the same format settings.  You can find them [here](https://github.com/oshi/oshi/tree/master/oshi-dist/src/main/resources/eclipse) if you use Eclipse, and I think IntelliJ can import eclipse settings (basically the "Eclipse Default" with spaces instead of tabs.)

If you can't easily format, don't worry, I'll just run a source format before merging.

oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
```diff
@@ -121,31 +121,39 @@ protected AbstractNetworkIF(NetworkInterface netint) {
...
+     */
+    protected static List<NetworkInterface> getAllNetworkInterfaces() {
+        try {
+            Enumeration<NetworkInterface> interfaces = NetworkInterface.getNetworkInterfaces();
+            return Collections.list(interfaces);
+        } catch (SocketException ex) {
+            LOG.error("Socket exception when retrieving interfaces: {}", ex.getMessage());
+        }
+
+        return new ArrayList<>();
```
dbwiddis, line 154: Use `Collections.emptyList()` here.

oshi-core/src/test/java/oshi/hardware/NetworksTest.java
```diff
@@ -37,17 +37,70 @@
...
      * @throws IOException
      *             Signals that an I/O exception has occurred.
      */
     @Test
-    public void testNetworkInterfaces() throws IOException {
+    public void testAllNetworkInterfaces() throws IOException {
         SystemInfo si = new SystemInfo();
 
-        for (NetworkIF net : si.getHardware().getNetworkIFs()) {
+        for (NetworkIF net : si.getHardware().getAllNetworkIFs()) {
```
dbwiddis, line 103: Given one list is the superset of the other, it seems like we don't need to repeat all the tests, just do the "all" version for the detail and do at least one test for the "non-local/virtual" version to show the result isn't loopback, etc.

dbwiddis (changes requested):
Thanks for the PR! It looks great.  A few changes I'd like:
 - I'm not really happy with the term "inet" used in the javadocs.  I'd prefer a more verbose term to represent the "not loopback or virtual" nature.
 - Rather than a getX() and getAllX() pair, let's overload the methods. The new method will add a boolean argument, the old method will keep the no-arg version.  In the Abstract object you can implement the no-arg delegating to the arg version with `false`.  Then in the platform-specific code you only need to implement one method, with a boolean argument you can switch on.

dbwiddis:
Interesting that appveyor fails now but didn't before.  Even more interesting is that the failure, `getMTU()`, is in the Java class.  I wonder if it's negative or just a very large (unsigned) value being interpreted as negative.  Perhaps that test should be deleted.

.editorconfig
zalintyre: This is not really IDE specific configuration. You can use .editorconfig files to configure almost all IDEs: https://editorconfig.org/

You can match the config to your code style guidelines. So it's easier for contributors to comply to your code style - it's rather difficult to setup your IDE for multiple different projects.

I can update the editorconfig with some default / recommended values or remove it entirely, if you want.

oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
zalintyre: I did not overload the method on purpose, following "Clean Code"'s statement about boolean parameters: https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3

If you want, I can change it, though.

.editorconfig
dbwiddis: I'm ok with it as long as the settings are clearly documented with comments explaining them.

oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
dbwiddis: Thanks for that article link.  I did spend some time reading it and considering this issue, and present the following counter for discussion.

While the article makes great points, it's dealing primarily with mutually exclusive control flows.  That, however, is not the case here. We are using overlapping/nearly-identical flows (iterating interfaces), with the only difference being a filtering parameter.  We are not making the code any more complex than it already is; the existing code includes this conditional:
```
if (!netint.isLoopback() && netint.getHardwareAddress() != null) { .... }
```
All you'd do is add the boolean parameter (or its opposite) to that conditional.

Another consideration is API consistency.  We use exactly the pattern I'm proposing in iterating Filesystems (see `FileSystem.getFileStores(boolean localOnly)`).   Of course, that's now complicated by the fact that there's a request to not filter the overlay filesystem types ( #1339 ), so perhaps a boolean is a poor choice there, or we may need a second parameter (probably a collection of strings).   Also, the `OperatingSystem` class has a default `getProcesses()` which gets all processes, delegating to `getProcesses(int limit, ProcessSort sort)` for the filtering/sorting/size limiting output.  I see this as a very similar type of application.

More importantly, the API is designed to transparently work with the Jackson project's `ObjectMapper` (see the `Json` class in the `oshi-demo` project).  By default, this class automatically, via reflection, assumes a `get*()` method maps directly to a field.  Perhaps using `get*()` is a bad choice to begin with, but at this point we're stuck with the API. Having two no-arg `get*()` methods will duplicate the JSON output.  The mapper does ignore the "getters" with arguments, and would ignore a method name that doesn't look like a getter (which is the reason `queryNetworkInterfaces()` is not `getNetworkInterfaces()`.)  

So I'm not overly opposed to you using two separate methods if you make the "full method" a `query*()` one, and then filter from those results in the `get*()` method.  That'd be an implementation detail.   However, given the similarity to the existing API implementations, I'd prefer to just have a boolean filter parameter on whether to include the loopback interfaces.

.editorconfig
dbwiddis: After reading the website, I'm actually rather fond of this... also add 4-space indent, and if you want to peruse the eclipse settings in the `oshi-dist` project under the resources and copy equivalenet ones, that'd be helpful.  I note that file is compatible with Eclipse (with plugin) but I don't see it referencing IntelliJ IDEA?

zalintyre: IntelliJ doesn't need a plugin since quite a few versions. In addition to the standard editorconfig settings, IntelliJ supports hundreds of own configuration keys starting with `ij_`, like the two you can already see in the pull request.

I will add a few default settings then!

oshi-core/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
dbwiddis: Also, complicating things, in the existing method there are two filters, the loopback and the virtual (whether a hardware address exists).  So perhaps boolean isn't the best filtering, but open to ideas on how to make this both flexible and not overly complex (I don't want two booleans).  I do think an "unfiltered" output vs. a "most common filters" with a boolean switch is a good compromise.

.editorconfig
dbwiddis: I installed the Eclipse plugin. I don't see a way to export my existing eclipse file.

Spitballing without any experience, this is what I'd guess is needed. Feel free to add any  you think are missing.
```
# top-most EditorConfig file
root = true

# Unix-style newlines with a newline ending every file
[*]
end_of_line = lf
insert_final_newline = true

# Matches multiple files with brace expansion notation
# Set default charset
[*.{java,xml}]
charset = utf-8

# 4 space indentation
[*.{java,xml}]
indent_style = space
indent_size = 4

# Matches the exact files either package.json or .travis.yml
[{package.json,.travis.yml}]
indent_style = space
indent_size = 2
```

zalintyre: I pushed a suggestion for the editorconfig.

dbwiddis:
FYI, the macos failures may be a separate bug that I'm hunting down in the tcpstats structure. You can ignore them.

zalintyre:
I refactored the getNetworks() method. I tried to keep the API stable, though.

dbwiddis (commented):
Looks good and works.  I have two more requests and this will be ready to merge!
 - DRY: just implement the no-arg to "false" delegation once in the AbstractHAL, and save yourself 10 other method calls.
 - Move the boolean test outside of the loop (only test once) and move the exception handling to where it's thrown.  This is logically the same as what you had, but lets us simplify the method.

oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
```diff
@@ -116,36 +116,48 @@ protected AbstractNetworkIF(NetworkInterface netint) {
...
-                NetworkInterface netint = interfaces.nextElement();
-                try {
-                    if (!netint.isLoopback() && netint.getHardwareAddress() != null) {
-                        result.add(netint);
-                    }
-                } catch (SocketException ex) {
-                    LOG.error("Socket exception when retrieving interface \"{}\": {}", netint.getName(),
-                            ex.getMessage());
+        for (NetworkInterface iface : interfaces) {
+            try {
```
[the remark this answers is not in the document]
zalintyre: Done :)

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHardwareAbstractionLayer.java
```diff
@@ -85,6 +85,11 @@ public Sensors createSensors() {
         return LinuxNetworkIF.getNetworks();
```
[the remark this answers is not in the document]
zalintyre: Done :)

dbwiddis:
Looks good with the latest commit!  I'll do some final cleanup and merge!

dbwiddis (approved):

PR #1353, opened 2020-10-07
Passes default locale setting when executing commands.

### LINKED ISSUE #1347: runNative locale issues on linux
This code will not work as expected if locale in not english or C
https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java#L123

Proposed fix: set env. variable `LC_ALL=C`
I can't make PR because i don't have sane internet right now.

dmitraver:
Closes #1347 

Passes locale when executing commands on different platforms.

### DIFF
--- a/oshi-core/src/main/java/oshi/util/ExecutingCommand.java
+++ b/oshi-core/src/main/java/oshi/util/ExecutingCommand.java
@@ -34,6 +34,8 @@ import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import oshi.PlatformEnum;
+import oshi.SystemInfo;
 import oshi.annotation.concurrent.ThreadSafe;
 
 /**
@@ -81,7 +83,16 @@ public final class ExecutingCommand {
     public static List<String> runNative(String[] cmdToRunWithArgs) {
         Process p = null;
         try {
-            p = Runtime.getRuntime().exec(cmdToRunWithArgs);
+            PlatformEnum platform = SystemInfo.getCurrentPlatformEnum();
+            if (platform == PlatformEnum.WINDOWS) {
+                p = Runtime.getRuntime().exec(cmdToRunWithArgs, new String[]{"LANGUAGE=C"});
+            } else if (platform == PlatformEnum.LINUX || platform == PlatformEnum.FREEBSD
+                    || platform == PlatformEnum.SOLARIS || platform == PlatformEnum.AIX
+                    || platform == PlatformEnum.MACOSX) {
+                p = Runtime.getRuntime().exec(cmdToRunWithArgs, new String[]{"LC_ALL=C"});
+            } else {
+                p = Runtime.getRuntime().exec(cmdToRunWithArgs);
+            }
         } catch (SecurityException | IOException e) {
             LOG.trace("Couldn't run command {}: {}", Arrays.toString(cmdToRunWithArgs), e.getMessage());
             return new ArrayList<>(0);
--- a/oshi-core/src/main/java/oshi/util/ExecutingCommand.java
+++ b/oshi-core/src/main/java/oshi/util/ExecutingCommand.java
@@ -47,7 +47,20 @@ public final class ExecutingCommand {
 
     private static final Logger LOG = LoggerFactory.getLogger(ExecutingCommand.class);
 
+    private static final String[] DEFAULT_ENV = getDefaultEnv();
+
     private ExecutingCommand() {
+    }
+
+    private static String[] getDefaultEnv() {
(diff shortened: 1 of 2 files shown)


oshi-core/src/main/java/oshi/util/ExecutingCommand.java
```diff
@@ -81,7 +83,16 @@ private ExecutingCommand() {
...
         try {
-            p = Runtime.getRuntime().exec(cmdToRunWithArgs);
+            PlatformEnum platform = SystemInfo.getCurrentPlatformEnum();
+            if (platform == PlatformEnum.WINDOWS) {
+                p = Runtime.getRuntime().exec(cmdToRunWithArgs, new String[]{"LANGUAGE=C"});
+            } else if (platform == PlatformEnum.LINUX || platform == PlatformEnum.FREEBSD
+                    || platform == PlatformEnum.SOLARIS || platform == PlatformEnum.AIX
+                    || platform == PlatformEnum.MACOSX) {
+                p = Runtime.getRuntime().exec(cmdToRunWithArgs, new String[]{"LC_ALL=C"});
+            } else {
```
dbwiddis, line 93: This `else` case will never happen. It's either Windows or *nix. You can simplify the previous condition to be the "else" without testing the platform.

oshi-core/src/main/java/oshi/util/ExecutingCommand.java
```diff
@@ -81,7 +83,16 @@ private ExecutingCommand() {
     public static List<String> runNative(String[] cmdToRunWithArgs) {
         Process p = null;
         try {
-            p = Runtime.getRuntime().exec(cmdToRunWithArgs);
+            PlatformEnum platform = SystemInfo.getCurrentPlatformEnum();
+            if (platform == PlatformEnum.WINDOWS) {
+                p = Runtime.getRuntime().exec(cmdToRunWithArgs, new String[]{"LANGUAGE=C"});
```
dbwiddis, line 88: Let's make this string array a constant so we don't have to instantiate it every time.

oshi-core/src/main/java/oshi/util/ExecutingCommand.java
```diff
@@ -81,7 +83,16 @@ private ExecutingCommand() {
...
+            } else if (platform == PlatformEnum.LINUX || platform == PlatformEnum.FREEBSD
+                    || platform == PlatformEnum.SOLARIS || platform == PlatformEnum.AIX
+                    || platform == PlatformEnum.MACOSX) {
+                p = Runtime.getRuntime().exec(cmdToRunWithArgs, new String[]{"LC_ALL=C"});
```
dbwiddis, line 92: Let's make this string array a constant too.  Actually you can probably just define a single constant `EXEC_ENV` (or similar) and use the platform conditional to decide which array it is. The platform won't change during execution.  Then you can just use that as the second arg here with no conditional.

dbwiddis (changes requested):
Thanks. This works but I've left some comments on improving it.

Can you also add a change log entry under 5.3.0?

XakepSDK:
Maybe helper method? `ENVP_WINDOWS` and `ENVP_UNIX` constants.
```java
private static String[] getDefaultEnv() {
    if (platform == PlatformEnum.WINDOWS) {
        return ENVP_WINDOWS;
    } else if (platform != PlatformEnum.UNKNOWN) {
        return ENVP_UNIX;
    } else {
        return null;
    }
}
```

dbwiddis:
> Maybe helper method?

In the direction I was thinking but this will not change during operation, so we just need a one-time constant (So using this helper method but defining a constant as the result would work).

oshi-core/src/main/java/oshi/util/ExecutingCommand.java
dmitraver: I actually thought about checking for windows and treat everything else as *nix but then found that there is also UNKNOWN value in the enum so wanted to make sure that there are no surprises when new not windows and not unix platform is added but if you say its fine I will change it 👍

dmitraver:
Pushed review fixes.

dmitraver:
Ops, haven't seen your comment about changelog, adding it.

dmitraver:
@dbwiddis updated changelog and rebased.

dbwiddis:
This looks good!  I'm headed to bed but will get this merged tomorrow. Thanks for your contribution!

oshi-core/src/main/java/oshi/util/ExecutingCommand.java
dbwiddis: It is in the enum, but also throws an exception at SystemInfo instantiation.

PR #1356, opened 2020-10-11
Add messages to unit test assertions

j4ckofalltrades:
Update test assertions for classes in the oshi-core/src/test/software.os
package.

Refs #1215


dbwiddis:
Thanks, looks good!  Will format code and merge!

PR #1357, opened 2020-10-13
Add messages to unit test assertions.

J-Jimmy:
I've added some messages to unit test assertions as per #1215.
Honestly I got bored after this bit, however I don't see myself able to contribute to the other Issues...


dbwiddis:
Thanks for the contribution!  Yes, this category was meant for first-timers, and some of the other ones can be a bit more challenging.  I think there's a good solution to #1339 refactoring the "mount" code to an external "driver" class offering the desired functionality, if you wanted to tackle that.

PR #1359, opened 2020-10-14
make suppressed types/paths per globalconfig and set on startup if not set before

J-Jimmy:
More like this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 5.3.1 (in progress)
 ================
-* Your contribution here
 * [#1359](https://github.com/oshi/oshi/pull/1359): Set suppressed network filesystem types and pseudo filesystem types via config - [@J-Jimmy](https://github.com/J-Jimmy).
+* [#1362](https://github.com/oshi/oshi/pull/1362): Correct invalid Windows processor bitmask logic on 64th core - [@J-Jimmy](https://github.com/J-Jimmy).
+* Your contribution here
 
 5.3.0 (2020-10-11)
 ================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 5.3.1 (in progress)
 ================
 * Your contribution here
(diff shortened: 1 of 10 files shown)


dbwiddis:
On the right track, but way too much new code.  Most of the structure to do this is already there.   All you should need to do is:
 - Move the strings defining the different filesystems / paths from `AbstractFileSystem` into the config file `oshi.properties` (`src/main/resources`).  So, for example you'd have:

```
# Filesystem types which do not display in a "local" filesystem search
oshi.filesystem.network.types=afs,cifs,smbfs,sshfs,ncpfs,ncp,nfs,nfs4,gfs,gds2,glusterfs
```
 - Change the constant declarations in `AbstractFileSystem` to pull from the `GlobalConfig`.  Declare a public constant in the class matching the config file, e.g., `OSHI_FILESYSTEM_NETWORK_TYPES="oshi.filesystem.network.types"` and use that constant in `GlobalConfig.get()`.

The config will read in the default from the config file, optionally change it shortly after startup using `GlobalConfig.set()`, and the first time any user instantiates a `FileSystem` object it will create those string arrays from the current value in the config.

So most of the changes are just in `AbstractFileSystem` and `oshi.properties`.  Possibly update the javadocs to reference the functionality.

Do similar for the pseudo FS types which is where "overlay" is being removed.

See the top of file `AbstractOperatingSystem` for an example of how `oshi.os.unix.whoCommand` is defined similarly, or `WindowsOSProcess` for `oshi.os.windows.commandline.batch`.  The difference here is instead of a boolean you'll be reading in a comma-delimited string and parsing it (`split()`) into a String array.

J-Jimmy:
I've now changed it.
Sorry for taking so much of your time, but seeing how you would approach things definitely helps to learn for future tasks.

dbwiddis:
Thanks, this looks good!  I'll want to run some tests but expect to merge tonight.

J-Jimmy:
I'm not sure about this build exception:
"Unresolveable build extension: Plugin org.apache.maven.wagon:wagon-ssh:3.4.0 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.wagon:wagon-ssh:jar:3.4.0 @"
It works locally...?

dbwiddis:
Occasionally the CI containers have network issues and don't download things. I've rerun tests and they're all good.

Issue #1361, opened 2020-10-15
Incorrect CPU usage reported for Windows based processor group setup


RajeshPutta87:
I am from AppDynamics. We are using OSHI 5.1.0 to retrieve CPU, Memory, Processor, Process information and the environment in use is name=Windows, release=Microsoft Windows Server 2016 Standard, architecture=64-bit

Environment in test is having
64 physical cores
128 logical cores/logical processors

AppDynamics customer had created two processor groups, each with 64 logical processor. Issue is that using OSHI we are detecting CPU activity on the first “processor group”, i.e. 64 logical processors. Generally a instance of a service in run in each processor group – so that means service instances are running in each of the two processor groups on this machine… but OSHI is computing CPU activity on only one processor group.

Attached are the results of client's testing of the 64 core issue…  the first screenshot shows the CPU usage (57%) for a task that ran on the second processor group  (cores 64 – 127) for about 1.5 minutes from 3:45pm to 3:47pm.
The second screenshot shows the CPU usage (also 57%) for the same task run on the first processor group (cores 0 – 63) also for about 1.5 minutes from 4:00p to 4:02pm.

The third screenshot show the AppDynamics using OSHI shows a max cpu trend from 3:40pm to 4:04pm and doesn't correctly reflect the CPU usage.

I also referred to https://github.com/oshi/oshi/issues/785. This looks to be relevant to this issue.

![image009](https://user-images.githubusercontent.com/72386265/96132255-cfce4b80-0f17-11eb-826b-461d0912d2e0.jpg)

Can you please look into the issue.

![image007](https://user-images.githubusercontent.com/72386265/96132078-939aeb00-0f17-11eb-9435-505328caa22a.png)
![image008](https://user-images.githubusercontent.com/72386265/96132111-9bf32600-0f17-11eb-9a86-236cc9ae1899.png)

Thanks
Rajesh

RajeshPutta87:
[Uploading processor.txt…]()

RajeshPutta87:
[processor.txt](https://github.com/oshi/oshi/files/5385334/processor.txt)

dbwiddis:
To troubleshoot this more I'll probably need some information from the relevant machine.  Will it be possible to do so?  I'd need ideally both, but at least one of:
 - a view of the performance counters for "Processor Information", specifically the format of the instances which should identify the processor groups uniquely
 - output of the query `wmic path Win32_PerfRawData_Counters_ProcessorInformation`

[...]

PR #1362, opened 2020-10-15
Logical cores iteration

J-Jimmy:
Basically copied from you comment.....
Seems kinda hacky to get a PR that way.....

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 5.3.1 (in progress)
 ================
 * Your contribution here
+* [#1362](https://github.com/oshi/oshi/pull/1362): 63.rd core is wrongly calculated due to too high numbers - [@J-Jimmy](https://github.com/J-Jimmy).
 
 5.3.0 (2020-10-11)
 ================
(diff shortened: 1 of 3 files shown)


dbwiddis:
It's not hacky to do an bug fix!   But you need a changelog entry and probably a better title for this PR.

Also I'm going to hold off on merging this while resolving #1361 which is probably in that same iteration.  I suspect using `i` from the bitmask should be replaced by an incrementing (from 0) counter for each core instead, and it'd be nice to fix both problems in one PR.

J-Jimmy:
I'm really not following all this code about CPU cores....
And I have no environment to test/reproduce and analyze the behaviour in order to fix it.

dbwiddis:
I don't have a test environment either but I understand the problem well enough to describe the fix, and the user can test it.

I'll go ahead and merge this.

dbwiddis:
Or not.  Sorry, as I dig into this I keep finding more.  The fix you did so far applies for the pre-Win7 branch of the code.

Same fix needs to be applied in `getMatchingPackage()`, `getMatchingNumaNode()`, and `getMatchingCore()`.  Everywhere you see `... mask ... & (1L << lp)) > 0` should be `... mask ... & (1L << lp)) != 0`

dbwiddis:
And also in `getLogicalProcessorInformationEx()`

PR #1368, opened 2020-10-16
Backup for process name and state

J-Jimmy:
Retrieve process name and state from stat string if status map doesn't contain them already.
(I checked for Uid and Gid too, but it seems they are not contained in the string)

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
@@ -358,12 +358,11 @@ public class LinuxOSProcess extends AbstractOSProcess {
             return;
         }
 
-        int nameStart = stat.indexOf("(");
-        int nameEnd = stat.indexOf(")");
+        int nameStart = stat.indexOf('(');
+        int nameEnd = stat.indexOf(')');
         if (Util.isBlank(status.get("Name")) && nameStart > 0 && nameStart < nameEnd) {
             // remove leading and trailing parentheses
-            String statName = stat.substring(nameStart, nameEnd + 1);
-            statName = statName.substring(1, statName.length() - 1);
+            String statName = stat.substring(nameStart + 1, nameEnd);
             status.put("Name", statName);
         }
 
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
@@ -360,20 +360,16 @@ public class LinuxOSProcess extends AbstractOSProcess {
 
         int nameStart = stat.indexOf("(");
         int nameEnd = stat.indexOf(")");
-        if (Util.isBlank(status.get("Name")) && nameStart < nameEnd) {
+        if (Util.isBlank(status.get("Name")) && nameStart > 0 && nameStart < nameEnd) {
             // remove leading and trailing parentheses
             String statName = stat.substring(nameStart, nameEnd + 1);
             statName = statName.substring(1, statName.length() - 1);
             status.put("Name", statName);
         }
 
-        // get the rest of the string from after the name
-        // (the name may contain whitespaces and destroy the index)
-        String statEnd = stat.substring(nameEnd + 2);
-        String[] statSplit = ParseUtil.whitespaces.split(statEnd);
         // As per man, the next item after the name is the state
-        if (Util.isBlank(status.get("State")) && statSplit.length > 0) {
-            String statState = statSplit[0];
+        if (Util.isBlank(status.get("State")) && nameEnd > 0 && stat.length() > nameEnd + 2) {
+            String statState = String.valueOf(stat.charAt(nameEnd + 2));
             status.put("State", statState);
         }
     }
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
@@ -53,6 +53,7 @@ import oshi.software.os.OSThread;
 import oshi.util.ExecutingCommand;
 import oshi.util.FileUtil;
 import oshi.util.ParseUtil;
+import oshi.util.Util;
 import oshi.util.platform.linux.ProcPath;
 
 @ThreadSafe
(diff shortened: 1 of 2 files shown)


oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -339,6 +343,38 @@ public boolean updateAttributes() {
...
+     * @param status
+     *            status map to fill.
+     * @param stat
+     *            string to read from.
+     */
+    private static void getMissingDetails(Map<String, String> status, String stat) {
+        if (status == null || stat == null) {
+            return;
+        }
+        String[] statSplit = stat.split(" ");
```
dbwiddis, line 359: For splitting on whitespace, use `Parseutil.whitespace.split()` which precompiles this.  However, I believe some filenames can have spaces in them, so this won't work as easily as you've implemented. You need to find the open and close parens and read the values between them. `String.indexOf()` for `(` and `)` should give you enough info to grab all the appropriate substrings without using split.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -339,6 +343,38 @@ public boolean updateAttributes() {
...
+        if (status.get("Name") == null || status.get("Name").trim().isEmpty()) {
+            // As per man, the second item is the name
+            if (statSplit.length > 1) {
```
dbwiddis, line 362: While I'm suggesting changing away from split, this could have been included in the above conditional. But it won't work for names with spaces in them.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -339,6 +343,38 @@ public boolean updateAttributes() {
...
+                String statName = statSplit[1];
+                // remove leading and trailing parentheses
+                statName = statName.substring(1, statName.length() - 1);
+                status.put("Name", statName);
+            }
+        }
+        if (status.get("State") == null || status.get("State").trim().isEmpty()) {
+            // As per man, the third item is the state
+            if (statSplit.length > 2) {
```
dbwiddis, line 371: While I'm suggesting changing away from split, this could have been included in the above conditional.  But it won't work for names with spaces in them.

dbwiddis (commented):
I left a commit to fix the checkstyle issues CI was complaining about, and changed `name` and `state` not to conflict with the class variables.   You need to rework the parsing logic, though.  Filenames (and thus executables) can have spaces in the name.

J-Jimmy:
I can respect spaces in the name if I extract it like this.
However to get the state, I do need to split by whitespaces....There just is nothing else. (At least on my testsystem)

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -339,6 +344,40 @@ public boolean updateAttributes() {
...
+     *            string to read from.
+     */
+    private static void getMissingDetails(Map<String, String> status, String stat) {
+        if (status == null || stat == null) {
+            return;
+        }
+
+        int nameStart = stat.indexOf("(");
+        int nameEnd = stat.indexOf(")");
+        if (Util.isBlank(status.get("Name")) && nameStart < nameEnd) {
```
dbwiddis, line 363: I believe the strict `nameStart < nameEnd` should catch the case where there's no `( ... name ... )` in the String.  However, just to be completely 100% sure here, we should probably ensure `nameStart > 0`.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -339,6 +344,40 @@ public boolean updateAttributes() {
...
+            // remove leading and trailing parentheses
+            String statName = stat.substring(nameStart, nameEnd + 1);
```
dbwiddis, line 365: shouldn't this be `nameStart + 1` to exclude the leading parenthesis and `nameEnd` (which is exclusive) to exclude the ending parenthesis?

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -339,6 +344,40 @@ public boolean updateAttributes() {
...
+            statName = statName.substring(1, statName.length() - 1);
+            status.put("Name", statName);
+        }
+
+        // get the rest of the string from after the name
+        // (the name may contain whitespaces and destroy the index)
+        String statEnd = stat.substring(nameEnd + 2);
+        String[] statSplit = ParseUtil.whitespaces.split(statEnd);
```
dbwiddis, line 373: I think this is unnecessary. You only need the single character, so `stat.charAt(nameEnd + 2)` should return this.  You'll need to do a length check, of course.

dbwiddis (commented):
Looking good. A few more requests and it'll be good to go!

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
J-Jimmy: If there is no name in the string, both would be index of -1 and therefore the check would turn to false.
I will add the extra check, but can't imagine a case where its necessary.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
J-Jimmy: I did output the original name from the status map and the name I extract here, the indexes are right as they are (on Ubuntu 20.04 LTS with Java 1.8_265)

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
J-Jimmy: Added the check

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -339,6 +344,36 @@ public boolean updateAttributes() {
...
+     *            status map to fill.
+     * @param stat
+     *            string to read from.
+     */
+    private static void getMissingDetails(Map<String, String> status, String stat) {
+        if (status == null || stat == null) {
+            return;
+        }
+
+        int nameStart = stat.indexOf("(");
```
dbwiddis, line 361: Scrutinzer says using the character `'('` is faster than `"("`.  I'm not sure it matters but why not? :)

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -339,6 +344,36 @@ public boolean updateAttributes() {
...
+        int nameEnd = stat.indexOf(")");
```
dbwiddis, line 362: But [benchmarks show no real difference](https://stackoverflow.com/questions/33646781/java-performance-string-indexofchar-vs-string-indexofsingle-string#:~:text=1%20Answer&text=As%20you%20can%20see%2C%20indexOfChar,but%20somewhat%20slower%20for%20indirect.) so I don't care that much.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
dbwiddis: Ah, I see what you're doing. In this line you're capturing the full string with both parentheses, but then in the next line you're starting the substring from 1, and going to length-1 (which, since it's exclusive, has length-2 as last character), effectively removing the two parenthesis.  If you just use `start+1, end` on this line, you don't need a second line.

dbwiddis (commented):
I can make these fixes and merge if you'd like, but wanted to at least give feedback why.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
J-Jimmy: Oh yeah, thats still in because I had the string split at first, obviously can be put together. I just woke up so I could do it in half an hour.

dbwiddis:
Looks great! Merging.

PR #1369, opened 2020-10-19
Add messages to unit test assertions

Spotts9:
https://github.com/oshi/oshi/issues/1215


dbwiddis:
Thanks for the commit!  Actually, the message is what happens when the test fails, so it should read that there *were* errors found.

dbwiddis:
Looks good, thanks!

PR #1376, opened 2020-10-25
Addition of assertion messages for GlobalConfigTest

singloon:
Addition of assertion messages for GlobalConfigTest, encapsulated in the GlobalConfigAsserter class. #1215 Feedback would much appreciated.


dbwiddis:
Thanks! I like the refactoring using `assertThat()`.  Will merge.

PR #1380, opened 2020-10-27
Fix exception sorting child process list on Windows

agaponik:
Fix for WindowsOperatingSystem.getChildProcesses throwing exception, trying to sort unmodifiableList:

java.lang.UnsupportedOperationException
	at java.util.Collections$UnmodifiableList.sort(Collections.java:1331)
	at oshi.software.common.AbstractOperatingSystem.processSort(AbstractOperatingSystem.java:155)
	at oshi.software.os.windows.WindowsOperatingSystem.getChildProcesses(WindowsOperatingSystem.java:296)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 5.3.3 (in progress)
 ================
 * [#1379](https://github.com/oshi/oshi/pull/1379): Add `squashfs` as a pseudo file system so it will be skipped by default - [@mprins](https://github.com/mprins).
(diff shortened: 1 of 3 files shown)


dbwiddis:
Nice catch and thanks for this fix.   Two requests:
 * Looks like this was a regression with the 5.X version and went unnoticed for 6 months. That points to insufficient test cases.  Can you please add a testcase in `OperatingSystemTest` that has a non-null sorting method that should have caught this?
 * Please add a changelog entry.

PR #1388, opened 2020-11-01
support systemctl list-unit-files return 3 columns

Szwendacz99:
I was getting error on maven test:

[ERROR] oshi.software.os.OperatingSystemTest.testGetServices  Time elapsed: 0.443 s  <<< FAILURE!
java.lang.AssertionError: 
There should be at least 1 stopped service
Expected: is a value greater than <0>
     but: <0> was equal to <0>
        at oshi.software.os.OperatingSystemTest.testGetServices(OperatingSystemTest.java:339)

and found out that getServices() in LinuxOperatingSystem class was trying to get current list of services from systemd assuming that systemctl list-unit-files returns just two columns of data while on my version (systemd 246 (246.6-1-arch)) it returns 3 columns: " 'UNIT FILE'          'STATE'           'VENDOR PRESET' ", so I updated that function so it will handle 3 columns, and after my small change it seems to work good now, maven test passes.

Tested on latest Arch Linux with systemd 246 (246.6-1-arch)

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
@@ -562,7 +562,7 @@ public class LinuxOperatingSystem extends AbstractOperatingSystem {
         List<String> systemctl = ExecutingCommand.runNative("systemctl list-unit-files");
         for (String str : systemctl) {
             String[] split = ParseUtil.whitespaces.split(str);
-            if (split.length == 2 && split[0].endsWith(".service") && "enabled".equals(split[1])) {
+            if ((split.length == 2 || split.length == 3) && split[0].endsWith(".service") && "enabled".equals(split[1])) {
                 // remove .service extension
                 String name = split[0].substring(0, split[0].length() - 8);
                 int index = name.lastIndexOf('.');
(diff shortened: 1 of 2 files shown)


Szwendacz99:
Here is part of changelog when 3rd column was introduced in systemd:
" * "systemctl list-unit-files" has been updated to show a new column
          with the suggested enablement state based on the vendor preset files
          for the respective units."
link: https://github.com/systemd/systemd/blob/master/NEWS

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -562,7 +562,7 @@ private static String filenameToFamily(String name) {
         List<String> systemctl = ExecutingCommand.runNative("systemctl list-unit-files");
         for (String str : systemctl) {
             String[] split = ParseUtil.whitespaces.split(str);
-            if (split.length == 2 && split[0].endsWith(".service") && "enabled".equals(split[1])) {
+            if ((split.length == 2 || split.length == 3) && split[0].endsWith(".service") && "enabled".equals(split[1])) {
```
dbwiddis, line 565: Since we only use the first two columns, how about just `split.length >= 2`?  Then they can add even more columns later.

dbwiddis (commented):
I like it when unit testing helps identify bugs.   Thanks for the fix.  Can you also add a changelog entry?

dbwiddis:
Thanks! Looks good, I'll merge this as soon as CI tests finish.

PR #1411, opened 2020-11-15
Add mapping of device and volume to be used for getting uuid

tausiflife:


oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -65,14 +65,29 @@
...
         // Map uuids with device path as key
         Map<String, String> uuidMap = new HashMap<>();
         File uuidDir = new File("/dev/disk/by-uuid");
         if (uuidDir.listFiles() != null) {
             for (File uuid : uuidDir.listFiles()) {
                 try {
-                    // Store UUID as value with path (e.g., /dev/sda1) as key
+                    // Store UUID as value with path (e.g., /dev/sda1) as key and also as volumes as key
                     uuidMap.put(uuid.getCanonicalPath(), uuid.getName().toLowerCase());
+                    if (volumeDeviceMap.get(uuid.getCanonicalPath()) != null) {
```
dbwiddis, line 88: Any reason not to use `containsKey()` here to avoid creating an object twice?

dbwiddis (approved):
Thanks! Looks good, one comment inline.  Don't forget the changelog.

oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
tausiflife: fixed.

tausiflife:
> Thanks! Looks good, one comment inline. Don't forget the changelog.

ok will add it :)

oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
dbwiddis: Also we're calling `getCanonicalPath()` multiple times here (3), might be good to store that in a temp variable.

tausiflife: added a temp variable and used it everywhere.

tausiflife:
@dbwiddis Will you be doing the merge or should I?

oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -65,14 +65,30 @@
 
     @Override
     public List<OSFileStore> getFileStores(boolean localOnly) {
+        //Map of volume with device path as key
+        Map<String, String> volumeDeviceMap = new HashMap<>();
+        File volumes = new File("/dev/mapper");
+        if (volumes.listFiles() != null) {
```
dbwiddis, line 71: Just noticed this on second review.  You call `listFiles()` twice, which is extra I/O.

dbwiddis:
Go ahead and squash-and-merge if tests pass after the `listFiles()` change.

dbwiddis (approved):

oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
tausiflife: fixed this one and also the below uuid call.

dbwiddis:
Thanks for handling this!

tausiflife:
Anytime :-) I will try my best to help you and this project whenever i can.

Issue #1145, opened 2020-03-20
maven-javadoc-plugin 3.2.0 breaks release


dbwiddis:
I was repeatedly unable to successfully execute `mvn release:perform` with the latest version of the javadoc plugin.  Rolling back to 3.1.1 caused no problems.  I can't see anything obvious about what might have changed and would appreciate the help of someone more experienced in maven trying to troubleshoot this.

It works fine on the `oshi-core` artifact but fails on the shaded version.

Log right before the errors posted below.

```
    Building index for all the packages and classes...
    Generating /Users/danielwiddis/git/oshi/target/checkout/oshi-core-shaded/target/apidocs/overview-tree.html...
    Generating /Users/danielwiddis/git/oshi/target/checkout/oshi-core-shaded/target/apidocs/deprecated-list.html...
    Generating /Users/danielwiddis/git/oshi/target/checkout/oshi-core-shaded/target/apidocs/index.html...
    Generating /Users/danielwiddis/git/oshi/target/checkout/oshi-core-shaded/target/apidocs/index-all.html...
    Building index for all classes...
    Generating /Users/danielwiddis/git/oshi/target/checkout/oshi-core-shaded/target/apidocs/allclasses-index.html...
    Generating /Users/danielwiddis/git/oshi/target/checkout/oshi-core-shaded/target/apidocs/allpackages-index.html...
    Generating /Users/danielwiddis/git/oshi/target/checkout/oshi-core-shaded/target/apidocs/overview-summary.html...
    Generating /Users/danielwiddis/git/oshi/target/checkout/oshi-core-shaded/target/apidocs/help-doc.html...
    1 error
    [INFO] ------------------------------------------------------------------------
    [INFO] Reactor Summary for Operating System and Hardware Information 4.5.1:
    [INFO] 
    [INFO] Operating System and Hardware Information .......... SUCCESS [01:16 min]
    [INFO] oshi-core .......................................... SUCCESS [04:42 min]
    [INFO] oshi-core-shaded ................................... FAILURE [  6.868 s]
    [INFO] oshi-demo .......................................... SKIPPED
    [INFO] oshi-dist .......................................... SKIPPED
    [INFO] ------------------------------------------------------------------------
    [INFO] BUILD FAILURE
    [INFO] ------------------------------------------------------------------------
    [INFO] Total time:  06:05 min
    [INFO] Finished at: 2020-03-19T16:40:04-07:00
    [INFO] ------------------------------------------------------------------------
[...]

PR #1428, opened 2020-11-29
[pom] Bump javadoc plugin to 3.2.0

hazendaz:
@dbwiddis Not sure if there was a reason this one was skipped or why dependabot v2 is not suggesting it.  I did fix a javadoc issue tonight with jdk 11 so possibly that was one reason it was missed or some odd mix up in dependabot.  Anyway, this should work fine.  I haven't run into any issues with latest.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -147,7 +147,7 @@
 		<maven-shade-plugin.version>3.2.4</maven-shade-plugin.version>
 		<!-- reporting plugins -->
 		<maven-checkstyle-plugin.version>3.1.1</maven-checkstyle-plugin.version>
-		<maven-javadoc-plugin.version>3.1.1</maven-javadoc-plugin.version>
+		<maven-javadoc-plugin.version>3.2.0</maven-javadoc-plugin.version>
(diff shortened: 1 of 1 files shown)


dbwiddis:
See #1145.  I've intentionally suppressed this update.  If you can fix that issue, by all means, update! :)

dbwiddis (commented):
I intentionally suppressed dependabot updates for 3.2.x for this plugin based on issues reported #1145.  Having spent several hours trying various workarounds and fixes, I decided "don't update" was the easiest way forward.   

I'm ok with updating if you want to handle the next release and troubleshooting the associated issues relating to javadocs (from memory, the shade plugin appears to see the module name from the main build and assume docs are in JDK9+ paths when they are in JDK8 paths).

However, I was hoping time, and any issues fixed before 3.3.x would be of more use before investing troubleshooting time.

hazendaz:
Ah, thanks.  I remember the issue now :)  I'll take a look into this.  For now I'll close this PR.

PR #1433, opened 2020-11-30
[ci] Run on pull request

hazendaz:


dbwiddis:
OK, so what I'd really like is:
 - full matrix (3 x 4) of all pushes to `master`
 - limited tests (3 OS's, one JDK only) of pull requests

Looking at examples it seems I could do something like this:

```
name: Master CI
on:
  push:
    branches:
      - master
jobs:
 ... etc. full matrix...
```
and
```
name: PR CI
on:
  pull_request:
jobs:
 ... etc., limited matrix ...
```

PR #1438, opened 2020-12-01
[pom] Use suggested sonatype id as 'ossrh' for both

hazendaz:
This ties to the maven settings file.  Using a different name on both would require duplicate entries for the account data.  The suggested entry at least check (few years) was to call it ossrh.  Either way, its duplicate settings or consistent name.  Please do make sure to adjust your local as well.  In fact, there makes sense to have duplicate entries for all 3 (2 original plus this) as many teams use various older setups.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,12 +93,12 @@
 			<url>github:ssh://oshi.github.io/oshi/</url>
 		</site>
 		<snapshotRepository>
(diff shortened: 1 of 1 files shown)


hazendaz:
@dbwiddis This fixes sonatype.  Do note it will impact your process slightly so you will need to add in your settings file the new entry.  Currently you have 1 or 2 based on the naming.

PR #1444, opened 2020-12-05
[actions] Add coveralls

hazendaz:


hazendaz:
@dbwiddis Coveralls curiousy of Jeff Butler's hard work over at mybatis :)

.github/workflows/coveralls.yaml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: 8
+      - name: Report Coverage to Coveralls for Pull Requests
+        if: github.event_name == 'pull_request'
+        run: ./mvnw test jacoco:report coveralls:report -q -Dlicense.skip=true -DrepoToken=$GITHUB_TOKEN -DserviceName=github -DpullRequest=$PR_NUMBER
```
dbwiddis, line 33: Can't we just add these lines to the CI and PR workflows to run conditionally?  

I combined the PR and CI workflows in the version I installed on JNA and just had an "optional" step that only ran on the first matrix item:
```
      - name: Checkstyle
        if: contains(matrix.os, 'ubuntu') && contains(matrix.java, '8')
        run: ant checkstyle
```
Seems we could do similarly here rather than create a whole new workflow.

hazendaz: Obsolutely.  I sort of like them separate as it seems more clear but that makes sense.  Do you want to merge this then work it from here or just rework from here?

dbwiddis: I'm already editing the two other files to add the Unix CI so it'll be easy to just copy over the lines from this PR into that one.  We did used to have the jacoco:report stuff in the earlier travis script (on an "if" for the first matrix) so it seems more consistent with what I'm used to.  I may also combine the two separate PR and CI  flows as well which means if this got merged it's just 3 files being combined into one.  SO... flip a coin. :)

dbwiddis: Can even just make it a separate "job" under the jobs instead of tying it to the matrix.  SO.... yeah, just leave this here, I'll combine into mine.

hazendaz: sounds good!

dbwiddis:
Incorporated into #1445

PR #1482, opened 2021-01-11
Added configuration properties to exclude/include filestores

majster-kat:
I have created initial implementation of filestore excludes/includes properties. In order to not mess with existing filesystems I have created new  AbstractUnixFileSystem which is inherited only by SolarisFileSystem. 
Next filesystems will continue bit later.

Also names of properties loaded from oshi.properties are calculated using `SystemInfo.getCurrentPlatformEnum().name().toLowerCase();` in `AbstractUnixFileSystem.parseFSConfig`. Disadvantage is that full property names are not used in code and IntelliJ Idea marks all new properties in oshi.properties as unused. 

I have created also some tests and tested all on live Solaris OS. 

If you agree with my approach, then I will continue with remaining Filesystems and will also review all code again (add missing comments, review comments, remove System.out lines :))


dbwiddis:
Thanks for the PR!  Some thoughts:
 - While automatic creation of the config parameter is nice and possibly a generally good idea to do for *all* config parameters (possibly in the future, in conjunction with automatic code generation, modularization, and other "design from the ground up" choices) I think I'd prefer in this case to just have `String` constants in each of the OS FileSystem classes.  The benefit of this is that users who wanted to override the config programmatically could just use that constant, instead of repeating the same code you have in the `parseFSConfig()` method in their own implementations.  (Also, the `MACOSX` enum eventually needs to become `MACOS` and the less code we add with the old name the better.)
 - I like the `FileSystemUtil` class for the parsing and matching methods, but I think `isFileStoreExcluded()` would be better as a `protected` member of each FileSystem class's parent.  You could reduce the argument list to two, as those four include/exclude lists could be in the abstract parent (they already are!), but populated by an `abstract` method implemented in each individual class (reading the class-specific constant name).
 - I'm not sure we need a separate abstract parent just for the Unix classes, but I'm not strongly opposed, either.  I guess it makes sense given my previous bullet so we don't have to have a "no op" windows implementation.  
 - Don't forget that you need to explicitly allow the `/` mount path, either hardcoded or via config.
 - I like how you've done `GLOB_PREFIX`, adding it by default if it's not already there. I would suggest also allowing a `REGEX_PREFIX`, even if we're not using it, since the capability to use it is built in.

Other minor cleanup comments you may already have thought of:
 - See CI comments on unused constants and imports.
 - Go ahead and call this a "new feature" in the change log.

dbwiddis:
I went ahead and made some of the tweaks I suggested above and implemented for all the OS's.
 - After trying to manually define all the configs I decided against it and stuck with your automatic generation.
   - I did tune it a little bit :) 
 - While we could move the inclusion from the util to the abstract unix class, that removes the ability to test it, so I left your arrangement as is.
   - I did tune it a little bit :)
 - I updated to explicitly include `/` and added `regex:` as I suggested

Let me know what you think. This is probably good to merge at this point.

majster-kat:
Hi, sorry for staying quiet, I was quite busy at work :)
Anyway yesterday I tried different approach (to get property names from abstract methods, implemented in inherited classes), but the problem with this solution is that method getFileStoreMatching is static.

So I took another approach, instead of implementing all in parent class (AbstractUnixFileSystem), all is implemented in SolarisFileSystem (and others) with help of FileSystemUtil.

So basically this can be added to SolarisFileSystem:

```java
    public static final String OSHI_SOLARIS_FS_PATH_EXCLUDES = "oshi.os.solaris.filesystem.path.excludes";
    public static final String OSHI_SOLARIS_FS_PATH_INCLUDES = "oshi.os.solaris.filesystem.path.includes";
    public static final String OSHI_SOLARIS_FS_VOLUME_EXCLUDES = "oshi.os.solaris.filesystem.volume.excludes";
    public static final String OSHI_SOLARIS_FS_VOLUME_INCLUDES = "oshi.os.solaris.filesystem.volume.includes";

    private static final List<PathMatcher> FS_PATH_EXCLUDES = FileSystemUtil.loadAndParseFileSystemConfig(OSHI_SOLARIS_FS_PATH_EXCLUDES);
    private static final List<PathMatcher> FS_PATH_INCLUDES = FileSystemUtil.loadAndParseFileSystemConfig(OSHI_SOLARIS_FS_PATH_INCLUDES);
    private static final List<PathMatcher> FS_VOLUME_EXCLUDES = FileSystemUtil.loadAndParseFileSystemConfig(OSHI_SOLARIS_FS_VOLUME_EXCLUDES);
    private static final List<PathMatcher> FS_VOLUME_INCLUDES = FileSystemUtil.loadAndParseFileSystemConfig(OSHI_SOLARIS_FS_VOLUME_INCLUDES);

...
// And then later in code to check if FS should be excluded:
FileSystemUtil.isFileStoreExcluded(path, volume,
                                                          FS_PATH_EXCLUDES,
                                                          FS_PATH_INCLUDES,
                                                          FS_VOLUME_EXCLUDES,
                                                          FS_VOLUME_INCLUDES)

```

It's actually quite simple, no need to modify any parent classes. And if some FS does not support it (for example WindowsFileSystem) then it simply won't be added to the class :)

What do you think?

dbwiddis:
> What do you think?

I like that approach. I started down that road with the public constants and tried to go back and forth to the parent with abstract methods to get those names, which required public implementations, all just to avoid repeating constants... that you could just put in each file!   Especially since the parent just calls out to the util anyway.   Don't know why I didn't think of doing what you just did, it's nice and clean.

dbwiddis:
Updated with your suggestion.  Let me know if you think this is good to merge now.

majster-kat:
It looks ok to me, I just did a small cleanup in FileSystemUtilTest. I think it's good to merge.

dbwiddis:
So of course I merge and it fails CI on AIX...
```
Caused by: java.util.regex.PatternSyntaxException: 
Explicit 'name separator' in class near index 2
[!/]**
  ^
```
Oops. :)

PR #1503, opened 2021-01-21
Fix bitness mismatch on 32-bit OpenBSD

dbwiddis:
The `sysctl` for cpu ticks returns a native long for system (`kern.cptimes`) and 64-bit long for processor (`kern.cptimes2`) but were hardcoded as `long` in both cases. This caused an exception on 32-bit openbsd. Which is also added to the CI matrix in this PR.


oshi-core/src/main/java/oshi/hardware/platform/unix/openbsd/OpenBsdCentralProcessor.java
```diff
@@ -241,14 +244,33 @@ protected long queryInterrupts() {
...
                 return new CpTime(m).cpu_ticks;
             } else if (m.size() == 6 * Native.LONG_SIZE) {
                 return new CpTimeNew(m).cpu_ticks;
             }
         }
+        return new NativeLong[0];
+    }
+
+    /**
+     * Parse memory buffer returned from sysctl kern.cptime2 to array of 5 or 6
```
mprins, line 259:
```suggestion
     * Parse memory buffer returned from sysctl kern.cp_time to array of 5 or 6
```

oshi-core/src/main/java/oshi/hardware/platform/unix/openbsd/OpenBsdCentralProcessor.java
```diff
@@ -241,14 +244,33 @@ protected long queryInterrupts() {
...
+     * longs depending on version
+     *
+     * @param m
+     *            A buffer containing a long array
+     * @return The long array
+     */
+    private static long[] cpTime2ToTicks(Memory m) {
```
mprins, line 266: it's a private method so not to important, but since it's parsing `kern.cp_time` you may want to change this method name

mprins (approved):
LGTM, 
works on my VM, will give it a shot on my laptop later today, as that has actual graphics

oshi-core/src/main/java/oshi/hardware/platform/unix/openbsd/OpenBsdCentralProcessor.java
dbwiddis: Good point.   There's a bit of inconsistency in the source.
 - `kern.cptime` name points to `KERN_CPTIME` constant points to internal code with a `cp_time` array of native-sized `long`
 - `kern.cptimes` name ponits to `KERN_CPTIME2` constant points to internal BSD `sysctl_cptime2()` pointing to internal code with a structure containing `spc_cp_time` array of 64-bit `long`.

I initially used the structure approach to make array access easy but that's not really needed and we now have 4 pseudo-structures. Let me just rewrite it to return arrays directly from the memory buffer.   Still need a separate method for the 32/64-bit vs. 64-bit-only approaches.

mprins (approved):

PR #1524, opened 2021-02-06
Fix Windows USB serial number retrieval

### LINKED ISSUE #1522: UsbDevice.serialNumber is always empty on windows
I used OSHI 4.0.0 and i have updated to OSHI 5.4.1. The serialNumber attribute of an instance of UsbDevice is always empty on windows.

It is a regression introduced by 00ba7d66a476cc76361efd8bfc7029d8135cfdb6.

The serialNumber is computed at line [154 of WindowsUsbDevice](https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsUsbDevice.java#L154). But this computation is performed only if the  `pnpDeviceId` is in the `deviceStringCache` map. This map is filled too late (see [line 164](https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsUsbDevice.java#L164)).

ymortier:
Fixes #1522


ymortier:
I removed the check as adviced in #1522.

dbwiddis:
Thanks for your contribution! Looks great.

PR #1526, opened 2021-02-06
Make process filtering and sorting more flexible

### LINKED ISSUE #1514: Filtering the Process list by name or group (or anything)
Hello,

Is there currently a way to filter what processes to query in terms of the process name or the process group or something?

Based on the oshi documentation, we can only use:
- [getProcesses()](https://oshi.github.io/oshi/apidocs/oshi/software/os/OperatingSystem.html#getProcesses()) - get all of the processes
- [getProcesses​(int limit, OperatingSystem.ProcessSort sort)](https://oshi.github.io/oshi/apidocs/oshi/software/os/OperatingSystem.html#getProcesses(int,oshi.software.os.OperatingSystem.ProcessSort)) - get X processes sorted using criteria ProcessSort
- [getProcesses​(java.util.Collection<java.lang.Integer> pids)](https://oshi.github.io/oshi/apidocs/oshi/software/os/OperatingSystem.html#getProcesses(java.util.Collection)) - get processes based on the supplied pids.

The current use case is that a user wants to monitor a specific set of processes only. The first API returns too much information (especially for busy systems), and the second and third ones cannot be used for that use case.

Thanks!

dbwiddis:
Fixes #1514

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/OperatingSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/OperatingSystem.java
@@ -26,12 +26,18 @@ package oshi.software.os;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Objects;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 import oshi.annotation.concurrent.Immutable;
 import oshi.annotation.concurrent.ThreadSafe;
+import oshi.driver.unix.Who;
+import oshi.driver.unix.Xwininfo;
 import oshi.software.os.OSProcess.State;
 import oshi.util.Constants;
+import oshi.util.ExecutingCommand;
+import oshi.util.ParseUtil;
 import oshi.util.Util;
 
 /**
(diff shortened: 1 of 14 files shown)


dbwiddis:
Co-maintainers: request feedback on API and documentation (focus on OperatingSystem interface).   If you also want to do code review, that's a bonus.

oshi-core/src/main/java/oshi/software/os/OperatingSystem.java
```diff
@@ -131,7 +238,10 @@
      *         specified. The list may contain null elements or processes with a
      *         state of {@link OSProcess.State#INVALID} if a process terminates
      *         during iteration.
+     * @deprecated Use {@link #getProcesses(Predicate, Comparator, int)} with
+     *             sorting constants from {@link ProcessSorting}.
      */
+    @Deprecated
     List<OSProcess> getProcesses(int limit, ProcessSort sort);
```
mprins, line 245: perhaps this can be a default method, so implementers can forget about it

oshi-core/src/main/java/oshi/software/os/OperatingSystem.java
```diff
@@ -174,9 +284,39 @@
      *         provided PID, sorted as specified. The list may contain null elements
      *         or processes with a state of {@link OSProcess.State#INVALID} if a
      *         process terminates during iteration.
+     * @deprecated Use {@link #getChildProcesses(int, Predicate, Comparator, int)}
+     *             with sorting constants from {@link ProcessSorting}.
      */
+    @Deprecated
     List<OSProcess> getChildProcesses(int parentPid, int limit, ProcessSort sort);
```
mprins, line 291: perhaps this can be a default method, so implementers can forget about it

mprins (approved):
perhaps the deprecated methods in `OperatingSystem` can be changed to `default` methods, effectively that is what the abstract class does now, so just a (design) thought.

LGTM

oshi-core/src/main/java/oshi/software/os/OperatingSystem.java
dbwiddis: Thanks, great suggestion.  Probably can be applied throughout other classes eventually.

dbwiddis: Well I tried to do this, but the implementation uses a private helper method common to the two deprecated methods (to convert the `ProcessSort` enum to the appropriate constant).  And private methods in interfaces is a JDK9+ feature.

So for now I'll just leave the implementation as-is.

dbwiddis: Or, actually, I had a nested class with the new constants which is a perfect place for a temporary private helper method to use those constants. Problem solved.

PR #1535, opened 2021-02-19
Fix NetworkIF.isConnectorPresent() on Windows

dornand:
Windows bit fields are in little-endian order.


dbwiddis:
Thanks, looks great! Can you also add a changelog entry?

PR #1541, opened 2021-02-22
Expose the alias of a network interface

dornand:
`NetworkIF.getIfAlias()` for Windows and Linux.  Relates to #1541

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/NetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/NetworkIF.java
@@ -62,11 +62,22 @@ public interface NetworkIF {
     String getDisplayName();
 
     /**
-     * Interface alias.
+     * The {@code ifAlias} as described in RFC 2863.
+     * <p>
+     * The ifAlias object allows a network manager to give one or more interfaces
+     * their own unique names, irrespective of any interface-stack relationship.
+     * Further, the ifAlias name is non-volatile, and thus an interface must retain
+     * its assigned ifAlias value across reboots, even if an agent chooses a new
+     * ifIndex value for the interface.
+     * <p>
+     * Only implemented for Windows and Linux.
      *
-     * @return The 'ifAlias' as described in RFC 2863.
+     * @return The {@code ifAlias} of the interface if available, otherwise the
+     *         empty string.
      */
-    String getIfAlias();
+    default String getIfAlias() {
+        return "";
+    }
 
     /**
      * The interface Maximum Transmission Unit (MTU).
--- a/oshi-core/src/main/java/oshi/hardware/NetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/NetworkIF.java
@@ -62,6 +62,13 @@ public interface NetworkIF {
     String getDisplayName();
 
     /**
+     * Interface alias.
+     *
+     * @return The 'ifAlias' as described in RFC 2863.
+     */
+    String getIfAlias();
+
+    /**
      * The interface Maximum Transmission Unit (MTU).
      *
      * @return The MTU of the network interface.
--- a/oshi-core/src/main/java/oshi/hardware/NetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/NetworkIF.java
@@ -159,7 +159,9 @@ public interface NetworkIF {
      *
      * @return the ifType
      */
-    int getIfType();
+    default int getIfType() {
+        return 0;
+    }
 
     /**
      * (Windows Vista and higher only) The NDIS physical medium type. This member
@@ -168,7 +170,9 @@ public interface NetworkIF {
      *
      * @return the ndisPhysicalMediumType
      */
-    int getNdisPhysicalMediumType();
+    default int getNdisPhysicalMediumType() {
+        return 0;
+    }
 
     /**
      * (Windows Vista and higher) Set if a connector is present on the network
@@ -179,7 +183,9 @@ public interface NetworkIF {
      * @return {@code true} if there is a physical network adapter (Windows) or a
      *         connected cable (Linux), false otherwise
      */
-    boolean isConnectorPresent();
+    default boolean isConnectorPresent() {
+        return false;
+    }
 
     /**
      * <p>
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
@@ -43,6 +43,7 @@ import org.slf4j.LoggerFactory;
 
 import oshi.annotation.concurrent.ThreadSafe;
 import oshi.hardware.NetworkIF;
+import oshi.util.Constants;
 import oshi.util.FileUtil;
 import oshi.util.FormatUtil;
 import oshi.util.ParseUtil;
@@ -58,6 +59,8 @@ public abstract class AbstractNetworkIF implements NetworkIF {
     private static final String OSHI_VM_MAC_ADDR_PROPERTIES = "oshi.vmmacaddr.properties";
 
     private NetworkInterface networkInterface;
+    private String name;
+    private String displayName;
     private int mtu;
     private String mac;
     private String[] ipv4;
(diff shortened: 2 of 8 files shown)


dbwiddis:
Note to self: for macOS, looks like [SCNetworkInterfaceCopyAll](https://developer.apple.com/documentation/systemconfiguration/1517090-scnetworkinterfacecopyall?language=objc) can enumerate interfaces. Not sure whether any of the other commands include the alias though.

dbwiddis:
Looking more into this, what is the difference between the Alias and the "DisplayName" we already use?

dornand:
On Windows i get:
Name: eth1
Display name: Intel(R) Ethernet Connection (4) I219-V
Alias: Ethernet

`ipconfig` is using the alias here:
Ethernet adapter _Ethernet_:

   Connection-specific DNS Suffix  . : 
   Link-local IPv6 Address . . . . . : fe80::2916...
   IPv4 Address. . . . . . . . . . . : 10.11...
   Subnet Mask . . . . . . . . . . . : 255.255.255.0
   Default Gateway . . . . . . . . . : 10.11...

Also the network adapter is displayed as "Ethernet" in the control center.

dornand:
Actually on Windows `NetworkInterface.getDisplayName()` is returning the device description (=MIB_IF_ROW2.Description), instead of the "real" display name, which is the alias.

dornand:
In general though, the alias is the custom name, given to a network interface by an administrator.
So `NetworkIF.getAlias()` could be useful, where such custom names exist.

dbwiddis:
Thanks for the details.  There's probably a default version we can produce parsing the output of `ifconfig` on *nix machines.  I'll research more tonight.

dbwiddis:
Interestingly, looking at the JDK source code, the [description](https://github.com/openjdk/jdk/blob/master/src/java.base/windows/native/libnet/NetworkInterface.c#L265) is fetched, probably because they chose the older [MIB_IFROW](https://docs.microsoft.com/en-us/windows/win32/api/ifmib/ns-ifmib-mib_ifrow) structure for compatibility.  All the other implementations [just use the same value for name and displayName](https://github.com/openjdk/jdk/blob/master/src/java.base/unix/native/libnet/NetworkInterface.c#L719).

I'm still confused, though, it seems you can assign [multiple aliases](https://linuxtechlab.com/ip-aliasing-multiple-ip-single-nic/) to a NIC. How do we handle that (or how does `sysfs` handle it)?  Or is the `IfAlias` we're mapping completely different from an "IP Alias"?  In which case perhaps it should be named `getIfAlias()`.

oshi-core/src/main/java/oshi/hardware/NetworkIF.java
```diff
@@ -61,6 +61,13 @@
      */
     String getDisplayName();
 
+    /**
+     * Interface alias.
```
dbwiddis, line 65: Let's me more verbose and describe the IfAlias per the RFC so it's clear what we're talking about.

oshi-core/src/main/java/oshi/hardware/NetworkIF.java
```diff
@@ -61,6 +61,13 @@
...
+     *
+     * @return The alias of the network interface.
+     */
+    String getAlias();
```
dbwiddis, line 69: I think `getIfAlias()` will make it more clear what this means.

oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
```diff
@@ -229,6 +229,12 @@ public boolean isConnectorPresent() {
         return false;
     }
 
+    @Override
+    public String getAlias() {
+        // default
+        return "";
```
dbwiddis, line 235: Is there a default parsing of `ifconfig` we can put in here that's reasonably consistent on *nix machines?

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsNetworkIF.java
```diff
@@ -179,6 +185,7 @@ public boolean updateAttributes() {
             this.collisions = ifRow.OutDiscards; // closest proxy
             this.inDrops = ifRow.InDiscards; // closest proxy
             this.speed = ifRow.ReceiveLinkSpeed;
+            this.alias = new String(ifRow.Alias).trim();
```
dbwiddis, line 188: Use JNA's `Native.toString(char[])` here.

dbwiddis (commented):
Some suggestions while I research the other platforms.

dornand:
I think IP Aliasing is a completely different thing. That's why we have `String[] getIPv4addr()`, to map several IP addresses.

dbwiddis:
Yeah. I've spent way too much time going down rabbit holes on this.  My observation:
 - Windows has 3 string fields, the name, alias, and description
 - The JDK uses the description for the "displayName" and ignores the alias (for compatibility) on Windows
 - The JDK simply duplicates the name into displayName for all other OS's
 - macOS does have a "LocalizedDisplayName" accessible in the System Configuration framework which isn't really an ifAlias. Would be nice to override the display name for macOS though.
 - Other than that there's no easy/consistent/fast way to get to it.

I'll see if I can work up the macOS change to the display name tomorrow, other than that this is probably as good as we're going to get.  Probably want to tweak the toString.

dbwiddis:
Oh, is there a "description" field in Linux we can put in the displayName field?

dornand:
i'll have to check later. could not find "ifDescr" exposed in sysfs.

i found a discussion suggesting to use the "pci_lookup_name" for ifDescr:
[https://sourceforge.net/p/net-snmp/mailman/message/26092573/](https://sourceforge.net/p/net-snmp/mailman/message/26092573/)

dbwiddis:
That's probably overkill.  Another option is `lspci` for example
```
00:05.0 Ethernet controller: Red Hat, Inc. Virtio network device
```
PCI bus 0, Slot 5 `enp0s5`: `/sys/class/net/enp0s5`

But I don't like command lines when I can avoid them.

dbwiddis:
I think I may have a native solution for the name using udev.  
```
$ udevadm info /sys/class/net/enp0s5
...
E: ID_MODEL_FROM_DATABASE=Virtio network device
E: ID_MODEL_ID=0x1000
...
E: ID_VENDOR_FROM_DATABASE=Red Hat, Inc.
E: ID_VENDOR_ID=0x1af4
...
```
However that may just be the same "pci_lookup_name" lookup you described earlier from the model and vendor ID which I expect are keys, while the DB lookup is probably done internally in `udevadm` and not stored...

dbwiddis:
See code in `LinuxHWDiskStore` starting at line 167 as an example, but use the sysfs path and try the `ID_MODEL_FROM_DATABASE` property value above.

dbwiddis:
I'm good to merge this unless you think there is a need for any other changes.  I'd appreciate you testing on your own Linux as well.  It was pretty trivial to grab the vendor and model strings from udev, simulating the `pci_lookup_name` approach.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxNetworkIF.java
```diff
@@ -162,6 +194,7 @@ public boolean updateAttributes() {
         String collisionsPath = String.format("/sys/class/net/%s/statistics/collisions", getName());
         String rxDropsPath = String.format("/sys/class/net/%s/statistics/rx_dropped", getName());
         String ifSpeed = String.format("/sys/class/net/%s/speed", getName());
+        String ifAlias = String.format("/sys/class/net/%s/ifalias", getName());
```
dbwiddis, line 197: CI complains this local variable name shadows the class field name.  Must be a popular name. :)

dornand:
Ah, now we need to test those default methods separately.
Should I just make a NetworkIFTest?

dbwiddis:
Not sure what you mean about the testing? Should give identical results to the previous version?   (We mostly just test the API... yes I know that skips a lot... we'll do it right on the redesign from scratch! :grin: )

dornand:
Just noticed the failing coverage check  :)

dbwiddis:
Yeah it's annoying but I can't find a way to set a threshold. It counts comments too!  We can ignore it -- you added the necessary test for the new API.

dbwiddis:
Coverage is really only reported on the linux branch but we run all the tests on all 7 supported OS's. 

So are we good to merge this? Anything else you need to add?

dornand:
I'm ok to merge. Thanks for the support!

PR #1545, opened 2021-02-24
Fix a NPE in NetworkIF.toString() on Windows.

dornand:
ifAlias was not initialized properly on Windows before Vista.


oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsNetworkIF.java
```diff
@@ -65,7 +65,7 @@
     private long collisions;
     private long speed;
     private long timeStamp;
-    private String ifAlias;
+    private String ifAlias = "";
```
dbwiddis, line 68: This works but it might make more sense to explicitly set it in the pre-vista branch (around line 210) to correspond directly to the line 189 setting.

PR #1546, opened 2021-02-24
Expose network interface operational status

dornand:
Expose the `ifOperStatus` as described in [RFC 2863](https://tools.ietf.org/html/rfc2863) as `NetworkIF.getIfOperStatus()`.

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/NetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/NetworkIF.java
@@ -24,6 +24,7 @@
 package oshi.hardware;
 
 import java.net.NetworkInterface;
+import java.util.Arrays;
 
 import oshi.annotation.concurrent.ThreadSafe;
 
@@ -70,13 +71,24 @@ public interface NetworkIF {
      * its assigned ifAlias value across reboots, even if an agent chooses a new
      * ifIndex value for the interface.
      * <p>
-     * Only implemented for Windows and Linux.
+     * Only implemented for Windows (Vista and newer) and Linux.
      *
      * @return The {@code ifAlias} of the interface if available, otherwise the
      *         empty string.
      */
     default String getIfAlias() {
         return "";
+    }
+
+    /**
+     * The {@code ifOperStatus} as described in RFC 2863.
+     * <p>
+     * Only implemented for Windows (Vista and newer) and Linux.
+     *
+     * @return The current operational state of the interface.
+     */
+    default IfOperStatus getIfOperStatus() {
+        return IfOperStatus.UNKNOWN;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


oshi-core/src/main/java/oshi/hardware/IfOperStatus.java
```diff
@@ -0,0 +1,88 @@
+/**
```
dbwiddis, line 1: There's an extra `*` here!   :-)

More seriously, though.  Since this `Enum` is associated very specifically with the `NetworkIF` interface, I'd like it as a nested class there, similar to the `State` enum in `OSProcess`.

oshi-core/src/main/java/oshi/hardware/IfOperStatus.java
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Some component is missing
+     */
+    NOT_PRESENT(6),
+    /**
+     * Down due to state of lower-layer interface(s).
+     */
+    LOWER_LAYER_DOWN(7);
+
+    int value;
```
dbwiddis, line 65: This should be `private`.

oshi-core/src/main/java/oshi/hardware/NetworkIF.java
```diff
@@ -79,6 +79,15 @@ default String getIfAlias() {
         return "";
     }
 
+    /**
+     * The {@code ifOperStatus} as described in RFC 2863.
```
dbwiddis, line 83: Add some caveats about Windows and Linux only, similar to the `getIfAlias()` ones.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxNetworkIF.java
```diff
@@ -211,7 +219,29 @@ public boolean updateAttributes() {
         // speed may be -1 from file.
         this.speed = speedMiB < 0 ? 0 : speedMiB << 20;
         this.ifAlias = FileUtil.getStringFromFile(ifAliasPath);
+        this.ifOperStatus = mapIfOperStatus(FileUtil.getStringFromFile(ifOperStatusPath));
 
         return true;
     }
+
+    private IfOperStatus mapIfOperStatus(String operState) {
```
dbwiddis, line 227: (Preference) I'd avoid using "map" unless the return value is a `Map<>`.  I think `parse...` may be better?

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxNetworkIF.java
```diff
@@ -211,7 +219,29 @@ public boolean updateAttributes() {
...
+        switch (operState) {
+            case "up":
+                return IfOperStatus.UP;
+            case "down":
+                return IfOperStatus.DOWN;
+            case "testing":
+                return IfOperStatus.TESTING;
+            case "unknown":
```
dbwiddis, line 235: You could move this case right above `default:` and combine it.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsNetworkIF.java
```diff
@@ -187,6 +194,7 @@ public boolean updateAttributes() {
             this.inDrops = ifRow.InDiscards; // closest proxy
             this.speed = ifRow.ReceiveLinkSpeed;
             this.ifAlias = Native.toString(ifRow.Alias);
+            this.ifOperStatus = IfOperStatus.byValue(ifRow.OperStatus);
```
dbwiddis, line 197: It looks to me like the Windows value here is 0-indexed while the `byValue()` method assumes 1-indexed. Can you test/confirm?

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsNetworkIF.java
```diff
@@ -208,6 +216,8 @@ public boolean updateAttributes() {
             this.collisions = ParseUtil.unsignedIntToLong(ifRow.dwOutDiscards); // closest proxy
             this.inDrops = ParseUtil.unsignedIntToLong(ifRow.dwInDiscards); // closest proxy
             this.speed = ParseUtil.unsignedIntToLong(ifRow.dwSpeed);
+            this.ifAlias = ""; // not supported by MIB_IFROW
+            this.ifOperStatus = IfOperStatus.UNKNOWN; // TODO find a mapping
```
dbwiddis, line 220: Don't think this is a TODO.  I think it's safe to ignore doing anything extraordinary for XP / Win Server 2003 and just say not supported.

dbwiddis (changes requested):
Looks good. A few requests inline.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsNetworkIF.java
dbwiddis: Actually the Windows ones are properly indexed, so no problem here.  From the SDK:
```
//
// OperStatus values from RFC 2863
//
typedef enum {
    IfOperStatusUp = 1,
    IfOperStatusDown,
    IfOperStatusTesting,
    IfOperStatusUnknown,
    IfOperStatusDormant,
    IfOperStatusNotPresent,
    IfOperStatusLowerLayerDown
} IF_OPER_STATUS;
```

dbwiddis:
LGTM. I'll merge once you click the "Ready for review" button. :)

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsNetworkIF.java
dornand: Yes, i checked here: [https://docs.microsoft.com/en-us/windows/win32/api/netioapi/ns-netioapi-mib_if_row2](https://docs.microsoft.com/en-us/windows/win32/api/netioapi/ns-netioapi-mib_if_row2).

Also did some testing on my machine. At least UP and DOWN are mapped correctly.

dbwiddis (approved):

dbwiddis:
Actually one last change. Scrutinizer suggests making the `parseIfOperStatus` method `static`. I'll grab that quickly and then merge.

dornand:
Yes, static makes sense there.

PR #1556, opened 2021-03-04
Refactored tree traversal to nonrecursive BFS

jspinn:


oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
```diff
@@ -128,6 +131,11 @@ private int queryPlatformBitness() {
 
     protected abstract List<OSProcess> queryDescendantProcesses(int parentPid);
 
+    private static Set<Integer> getChildren(Map<Integer, Integer> parentPidMap, int parentPid) {
+        return parentPidMap.entrySet().stream().filter(e -> e.getValue().equals(parentPid))
```
dbwiddis, line 135: Might be useful to screen out the parent pid == pid case in this filter so you don't have to handle it separately later.

oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
```diff
@@ -136,29 +144,14 @@ private int queryPlatformBitness() {
...
-        // Recurse
-        if (recurse) {
-            for (int pid : childPids) {
-                if (pid != parentPid && !descendantPids.contains(pid)) {
-                    addChildrenToDescendantSet(allProcs, pid, descendantPids, true);
-                }
-            }
-        }
+    protected static Set<Integer> getChildrenOrDescendants(Collection<OSProcess> allProcs, int parentPid, boolean allDescendants) {
+        Map<Integer, Integer> parentPidMap = allProcs.stream().collect(Collectors.toMap(OSProcess::getProcessID, p -> p.getParentProcessID()));
```
dbwiddis, line 153: I knew there was some sort of stream map thing. Thanks for reminding me what it was. I like this approach. Is it not possible to do `toMap(OSProcess::getProcessID, OSProcess::getParentProcessID)`?  Also, would be good to add a filter before entering in the map to verify parent start time is equal to or before child start time.

dbwiddis: Also, using this for the Solaris branch is actually probably more efficient (a single `ps` command line) than the existing multiple `pgrep` command lines (at the cost of the extra ps parsing for everything else).   Might be useful in the "descendants" case in Solaris to just call this method, keeping the `pgrep` option only for children.

oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
```diff
@@ -170,29 +163,36 @@ protected static void addChildrenToDescendantSet(Collection<OSProcess> allProcs,
...
+     * @param allDescendants
      *            If false, only gets immediate children of this process. If true,
      *            gets all descendants.
+     * @return Set of children or descendants of parentPid
      */
-    protected static void addChildrenToDescendantSet(Map<Integer, Integer> parentPidMap, int parentPid,
-            Set<Integer> descendantPids, boolean recurse) {
+    protected static Set<Integer> getChildrenOrDescendants(Map<Integer, Integer> parentPidMap, int parentPid,
+            boolean allDescendants) {
+        Set<Integer> descendantPids = new HashSet();
```
dbwiddis, line 173: Make sure you're Java 8 compatible.  I think this and the next line needs the diamond operator (`new HashSet<>()`).

oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
```diff
@@ -170,29 +163,36 @@ protected static void addChildrenToDescendantSet(Collection<OSProcess> allProcs,
...
+        Queue<Integer> queue = new LinkedList();
```
dbwiddis, line 174: Not a big deal, but I think `ArrayDeque` may be a faster and more memory efficient `Queue` implementation.  

Rule of thumb: nearly always*, array implementations have better performance than linked lists.
* Unless you have a need to remove elements while iterating, which is not the case here.

oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
```diff
@@ -170,29 +163,36 @@ protected static void addChildrenToDescendantSet(Collection<OSProcess> allProcs,
...
-                    addChildrenToDescendantSet(parentPidMap, pid, descendantPids, true);
+        // Add all descendents
+        if (allDescendants) {
+            queue.addAll(descendantPids);
+            while (!queue.isEmpty()) {
+                currPid = queue.remove();
+                // Collect current process's children
+                childPids = getChildren(parentPidMap, currPid);
+                for (int pid : childPids) {
+                    if (pid != parentPid && !descendantPids.contains(pid)) {
```
dbwiddis, line 188: Per my earlier comment if you filter out the parent in the `getChildren()` method you can skip it here.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -182,16 +182,15 @@ public OSProcess getProcess(int pid) {
             }
         } else {
             // Only put descendants in
```
dbwiddis, line 184: Your changes work for this but looking at the whole method there's opportunity for more improvement.  You could make this branch the top one (`if (parentPid >= 0)`) and just do `return queryProcessList(getChildrenOrDescendants...)`. With the return no need for `else`.  Then handle the other (`-1`) case by declaring the set and adding to it as is currently done.  I see a bug, we shouldn't add the default `0` to the list of children.  Can you fix that please? :)

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOperatingSystem.java
```diff
@@ -129,32 +130,49 @@ public OSProcess getProcess(int pid) {
...
         }
         return getProcessListFromPS("ps -o s,pid,ppid,user,uid,group,gid,nlwp,pri,vsz,rss,etime,time,comm,args -p "
                 + String.join(",", descendantPids), -1);
     }
 
-    private static void addChildrenToDescendantSet(String parentPid, Set<String> descendantPids, boolean recurse) {
-        // Get list of children
+    private static Set<String> getChildren(String parentPid) {
         Set<String> childPids = new HashSet<>();
         for (String s : ExecutingCommand.runNative("pgrep -P " + parentPid)) {
```
dbwiddis, line 143: see earlier comments about only using this for the one-generation branch and using the superclass method for all descendants.

dbwiddis (changes requested):
Thanks for this submission!  Great work.  I've done an initial review just reading the diff and left some comments inline.   I'll do some testing later.

oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
jspinn: I think I've got all of the other fixes sorted out besides these two.  For the start time filtering, I'm thinking I just need to change it to `allProcs.stream().filter( p -> p.getStartTime() >= ???  ).collect(Collectors.toMap(OSProcess::getProcessID, OSProcess::getParentProcessID))` but I'm not sure how to get the parent's start time.

For the Solaris part, would this change be correct (leaving everything else the same)?

```       
if (allDescendants) {
        if (allDescendants) {
            Collection<OSProcess> allProcs = getProcessListFromPS("ps -eo s,pid,ppid,user,uid,group,gid,nlwp,pri,vsz,rss,etime,time,comm,args", -1);
            descendantPids = getChildrenOrDescendants(allProcs, Integer.parseInt(parentPid), allDescendants)
                    .stream().map(String::valueOf).collect(Collectors.toSet());
        }
} 
```

dbwiddis: RE: start time... things are always [easier to say than to do](https://xkcd.com/1425/), eh?

Follow along with my train of thought here: 
 - It seems we'd need to have already mapped the processes to their start times using a similar iteration but with start time instead of parent pid.  
 - But rather than iterating twice we could just iterate once here and create a map of pid to OSProcess
 - This would change the map parameter passed to `getChildrenOrDescendants` to `Map<Integer, OSProcess>`
 - This would require changing the map generation in `LinuxOperatingSystem.getParentPidsFromProcFiles()` and `WindowsOperatingSystem.getParentPidsFromSnapshot()` to include start time.  But this is not trivial on Linux, and isn't in the snapshot on Windows.  (And in Linux it'd be faster to compare up times.)

So, nevermind.  Just leave it for now, maybe I'll make that another improvement at a later time.

RE... Solaris.  You've got two nested if's asking the same thing, but ignoring that what you have works but duplicates code.  Just call `queryAllProcesses()`.  So I think it'd just be:
```
if (allDescendants) {
    descendantPids = getChildrenOrDescendants(queryAllProcesses(), Integer.parseInt(parentPid), true)
                    .stream().map(String::valueOf).collect(Collectors.toSet());
}
```
Except also, don't directly use `parseInt` that could throw an exception, use ParseUtil.parseIntOrDefault() with a negative pid default that will not match any children.

Also I recall some obscure fact comparing `String.valueOf()` vs. `Integer.toString()` and localization/number format differences and preferring one over the other. I'll have to look that up.

dbwiddis: Ok, either of those is identical. `String.valueOf()` simply calls `Integer.toString()` directly (and one would assume the JIT compiler would produce identical bits).   The obscure thing I was thinking of was using `%d` in `String.format()` which can produce odd results. Using `%s` invokes `Integer.toString()` and produces consistent output (e.g., 1234 not 1,234).

jspinn: Woops sorry I didn't realize I duplicated the if when I copy/pasted the code here!

Using `queryAllProcesses()` wasn't working because of the issue of calling a non-static method from a static method, so that's why I did it that way.  What else can I do to get around this?

Sounds good, I'll change the `parseInt` part. Thanks!

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
jspinn: My first instinct to fix the default 0 bug was to replace it with `Integer.parseInt()` but from seeing your other reply I'm guessing this isn't correct because it could throw an exception. Should I be changing it to `ParseUtil.parseIntOrDefault(procFile.getName(), -1)` instead?

jspinn: Or actually should I just add an if statement to only add it if it's not 0?

oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
dbwiddis: Just remove the static modifier.  :)

dbwiddis: Or create a common static method that both call.  The only reason that is not static is it implements an abstract method in the superclass.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
dbwiddis: In some OS's (Windows) 0 is a valid PID.  In *nix, I think 1 is the minimum, so 0 could be safe here in later code but still shouldn't be added.  Any negative number should also work*, but note that a few methods use -1 as the input PID to mean "get all the PIDs" so you'll have to be careful.

But as you just said, if you separate the retrieving of the value and an if statement adding it to the set, then you can use any value.

* Actually if windows has a process ID higher than Integer.MAX_VALUE it would show up as negative!  But this is Linux so that doesn't matter.

jspinn: Thanks for all the information and help! Last question (for now at least), when I'm finished with my changes am I supposed amend and push my first commit or push a new commit?

dbwiddis: Either is fine.  GitHub has a Squash and merge” option for me to use.  

Some other projects prefer you squash and set your real name and email as well.  I’m not that picky. :)

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOperatingSystem.java
```diff
@@ -129,32 +131,43 @@ public OSProcess getProcess(int pid) {
 
     @Override
     public List<OSProcess> queryDescendantProcesses(int parentPid) {
-        Set<String> descendantPids = new HashSet<>();
-        addChildrenToDescendantSet(Integer.toString(parentPid), descendantPids, true);
+        Set<String> descendantPids = getChildrenOrDescendants(Integer.toString(parentPid), true);
```
dbwiddis, line 134: I missed this in the earlier discussion.  There's no need to convert `parentPid` to a String here just to convert it back in the later method.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOperatingSystem.java
```diff
@@ -129,32 +131,43 @@ public OSProcess getProcess(int pid) {
...
-            if (!pid.equals(parentPid) && !descendantPids.contains(pid)) {
+            if (!pid.equals(parentPid)) {
                 childPids.add(pid);
             }
         }
+        return childPids;
+    }
+
+    private static Set<String> getChildrenOrDescendants(String parentPid, boolean allDescendants) {
+        Set<String> descendantPids = new HashSet<>();
```
dbwiddis, line 158: You don' need to create a queue just to add the results of `getChildren()`.  Just return that set directly in the case where `allDescendants` is false.  SO this whole method would be simply:
```
// note change of argument to int
private static Set<String> getChildrenOrDescendants(int parentPid, boolean allDescendants) {
    if (allDescendants) {
        // no need to parse parentPid to int in this call
        return getChildrenOrDescendants( ... );
    }
    // change this to take an int argument
    return getChildren(parentPid);
}
```

dbwiddis (commented):
Looks really good!  The main parts look great, just some final cleanup/simplification in the Solaris version.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOperatingSystem.java
dbwiddis: Actually it's even simpler.  Just call `getChildren` from `queryChildProcesses()` and call `getChildrenOrDescendants()` directly from `queryDescendantProcesses()`. No need for this intermediate method to combine the approaches and split them out again, so it can be deleted.

dbwiddis (approved):

dbwiddis:
Sanity check passes:
```
parent->pid->children [descendants]
1->3901-->3913 [ 41600 41599 3916 3913 ]
3901->3913-->3916 [ 41600 41599 3916 ]
3913->3916-->41600 41599 [ 41600 41599 ]
```

dbwiddis:
Thanks for your contribution!  Very much appreciated.

jspinn:
Thank you for all your help to make my contribution!

PR #1582, opened 2021-03-24
Fix mapping of filestores with logical volumes

tausiflife:


dbwiddis:
Looks great. Probably worthy of a change log entry.  Go ahead and merge!

dbwiddis (approved):

PR #1584, opened 2021-03-27
Add logical volume group information

tausiflife:
1. Its name
2. Its physical volumes
3. Its logical volumes and what physical volume they are using.


oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -273,4 +280,73 @@ private static long getFileDescriptors(int index) {
...
+     *
+     * Instantiates a list of {@link VolumeGroup} objects,
+     * representing a storage pool, device, partition, volume, concrete file system
+     * or other implementation specific means of file storage.
+     *
+     * @return A list of {@link VolumeGroup} objects or an empty
+     *      *         array if none are present.
+     */
+
+    public List<VolumeGroup> getLogicalVolumeGroups() {
```
dbwiddis, line 295: Per discussion on #1578, I think this might be better on the Hardware Abstraction Layer.  Interface implementation should be `default` returning `Collections.emptyList()`.

oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -273,4 +280,73 @@ private static long getFileDescriptors(int index) {
...
+        Map<String, Map<String, List<String>>> logicalVolumesMap = new HashMap<>();
+        Map<String, Set<String>> physicalVolumesMap = new HashMap<>();
```
dbwiddis, line 297: I think there's value in having this as the same map setup as the logical volume map, e.g., quickly identifying "unused" physical volumes, and it doesn't cost us anything in the iteration to add to both maps when we find a linkage.

oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -273,4 +280,73 @@ private static long getFileDescriptors(int index) {
...
+                                    physicalVolumesMap.put(vgName, pvSetForGroup);
+
+                                    String lvName = device.getPropertyValue(DM_LV_NAME);
+                                    File slavesDir = new File(syspath + "/slaves");
+                                    File[] slaves = slavesDir.listFiles();
+                                    if (slaves != null) {
+                                        for (File f : slaves) {
+                                            String pvName = f.getName();
+                                            lvMapForGroup.computeIfAbsent(lvName, k -> new ArrayList<>()).add(pvName);
+                                            pvSetForGroup.add(pvName);
```
dbwiddis, line 330: Per above comment, if we keep a pvMap we can just add the lv here very simply.

oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -273,4 +280,73 @@ private static long getFileDescriptors(int index) {
...
+                    String syspath = entry.getName();
+                    Udev.UdevDevice device = udev.deviceNewFromSyspath(syspath);
+                    if (device != null) {
+                        try {
+                            // devnode is what we use as name, like /dev/sda
+                            String devnode = device.getDevnode();
+                            if (devnode != null && devnode.startsWith("/dev/dm")) {
+                                String uuid = device.getPropertyValue(DM_UUID);
+                                if (uuid != null && uuid.startsWith("LVM-")) {
+                                    String vgName = device.getPropertyValue(DM_VG_NAME);
```
dbwiddis, line 315: We might want to check for null/empty here.

oshi-core/src/main/java/oshi/software/os/linux/VolumeGroup.java
```diff
@@ -0,0 +1,57 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package oshi.software.os.linux;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class VolumeGroup {
```
dbwiddis, line 30: We should probably use the full name `LogicalVolumeGroup` here.

oshi-core/src/main/java/oshi/software/os/linux/VolumeGroup.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * @param name Name of the volume group
+     * @param lvMap Logical volumes derived from this volume group and the physical volumes its mapped to.
+     * @param pvSet Set of physical volumes this volume group consists of.
+     */
+    public VolumeGroup(String name, Map<String, List<String>> lvMap, Set<String> pvSet) {
+        this.name = name;
+        this.lvMap = lvMap;
+        this.pvSet = pvSet;
```
dbwiddis, line 43: Funny that scrutinizer complains about the set and not the map.  Still, we should probably wrap the map with `Collections.unmodifiableMap()` or the immutable version of it, and also add the appropriate `@ThreadSafe` or `@Immutable` annotation.

dbwiddis (commented):
Looks good so far. Per discussion on #1578, I think I'd still like to have the PVs be a map, and we also need another way to identify PVs which are not (yet) mapped to an LV, such as the `pvs` command or its native equivalent.

dbwiddis:
OK I made a few commits implementing some of my suggestions.

Still unresolved questions in my mind:
1. We use a `Set` for the PVs but then in the LV map we use a `List`.  These should probably be the same type (and probably a `Set`).
2. I still think it would be useful to have a set of PV-to-LV, but I do see how it's somewhat redundant, so I'm ok with not including that.
3. In the case where we don't have information on which LV maps to which PV (macOS at least) do we want to keep it an empty set for the map (leaving the choice to the user), or return the set of all PVs?

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
```diff
@@ -66,6 +67,15 @@
     private static final String ID_MODEL = "ID_MODEL";
     private static final String ID_SERIAL_SHORT = "ID_SERIAL_SHORT";
 
+    private static final String DM_UUID = "DM_UUID";
+    private static final String DM_VG_NAME = "DM_VG_NAME";
+    private static final String DM_LV_NAME = "DM_LV_NAME";
+    private static final String LOGICAL_VOLUME_GROUP = "Logical Volume Group";
+    private static final String DEV_NAME = "DEV_NAME";
+    private static final String DEV_PATH = "DEV_PATH";
+    private static final String DEV_LOCATION = "/dev";
```
dbwiddis, line 76: I think you'll find setting this constant to include the trailing slash will save you time elsewhere. :)

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
```diff
@@ -270,6 +300,27 @@ private static void computeDiskStats(LinuxHWDiskStore store, String devstat) {
         store.transferTime = devstatArray[UdevStat.ACTIVE_MS.ordinal()];
     }
 
+    private static String getPartitionNameForDmDevice(String vgName, String lvName) {
+        return new StringBuilder().append(DEV_LOCATION).append("/").append(vgName).append("/").append(lvName).toString();
```
dbwiddis, line 304: Minor point: some of our CI will say the append will be more efficient with a character rather than single character string, so I'd suggest single quotes on the slashes here.  Also if you edit DEV_LOCATION to include the trailing slash you won't need it here.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
```diff
@@ -66,6 +67,15 @@
...
     private static final String ID_SERIAL_SHORT = "ID_SERIAL_SHORT";
 
+    private static final String DM_UUID = "DM_UUID";
+    private static final String DM_VG_NAME = "DM_VG_NAME";
+    private static final String DM_LV_NAME = "DM_LV_NAME";
+    private static final String LOGICAL_VOLUME_GROUP = "Logical Volume Group";
+    private static final String DEV_NAME = "DEV_NAME";
+    private static final String DEV_PATH = "DEV_PATH";
+    private static final String DEV_LOCATION = "/dev";
+    private static final String DEV_MAPPER = DEV_LOCATION + "/mapper";
```
dbwiddis, line 77: Similarly, add slash after mapper and remove before if you've added it to DEV_LOCATION....

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
```diff
@@ -270,6 +300,27 @@ private static void computeDiskStats(LinuxHWDiskStore store, String devstat) {
         store.transferTime = devstatArray[UdevStat.ACTIVE_MS.ordinal()];
     }
 
+    private static String getPartitionNameForDmDevice(String vgName, String lvName) {
+        return new StringBuilder().append(DEV_LOCATION).append("/").append(vgName).append("/").append(lvName).toString();
+    }
+
+    private static String getMountPointOfDmDevice(String vgName, String lvName) {
+        return new StringBuilder().append(DEV_MAPPER).append("/").append(vgName).append("-").append(lvName).toString();
```
dbwiddis, line 308: Similar to above comment, use single quotes for single characters and possibly include the first slash in the DEV_MAPPER string

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
```diff
@@ -270,6 +300,27 @@ private static void computeDiskStats(LinuxHWDiskStore store, String devstat) {
...
+    }
+
+    private static String getDependentNamesFromHoldersDirectory(String sysPath) {
+        File holdersDir = new File(sysPath + "/holders");
+        File[] holders = holdersDir.listFiles();
+        StringBuilder dependentNames = new StringBuilder();
+        if (holders != null) {
+            for (File f : holders) {
+                String devName = f.getName();
+                dependentNames.append(DEV_LOCATION).append("/").append(devName).append(" ");
```
dbwiddis, line 318: I'm not sure you'll ever use `/dev` without the trailing slash. :)

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
```diff
@@ -270,6 +300,27 @@ private static void computeDiskStats(LinuxHWDiskStore store, String devstat) {
...
+    }
+
+    private static String getMountPointOfDmDevice(String vgName, String lvName) {
+        return new StringBuilder().append(DEV_MAPPER).append("/").append(vgName).append("-").append(lvName).toString();
+    }
+
+    private static String getDependentNamesFromHoldersDirectory(String sysPath) {
+        File holdersDir = new File(sysPath + "/holders");
+        File[] holders = holdersDir.listFiles();
+        StringBuilder dependentNames = new StringBuilder();
```
dbwiddis, line 314: Consider a `StringJoiner` here to avoid having to deal with that extra delimiter.   `new StringJoiner(" ");` and then `.add()` the `f.getNames()` to it.  Or, alternately, `return Arrays.stream(holders).map(f -> f.getName()).collect(Collectors.joining(" "));`

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxLogicalVolumeGroup.java
```diff
@@ -0,0 +1,122 @@
...
+                                        File slavesDir = new File(syspath + "/slaves");
+                                        File[] slaves = slavesDir.listFiles();
+                                        if (slaves != null) {
+                                            for (File f : slaves) {
+                                                String pvName = f.getName();
+                                                lvMapForGroup.computeIfAbsent(lvName, k -> new ArrayList<>())
+                                                        .add(pvName);
+                                                // Backup to add to pv set if pvs command failed
+                                                //Added /dev to remove duplicates like /dev/sda1 and sda1
+                                                pvSetForGroup.add("/dev/" + pvName);
```
dbwiddis, line 101: Either don't prepend `/dev/` here or prepend it to the `pvName` in the `lvMap` above.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxLogicalVolumeGroup.java
```diff
@@ -0,0 +1,122 @@
...
+                                        physicalVolumesMap.computeIfAbsent(vgName, k -> new HashSet<>());
+                                        Set<String> pvSetForGroup = physicalVolumesMap.get(vgName);
+
+                                        File slavesDir = new File(syspath + "/slaves");
+                                        File[] slaves = slavesDir.listFiles();
+                                        if (slaves != null) {
+                                            for (File f : slaves) {
+                                                String pvName = f.getName();
+                                                lvMapForGroup.computeIfAbsent(lvName, k -> new ArrayList<>())
+                                                        .add(pvName);
```
dbwiddis, line 98: Either prepend `/dev/` to this `pvName` or don't prepend it to the one below.

dbwiddis (commented):
This looks really good!  Output you posted in the issue looks great, thanks for working on this.

I've left a bunch of extremely minor nitpick comments/suggestions.

dbwiddis:
Windows port done, in progress. Need to add javadocs to the driver class, and change list to set for the LV map.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxLogicalVolumeGroup.java
tausiflife: I have added `/dev` here because the `pvs -o vg_name,pv_name` adds values like `/dev/sda1` but the slaves directory adds only `sda1`.

tausiflife: ok

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
tausiflife: your way is much better than mine :)

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
tausiflife: ok

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
tausiflife: noted.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
tausiflife: yes makes it much easier.

dbwiddis:
This looks really good!   I'll review the code again this evening but I think this is ready to merge.

dbwiddis (approved):

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
dbwiddis: Actually it's even better with a lambda:
```
Arrays.stream(holders).map(File::getName).collect(Collectors.joining(" "));
```

PR #1605, opened 2021-04-14
Cache localized perf counter object strings

dbwiddis:
Profiling identifies the largest significant non-native CPU time is associated with the localization of PDH counter strings.  The localization code is inefficient, an O(n<sup>2</sup>) process where every counter string is individually localized by querying the entire list of strings to fetch an index, which is then passed to native code to query the localized value.

These localized strings don't change so it makes sense to cache them.  It's still slightly inefficient for the first insertion but will be non-blocking O(1) lookup once the string has been initially localized.


mprins (approved):
untested as I don't run windows, but makes sense.

PR #1607, opened 2021-04-15
More efficiently test for local network interfaces

### LINKED ISSUE #1606: Loopback network interface test is redundant
Profiling shows the largest CPU impact in OSHI is associated with network interface queries.  Running all queries from `SystemInfoTest` in a loop results in 1/3 of CPU time being associated with the `isLocalInterface()` method, split equally between calls to (JDK methods) `networkInterface.isLoopback()` and `networkInterface.getHardwareAddress()`.  

This delay is evident in logs: querying network interfaces while excluding local addresses (on Windows) takes 2.1 seconds out of a total of 7 seconds to run `SystemInfoTest`.

These calls should be cached or memoized in some manner.

dbwiddis:
Fixes #1606


dbwiddis:
While this was a fun attempt for loopback, we need to also test for no mac address, which increases the complexity well beyond any incremental benefit.  Closing this in favor of a simpler solution reducing redundant checks.

Issue #1608, opened 2021-04-15
LinuxOSProcess#getOpenFiles returns one more than expected


slaha:
Hi all,

I think there is a bug in `LinuxOSProcess` when getting number of opened files. The method `LinuxOSProcess#getOpenFiles` executes `ls -f /proc/{pid}/fd`, gets the output as `List<String>` and returns size of the List subtracted by one. But in the List there are also two entries which may not be counted: `.` and `..`. So the proper number to subtract is 2 instead of 1.

Another approach may be executing `ls -l` instead of `ls -f` and subtract just 1 (the "total" header).

I have created simple PoC:

```java
public class FD {

    private static final String LS_F_PROC_PID_FD = "ls -f " + ProcPath.PID_FD;
    private static final String LS_L_PROC_PID_FD = "ls -l " + ProcPath.PID_FD;

    public static void main(String[] args) {
        int pid = Integer.parseInt(args[0]);

        List<String> ls_f = ExecutingCommand.runNative(String.format(LS_F_PROC_PID_FD, pid));
        List<String> ls_l = ExecutingCommand.runNative(String.format(LS_L_PROC_PID_FD, pid));

        System.out.printf("Command '%s' returned %d opened files:%n%s%n%n", LS_F_PROC_PID_FD, ls_f.size(), lines(ls_f));
        System.out.println();
        System.out.printf("Command '%s' returned %d opened files:%n%s%n%n", LS_L_PROC_PID_FD, ls_l.size(), lines(ls_l));
    }

    private static String lines(List<String> lines) {
        StringBuilder sb = new StringBuilder();
        for (int i = 0; i < lines.size(); i++) {
            sb.append(i + 1).append('\t').append(lines.get(i)).append(System.lineSeparator());
        }
        return sb.toString();
    }
```

which gives this output:

```
Command 'ls -f /proc/%d/fd' returned 6 opened files:
1	.
2	..
3	0
4	1
5	2
6	4


Command 'ls -l /proc/%d/fd' returned 5 opened files:
1	total 0
2	lrwx------ 1 root root 64 Apr 15 09:17 0 -> /dev/pts/4
3	lrwx------ 1 root root 64 Apr 15 09:17 1 -> /dev/pts/4
4	lrwx------ 1 root root 64 Apr 15 09:14 2 -> /dev/pts/4
5	lrwx------ 1 root root 64 Apr 15 09:17 4 -> /etc/fstab
```

[...]

PR #1609, opened 2021-04-15
LinuxOSProcess#getOpenFiles returns one more than expected #1608

slaha:
An attempt to fix https://github.com/oshi/oshi/issues/1608


dbwiddis (approved):
This looks great!   

While this fixes this specific bug, there's some repetition in the `ProcessStat` file for multiple methods that fetch a list of numeric files.  I was just doing this (unaware you were doing it) and the result is in https://github.com/dbwiddis/oshi/commit/7887bd33bb714acb09ca276d2a41f11fbe9df07b ... take a look and see if you can do the same?

dbwiddis:
(Actually in my commit, the thread list missed excluding the process itself from the list, so that'd need to also be updated, if you're up to it.)

slaha:
@dbwiddis I will take a look at this

dbwiddis (approved):
Looks great.  I'll merge after the CI tests complete.

dbwiddis:
Oops, one more request: can you make a change log entry? :)

slaha:
Done (at least I hope so)

dbwiddis:
Thanks for the bug report and fix!

PR #1610, opened 2021-04-16
Improve performance AbstractNetworkIF::isLocalInterface

barddoo:
Remove redundant check for isLocalInterface. JDK already [does that](https://github.com/openjdk/jdk/blob/3423f3e1f5a2120e8f761a238c2929c44957760d/src/java.base/share/classes/java/net/NetworkInterface.java#L526) so it improves performance.
This PR solves https://github.com/oshi/oshi/issues/1606


dbwiddis:
Looks great, thanks! I'll merge when CI finishes.

PR #1626, opened 2021-05-01
Add getIndex() in NetworkIF and demo of Interfaces

angelyouyou:
Add getIndex() in NetworkIF and demo of Interfaces

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
@@ -61,6 +61,7 @@ public abstract class AbstractNetworkIF implements NetworkIF {
     private NetworkInterface networkInterface;
     private String name;
     private String displayName;
+    private int index;
     private int mtu;
     private String mac;
     private String[] ipv4;
@@ -100,6 +101,7 @@ public abstract class AbstractNetworkIF implements NetworkIF {
         try {
             this.name = networkInterface.getName();
             this.displayName = displayName;
+            this.index = networkInterface.getIndex();
             // Set MTU
             this.mtu = networkInterface.getMTU();
             // Set MAC
@@ -190,6 +192,11 @@ public abstract class AbstractNetworkIF implements NetworkIF {
     @Override
     public String getName() {
         return this.name;
+    }
+
+    @Override
+    public int getIndex() {
+        return this.index;
     }
 
     @Override
--- a/oshi-demo/src/main/java/oshi/demo/OshiGui.java
+++ b/oshi-demo/src/main/java/oshi/demo/OshiGui.java
@@ -35,6 +35,7 @@ import javax.swing.WindowConstants;
 import oshi.SystemInfo;
 import oshi.demo.gui.Config;
 import oshi.demo.gui.FileStorePanel;
+import oshi.demo.gui.InterfacePanel;
 import oshi.demo.gui.MemoryPanel;
 import oshi.demo.gui.OsHwTextPanel;
 import oshi.demo.gui.OshiJPanel;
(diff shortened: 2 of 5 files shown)


oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
```diff
@@ -170,6 +164,11 @@ public String getName() {
         return this.networkInterface.getName();
     }
 
+    @Override
+    public int getIndex() {
+        return this.networkInterface.getIndex();
```
dbwiddis, line 169: This works, but looking at `AbstractNetworkInterface` we already grab many of the other relevant values (`name`, `mtu`, etc.) when we construct the object and it makes sense to add a `private int index` variable, populate it in the constructor, and simply return that.

oshi-demo/src/main/java/oshi/demo/OshiGui.java
```diff
@@ -85,6 +72,7 @@ private void init() {
         new Thread(new AddMenuBarTask("FileStores", 'F', "FileStore Usage", new FileStorePanel(si))).start();
         new Thread(new AddMenuBarTask("Processes", 'P', "Processes", new ProcessPanel(si))).start();
         new Thread(new AddMenuBarTask("USB Devices", 'U', "USB Device list", new UsbPanel(si))).start();
+        new Thread(new AddMenuBarTask("Interfaces", 'I', "Interface list", new InterfacePanel(si))).start();
```
dbwiddis, line 75: Thanks for adding to the demo! I'm wondering if "Network Interfaces" would be a better title (with `N` mnemonic) although I don't have a strong opinion.

oshi-demo/src/main/java/oshi/demo/gui/InterfacePanel.java
```diff
@@ -0,0 +1,125 @@
...
+        Object[][] intfArr = new Object[intfList.size()][COLUMNS.length];
+
+        for (Entry<NetworkIF, Integer> e : intfList) {
+            NetworkIF intf = e.getKey();
+
+            intfArr[i][0] = intf.getName();
+            intfArr[i][1] = intf.getIndex();
+            intfArr[i][2] = intf.getSpeed();
+            intfArr[i][3] = getIPAddressesString(intf.getIPv4addr());
+            intfArr[i][4] = getIPAddressesString(intf.getIPv6addr());
```
dbwiddis, line 91: Is there room to add the MAC address?

oshi-demo/src/main/java/oshi/demo/gui/InterfacePanel.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private static String getIPAddressesString(String[] ipAddressArr) {
+        StringBuilder sb = new StringBuilder();
+        boolean first = true;
+
+        for (String ipAddress : ipAddressArr) {
+            if (first) {
+                first = false;
+            } else {
+                sb.append(IP_ADDRESS_SEPARATOR);
```
dbwiddis, line 118: I wonder if a "new line" might be a better delimiter for this type of display.

dbwiddis (commented):
Thanks for this contribution. Looks great and the addition to the demo is fantastic.  I've left a few suggestions inline.

oshi-demo/src/main/java/oshi/demo/OshiGui.java
dbwiddis: We could also put some of the "Network Params" results in the top of a "Network" pane.

dbwiddis:
Since I'm going to cut a release tonight I went ahead and made the tweaks I suggested and I'll merge shortly.  Thanks for your contribution, especially the demo!

One bit of feedback:  you had a very old fork of OSHI as your base, somewhere around 5.3.8.  In the future it's best to try to start from the most recent upstream in order to minimize conflicts.

PR #1635, opened 2021-05-08
Enhance network and interface demo.

angelyouyou:
1.Add NetworkPanel demo;
2.Move NetworkParams demo to NetworkPanel;
3.Add IPUtil.java.

### DIFF
--- a/oshi-core/pom.xml
+++ b/oshi-core/pom.xml
@@ -82,5 +82,11 @@
 			<version>${code-assert.version}</version>
 			<scope>test</scope>
 		</dependency>
-	</dependencies>
+        <dependency>
+            <groupId>com.github.seancfoley</groupId>
+            <artifactId>ipaddress</artifactId>
+            <version>${ipaddress.version}</version>
+            <scope>compile</scope>
+        </dependency>
+    </dependencies>
 </project>
--- a/oshi-core/src/main/java/oshi/util/IPUtil.java
+++ b/oshi-core/src/main/java/oshi/util/IPUtil.java
@@ -1,0 +1,259 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2010 - 2021 The OSHI Project Contributors: https://github.com/oshi/oshi/graphs/contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package oshi.util;
+
+import inet.ipaddr.IPAddress;
+import inet.ipaddr.IPAddressNetwork;
+import inet.ipaddr.IPAddressString;
+import inet.ipaddr.ipv4.IPv4Address;
+import inet.ipaddr.ipv6.IPv6Address;
+import oshi.annotation.concurrent.ThreadSafe;
+
+import java.math.BigInteger;
+
+/**
+ * Formatting utility for appending units or converting IP address types.
+ */
+@SuppressWarnings({"AlibabaClassNamingShouldBeCamel", "unused"})
+@ThreadSafe
+public final class IPUtil {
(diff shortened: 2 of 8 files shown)


oshi-core/src/main/java/oshi/util/IPUtil.java
```diff
@@ -0,0 +1,259 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package oshi.util;
+
+import inet.ipaddr.IPAddress;
```
mprins, line 26: why not use java built-in `java.net.InetAddress`

oshi-core/pom.xml
```diff
@@ -82,5 +82,11 @@
 			<version>${code-assert.version}</version>
 			<scope>test</scope>
 		</dependency>
-	</dependencies>
+        <dependency>
+            <groupId>com.github.seancfoley</groupId>
```
mprins, line 86: I don't think adding this as a core dependency is a good idea, I'm not convinced it is necessary at all as java has built-in objects for [java.net.InetAddress](https://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html) etc.

oshi-demo/src/main/java/oshi/demo/OshiGui.java
```diff
@@ -23,25 +23,11 @@
...
-import oshi.demo.gui.Config;
-import oshi.demo.gui.FileStorePanel;
-import oshi.demo.gui.InterfacePanel;
-import oshi.demo.gui.MemoryPanel;
-import oshi.demo.gui.OsHwTextPanel;
-import oshi.demo.gui.OshiJPanel;
-import oshi.demo.gui.ProcessPanel;
-import oshi.demo.gui.ProcessorPanel;
-import oshi.demo.gui.UsbPanel;
+import oshi.demo.gui.*;
```
mprins, line 27: please stay ways from wildcard imports

oshi-demo/src/main/java/oshi/demo/gui/NetworkPanel.java
```diff
@@ -0,0 +1,152 @@
...
+import java.awt.*;
+import java.util.List;
+import java.util.*;
+import java.util.Map.Entry;
+
+/**
+ * Displays network parameters.
+ *
+ */
+@SuppressWarnings("AlibabaLowerCamelCaseVariableNaming")
```
mprins, line 46: this is not a known check in this project

oshi-core/src/main/java/oshi/util/IPUtil.java
```diff
@@ -0,0 +1,259 @@
...
+import inet.ipaddr.ipv4.IPv4Address;
+import inet.ipaddr.ipv6.IPv6Address;
+import oshi.annotation.concurrent.ThreadSafe;
+
+import java.math.BigInteger;
+
+/**
+ * Formatting utility for appending units or converting IP address types.
+ */
+@SuppressWarnings({"AlibabaClassNamingShouldBeCamel", "unused"})
```
mprins, line 38: this is not a known check in this project

pom.xml
```diff
@@ -136,6 +136,7 @@
 		<tidy-maven-plugin.version>1.1.0</tidy-maven-plugin.version>
 		<license-maven-plugin.version>4.1</license-maven-plugin.version>
 		<code-assert.version>0.9.15</code-assert.version>
+        <ipaddress.version>5.3.3</ipaddress.version>
```
mprins, line 139: formatting

pom.xml
```diff
@@ -412,12 +413,12 @@
 						</archive>
 					</configuration>
 					<executions>
-						<!-- ! here we override the super-pom attach-sources execution id which 
-							! calls sources:jar goal. That goals forks the lifecycle, ! causing the generate-sources 
-							phase to be called twice for the install goal. ! Starting with Maven 3.4.0 
+						<!-- ! here we override the super-pom attach-sources execution id which
```
mprins, line 416: please don't add unrelated (whitespace) changes; they make review a lot more work

mprins (changes requested):

mprins:
Please make sure the full build passes locally before marking a PR ready;

dbwiddis (changes requested):
Thanks for your submission!  I really appreciate the contributions you're adding to the demo project.

I've left several comments inline and I see @mprins has already reviewed commenting on some of the same things.

pom.xml
dbwiddis: This one's interesting as it's a "tab" vs. "space" thing.  Seems we have default tab indents for XML files in the project, which I should probably switch to spaces to prevent things like this.

oshi-core/src/main/java/oshi/util/IPUtil.java
dbwiddis: Reviewing the util I think the main purpose of this is shortening IPv6 addresses.  There are [many ways to do this](https://stackoverflow.com/questions/7043983/ipv6-address-into-compressed-form-in-java) without a separate dependency... I've done so myself in a different project.

dbwiddis: In fact I did so in this project at some point using a regex, and it's in `ParseUtil.parseUtAddrV6toIP()`.  That should/could be refactored out into a public helper method if needed, the rest of the functions from this dependency aren't needed.

angelyouyou:
@dbwiddis Thanks for your suggestions, I will modify the code.First I will close the PR.

dbwiddis:
You generally do not need to close a PR. You can just push additional commits to the branch for the PR.  This is preferred as we can keep track of the code review requests and resolve them.

If you do submit a new PR, please use a different branch name than `master`.  It makes it more difficult for me to locally review your changes when your branch has the same name as one of my local branches.

PR #1637, opened 2021-05-10
Remove non-Swing concurrency from demo GUI

dbwiddis:
I initially intended to run GUI frames in their own threads but the way I did the implementation the only thing that actually was multithreaded was adding buttons to the menu -- ironically the one thing that shouldn't have been multithreaded because Swing isn't thread safe!

This moves the threading into the actual panels allowing for faster startup.


dbwiddis:
Converting this to a draft. There still are multithreading issues.

dbwiddis:
Removed threading attempts that don't use proper [Swing Concurrency](https://docs.oracle.com/javase/tutorial/uiswing/concurrency/index.html). For now the demo is single-threaded (it effectively already was).

PR #1653, opened 2021-05-31
Revert "Bump maven-javadoc-plugin from 3.1.1 to 3.3.0"

mprins:
Reverts oshi/oshi#1652

because of https://github.com/oshi/oshi/issues/1145#issuecomment-851026550


dbwiddis:
Thanks. At some point I'll try to figure out what's happening or at least create a minimal reproducible example of the bug.

That point is not today. Or this month. Or likely this year.

Issue #1618, opened 2021-04-27
Improve user choice in handling exceptions with more handlers


dbwiddis:
We've had lengthy discussions on handling exception behavior, for example this discussion: https://github.com/oshi/oshi6/issues/2 also related to requests by @rnaval in #1341 and @joshiste in #1589.

I've deferred most of these conversations to the possibility that in some future version we may completely change the API introducing `Optional` or wrapper objects with more info, but the longer we go the less I think that's a good direction.  I do think there's something we can do within the current API, however and that's to expand the usage of designed-for-extension handler objects.

For example, the [`WmiQueryHandler`](https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/util/platform/windows/WmiQueryHandler.java) class is designed for users to be able to override some behavior (e.g., COM initialization if they have done so in their own app).  It also includes a [`handleComException()`](https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/util/platform/windows/WmiQueryHandler.java#L195) method called from within a catch block that by default logs (at WARN level) the exception, but a user could change that behavior if they chose; perhaps a different level of logging, or re-throwing the exception.

I'd like the community's help in identifying good candidates (like #1589) to to add handlers for, as well as suggestions/ideas on best practices so we do it right the first time.

dbwiddis:
Will be incorporated in #1694

Issue #1366, opened 2020-10-16
AIX Process list relies on running external application


dazhitchman:
The AIXOSProcess class obtains a list of running processes by calling the "ps" command.  Is there a reason it doesn't walk the /proc path like the LinuxOSProcess ?
The /proc has been available in AIX since 5.2 from what I can see.
Issues with current reliance is 
- calling an external binary incurs overhead

- path the ps isn't fully qualified (e.g. /usr/bin/ps ) so it relies on PATH and hence is vulnerable to exploitation

- Although unlikely, the user running the application may not have permission to invoke ps (e.g. running in a restricted shell)

If there is a valid reason for this implementation then that's fine otherwise I propose using the same method/impl as oshi.software.os.linux.LinuxOperatingSystem:getProcesses(int limit, ProcessSort sort)

dbwiddis:
While there is a `proc` filesystem and it's sufficient for enumerating processes and their threads, it does not provide nearly the level of detail that Linux systems does.  (The same issue exists in Solaris.). Here's what I see on the AIX 7.1 machine that I have (non-root) access to:
```
~$ ls /proc
0        1376316  1966158  2556020  3014760  3604636  4194432  4718738  5308582  589842   655380   7209002  7799016  8257676  9044002
1        1507580  2031818  262152   3080320  3670026  4259886  4784256  5439654  5964002  6619166  720918   7864482  8323254  9109674
1048684  1573112  2162760  2621542  3145828  3735678  4325518  4915366  5505132  6029502  6684698  7274594  786456   8454300  917532
1114238  1704024  2228292  2687086  327690   3866754  4390936  4980792  5570782  6160394  6750382  7405628  7930002  8519826  9371730
1179736  1769572  2293876  2752656  3342438  393228   4456608  5046450  5636144  6226152  6815782  7536640  7995490  851994   sys
1245226  1835104  2359466  2818188  3408050  3932364  458766   5177558  5701770  6291662  6881328  7602260  8061088  8585314  version
131076   1900624  2425034  2883698  3473568  3997858  4587676  524304   5767398  6357140  7078108  7667926  8126594  8650754
1310858  196614   2490590  2949214  3539054  4063248  4653198  5243072  5832902  6553754  7143632  7733264  8192128  8781990
~$ ls /proc/393228
cred  lwp  psinfo  sigact  status
~$ cat /proc/393228/status
cat: /proc/393228/status: The file access permissions do not allow the specified action.
[...]

PR #1654, opened 2021-06-05
API for process arguments and environment

### LINKED ISSUE #369: API to read a process' environment variables?
For linux, this is easy (/proc/<pid>/environ). But it would be nice to have something like that in a cross-platform manner, esp for Windows. I tag and find stray processes in maven builds by giving them specific environment variables that let me identify them somewhat reliably.

basil:
Fixes #369. Posting this draft for initial feedback on the API and the overall strategy. The Linux implementation is shamelessly copied from the (MIT-licensed) code in Jenkins core. If this approach of moving code from Jenkins core into this project is deemed suitable, I'll look into moving over the Jenkins implementations for macOS, AIX, and Solaris and writing new implementations for the BSDs.

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/OSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSProcess.java
@@ -23,6 +23,7 @@
  */
 package oshi.software.os;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
@@ -78,20 +79,27 @@ public interface OSProcess {
     String getCommandLine();
 
     /**
-     * Gets the command-line arguments of the process.
+     * Makes a best effort attempt to get a list of the the command-line arguments
+     * of the process.
      *
-     * <p>On Windows, where the OS models command-line arguments as a single string, this method
-     * computes the approximated tokenization.
+     * @return A list of Strings representing the arguments. May return an empty
+     *         list if there was a failure (for example, because the process is
+     *         already dead or permission was denied).
      */
-    List<String> getArguments();
+    default List<String> getArguments() {
+        return Collections.emptyList();
+    }
 
     /**
      * Obtains the environment variables of the process.
      *
-     * @return an empty map if there was a failure (for example, because the process is already dead
-     *     or permission was denied).
+     * @return A map representing the environment variables and their values. May
+     *         return an empty map if there was a failure (for example, because the
+     *         process is already dead or permission was denied).
      */
-    Map<String, String> getEnvironmentVariables();
+    default Map<String, String> getEnvironmentVariables() {
+        return Collections.emptyMap();
+    }
 
     /**
      * Gets the current working directory for the process.
--- a/oshi-core/src/main/java/oshi/software/os/OSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSProcess.java
@@ -58,18 +58,21 @@ public interface OSProcess {
      * if available to be determined.
      * <p>
      * The format of this string is platform-dependent and may require the end user
-     * to parse the result.
+     * to parse the result or use {@link #getArguments()}.
      * <p>
      * On Linux and macOS systems, the string is null-character-delimited, to permit
-     * the end user to parse the executable and arguments if desired. Further, the
-     * macOS variant may include environment variables which the end user may wish
-     * to exclude from display.
+     * the end user to parse the executable and arguments if desired.
      * <p>
      * On Solaris, the string is truncated to 80 characters.
      * <p>
-     * On Windows, by default, performs a single WMI query for this process, with
-     * some latency. If this method will be frequently called for multiple
-     * processes, see the configuration file to enable a batch query mode leveraging
+     * On Windows, attempts to retrieve the value from process memory, which
+     * requires that the process be owned by the same user as the executing process,
+     * or elevated permissions, and additionally requires the target process to have
+     * the same bitness (e.g., this will fail on a 32-bit process if queried by
+     * 64-bit and vice versa). If reading process memory fails, by default, performs
+     * a single WMI query for this process, with some latency. If this method will
+     * be frequently called for multiple processes, see the configuration file to
+     * enable a batch query mode leveraging
      * {@link Win32ProcessCached#getCommandLine} to improve performance, or setting
      * that parameter via {@link GlobalConfig#set(String, Object)} before
      * instantiating any {@link OSProcess} object.
@@ -80,7 +83,8 @@ public interface OSProcess {
 
     /**
      * Makes a best effort attempt to get a list of the the command-line arguments
-     * of the process.
+     * of the process. Returns the same information as {@link #getCommandLine()} but
+     * parsed to a list.
      *
      * @return A list of Strings representing the arguments. May return an empty
      *         list if there was a failure (for example, because the process is
@@ -91,7 +95,8 @@ public interface OSProcess {
     }
 
     /**
-     * Obtains the environment variables of the process.
+     * Makes a best effort attempt to obtain the environment variables of the
+     * process. May require elevated permissions or same-user ownership.
      *
      * @return A map representing the environment variables and their values. May
      *         return an empty map if there was a failure (for example, because the
--- a/oshi-core/src/main/java/oshi/software/os/OSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSProcess.java
@@ -23,7 +23,6 @@
  */
 package oshi.software.os;
 
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
@@ -69,7 +68,7 @@ public interface OSProcess {
      * null-delimited behavior may change in future versions and should not be
      * relied upon; use {@link #getArguments()} instead.
      * <p>
-     * On Solaris, the string may be truncated to 80 characters if there was
+     * On AIX and Solaris, the string may be truncated to 80 characters if there was
      * insufficient permission to read the process memory.
      * <p>
      * On Windows, attempts to retrieve the value from process memory, which
@@ -97,9 +96,7 @@ public interface OSProcess {
      *         list if there was a failure (for example, because the process is
      *         already dead or permission was denied).
      */
-    default List<String> getArguments() {
-        return Collections.emptyList();
-    }
+    List<String> getArguments();
 
     /**
      * Makes a best effort attempt to obtain the environment variables of the
@@ -109,9 +106,7 @@ public interface OSProcess {
      *         return an empty map if there was a failure (for example, because the
      *         process is already dead or permission was denied).
      */
-    default Map<String, String> getEnvironmentVariables() {
-        return Collections.emptyMap();
-    }
+    Map<String, String> getEnvironmentVariables();
 
     /**
      * Makes a best effort attempt to obtain the current working directory for the
--- a/oshi-core/src/main/java/oshi/software/os/OSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSProcess.java
@@ -55,15 +55,22 @@ public interface OSProcess {
 
     /**
      * Gets the process command line used to start the process, including arguments
-     * if available to be determined.
+     * if available to be determined. This method generally returns the same
+     * information as {@link #getArguments()} in a more user-readable format, and is
+     * more robust to non-elevated access.
      * <p>
-     * The format of this string is platform-dependent and may require the end user
-     * to parse the result or use {@link #getArguments()}.
+     * The format of this string is platform-dependent, may be truncated, and may
+     * require the end user to parse the result. Users should generally prefer
+     * {@link #getArguments()} which already parses the results, and use this method
+     * as a backup.
      * <p>
      * On Linux and macOS systems, the string is null-character-delimited, to permit
-     * the end user to parse the executable and arguments if desired.
+     * the end user to parse the executable and arguments if desired. This
+     * null-delimited behavior may change in future versions and should not be
+     * relied upon; use {@link #getArguments()} instead.
      * <p>
-     * On Solaris, the string is truncated to 80 characters.
+     * On Solaris, the string may be truncated to 80 characters if there was
+     * insufficient permission to read the process memory.
      * <p>
      * On Windows, attempts to retrieve the value from process memory, which
      * requires that the process be owned by the same user as the executing process,
@@ -84,7 +91,7 @@ public interface OSProcess {
     /**
      * Makes a best effort attempt to get a list of the the command-line arguments
      * of the process. Returns the same information as {@link #getCommandLine()} but
-     * parsed to a list.
+     * parsed to a list. May require elevated permissions or same-user ownership.
      *
      * @return A list of Strings representing the arguments. May return an empty
      *         list if there was a failure (for example, because the process is
@@ -107,11 +114,10 @@ public interface OSProcess {
     }
 
     /**
-     * Gets the current working directory for the process.
+     * Makes a best effort attempt to obtain the current working directory for the
+     * process.
      *
      * @return the process current working directory.
-     *
-     *         On Windows, this value is only populated for the current process.
      */
     String getCurrentWorkingDirectory();
 
--- a/oshi-core/src/main/java/oshi/software/os/OSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSProcess.java
@@ -24,6 +24,7 @@
 package oshi.software.os;
 
 import java.util.List;
+import java.util.Map;
 
 import oshi.annotation.concurrent.ThreadSafe;
 import oshi.driver.windows.wmi.Win32ProcessCached;
@@ -75,6 +76,22 @@ public interface OSProcess {
      * @return the process command line.
      */
     String getCommandLine();
+
+    /**
+     * Gets the command-line arguments of the process.
+     *
+     * <p>On Windows, where the OS models command-line arguments as a single string, this method
+     * computes the approximated tokenization.
+     */
+    List<String> getArguments();
+
+    /**
+     * Obtains the environment variables of the process.
+     *
+     * @return an empty map if there was a failure (for example, because the process is already dead
+     *     or permission was denied).
+     */
+    Map<String, String> getEnvironmentVariables();
 
     /**
      * Gets the current working directory for the process.
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
@@ -36,9 +36,7 @@ import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
 import java.util.Collections;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Supplier;
@@ -77,12 +75,12 @@ public class LinuxOSProcess extends AbstractOSProcess {
     }
 
     private Supplier<Integer> bitness = memoize(this::queryBitness);
+    private Supplier<String> commandLine = memoize(this::queryCommandLine);
+    private Supplier<List<String>> arguments = memoize(this::queryArguments);
+    private Supplier<Map<String, String>> environmentVariables = memoize(this::queryEnvironmentVariables);
 
     private String name;
     private String path = "";
-    private String commandLine;
-    private List<String> arguments;
-    private Map<String, String> environmentVariables;
     private String user;
     private String userID;
     private String group;
@@ -105,41 +103,7 @@ public class LinuxOSProcess extends AbstractOSProcess {
 
     public LinuxOSProcess(int pid) {
         super(pid);
-        this.commandLine = FileUtil.getStringFromFile(String.format(ProcPath.PID_CMDLINE, pid));
-        this.arguments = parseArguments(FileUtil.readAllBytes(String.format(ProcPath.PID_CMDLINE, pid)));
-        this.environmentVariables = parseEnvironmentVariables(
-                FileUtil.readAllBytes(String.format(ProcPath.PID_ENVIRON, pid)));
         updateAttributes();
-    }
-
-    private static List<String> parseArguments(byte[] cmdline) {
-        List<String> arguments = new ArrayList<>();
-        int pos = 0;
-        for (int i = 0; i < cmdline.length; i++) {
-            byte b = cmdline[i];
-            if (b == 0) {
-                arguments.add(new String(cmdline, pos, i - pos));
-                pos = i + 1;
-            }
-        }
-        return Collections.unmodifiableList(arguments);
-    }
-
-    private static Map<String, String> parseEnvironmentVariables(byte[] environ) {
-        Map<String, String> environmentVariables = new LinkedHashMap<>();
-        int pos = 0;
-        for (int i = 0; i < environ.length; i++) {
-            byte b = environ[i];
-            if (b == 0) {
-                String line = new String(environ, pos, i - pos);
-                int sep = line.indexOf('=');
-                if (sep > 0) {
-                    environmentVariables.put(line.substring(0, sep), line.substring(sep + 1));
-                }
-                pos = i + 1;
-            }
-        }
-        return Collections.unmodifiableMap(environmentVariables);
     }
 
     @Override
@@ -154,17 +118,31 @@ public class LinuxOSProcess extends AbstractOSProcess {
 
     @Override
     public String getCommandLine() {
-        return this.commandLine;
+        return commandLine.get();
+    }
+
+    private String queryCommandLine() {
+        return FileUtil.getStringFromFile(String.format(ProcPath.PID_CMDLINE, getProcessID()));
     }
 
     @Override
     public List<String> getArguments() {
-        return this.arguments;
+        return arguments.get();
+    }
+
+    private List<String> queryArguments() {
+        return Collections.unmodifiableList(ParseUtil
+                .parseByteArrayToStrings(FileUtil.readAllBytes(String.format(ProcPath.PID_CMDLINE, getProcessID()))));
     }
 
     @Override
     public Map<String, String> getEnvironmentVariables() {
-        return this.environmentVariables;
+        return environmentVariables.get();
+    }
+
+    private Map<String, String> queryEnvironmentVariables() {
+        return Collections.unmodifiableMap(ParseUtil
+                .parseByteArrayToStringMap(FileUtil.readAllBytes(String.format(ProcPath.PID_ENVIRON, getProcessID()))));
     }
 
     @Override
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
@@ -57,7 +57,7 @@ import oshi.util.Util;
 import oshi.util.platform.linux.ProcPath;
 
 /**
- * OSProcess implemenation
+ * OSProcess implementation
  */
 @ThreadSafe
 public class LinuxOSProcess extends AbstractOSProcess {
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
@@ -106,10 +106,9 @@ public class LinuxOSProcess extends AbstractOSProcess {
     public LinuxOSProcess(int pid) {
         super(pid);
         this.commandLine = FileUtil.getStringFromFile(String.format(ProcPath.PID_CMDLINE, pid));
-        this.arguments =
-                parseArguments(FileUtil.readAllBytes(String.format(ProcPath.PID_CMDLINE, pid)));
-        this.environmentVariables =
-                parseEnvironmentVariables(FileUtil.readAllBytes(String.format(ProcPath.PID_ENVIRON, pid)));
+        this.arguments = parseArguments(FileUtil.readAllBytes(String.format(ProcPath.PID_CMDLINE, pid)));
+        this.environmentVariables = parseEnvironmentVariables(
+                FileUtil.readAllBytes(String.format(ProcPath.PID_ENVIRON, pid)));
         updateAttributes();
     }
 
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
@@ -36,6 +36,9 @@ import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Supplier;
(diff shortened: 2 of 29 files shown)


dbwiddis:
Thanks for starting this contribution! I think this will be a great addition.  I'm not sure why I decided it was out of scope earlier, perhaps because of the Windows implementation needing to dig into process memory, something I wasn't about to attempt based on a few snippets in a Stack Overflow answer.  

I like the addition and the new proposed API looks reasonable (I've tagged @mprins for review to offer an opinion on the API as well.)  Some general thoughts after a brief look at the API:
 - Later we'll have to be more verbose in the javadocs with caveats, "best effort" notation to advise users what they may need to double-check, etc.
 - Whether it was a good decision or bad one, I don't have OSHI throw exceptions, so the "unimplemented" ones should just return empty lists/maps.  Use a `default` method in the interface and then you don't need all the different OS implementations that throw `UnsupportedOperationException`.
 - I don't think it's necessary to make a freshly-generated list or map from a static method unmodifiable.  If each method call will produce a new copy, there's no harm in handing the user a mutable structure.

With regard to using code from Jenkins.  I want to respect the intent of the MIT license which requires the copyright notice to accompany "copies" or "substantial portions".  I hand-wave that at OSHI by referring to the "Contributors", which credits authors via the git history.  But that will show up as your commit, not the original author.   That said, I don't think it's an issue for the code I see so far, only the parsing routine looks "copied" and that's pretty much the only way to do it.  I think I've seen similar C code in the JDK's ProcessHandle implementation.   Would be nice to give attribution to the original author in a comment, though.

For the AIX port, do you have access to an AIX machine?  I got access to a server at Polarhome when I did the AIX port here, but it's been down for several months so I have no means to develop or test code there.

oshi-core/src/main/java/oshi/software/os/OSProcess.java
```diff
@@ -76,6 +77,22 @@
      */
     String getCommandLine();
 
+    /**
+     * Gets the command-line arguments of the process.
+     *
+     * <p>On Windows, where the OS models command-line arguments as a single string, this method
```
dbwiddis, line 83: I'm not sure what this means!  Make it clear what the end user should expect to see as the tokenization.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -101,9 +106,43 @@
...
         super(pid);
         this.commandLine = FileUtil.getStringFromFile(String.format(ProcPath.PID_CMDLINE, pid));
+        this.arguments =
+                parseArguments(FileUtil.readAllBytes(String.format(ProcPath.PID_CMDLINE, pid)));
+        this.environmentVariables =
+                parseEnvironmentVariables(FileUtil.readAllBytes(String.format(ProcPath.PID_ENVIRON, pid)));
         updateAttributes();
     }
 
+    private static List<String> parseArguments(byte[] cmdline) {
```
dbwiddis, line 116: This would be better moved to `ParseUtil` with a more generic name.  (Maybe `parseByteArrayToStringList`?)

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -101,9 +106,43 @@
...
+            byte b = cmdline[i];
+            if (b == 0) {
+                arguments.add(new String(cmdline, pos, i - pos));
+                pos = i + 1;
+            }
+        }
+        return Collections.unmodifiableList(arguments);
+    }
+
+    private static Map<String, String> parseEnvironmentVariables(byte[] environ) {
```
dbwiddis, line 129: Similar to above, put in `ParseUtil`.  (Maybe `parseKeyValueByteArrayToMap`?)

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -119,6 +158,16 @@ public String getCommandLine() {
         return this.commandLine;
     }
 
+    @Override
+    public List<String> getArguments() {
+        return this.arguments;
```
dbwiddis, line 163: OK now I see why you made the parsing methods return unmodifiable, you directly return them here.  I think there's a better way to do this.  I'm wondering if a memoizer would be a better implementation for these (as well as the command line).

oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
```diff
@@ -364,6 +364,42 @@ void testGetCommandLine() {
                 is(greaterThan(0)));
     }
 
+    @Test
+    void testGetArguments() {
+        int processesWithNonEmptyArguments = 0;
+
+        SystemInfo si = new SystemInfo();
+        OperatingSystem os = si.getOperatingSystem();
+        for (OSProcess process : os.getProcesses(null, null, 0)) {
```
dbwiddis, line 373: Why not create a `Predicate` for testing empty args and pass it to `getProcesses()`.  Then you just return the list size.  Yeah, overkill for a test method, I guess.  But still, behold the elegance!
```
int procsWithArgs = os.getProcesses(p -> !p.getArguments().isEmpty(), null, 0).size();
```
Better yet instead of asserting the size is greater than 0, roll up the above into:
```
assertThat("Processes with non-empty arguments should be 1 or higher",
    os.getProcesses(p -> !p.getArguments().isEmpty(), null, 0), is(not(empty())));
```

dbwiddis (commented):
Some specific code suggestions in addition to/clarifying my comments above.

mprins (commented):
shouldn't be too hard to get program arguments on OpenBSD, not sure about program environment, either one may need elevated user privileges for security reasons

oshi-core/src/main/java/oshi/software/os/OSProcess.java
```diff
@@ -76,6 +77,22 @@
...
+     * computes the approximated tokenization.
+     */
+    List<String> getArguments();
```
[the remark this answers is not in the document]
dbwiddis: I'm in favor of empty list, documented as such, unless there's a certain way to distinguish a program with no arguments/environment from one that you just don't have access to read, cross-platform.

> either one may need elevated user privileges for security reasons

Because I expect them to be blank/empty often, when run without elevation, on processes not belonging to the user.

dbwiddis:
I went ahead and cleaned up the API, Linux implementation, some utils, and did the macOS implementation (we already were fetching the bytes, just needed to put them in a map).

I'm still uncomfortable with the Windows implementation(s) as they all involve API functions which are stated may be removed in future versions of Windows, or are unreliable involving race conditions.  And I think require running as administrator.

Are you going to try to put in the Solaris and AIX ports?

dbwiddis:
Good News!  I got the Windows version working.  As a bonus, the same memory that provides the environment pointer also gets us CWD strings (that we previously didn't support) and Command Line strings (without the overhead of WMI).
```
        OSProcess p = os.getProcess(os.getProcessId());
        System.out.println("CWD: " + p.getCurrentWorkingDirectory());
        System.out.println("CMD: " + p.getCommandLine());
        System.out.println("ARG: " + p.getArguments());
        System.out.println("ENV: " + p.getEnvironmentVariables());
    }
```
Output:
```
CWD: Z:\git\oshi\oshi-core\
CMD: "C:\Program Files\Java\jre1.8.0_291\bin\javaw.exe" -Dfile.encoding=UTF-8 -classpath "C:\Program Files\Java\jre1.8.0_291\lib\resources.jar;C:\Program Files\Java\jre1.8.0_291\lib\rt.jar;C:\Program Files\Java\jre1.8.0_291\lib\jsse.jar;C:\Program Files\Java\jre1.8.0_291\lib\jce.jar;C:\Program Files\Java\jre1.8.0_291\lib\charsets.jar;C:\Program Files\Java\jre1.8.0_291\lib\jfr.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\access-bridge.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\cldrdata.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\dnsns.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\jaccess.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\jfxrt.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\localedata.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\nashorn.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\sunec.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\sunjce_provider.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\sunmscapi.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\sunpkcs11.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\zipfs.jar;Z:\git\oshi\oshi-core\target\classes;C:\eclipse\plugins\org.junit.jupiter.api_5.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.jupiter.engine_5.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.jupiter.migrationsupport_5.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.jupiter.params_5.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.platform.commons_1.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.platform.engine_1.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.platform.launcher_1.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.platform.runner_1.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.platform.suite.api_1.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.vintage.engine_5.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.opentest4j_1.0.0.v20180327-1502.jar;C:\eclipse\plugins\org.apiguardian_1.0.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit_4.12.0.v201504281640\junit.jar;C:\eclipse\plugins\org.hamcrest.core_1.3.0.v20180420-1519.jar;C:\Users\dbwiddis\.m2\repository\net\java\dev\jna\jna\5.8.0\jna-5.8.0.jar;C:\Users\dbwiddis\.m2\repository\net\java\dev\jna\jna-platform\5.8.0\jna-platform-5.8.0.jar;C:\Users\dbwiddis\.m2\repository\org\slf4j\slf4j-api\1.7.30\slf4j-api-1.7.30.jar" oshi.software.os.windows.WindowsOSProcess
ARG: [C:\Program Files\Java\jre1.8.0_291\bin\javaw.exe, -Dfile.encoding=UTF-8, -classpath, C:\Program Files\Java\jre1.8.0_291\lib\resources.jar;C:\Program Files\Java\jre1.8.0_291\lib\rt.jar;C:\Program Files\Java\jre1.8.0_291\lib\jsse.jar;C:\Program Files\Java\jre1.8.0_291\lib\jce.jar;C:\Program Files\Java\jre1.8.0_291\lib\charsets.jar;C:\Program Files\Java\jre1.8.0_291\lib\jfr.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\access-bridge.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\cldrdata.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\dnsns.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\jaccess.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\jfxrt.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\localedata.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\nashorn.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\sunec.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\sunjce_provider.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\sunmscapi.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\sunpkcs11.jar;C:\Program Files\Java\jre1.8.0_291\lib\ext\zipfs.jar;Z:\git\oshi\oshi-core\target\classes;C:\eclipse\plugins\org.junit.jupiter.api_5.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.jupiter.engine_5.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.jupiter.migrationsupport_5.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.jupiter.params_5.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.platform.commons_1.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.platform.engine_1.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.platform.launcher_1.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.platform.runner_1.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.platform.suite.api_1.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit.vintage.engine_5.1.0.v20180327-1502.jar;C:\eclipse\plugins\org.opentest4j_1.0.0.v20180327-1502.jar;C:\eclipse\plugins\org.apiguardian_1.0.0.v20180327-1502.jar;C:\eclipse\plugins\org.junit_4.12.0.v201504281640\junit.jar;C:\eclipse\plugins\org.hamcrest.core_1.3.0.v20180420-1519.jar;C:\Users\dbwiddis\.m2\repository\net\java\dev\jna\jna\5.8.0\jna-5.8.0.jar;C:\Users\dbwiddis\.m2\repository\net\java\dev\jna\jna-platform\5.8.0\jna-platform-5.8.0.jar;C:\Users\dbwiddis\.m2\repository\org\slf4j\slf4j-api\1.7.30\slf4j-api-1.7.30.jar, oshi.software.os.windows.WindowsOSProcess]
ENV: {LOCALAPPDATA=C:\Users\dbwiddis\AppData\Local, PROCESSOR_LEVEL=6, FP_NO_HOST_CHECK=NO, USERDOMAIN=DWA909, LOGONSERVER=\\DWA909, JAVA_HOME=C:\Program Files\Java\jre1.8.0_201\bin\javaw.exe, SESSIONNAME=Console, ALLUSERSPROFILE=C:\ProgramData, PROCESSOR_ARCHITECTURE=x86, PSModulePath=C:\Windows\system32\WindowsPowerShell\v1.0\Modules\, SystemDrive=C:, APPDATA=C:\Users\dbwiddis\AppData\Roaming, USERNAME=dbwiddis, CommonProgramFiles=C:\Program Files\Common Files, Path=C:/Program Files/Java/jre1.8.0_291/bin/client;C:/Program Files/Java/jre1.8.0_291/bin;C:/Program Files/Java/jre1.8.0_291/lib/i386;C:\Program Files\Parallels\Parallels Tools\Applications;C:\Program Files\Common Files\Oracle\Java\javapath;C:\ProgramData\Oracle\Java\javapath;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\eclipse;, PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC, OS=Windows_NT, COMPUTERNAME=DWA909, PROCESSOR_REVISION=9e0d, ComSpec=C:\Windows\system32\cmd.exe, ProgramData=C:\ProgramData, HOMEPATH=\Users\dbwiddis, SystemRoot=C:\Windows, TEMP=C:\Users\dbwiddis\AppData\Local\Temp, HOMEDRIVE=C:, PROCESSOR_IDENTIFIER=x86 Family 6 Model 158 Stepping 13, GenuineIntel, USERPROFILE=C:\Users\dbwiddis, TMP=C:\Users\dbwiddis\AppData\Local\Temp, ProgramFiles=C:\Program Files, PUBLIC=C:\Users\Public, NUMBER_OF_PROCESSORS=8, windir=C:\Windows}
```
Still a work in progress as I haven't handled the IsX64/IsWow checks, but this definitely looks doable.

basil:
This looks like it's coming along nicely! I'll try to keep helping to chip away at this as I have time.

The use of `CommandLineToArgvW` in the Windows implementation is very nice! Jenkins has its own quoted string tokenizer that splits apart the single string into an approximated tokenization. Using a native Windows API to do this seems cleaner.

I added a FreeBSD implementation for OSHI. I tested it on a 64-bit FreeBSD system running in a 64-bit Java process and inspecting both 64-bit and 32-bit processes. I think the OpenBSD implementation for OSHI will be similar modulo differences in the `sysctl` invocations (see [OpenBSD's documentation](https://man.openbsd.org/sysctl.2)).

For AIX and Solaris, I still think the easiest path forward is to copy the implementation from Jenkins (with proper attribution, of course). I have an illumos box to test on, but unfortunately I don't have an AIX box. On a side note, I don't like how the existing AIX and Solaris implementation in OSHI invokes `ps(1)` to get the list of all processes. These are `procfs`-based Unix platforms, so we should be able to more efficiently get the list of processes from `/proc` the same way we do for Linux in OSHI. Jenkins does it this way for Linux, AIX, and Solaris. While this work isn't strictly required for implementing this new feature, it is closely related. I think we might as well do that in this PR, if we're going to be testing this anyway. (This also gets me closer to my personal goal of using OSHI from Jenkins and getting Jenkins out of the process handling business.)

Similarly, I don't like how the existing FreeBSD implementation in OSHI invokes `ps(1)` to get the list of all processes. There is a `kern.proc.all` `sysctl` available for this. Jenkins has been using the `kern.proc.all` `sysctl` for macOS for a while, and I recently wrote and tested a FreeBSD implementation of the `kern.proc.all` `sysctl` for Jenkins (which was almost identical to the macOS implementation in Jenkins). I imagine it would be similar for OpenBSD. I think that using the `kern.proc.all` `sysctl` in OSHI for FreeBSD and OpenBSD (and possibly macOS?) might be desirable. Not only is it more efficient than `ps(1)`, but it would also match the implementation in Jenkins. However, the `kern.proc.all` `sysctl` API is a bit awkward, as you have to call it twice: once to get the size of the buffer you need and once to get the actual data. In between, the size required might have changed (for example, if new processes were spawned), so you have to do an awkward dance of increasing the buffer size and retrying. Still, it's better than `ps(1)`, which requires the expensive creation of a new process. While this work isn't strictly required for implementing this new feature, it is closely related. I think we might as well do that in this PR, if we're going to be testing this anyway. (This also gets me closer to my personal goal of using OSHI from Jenkins and getting Jenkins out of the process handling business.)

I noticed that Java 9+ has [`ProcessHandle`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ProcessHandle.html) and [`ProcessHandle.Info`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ProcessHandle.Info.html) classes that provides much of the information needed to implement `OperatingSystem` and `OSProcess` (but not the environment variables, unfortunately). Perhaps on the Java 11 branch, we'd want to try the Java Platform's implementation and fall back to our own implementation only if that doesn't work (or vice versa).

Another question: Should we try to preserve the order of the environment variable entries by using `LinkedHashMap` rather than `HashMap`? I think it would be nice to preserve the order of entries returned by the OS in the map that we return, but I don't think we should commit to any particular order in the API or that consumers should rely on any particular order being returned.

dbwiddis:
> I have an illumos box to test on, but unfortunately I don't have an AIX box.

I have an account on a Polarhome AIX server that I'm happy to share the login for. It's an older version but should be sufficient for this work.   Unfortunately the drive failed and it's been down since March.  Hoping it comes up soon!

> we should be able to more efficiently get the list of processes from /proc the same way we do for Linux in OSHI

There are key differences in access available to a non-root user.  I chose the `ps` approach primarily because it allows that access (owned by `root` but run by normal users) and, frankly, there's the question of how much work I want to put in to support an OS that few people use.  I've discovered crashing bugs in Solaris that were there for 2 years that imply nobody is using OSHI on Solaris.  (Heck, even OpenJDK isn't supporting it in the future.). 

That doesn't mean I wouldn't love improved code, of course... it's just something I would prefer users of those OS's provide.  My goals when porting to other OS's have been "provide the information and don't crash" without too much time spent on performance optimization.  See also #1366 discussing exactly what you're talking about for AIX.

> I think we might as well do that in this PR, if we're going to be testing this anyway.

I think I'd like to keep them separate for now, so this PR doesn't become overly broad, and breaking changes are more easily found and isolated to a single commit.  We certainly can target the same release, and we can design the implementations to be overlapping to ease transition later.

> I noticed that Java 9+ has ProcessHandle and ProcessHandle.Info classes that provides much of the information needed to implement OperatingSystem and OSProcess (but not the environment variables, unfortunately). Perhaps on the Java 11 branch, we'd want to try the Java Platform's implementation and fall back to our own implementation only if that doesn't work (or vice versa).

But it's also missing a lot.  I seriously considered just using that code in the java11 branch but there were too many gaps.  By the time I account for all the fallbacks it just adds to code bloat.  Sure, it's sufficient if you want fewer bits of information, but then... the users themselves have the ability to use the `ProcessHandle` API and not use OSHI.  So I don't see any harm in keeping the implementation separate.

I've used the JDK code as inspiration for JNA application of the same underlying code, though, so I don't think we're missing too much.

Longer term (actually mid-term now that JDK17 is EA) I plan on a whole new OSHI artifact built from the ground up leveraging lessons learned.  While the current version tries to provide pretty much everything to everyone, the new version will be more limited in scope (CPU, Memory, Processes to start, with community-driven additions after that), use FFI instead of JNA, and put performance at a premium.

> Another question: Should we try to preserve the order of the environment variable entries by using LinkedHashMap rather than HashMap? I think it would be nice to preserve the order of entries returned by the OS in the map that we return, but I don't think we should commit to any particular order in the API or that consumers should rely on any particular order being returned.

Entirely reasonable and I'll update the ParseUtil methods to do that as part of my next Windows commit.  I realized that need just now when trying to efficiently figure out how to conditionally remove the first environment variable which in my Windows tests produces the string "=::=::\" (empty key, and value "::=::\" in the map.  Other code I've seen checks for any of several special characters in the first key/value pair and it'd be nice to just check one.

> For AIX and Solaris, I still think the easiest path forward is to copy the implementation from Jenkins (with proper attribution, of course).

Alternately, I can look at the implementation, understand it, and write it myself. That's what I did with the Linux parsing routines and I think mine are slightly more efficient (less object creation, not that a few nanoseconds will matter).  Really the key thing to know is "where is this information stored". Not that I particularly want to avoid copying/attribution, happy to do that if it's truly a drop-in-place, but I'm guessing there's going to be enough tweaking to call it original/inspired by rather than derivative.

basil:
> Frankly, there's the question of how much work I want to put in to support an OS that few people use. I've discovered crashing bugs in Solaris that were there for 2 years that imply nobody is using OSHI on Solaris. (Heck, even OpenJDK isn't supporting it in the future.).

While I can't speak to AIX, the Solaris legacy is alive and well in the illumos community. While my company has switched from illumos to Linux and left the illumos community, I still have access to (somewhat dated) illumos boxes. And when Jenkins tried to drop Solaris-based OS support, members of the illumos community (working at several different companies) showed up to ask that it be kept. (Many of them use Jenkins to build illumos.) Those same members of the illumos community are also working with the OpenJDK community to keep illumos support alive and well in OpenJDK. My point is that demand for illumos support still exists. In a hypothetical future where Jenkins used OSHI, you'd inherit that installed user base.

> My goals when porting to other OS's have been "provide the information and don't crash" without too much time spent on performance optimization.

That is perfectly understandable. But my eventual goal is "get Jenkins out of the process business" without causing regressions for any existing Jenkins users. And Jenkins already uses `/proc` for AIX and Solaris, so I think using `ps(1)` would be at least a minor performance regression. I'll have to look into your point about root vs unprivileged access on illumos.

> Longer term (actually mid-term now that JDK17 is EA) I plan on a whole new OSHI artifact built from the ground up leveraging lessons learned.

A more explicit error handling and exception system might also be worth considering, if you haven't considered it already. Jenkins still supports Java 8, but the project will probably drop support for Java 8 sometime in the next year or so and start requiring Java 11. Jenkins has relatively simple needs: it needs to be able to list processes (including descendants) and to get arguments and environment variables. A slimmer artifact would increase the likelihood that the Jenkins core team would accept a move to OSHI. Out of curiosity, were you planning on replacing JNA with JNR?

dbwiddis:
> My point is that demand for illumos support still exists. In a hypothetical future where Jenkins used OSHI, you'd inherit that installed user base.

If only I inherited that community to write PRs. :grin: 

> my eventual goal is "get Jenkins out of the process business" 

And I'm happy for you, or any of your customers on various OS's, to help that transition, and even throw in a lot of work myself alongside.

dbwiddis:
> A more explicit error handling and exception system might also be worth considering, if you haven't considered it already.

Yep. A bit off-topic here, amply discussed in #1618 and issues linked therein. My current thought is all errors go to an overridable "handler" function which throws an exception by default. Users can customize the behavior however they want (logging, ignore, default values, etc.). But we'll cross that bridge when we come to it.

dbwiddis:
> Out of curiosity, were you planning on replacing JNA with JNR?

No, with JDK's bult-in FFI (Project Panama features).  The goal would be zero dependencies.

dbwiddis:
Recapping where we are:
 - Linux port is done
 - macOS port is done
 - Windows port is probably done, but I've only done dev/test on my Win7 x86 VM.  I'll do more thorough testing on my Win10 x64 machine later this week
 - It looks like FreeBSD code is done
 - OpenBSD code looks similar to FreeBSD (may be more "empty" results).  A project for another weekend.
 - AIX and Solaris are still to-do, but Jenkins has working code that can be used as a basis.  If you're doing those, great. If not, I'd appreciate a more direct link.  But I do have a 3-day weekend coming next week :)

basil:
I'll keep chipping away at this as I have time. Although I use Jenkins at my job, I'm not paid to work on Jenkins and generally contribute to Jenkins as a volunteer in my spare time. So things can take a while. The upcoming holiday does present a great opportunity, as you note. :-) The Jenkins code is all in the [`ProcessTree`](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/util/ProcessTree.java) class, and the Windows stuff is in the [`kohsuke/winp`](https://github.com/kohsuke/winp/) project on GitHub. I'll keep updating this PR branch as I make progress. I'll probably focus on setting up an OpenBSD development environment next so that I can port/test the FreeBSD implementation there. If it works, I might also open a PR to add the same implementation to the Jenkins `ProcessTree` class in addition to updating this PR.

I should note that I haven't formally proposed a transition to OSHI to the Jenkins core team yet (although I've mentioned the idea in a PR discussion). Today I realized that the Jenkins process handling implementation only uses environment variables as "cookies" to help identify processes to kill that aren't descendants of the parent that launched them (e.g. daemons). I'm not sure how necessary this functionality is: it might be of limited usefulness for the vast majority of Jenkins users. So perhaps the Jenkins core team might prefer an implementation that uses the Java 9+ `ProcessHandle` API for listing processes, getting their arguments, and recursively killing descendants, while dropping the environment-variable based "cookie" feature (perhaps leaving the existing Jenkins code in-place as an escape hatch for users who might encounter regressions with the new implementation and want to go back to the old working implementation by setting a Java property). If that's the way they want to go, though, it would still be about a year before Jenkins drops Java 8 support and the `ProcessHandle` APIs become available in Jenkins. I also need to investigate the matter of process termination. Jenkins is currently using some native code for Windows to send a Control-C and then wait for the process to die before forcibly terminating it. I need to look into whether the Java 9+ `ProcessHandle` implementation's `destroy​()` and `destroyForcibly​()` work the same way, or whether there might be a possible regression here.

My point is that Jenkins eventually adopting OSHI isn't certain: there are two possible paths forward, and a decision has not yet been made. One of those paths forward depends on Jenkins dropping support for Java 8, which hasn't happened yet. The other one depends on some improvements to OSHI, including but not limited to this PR. And both of those two paths forward would likely depend on me coding up the change to Jenkins' `ProcessTree` in a way that preserves the existing implementation as an escape hatch in case one of the hundreds of thousands of Jenkins users encounters a regression. Most likely the Jenkins core team would want to see draft PRs of both approaches before making a decisoin. Add to that my limited availability to work on all of this and you can see that it may very well stall or take a very long time.

All of that is to say that I wouldn't want to get your hopes up unfairly or to ask you to dedicate a lot of time to this idea for my sake. Your efforts maintaining this library are highly appreciated, and I want to respect your time. I think we both share the same goals: to write great software that gets widely used and to make a positive impact on the OSS community. And to have fun programming at the same time. :-)

dbwiddis:
Had a quick look at the FreeBSD implementation.  (Mainly because I just copied it to OpenBSD to see if it would work.). I think the whole parsing routine, while elegant, is unneeded.  Just take the memory object from the `sysctl` query, call `getByteArray()` and pass the resulting array to the existing ParseUtil functions I wrote for Linux.   I'll try that approach on OpenBSD sometime this week -- which looks like it'll be pretty much the same, just different values and MIB ordering for the sysctl calls.

dbwiddis:
Man I really had a hard time spelling delimiter today.  I blame the heat.

basil:
> I think the whole parsing routine, while elegant, is unneeded. Just take the memory object from the `sysctl` query, call `getByteArray()` and pass the resulting array to the existing `ParseUtil` functions I wrote for Linux.

Great suggestion! I implemented and tested this. Now the code is 22 lines shorter.

dbwiddis:
This is looking really good.

To avoid stepping on each other's toes:
 - I'll take care of OpenBSD, pretty much copying the FreeBSD code and changing the sysctl MIB values.  Looks like the BSDs (and macOS) have a nice portable sysctl approach.
   - One additional possible cleanup is removing the `args` option from the `ps` command since we're no longer using it (unless we want it as a backup) and reducing the array size by one.  Although if we're doing `ps` rewrite that can be deferred until we replace all of it...
 - The AIX and Solaris approaches look identical, reading an "as" file from the proc fs.  Since you're familiar with Solaris I'd suggest you pursue that one, then copying it to AIX should be pretty easy for one of us.

As far as overhauling `ps` let's make that our/your next project.

dbwiddis:
OpenBSD is done!  Sorry for all the commits -- debugging on a remote VM has its challenges. 

Confirmed there are a few system processes where the sysctl calls fail (no command line) and the ps provides a backup thanks to being owned by root, so I restored that "backup" option in the FreeBSD branch as well.

Just the Solaris and AIX implementations to go (and I think they're similar).

dbwiddis:
AIX and Solaris done.  

Have no way to test AIX but the implementation matches Solaris so I'm confident it will work.

Any chance you can test Solaris yourself, @basil?

dbwiddis:
OK, so the Solaris code works fine on the owning process but fails gloriously on various permissions errors since I haven't bothered to check any return codes.  And I'm going to test a more efficient way of reading the bytes.

dbwiddis:
Status report:  Solaris code is coming along nicely and is almost in its final state.  Still need to figure out something going on with the pointer sizes.  They don't always seem to match Native.POINTER_SIZE (e.g., under sudo I seem to get 4-byte offset values).  I'm not sure if it's JVM bitness (does sudo run a 32-bit java?) or process data model.  I think I can just do the math (evnp-argv)/(argc+1) to back into it, though.

After I get solaris working I'll do the same updates for AIX and see if there's anywhere I can reduce duplication.

dbwiddis:
Ready for final review.
I've tested thoroughly on macOS and Windows (x86 7 and x64 10).
@basil please test on Solaris and FreeBSD.
@mprins (optional) if you can test on OpenBSD it'd be helpful. I did on an OpenBSD VM.
@tausiflife (optional) another tester on Linux won't hurt :)

I plan to cut a 5.8.0 release this weekend.

PR #1665, opened 2021-06-12
Added windows to the list of opened modules

vatbub:


dbwiddis:
Thanks for this fix!  Don't know how I missed that package the last time (or how it passes tests... except I guess the CI servers don't have batteries...)

vatbub:
Funnily enough, I also didn't see that issue when I upgraded to version `5.7.0` and it seems to have worked nevertheless. I guess JNI is and will always be a big piece of magic...
Anyways, thanks for merging :)

dbwiddis:
Good timing too, as I logged on to cut a release so was able to squeeze this in.

PR #1675, opened 2021-06-30
Replace ps argument lists with enum

dbwiddis:
I really like this improvement for ps argument lists.  Much less error prone.  FreeBSD processes done, will do for other *nix PS lists for processes and threads.

@basil what do you think?


dbwiddis:
Thinking more on this... I don't like the use of `ordinal()` much.  So an array is not the ideal structure for the ps results.  An `EnumMap` would be ideal here, and it could be efficiently populated in value order in an iteration.  I'll try that...

basil (approved):
Very nice PR!

Issue #1670, opened 2021-06-21
Memory leak in getUsbDevices


botto:
I'm seeing a memory leak when repeatedly doing `getUsbDevices` running on a Linux machine.

The test program I used was:
```
import oshi.SystemInfo;
import oshi.hardware.HardwareAbstractionLayer;

public class main {
    private static final HardwareAbstractionLayer hal = new SystemInfo().getHardware();
    public static void main(String[] args) {
        while (true) {
            hal.getUsbDevices(false);
        }
    }
}
```

and the result I got over a a 22 min run. Those dips are me hitting the `Perform GC`
![image](https://user-images.githubusercontent.com/175029/122809470-0e49b080-d2c6-11eb-9d21-c57f979887c5.png)

dbwiddis:
I'm not sure this is evidence of a leak.  The memory goes on the stack from the creation of objects for the USB Devices, and the many strings used in the process of creating them, so if you can release it with GC it's not a leak -- a leak would eventually result in an out of memory error.

That said, I'm looking at the code in `LinuxUsbDevice` to confirm:
 - `udev` context object allocted line 86 is unreferenced line 151
 - `enumerate` enumerator allocated line 88 is unreferenced line 150
 - `devices` variable line 91 is just a pointer to structure in the existing enumerator and doesn't allocate anything. Same for the `dev_list_entry` in the loop.
 - `dev` object allocated line 111 is unreferenced line 147

Do you have other evidence of a memory leak?

botto:
Sorry, there was a test I missed before opening this. Setting a constrained maximum memory and letting it run. I've never got a OOM error so I'll close this.

PR #1678, opened 2021-07-06
Fix leaking udev reference in LinuxUsbDevice

mattmacleod:
This change fixes a leaking device reference in the Linux USB device implementation. When an interface device is encountered (i.e. devtype is "usb_device") the rest of the loop was skipped and the allocated device would never be unreferenced. This change adds an additional explicit unreference in this case.

This was the underlying leak causing the issue reported by @botto in #1670

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxUsbDevice.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxUsbDevice.java
@@ -44,6 +44,14 @@ import oshi.hardware.common.AbstractUsbDevice;
 @Immutable
 public class LinuxUsbDevice extends AbstractUsbDevice {
 
+    private static final String SUBSYSTEM_USB = "usb";
+    private static final String DEVTYPE_USB_DEVICE = "usb_device";
+    private static final String ATTR_PRODUCT = "product";
+    private static final String ATTR_MANUFACTURER = "manufacturer";
+    private static final String ATTR_VENDOR_ID = "idVendor";
+    private static final String ATTR_PRODUCT_ID = "idProduct";
+    private static final String ATTR_SERIAL = "serial";
+
     public LinuxUsbDevice(String name, String vendor, String vendorId, String productId, String serialNumber,
             String uniqueDeviceId, List<UsbDevice> connectedDevices) {
         super(name, vendor, vendorId, productId, serialNumber, uniqueDeviceId, connectedDevices);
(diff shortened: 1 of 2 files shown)


dbwiddis:
Gah, good find. I should have seen that when looking at #1670.

Your fix doesn't handle the case where `udev_device_new_from_syspath` returns `null` (which would be caught in the existing conditional check) and you'd try to unref that.  I suppose the native code wouldn't care but it's still somewhat sloppy.

This whole class could use a refactoring, if you wouldn't mind putting in a bit more work here:
 - I'd prefer to use a `try - finally` approach to defensively handle references with exceptions (and `continue` and `return`)
 - I wrote this code 5 years ago and since then I've contributed a more object-oriented front end to JNA. See the `LinuxHWDiskStore` class `getDisks()` method, code starting at line 168.

At a minimum please change your PR to do something like this pattern:
```
UdevDevice dev = Udev.INSTANCE.udev_device_new_from_syspath(udev, path);
if (dev != null) {
    try {
        // existing code between new and unref
    } finally {
        Udev.INSTANCE.udev_device_unref(dev);
    }
}
```

However, if you can refactor to the example in `LinuxHWDiskStore` it'd be even better.

mattmacleod:
Thanks for the pointers @dbwiddis – I have updated the PR to reflect the approach used in `LinuxHWDiskStore` as suggested. Confirmed that this still functions as expected in my use-case, and that it doesn't leak any more. Please let me know if you'd like any tweaks.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxUsbDevice.java
```diff
@@ -82,73 +90,76 @@ public LinuxUsbDevice(String name, String vendor, String vendorId, String produc
...
         Map<String, String> vendorIdMap = new HashMap<>();
         Map<String, String> productIdMap = new HashMap<>();
         Map<String, String> serialMap = new HashMap<>();
         Map<String, List<String>> hubMap = new HashMap<>();
 
-        // For each item enumerated, store information in the maps
-        for (UdevListEntry dev_list_entry = devices; dev_list_entry != null; dev_list_entry = Udev.INSTANCE
-                .udev_list_entry_get_next(dev_list_entry)) {
+        // Enumerate all usb devices and build information maps
+        Udev.UdevContext udev = Udev.INSTANCE.udev_new();
```
dbwiddis, line 105: Looking at this again, I see that it's possible to return null if the native call fails, but [of the four times we call this](https://github.com/oshi/oshi/search?q=udev_new), we only null check one of them.   No need to change this PR for that but I'll have to think about whether we should update all of them.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxUsbDevice.java
```diff
@@ -82,73 +90,76 @@ public LinuxUsbDevice(String name, String vendor, String vendorId, String produc
...
+        try {
+            UdevEnumerate enumerate = udev.enumerateNew();
```
dbwiddis, line 107: Similar to udev_new this can theoretically return null but probably never will.  Something to think about for a future fix.

dbwiddis (approved):
Thanks! This looks great and I'll go ahead and merge it.

PR #1680, opened 2021-07-15
Move supported operating system check out of SystemInfo constructor

KyongSik-Yoon:
#1671 not worked for me.
Run CI with snapshot version but I saw same error message below.
```
[ERROR] display as string  Time elapsed: 0.016 s  <<< ERROR!
java.lang.UnsupportedOperationException: Operating system not supported: JNA Platform type 2
Caused by: java.lang.UnsupportedOperationException: Operating system not supported: JNA Platform type 2
```
JNA could recognize current os in static block when failed on a constructor of SystemInfo class.
So I tried check unknown environment by JNA method instead of OSHI. At the result my CI passed successfully.
I'm not sure what is exact reason. But I just think that JNA and OSHI depends on many static initialization and it's make not easy to assume proper working.

In my opinion, static blocks and variables should be minimized.
But if OSHI have important reason to check unsupported environment in constructor this MR need to approved.

There is remained riddle why OSHI's static variable and JNA static method result don't have consistency.
Anyway, This change working for me.
Please consider enough.

### DIFF
--- a/oshi-core/src/main/java/oshi/SystemInfo.java
+++ b/oshi-core/src/main/java/oshi/SystemInfo.java
@@ -66,29 +66,25 @@ public class SystemInfo {
 
     // The platform isn't going to change, and making this static enables easy
     // access from outside this class
-    private static PlatformEnum currentPlatform;
+    private static final PlatformEnum currentPlatform = queryCurrentPlatform();
 
-    static {
-        initCurrentPlatform();
-    }
-
-    private static void initCurrentPlatform() {
+    private static PlatformEnum queryCurrentPlatform() {
         if (Platform.isWindows()) {
-            currentPlatform = WINDOWS;
+            return WINDOWS;
         } else if (Platform.isLinux()) {
-            currentPlatform = LINUX;
+            return LINUX;
         } else if (Platform.isMac()) {
-            currentPlatform = MACOS;
+            return MACOS;
         } else if (Platform.isSolaris()) {
-            currentPlatform = SOLARIS;
+            return SOLARIS;
         } else if (Platform.isFreeBSD()) {
-            currentPlatform = FREEBSD;
+            return FREEBSD;
         } else if (Platform.isAIX()) {
-            currentPlatform = AIX;
+            return AIX;
         } else if (Platform.isOpenBSD()) {
-            currentPlatform = OPENBSD;
+            return OPENBSD;
         } else {
-            currentPlatform = UNKNOWN;
+            return UNKNOWN;
         }
     }
 
@@ -109,9 +105,6 @@ public class SystemInfo {
      * {@link SystemInfo} object for future queries.
      */
     public SystemInfo() {
-        if (currentPlatform == null) {
-            initCurrentPlatform();
-        }
         if (getCurrentPlatform().equals(PlatformEnum.UNKNOWN)) {
             throw new UnsupportedOperationException(NOT_SUPPORTED + Platform.getOSType());
         }
--- a/oshi-core/src/main/java/oshi/SystemInfo.java
+++ b/oshi-core/src/main/java/oshi/SystemInfo.java
@@ -105,6 +105,9 @@ public class SystemInfo {
      * {@link SystemInfo} object for future queries.
      */
     public SystemInfo() {
+        // Intentionally empty, here to enable the constructor javadoc.
+        // Trying to access the static currentPlatform variable for OS check caused
+        // unexplained problems with initialization.
     }
 
     /**
--- a/oshi-core/src/main/java/oshi/SystemInfo.java
+++ b/oshi-core/src/main/java/oshi/SystemInfo.java
@@ -66,25 +66,25 @@ public class SystemInfo {
 
     // The platform isn't going to change, and making this static enables easy
     // access from outside this class
-    private static final PlatformEnum currentPlatform = queryCurrentPlatform();
+    private static final PlatformEnum currentPlatform;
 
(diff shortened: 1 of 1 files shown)


dbwiddis:
Thank you for testing and for the new proposed fix.   I don't disagree with the code, but until I understand exactly what the root cause of the error is, I am hesitant to accept it.  It could be some sort of race condition that your proposal appears to fix but doesn't.

I would really like to try to reproduce this myself.  Can you tell me what CI system you are using?  If this is not pure Java, what language is involved (is this Kotlin?). 

Based on the initial report I suspect having the code in the constructor may be the root cause, and your suggested fix may not be permanent.  There is probably a bug in some external program somewhere and it is important to try to reproduce it and fix it, rather than just trying workarounds.

dbwiddis:
> In my opinion, static blocks and variables should be minimized.

Generally true, but:
1. This is the top-level entry point into OSHI. It will always be called.
2. The operating system being used will never change. There is no reason to recalculate which operating system is in use every time you re-instantiate a SystemInfo object.  

The existing code follows the order of initialization of static variables/execution of static blocks in Java.  If non-Java code is not observing this convention, there is a bug in that non-Java code that should be fixed, and fixing that code requires finding a reproducible example.   If it is not a bug but a "feature" of that other code (e.g., Kotlin treats default constructors differently) then the reason that language violates Java conventions should be completely understood so it can be properly defended against.

KyongSik-Yoon:
Yes, That's important to find root cause.
I'm using GitLab CI on premise.
My application has both java and kotlin code.
CI runs maven test task that implemented using kotlin and [kotest](https://github.com/kotest/kotest) framework.
It also happened when run maven install or test task without CI.
Maven install and test task run entire unit test of our product.
An exception only occurred in this phase but if run only single unit test it's happened.

KyongSik-Yoon:
I was test on more thing.
An exception also happened with written by java code has same logic.
It just called constructor of SystemInfo class.

KyongSik-Yoon:
> > In my opinion, static blocks and variables should be minimized.
> 
> Generally true, but:
> 
> 1. This is the top-level entry point into OSHI. It will always be called.
> 2. The operating system being used will never change. There is no reason to recalculate which operating system is in use every time you re-instantiate a SystemInfo object.
> 
> The existing code follows the order of initialization of static variables/execution of static blocks in Java. If non-Java code is not observing this convention, there is a bug in that non-Java code that should be fixed, and fixing that code requires finding a reproducible example. If it is not a bug but a "feature" of that other code (e.g., Kotlin treats default constructors differently) then the reason that language violates Java conventions should be completely understood so it can be properly defended against.

I agree with efficiency is very important.
But when I see SystemInfo class I couldn't understand why this class provide two way for access of fields or method.
I think this problem will not happened if SystemInfo class doesn't provide access way using instance for getHardware method.
It's already existed static variable in this class.

dbwiddis:
Thank you for the detailed description of the problem and the simplified test case.  I will try to reproduce it myself using that code.

dbwiddis:
Hi @KyongSik-Yoon I'm still trying to get to the root cause here. 

I have some questions about your testing setup.

You indicated "This error also happened when run maven install or test task without CI." and the debug output above shows only a single test running.  But in your earlier report you said "When I run single unit test. There is no problem."  Can you clarify what you mean by these two statements?
1. Are you running a multi-threaded test setup (one JVM, multiple threads, only one initialization of static final variables)?
2. Are you running a test setup with multiple forks/JVM instances (static final variables initialized once in each fork)?

Are you able to edit OSHI's code to add some log debug statements or use a debugger on the process and execute with that modified code?  The part where it looks like it's failing is where it checks `if (Platform.isWindows())` inside a static method.  It would be useful to print debug output for the result of that check by itself, and also the value of `Platform.getOsType()`.

dbwiddis:
So I've come to to the conclusion that it's probably Kotlin's "constructor first" behavior causing something strange here. I reverted to the OSHI 5.2.x setup for the 5.8.0 release.

I'd still like to try to definitively identify the problem, which may be a bug in how Kotlin interacts with the JVM/JLS.

KyongSik-Yoon:
It's working for me but I'm also wonder what's root cause.
I'll add some debug line to code and trace.

KyongSik-Yoon:
> Hi @KyongSik-Yoon I'm still trying to get to the root cause here.
> 
> I have some questions about your testing setup.
> 
> You indicated "This error also happened when run maven install or test task without CI." and the debug output above shows only a single test running. But in your earlier report you said "When I run single unit test. There is no problem." Can you clarify what you mean by these two statements?
> 
> 1. Are you running a multi-threaded test setup (one JVM, multiple threads, only one initialization of static final variables)?
> 2. Are you running a test setup with multiple forks/JVM instances (static final variables initialized once in each fork)?
> 
> Are you able to edit OSHI's code to add some log debug statements or use a debugger on the process and execute with that modified code? The part where it looks like it's failing is where it checks `if (Platform.isWindows())` inside a static method. It would be useful to print debug output for the result of that check by itself, and also the value of `Platform.getOsType()`.

1. Are you running a multi-threaded test setup (one JVM, multiple threads, only one initialization of static final variables)?
==> I was run maven on IDE(IntelliJ). It's default configuration is single thread base. It's just possible with -T option.

2. Are you running a test setup with multiple forks/JVM instances (static final variables initialized once in each fork)?
==> 
image
As you can see maven default fork count is one. I'm using maven without any additional configuration.

dbwiddis:
But the error occurs when running under Kotlin testing (`kotest`) but not Java testing (`junit`)?

PR #1683, opened 2021-07-19
Initialize static variable from static method instead of static block - rollback

KyongSik-Yoon:
@dbwiddis I'm so sorry. It's my fault.
Before some day, I wrote some unit test for unsupported environment as blow for ensure unsupported platform.
At that time this class didn't have check logic.
```java
ReflectionUtil.setFinalFieldValue(
  SystemInfo::class.java,
  SystemInfo::class.java.getDeclaredField("currentPlatform"),
  PlatformEnum.UNKNOWN
)
OSCompatibleCentralProcessor.getProcessorIfPossible() shouldBe null
```
I should have been recovered currentPlatform forcibly before.
After remove above code, everything worked expectedly.
This MR just include rollback to before.

I appreciate you hear my opinion although my mistake.
And I want say again sorry for my mistake.

### DIFF
--- a/oshi-core/src/main/java/oshi/SystemInfo.java
+++ b/oshi-core/src/main/java/oshi/SystemInfo.java
@@ -143,22 +143,22 @@ public class SystemInfo {
 
     private static OperatingSystem createOperatingSystem() {
         switch (currentPlatform) {
-            case WINDOWS:
-                return new WindowsOperatingSystem();
-            case LINUX:
-                return new LinuxOperatingSystem();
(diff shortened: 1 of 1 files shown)


dbwiddis:
Ah, now it all makes sense.  Thanks for letting me know the cause so I don't have to worry about it any more. :)

I already updated to move the check out of the constructor back into the getters for hardware and software objects, so we can just leave it as is.  This PR is based on code before the latest release.

PR #1698, opened 2021-08-16
Shallow cloning on travis

YunLemon:
According to [https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth](url), Travis CI provide a way to shallow clone a repository. This has the obvious benefit of speed, since you only need to download a small number of commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,6 @@ addons:
 cache:
   directories:
     - $HOME/.m2
-
-# disable shallow clone
-git:
-  depth: false
     
 before_install:
   - |


dbwiddis:
The same comments I made at the similar PR on JNA apply here.
 - We only run one job on Travis
 - We could probably save 10s file transfer time on a 2+ minute build 
   - We do most builds on Github Actions which take longer, so we're never really waiting and the 10s improvement goes unnoticed
 - As outlined on the JNA PR, while overall build time is slower it can use more CPU (both on Github's side to create a custom pack file for the depth and on Travis side to unpack a non-optimized packfile).  Travis systems are more likely to be CPU-constrained than Bandwidth constrained.  It seems a better consumer of resources to be patient for 10 more seconds with an idle CPU.

I'll leave this open for other maintainers to voice any opinions.

dbwiddis:
Dug into git blame.  I intentionally added non-shallow cloning in https://github.com/oshi/oshi/commit/015466093de26e6605ebc7eed5072d82e9f15606 because it was required by our SonarQube CI, which checks for changes over a much longer history than CI.

This isn't really relevant any more, since we've moved the sonar builds to Github Actions, so it's probably OK to go back to a shallow clone for just the ARM build.

YunLemon:
@dbwiddis I have removed git configuration.

PR #1713, opened 2021-09-06
Add more applications and projects using OSHI

ghaseminya:


mprins (changes requested):
I don't see how a hosting provider qualifies as a project or application. 
I don;t think this should be merged

dbwiddis:
Hi @ghaseminya, thanks for this PR.  As @mprins says, you appear to be linking to a "service" and not linking to any "project" (generally open source) or "application" (installable or downloadable software).  

I'm guessing that any use of OSHI is in some sort of monitoring dashboard but that's not documented anywhere on the site. 

Can you clarify how OSHI is used and if users have any control over it?

ghaseminya:
Hi
Thanks for your attention.
Yes @dbwiddis , its right, Javahost didnt clarify that what is monitoring library, and we add a short description about what opensource library we use.
But we use oshi in monitoring pane page and this is that page:
Screenshot_2021-09-07 JavaHost - Hosting Dashboard 110370

dbwiddis:
Great, thanks for the explanation. Can you just edit your PR to include "monitoring" and either pane/page/dashboard or whatever you think fits best here?

dbwiddis (approved):

Issue #1424, opened 2020-11-26
Network MTU returned as -1 when a loopback interface has the maximum value 


mfujitaitrs:
On a Windows environment:

Software Loopback Interface's network MTU is currently reported as -1 which corresponds to unknown or unavailable.  
However, upon comparing against _Get-NetIPInterface_ and _netsh interface ipv4 show interface_, loopback's value for MTU is 4294967295. 

```
Get-NetIPInterface:

ifIndex InterfaceAlias                  AddressFamily NlMtu(Bytes) InterfaceMetric Dhcp     ConnectionState PolicyStore
------- --------------                  ------------- ------------ --------------- ----     --------------- -----------
1       Loopback Pseudo-Interface 1     IPv6            4294967295              75 Disabled Connected       ActiveStore
1       Loopback Pseudo-Interface 1     IPv4            4294967295              75 Disabled Connected       ActiveStore

netsh interface ipv4 show interface:

Idx     Met         MTU          State                Name
---  ----------  ----------  ------------  ---------------------------
  1          75  4294967295  connected     Loopback Pseudo-Interface 1
```
As seen [on this reference](https://social.technet.microsoft.com/Forums/office/en-US/eb1c8794-c5e5-4ee1-9b0d-298d10890e5d/what-is-the-mtu-size-of-loopback?forum=winservergen), it translates to 0xFFFFFFFF, the maximum possible value for an MTU. 
Given that Oshi returns -1, is there are way to distinguish if the loopback has an unknown MTU or has the max value?

dbwiddis:
The MTU is a carry-through from the JDK's `NetworkInterface` class, in which [`getMTU()`](https://docs.oracle.com/javase/8/docs/api/java/net/NetworkInterface.html#getMTU--) returns an `int`, which are always signed in Java.  This value is apparently unsigned 32-bit values at the native level (the JDK derives the value from the 32-bit `dwMtu` field in the [`MIB_IFROW`](https://docs.microsoft.com/en-us/windows/win32/api/ifmib/ns-ifmib-mib_ifrow) structure for that interface).

So -1 does not represent an error here; it faithfully reports the JDK's value, which is either correct or throws a SocketException, and reasonable to convert to a `long` using `mtu & 0xffff_ffffL` (OSHI provides `ParseUtil.unsignedIntToLong()` that does this, and this is probably the best way to handle it at the user level right now.

[...]

PR #1728, opened 2021-10-02
fixed return type of the NetworkInterface method getMTU() 

Simba-97:
fixed return type of the NetworkInterface method getMTU() from int to long in all its OS implementations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,665 +1,699 @@
 # 5.8.3 (in progress)
 
-* Your contribution here
+-   Your contribution here
 
 # 5.8.0 (2021-07-18), 5.8.1 (2021-08-22), 5.8.2 (2021-09-05)
 
 ##### New Features
-* [#1654](https://github.com/oshi/oshi/pull/1654): API for process arguments and environment - [@basil](https://github.com/basil) and [@dbwiddis](https://github.com/dbwiddis).
+
+-   [#1654](https://github.com/oshi/oshi/pull/1654): API for process arguments and environment - [@basil](https://github.com/basil) and [@dbwiddis](https://github.com/dbwiddis).
 
 ##### Bug fixes / Improvements
-* [#1673](https://github.com/oshi/oshi/pull/1673): Fix FreeBSD ps command arguments for context switches - [@basil](https://github.com/basil).
-* [#1675](https://github.com/oshi/oshi/pull/1675): Replace ps argument lists with enum - [@dbwiddis](https://github.com/dbwiddis).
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,665 +1,699 @@
 # 5.8.3 (in progress)
 
-* Your contribution here
+-   Your contribution here
 
 # 5.8.0 (2021-07-18), 5.8.1 (2021-08-22), 5.8.2 (2021-09-05)
 
 ##### New Features
-* [#1654](https://github.com/oshi/oshi/pull/1654): API for process arguments and environment - [@basil](https://github.com/basil) and [@dbwiddis](https://github.com/dbwiddis).
```
dbwiddis, line 8: Don't change the format of the entire file.  Is there a reason the existing format doesn't work well?

I don't see your addition here.

Please rebase to the latest version on master and add your entry to it.

oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
```diff
@@ -62,7 +62,7 @@
     private String name;
     private String displayName;
     private int index;
-    private int mtu;
+    private long mtu;
```
dbwiddis, line 65: It's not enough just to change the type here.   Change the line that assigns this variable:
```
this.mtu = networkInterface.getMTU();
```
That core Java method can return a negative value.  See #1424.  Use `ParseUtil.unsignedIntToLong()` to convert the Java value when assigning it to the instance field.

oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
```diff
@@ -75,10 +75,9 @@
      * Construct a {@link NetworkIF} object backed by the specified
      * {@link NetworkInterface}.
      *
-     * @param netint
-     *            The core java {@link NetworkInterface} backing this object.
-     * @throws InstantiationException
-     *             If a socket exception prevents access to the backing interface.
+     * @param netint The core java {@link NetworkInterface} backing this object.
+     * @throws InstantiationException If a socket exception prevents access to the
+     *                                backing interface.
```
dbwiddis, line 80: Try to avoid whitespace changes like this.  See the oshi-dist src/main/resources folder for an XML formatter file for Eclipse (it can be used with intelliJ also).

dbwiddis (changes requested):
See comments inline.

oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
dbwiddis: Also the unsignedInt parsing in the `toString()` can be removed once the reassignment is done.

Issue #1723, opened 2021-10-01
Remove deprecated code and change return types for 6.X


dbwiddis:
**It's time for Hacktoberfest!**

I'm marking this issue for [hacktoberfest](https://hacktoberfest.digitalocean.com/) and [first-timers-only](https://www.firsttimersonly.com/).  That means that I will only accept PRs for these changes from new contributors to open source as envisioned by the Hacktoberfest project; this can be your first commit, or perhaps a slightly more complex commit than your first if you're just getting started.  Experienced contributors, please leave this issue for those just getting started!

Don't worry if you don't know what to do. Ask for help and I'll hold your hand through this if you need me to. :-) 

As the OSHI 5.x major version has reached a rather stable point, I'm wanting to update API-breaking return types and remove deprecated code as part of a major version bump to 6.x.  You can submit a PR to help out!  

For each of these, please both submit the code to make the change and add appropriate documentation in the UPGRADING.md document.  If you do more than one of the below options, please do a separate PR for each one.

Plese reply to this issue if you're starting working on one of these to avoid anyone duplicating your effort.

- [x] Change the return type of the `NetworkInterface` method [`getMTU()`](https://github.com/oshi/oshi/blob/1aa0c1b9ed13bb5b38a274b440b70abb534c4412/oshi-core/src/main/java/oshi/hardware/NetworkIF.java#L114) from `int` to `long`.  This will require a "signed" `int` to (Java's unsigned) `long` conversion in each of the different OS implementations. `ParseUtil` has an [`unsignedIntToLong()`](https://github.com/oshi/oshi/blob/84a9afc849f52c88042ff1cbcc7d7d212e1b7af0/oshi-core/src/main/java/oshi/util/ParseUtil.java#L426) method you can use for this.

- [x] Change the `OperatingSystem` class method [`getServices()`](https://github.com/oshi/oshi/blob/c5c1b136a50fc82d1c215a9be7da277681dbdba8/oshi-core/src/main/java/oshi/software/os/OperatingSystem.java#L446) to return a `List<OSService>` rather than an array.  This will require chaning the return type for each of the different OS implementations.

- [x] Change the value returned from `MacOSProcess` and `LinuxOSProcess` method `getCommandLine()` from a null-delimited `String` to a space-delimited String.  Update the `OSProcess` javadoc for this method to remove the portion discussing null-terminated String on these two operating systems.  When you update UPGRADING.md add a note to use `getArguments()` as the preferred alternative.

[...]

PR #1729, opened 2021-10-02
Changed the return value of getCommandLine() method of linuxOSProcess and macOSProcess from null-delimited string to space-delimited string

prathamgandhi:
This is a fix to the third issue mentioned in https://github.com/oshi/oshi/issues/1723. The OSProcess javadoc and UPGRADING.md have been fixed as required.

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/OSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSProcess.java
@@ -63,8 +63,6 @@ public interface OSProcess {
      * {@link #getArguments()} which already parses the results, and use this method
      * as a backup.
      * <p>
-     * On Linux and macOS systems, the string is space-character-delimited.
-     * <p>
      * On AIX and Solaris, the string may be truncated to 80 characters if there was
      * insufficient permission to read the process memory.
      * <p>
--- a/oshi-core/src/main/java/oshi/software/os/OSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSProcess.java
@@ -63,10 +63,7 @@ public interface OSProcess {
      * {@link #getArguments()} which already parses the results, and use this method
      * as a backup.
      * <p>
-     * On Linux and macOS systems, the string is null-character-delimited, to permit
-     * the end user to parse the executable and arguments if desired. This
-     * null-delimited behavior may change in future versions and should not be
-     * relied upon; use {@link #getArguments()} instead.
+     * On Linux and macOS systems, the string is space-character-delimited.
      * <p>
      * On AIX and Solaris, the string may be truncated to 80 characters if there was
      * insufficient permission to read the process memory.
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
@@ -39,7 +39,6 @@ import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Stream;
 import java.util.stream.Collectors;
 import java.util.Arrays;
 import java.util.function.Supplier;
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
@@ -39,8 +39,10 @@ import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.function.Supplier;
+import java.util.stream.Stream;
 import java.util.stream.Collectors;
+import java.util.Arrays;
+import java.util.function.Supplier;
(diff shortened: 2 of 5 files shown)


oshi-core/src/main/java/oshi/software/os/mac/MacOSProcess.java
```diff
@@ -133,7 +135,7 @@ public String getPath() {
 
     @Override
     public String getCommandLine() {
-        return this.commandLine.get();
+        return Arrays.stream(commandLine.get().split("\0")).collect(Collectors.joining(" "));
```
dbwiddis, line 138: While this approach was appropriate for Linux, you can look at the `queryCommandLine()` method and see it simply joins `getArguments()` using the null.  So It's better to just join with a space in that method, and then you could return it directly here without further processing.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -118,7 +119,7 @@ public String getPath() {
 
     @Override
     public String getCommandLine() {
-        return commandLine.get();
+       return Arrays.stream(commandLine.get().split("\0")).collect(Collectors.joining(" "));
```
dbwiddis, line 122: Since the commandline won't change as long as the process is active, this parsing can be done in the memoized `queryCommandLine()` method so it only has to be done once.

oshi-core/src/main/java/oshi/software/os/OSProcess.java
```diff
@@ -63,10 +63,7 @@
      * {@link #getArguments()} which already parses the results, and use this method
      * as a backup.
      * <p>
-     * On Linux and macOS systems, the string is null-character-delimited, to permit
-     * the end user to parse the executable and arguments if desired. This
-     * null-delimited behavior may change in future versions and should not be
-     * relied upon; use {@link #getArguments()} instead.
+     * On Linux and macOS systems, the string is space-character-delimited.
```
dbwiddis, line 66: All the OS's are now space-character delimited so you don't need to call out Linux and macOS separately here.  I'd just completely delete this section since the paragraph above indicates the necessary information for the user.

dbwiddis (changes requested):
Thanks for this PR.  A few comments inline -- since we're using memoizers to only fetch the information once the processing should be done in the memoized query method.

oshi-core/src/main/java/oshi/software/os/mac/MacOSProcess.java
prathamgandhi: All right, I've fixed it and understood how it improves performance. Could you maybe explain in brief why linux and mac do this differently? Like what section makes it platform-dependent and if it improves performance in both the cases, just for learning's sake.

dbwiddis: Linux provides the command line arguments via a null-delimited string in the `proc` filesystem, specifically `/proc/[pid]/cmdline`.  (See [man proc](https://man7.org/linux/man-pages/man5/proc.5.html).)

macOS provides the command line arguments via `sysctl` call (`KERN_PROCARGS2`) in a null-delimited buffer.  We can iterate that buffer to fetch the strings.

Before we (recently) added methods to fetch the process arguments separately,  and because of the native null-delimited strings, these two OS's provided the ability to parse the arguments without having to do special treatment of spaces in names, etc.  Rather than "lose information" by replacing the nulls with spaces, we provided users the full null-delimited string to parse to get the argument list themselves.

For the rest of the Unix-based implementations we just fetched the space-delimited argument string from the `ps` output (which is truncated to 80 characters on Solaris), and we got a space-delimited string from Windows via WMI (with additional parsing complexities using quotation marks).  Parsing space-delimited command lines is not trivial, and rather than try to parse all these to a list we just returned them as-is to the user and let them do the parsing.

With the addition of a separate list of command line arguments, user-level parsing of `getCommandLine()` is no longer necessary, so the goal of this fix was just to simplify the getCommandLine output to the same common "user readable" space-delimited output on all OS's.  Users who wanted to simply display a user-readable string had to do special cases for macOS and Linux to process the nulls. They will no longer have to do so.

There is still some bit of platform dependence in the output.  While all the other Unix implementations now combine the argument list, that approach fails if you're not running with elevated permissions and querying a process you don't own.  For those cases, the `ps`-based parsing does still work (since `ps` is owned by root) and is provided as a "backup" option (so for example on Solaris not running elevated, you'll get a truncated 80-character output for some processes and full output on others that you own).  Bottom line:  the method should now be treated only as a cosmetic "best effort output" and anyone looking to actually parse the output should use `getArguments()`.

dbwiddis (approved):

oshi-core/src/main/java/oshi/software/os/mac/MacOSProcess.java
prathamgandhi: Ohhh, I understand now. Thanks for the explanation and all the help. Learnt a lot from this.

PR #1732, opened 2021-10-03
Fixed return type of the NetworkInterface method getMTU() in all its OS implementations

Simba-97:


adrian-kong:
i think u need to add L to NetworkTests when it checks whether getMTU==0 need to be 0L?

dbwiddis (changes requested):
As @adrian-kong mentioned you need to update the test to explicitly use a long type for the zero.

Also you have the same issue on this PR as the previous one where you completely reformatted the change log.   Revert your file to the existing one on the master branch and then just add a single line.

dbwiddis:
Also if you have changes to your PR  you can simply push to your branch, you don't need to close the PR and open a new one.

Simba-97:
@dbwiddis Can you check now please? Please let me know what am I doing wrong.

dbwiddis:
OK, a few things.
 - I think you've correctly fixed the `long` type of the 0, but there is so much whitespace change in the diff, it's really hard to review.  
   - Please use a code formatter to match the standard OSHI code format.  You can look under `oshi-dist/src/main/resources/eclipse` to see an xml file you can load as a formatter for Eclipse, and it should also be able to be used with IntelliJ with an appropriate add-in.
 - You'll notice the MTU test was moved down to the "Local networks" section with a comment about windows returning negative. This won't happen anymore with the code fix, so that test can actually be moved back up to where the rest of the tests are
 - You still have a massive change of every single line's format in the CHANGELOG.  Don't do this; just add one line matching the existing format.

Issue #1727, opened 2021-10-01
Write JUnit tests for Driver Classes


dbwiddis:
**It's time for [Hacktoberfest](https://hacktoberfest.digitalocean.com/)!**

I'm marking this issue for [hacktoberfest](https://hacktoberfest.digitalocean.com/).  That means that I will only accept PRs for these changes from new(ish) contributors to open source as envisioned by the Hacktoberfest project.  Experienced contributors, please leave this issue for those just getting started!

These tasks are slightly more complex than some first-time contributions and targeted for slightly more experienced Java coders who are new to git and/or GitHub who would like a more challenging project than a trivial change.  But they are still very accessible to beginners.

OSHI's future design envisions the majority of native code executed in a "driver" class (see the packages under `oshi.driver` for examples).  These are mostly stateless `static` methods that can be called by the appropriate platform-specific implementation.  Drivers have been used for most recent code added in the last year or two.

Currently, all of OSHI's testing is at the API level (Integration Testing) which means that when there are fallbacks for failures, either the fallback never gets tested, or the primary test fails gracefully, the fallback works, and tests don't fail.  While many of these types of bugs have been fixed after bug reports, I'm sure there are probably a few bugs remaining in driver classes that we won't find without testing.  You could help us find those bugs!

The task: Write tests execute driver class methods and verify expected output per the API.

There are dozens of these drivers available for new committers to work on.  Pick one or a few and have at it!  Once the majority have been written I'll update this issue with those still remaining.

Steps you'll take:

1. Pick a driver class in the `oshi-core/src/main/java` path that has not yet been worked on.  You can navigate the `oshi.driver.*` package tree.  Right now all are available; as more get written I'll update this section with remaining ones.

2. Create an identically named class with the word `Test` appended to it in the `oshi-core/src/test/java` tree.  For example, `oshi.driver.linux` package, `Devicetree.java` class in `src/main/java` should have a corresponding `oshi.driver.linux` package, `DevicetreeTest.java` class.

3. In the test class, write test methods to exercise each method in the driver.   For example, for the `DeviceTree` class in the previous step, there is a `queryModel()` method.  Write a corresponding `testQueryModel()` test method.  

4. Write JUnit 5 compatible tests within the method.  Look at other OSHI test classes for examples; generally you'll want the `@Test` annotation and an assert method associated with the output, which should correspond with the documentation (non-null, etc.)
[...]

PR #1734, opened 2021-10-05
LshwTest and DmidecodeTest for Linux Driver

adrian-kong:
Added some more tests for #1727

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * [#1724](https://github.com/oshi/oshi/pull/1724): Removed deprecated MACOSX value from PlatformEnum and SystemInfo and removed the getCurrentPlatformEnum() method - [@Novaenn](https://github.com/Novaenn).
 * [#1725](https://github.com/oshi/oshi/pull/1725): Removed deprecated process sorting methods from the OperatingSystem class - [@varnaa](https://github.com/varnaa).
 * [#1729](https://github.com/oshi/oshi/pull/1729): Changed the return value of linuxOSPRocess and macOSProcess method getCommandLine() from null-delimited string to space-delimited string - [@prathamgandhi](https://github.com/prathamgandhi).
-* [#1730](https://github.com/oshi/oshi/pull/1730): Changed the return value of getServices() in OperatingSystem - [@adrian-kong](https://github.com/adrian-kong).
+* [#1730](https://github.com/oshi/oshi/pull/1730): Changed the return value of getServices() from OSService[] to List<OSService> in OperatingSystem - [@adrian-kong](https://github.com/adrian-kong).
 
 # 5.8.0 (2021-07-18), 5.8.1 (2021-08-22), 5.8.2 (2021-09-05)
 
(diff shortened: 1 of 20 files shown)


adrian-kong:
i'm not sure what to check so i just checked whether they were null or non empty
also for Dmidecode#queryUUID, can't you run command "dmidecode -s system-uuid" or "dmidecode -s system-serial-number" so you don't have to parse the string itself + splitting + trimming the string?

adrian-kong:
also, would java version be updated? the old switch statements are very ugly

dbwiddis:
> i'm not sure what to check so i just checked whether they were null or non empty 

That's reasonable. For drivers that are documented to return null, the null is usually handled.  So we're looking to execute code that might throw some sort of exception.

> also for Dmidecode#queryUUID, can't you run command "dmidecode -s system-uuid" or "dmidecode -s system-serial-number" so you don't have to parse the string itself + splitting + trimming the string?

I haven't looked into the details of this, but I do want to comment that I am thrilled to have a second set of eyes on this code.  If those commands/switches are sufficiently standard across distributions, I'm all for improving efficiency.  I know in some cases we run one command to pull several bits of information out of it, but if there's a way to filter/shorten the return to only get the information we want without added parsing, that's fine and I welcome a fix.

adrian-kong:
yea i searched up dmidecode before and some people use `dmidecode` -s system-serial-number` and `dmidecode -t system | grep "Serial Number"` so im not sure the difference but it looked like less splitting the string with system-serial-number.
however apparently some people get random things as output using dmidecode:
https://community.hpe.com/t5/System-Administration/How-to-get-Serial-number-in-Linux-w-o-using-dmidecode/m-p/6421524/highlight/true#M54558

dbwiddis (approved):

dbwiddis:
>  however apparently some people get random things as output using dmidecode

TBH, System Serial Number is a huge headache. Lots of issues in history over it.  Linux decided they wanted to obscure that data for privacy reasons so it's not easily available without elevated permissions or other user cooperation.  Which is a win for privacy advocates, but IMO a bit aggressive when there are far easier unique ID features available.

dbwiddis:
Uh, anyway, I'm ok merging this PR unless you wanted to change the underlying code?

adrian-kong:
> Uh, anyway, I'm ok merging this PR unless you wanted to change the underlying code?

should i change to dmidecode -s system-serial-number and the uuid for brevity or leave as it is?

dbwiddis:
> should i change to dmidecode -s system-serial-number and the uuid for brevity or leave as it is? it might change the output of the methods `queryUUID()` and `querySerialNumber()` since it would no longer check if the string contains the marker. i think there would be no null output instead only empty string outputs

I think I'm leaning toward leaving it as-is, mainly because these are backups and should be as robust as possible.   If running under elevated permissions, we should never get here because the sysfs command should work.
```
    private static String querySerialNumber() {
        String result = null;
        if ((result = Sysfs.queryProductSerial()) == null && (result = Dmidecode.querySerialNumber()) == null
                && (result = Lshal.querySerialNumber()) == null && (result = Lshw.querySerialNumber()) == null) {
            return Constants.UNKNOWN;
        }
        return result;
    }
```
Plus we rely on the `null` return as a fallback.

So I'm in the "If it ain't broke don't fix it" camp, particularly for "backups".

adrian-kong:
ok epic i agree with that motto

dbwiddis:
OK, so I merged this, but Coverity is complaining about null dereferences in the tests.

Not an overly big deal as a NPE causes a failed test, but I'm wondering if there's a way we can restructure the tests to avoid the null dereference.  Does order matter on anyOf?

dbwiddis:
Or we can just null check the test, essentially accomplishing the "don't fail for null" goal much more cleanly.
```
if (foo != null) {
   assertThat(foo, is(something));
}
```

adrian-kong:
> Or we can just null check the test, essentially accomplishing the "don't fail for null" goal much more cleanly.
> 
> ```
> if (foo != null) {
>    assertThat(foo, is(something));
> }
> ```

yep that's a simple and easy solution

dbwiddis:
Do you want to handle it then? :) Here's a list:
image

PR #1737, opened 2021-10-06
Removed nullValue() in tests for Coverity

adrian-kong:
a pr a day keeps the doctors away


adrian-kong:
#1734

dbwiddis:
> also for `WmiQueryHandler#createInstance()`, other methods refer to it when its null since it's caught inside this method, should callers wrap Object.requireNonNull() ?

Possibly.

All the possible exceptions are generally ones that should be found immediately by the developer trying to use a "bad" class.  Perhaps it would be better to not catch/log the exceptions (and eventually throw a NPE) but in any case, if you're getting a null after creating a custom class, you broke something and you should fix it.

Let's discuss that in a different issue/PR if you have a better idea, though.  I think this one's good to merge....

PR #1744, opened 2021-10-17
[config] Update build config

hazendaz:
adopt is dead.  They renamed it to something else.  Zulu is microsofts.  Updating maven and jdk otherwise.


dbwiddis:
Thanks for the update and also the JDK version bump that was on my to-do list for a while...

PR #1746, opened 2021-10-18
Introduced mockito and added tests for ThreadInfo & Who for Mac driver

iamdebojyoti:
In this commit we added,

1.  Mockito as a dependency
2.  Wrote two test classes with mocking

### DIFF
--- a/oshi-core/pom.xml
+++ b/oshi-core/pom.xml
@@ -21,7 +21,7 @@
     </scm>
 
     <properties>
-        <!-- Users of the Spring Boot Starter Parent should include this property 
+        <!-- Users of the Spring Boot Starter Parent should include this property
             in their POM -->
         <jna.version>5.9.0</jna.version>
         <main.basedir>${project.parent.basedir}</main.basedir>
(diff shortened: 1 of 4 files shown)


dbwiddis:
> never saw 'junit-platform-maven-plugin' before. While in here already, that works better than surefire? Not sure if that is an issue or not.

@hazendaz It's a super-easy way of writing tests that execute in both modular and non-modular environments.  See [this blog](https://sormuras.github.io/blog/2018-09-11-testing-in-the-modular-world.html).

dbwiddis:
> I see hamcrest being used. Not sure what is bringing that in but junit 5 doesn't require that any longer. 

Probably me explicitly wanting to use some matchers in it that aren't in Junit 5.

> I would suggest using assertj-core instead. Its entirely drop in place and aggressively supported.

This suggestion feels like deja vu, and I think I attempted this given the "drop in place" comment, and it wasn't.  Just about every assertion would need to be rewritten.  There are scripts to help with that but they don't handle the optional comment string we use.

I am not opposed to a complete redesign of the testing system and completely rewriting all the tests in the best system.  I'll even do a lot of the work.   I do not want to do the design, though, so I need someone to step up and plan this thing.

dbwiddis:
I'm closing this PR as it has test failure issues and appears to have been abandoned (no update in >30 days).   Please consider the comments on #1794 if you want to resubmit any Mockito dependencies (i.e., plan to do a lot more than just one, if this is something you want.)

PR #1772, opened 2021-11-14
Fix enums should be immutable by add the final modifier to enum fields

LENHUTDUY-8120:
Please review this and merge if possible. Thank you!


dbwiddis (approved):
This looks great!   Thanks for your contribution.

PR #1776, opened 2021-11-17
Added a parameter StandardCharsets.UTF_8 to each of the above 3 lines

VedantPol:


oshi-core/src/main/java/oshi/util/ParseUtil.java
```diff
@@ -1344,7 +1344,7 @@ public static String removeLeadingDots(String dotPrefixedStr) {
                 }
                 // Otherwise add string and reset start
                 // Intentionally using platform default charset
-                strList.add(new String(bytes, start, end - start));
+                strList.add(new String(bytes,StandardCharsets.UTF_8));
```
dbwiddis, line 1347: You still need to limit the string length with `start, end-start`.  You'll ad the charset as a fourth parameter.  Use this constructor:  https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#%3Cinit%3E(byte%5B%5D,int,int,java.nio.charset.Charset)

oshi-core/src/main/java/oshi/util/ParseUtil.java
```diff
@@ -1378,11 +1378,11 @@ public static String removeLeadingDots(String dotPrefixedStr) {
                 }
                 // Otherwise add string (possibly empty) and reset start
                 // Intentionally using platform default charset
-                strMap.put(key, new String(bytes, start, end - start));
+                strMap.put(key, new String(bytes,StandardCharsets.UTF_8));
```
dbwiddis, line 1381: same comment as line 1347

oshi-core/src/main/java/oshi/util/ParseUtil.java
```diff
@@ -1378,11 +1378,11 @@ public static String removeLeadingDots(String dotPrefixedStr) {
...
                 key = null;
                 start = end + 1;
             } else if (bytes[end] == '=' && key == null) {
-                key = new String(bytes, start, end - start);
+                key = new String(bytes, start, end - start,StandardCharsets.UTF_8);
```
dbwiddis: This is correct and the way you should do the other two.

dbwiddis (changes requested):
Good job submitting a PR.   One of the three changes was correct, please fix the other two.

dbwiddis:
You don't need to close the PR to fix it. Just push changes to your new branch. But I see you've opened #1777 so I'll go there. :)

PR #1800, opened 2021-11-26
adding final to the enum

jamal-eldeen-wari:


oshi-core/src/main/java/oshi/PlatformEnum.java
```diff
@@ -81,19 +81,19 @@
...
     }
 
     /**
      * Gets the friendly name of the platform
      *
      * @return the friendly name of the platform
      */
-    public String getName() {
-        return this.name;
+    public String getsName() {
```
dbwiddis, line 95: This is an API breaking change.  We don't do these without good reason.

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
```diff
@@ -288,18 +288,18 @@
...
+            this.jIndex = value;
         }
 
         /**
          * @return The integer index of this ENUM in the processor tick arrays, which
          *         matches the output of Linux /proc/cpuinfo
          */
-        public int getIndex() {
-            return index;
+        public int getjIndex() {
```
dbwiddis, line 301: API breaking change.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -421,14 +421,14 @@ private static void getMissingDetails(Map<String, String> status, String stat) {
         PPID(4), MINOR_FAULTS(10), MAJOR_FAULTS(12), USER_TIME(14), KERNEL_TIME(15), PRIORITY(18), THREAD_COUNT(20),
         START_TIME(22), VSZ(23), RSS(24);
 
-        private final int order;
+        private final int JOrder;
 
-        public int getOrder() {
-            return this.order;
+        public int getJOrder() {
```
dbwiddis, line 426: API breaking change.

dbwiddis (changes requested):
I am completely unclear on why any of this needs to be done.  Also changing the API requires a major version bump and there's absolutely no reason to do so in this case.

And nothing here is adding final to enums. That PR was completed in #1772

PR #1801, opened 2021-11-26
change enum to final and change variable name

QusaiTA:


dbwiddis:
I'm not clear on the reasoning behind these variable renames. The first one with an uppercase letter doesn't seem to fit with coding conventions.  What convention/best practice is motivating these changes?

Issue #1797, opened 2021-11-25
Replace code-assert with alternative checkstyle test


dbwiddis:
> I also think we should get rid of `guru.nidi:code-assert` alltogether as it has a humongous dependency tree and it's only used in one testcase:
> 
> _Originally posted by @mprins in https://github.com/oshi/oshi/issues/1796#issuecomment-979039947_

The one testcase code-assert is used in is completely intended to give a runtime test enforcement of checkstyle issues.   Prior to using this, we were only finding these issues in the Appveyor builds, and when I say "finding" I mean failing builds and spending 10 minutes poring over the log to try to figure out the issue, instead of having it print out nicely.

I'm okay with removing code-assert if we properly configure the checkstyle plugin to fail the build on warnings, see [here, for example](https://stackoverflow.com/questions/52006084/maven-checkstyle-plugin-doesnt-fail-during-build-even-though-failsonerror-is/52006315).  Other potential options are using one of our existing CI integrations that integrates with PRs, or a Github Action that tests checkstyle, like [this one](https://github.com/jwgmeligmeyling/checkstyle-github-action)

Issue #1794, opened 2021-11-25
Release 5.8.4 not recommended - expect 5.8.5 soon


dbwiddis:
I went through the process of creating release 5.8.4 today.  Unfortunately once you release something to the Central Repository you can't delete it, so it's out there.   I'm already working on a 5.8.5, so don't hurry to update.

**TLDR:**  

'll be removing the `FsstatTest` introduced in #1743, and all dependencies on Mockito.  And an accidental dependency on Apache Commons StringUtils in #1747 that works when we have the codeassert dependency, but we remove that for Java 11 due to incompatibilities.

**Details:**

FYI: @squid-ney and @iamdebojyoti

There was a recent discussion in #1743 and #1746 about adding a dependency on Mockito.  

I went ahead and accepted it in #1743 without doing a careful enough review.
 - `mockito-core` was not properly scoped as a test dependency, so it's a compile-time dependency in the latest release, even though it's not used
 - there's an unknown incompatibility between the mockito dependency tree and our current Junit testing, when testing in Java 11.  I don't know what it is, and I'm not going to spend any time debugging it.  
 - Mockito's dependency tree includes a dependency on Objenesis, which uses classes in sun.misc.Unsafe which I'd prefer not to use without good reason.  A single test class using the dependency is not a good enough reason in my book. :)

#1746 still hasn't been merged, but is experiencing test failures that I don't understand.   

Bottom line: If we are going to include Mockito, it's probably going to be part of a complete rewrite of our testing system, keeping modularity in mind.  And that complete rewrite will be done by someone other than me.

dbwiddis:
OK, so after removing Mockito I still had the issues, and realized it was many of the new tests incorrectly using `org.junit.Test` instead of the Jupiter API version.   Gah.  I hate modules.   So I might try putting that back.  Later, though.  Too many wasted hours tonight.

mprins:
>  new tests incorrectly using org.junit.Test

I think we should add an enforcer rule to prevent `junit:junit` sneaking it's way in, Ive done some junt 4 to 5 migrations and found out that the transitive dependency on `junit:junit` in many projects is not properly scoped and even when it is an ide (and maven) will pick up and use it without qualm
Having an enforcer rule can prevent that by early failure - making it necessary to add proper excludes to the dependency

hazendaz:
Use this -> https://github.com/skuzzle/restrict-imports-enforcer-rule, works well to block the import usage.  You could also block junit4 as well as noted.

[...]

PR #1802, opened 2021-11-27
Disallow imports not on module path

### LINKED ISSUE #1799: Block imports not on module path
We want to avoid importing classes to the project from any packages not included as top-level dependencies (which are also included in the module descriptor for the java11 branch).  This aligns with the Java Modules need for an explicit `requires transitive` statement from these dependencies, and helps keep surprises with java11 releases to a minimum.

Consider the below maven plugin if it's easy to use, or find another way to block imports that aren't explicitly listed as dependencies in the pom, or specifically included (with associated `requires transitive` statement).

> Use this -> https://github.com/skuzzle/restrict-imports-enforcer-rule, works well to block the import usage.  You could also block junit4 as well as noted.
> 
> _Originally posted by @hazendaz in https://github.com/oshi/oshi/issues/1794#issuecomment-978969292_

dbwiddis:
Fixes #1799 

(Aside: I don't know why my formatter adds a space inside wrapped XML comments and have spent more time than I would care to trying to fix that, and failing.)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project
+    xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>com.github.oshi</groupId>
@@ -104,6 +107,7 @@
         <maven-deploy-plugin.version>3.0.0-M1</maven-deploy-plugin.version>
         <nexus-staging-plugin.version>1.6.8</nexus-staging-plugin.version>
         <maven-enforcer-plugin.version>3.0.0</maven-enforcer-plugin.version>
+        <restrict-imports-enforcer-rule.version>1.4.0</restrict-imports-enforcer-rule.version>
         <maven-gpg-plugin.version>3.0.1</maven-gpg-plugin.version>
         <maven-install-plugin.version>3.0.0-M1</maven-install-plugin.version>
         <maven-resources-plugin.version>3.2.0</maven-resources-plugin.version>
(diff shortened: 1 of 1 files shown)


mprins (commented):
+0 It seems like a lot of work, and it still won't catch the case where a fully qualified classname is used in the code (and thus no import)

I've never used the [banTransitiveDependencies](http://maven.apache.org/enforcer/enforcer-rules/banTransitiveDependencies.html) but it seems like that would catch most of this as well.

dbwiddis:
> +0 It seems like a lot of work

Well I've already done the work, heh.

> and it still won't catch the case where a fully qualified classname is used in the code (and thus no import)

Yeah, but hopefully that'll be a bit more obvious to us when reviewing code. 

> I've never used the [banTransitiveDependencies](http://maven.apache.org/enforcer/enforcer-rules/banTransitiveDependencies.html) but it seems like that would catch most of this as well.

That's a totally different use case.  That actually prevents our dependencies from using their own dependencies.  (So we'd have to exclude those transitives in the `<dependency>` section.

I recognize it's currently a little unwieldy, but:
 - While I named this PR to ban transitive dependencies, what I *really* want is to restrict dependency packages to exactly the contents (which are exported) of the module descriptors of our dependencies.  This is actually even more restrictive than just the dependencies (which is good)
 - This is a package-level enforcer which is therefore precisely what I'm looking for; the other proposed solutions are dependency-level
 - I've filed an issue on the site about the duplication required by "static foo", and in version 2 that part will be a single switch (defaulting to true) which greatly simplifies it.
 - It's a little broad in including all of `java.*` rather than just the `java.base` module packages (which I'm not going to spam), but that's also a potential future improvement to import from an external file.

dbwiddis:
I do see that this package-level functionality is in checkstyle as well, so we could perhaps go that route instead.  Seems to be roughly the same amount of work.  I'll leave this PR here while trying to tackle this in #1797.

> I think code inspection is better done using checkstyle, eg. using https://checkstyle.sourceforge.io/config_imports.html#ImportControl
as that will catch class use and isn't limited to `imports`  as noted in https://github.com/skuzzle/restrict-imports-enforcer-rule#limitation
> 
> _Originally posted by @mprins in https://github.com/oshi/oshi/issues/1794#issuecomment-979043483_

dbwiddis:
So, I configured checkstyle using the default rules which is all well and good.  Had two ways to go:
 - use the default and add suppressions (I did this)
 - copy the default and remove the ones I don't want (started, and abandoned this as basically unmaintainable)

Now I looked at how to do this to add the ImportControl rule, and everything I read says I can't just add it to the existing default (`sun_checks.xml`), I have to basically copy the default and add to it.  What?  Seriously?

Surely I'm missing something. Importing another config and adding to it is not exactly a corner use case.

hazendaz:
I think the sun checks are rather old.  There is a google one as well they include that works well for us.  Not too hard maintain as their jar as the file(s).  So on any new update, if something doesn't work, just compare the file from latest checkstyle with one in repo and have block comments around custom items so its clear.  Yes that is extra work and checkstyle verioning never tells you when it will hard break.

dbwiddis:
> I think the sun checks are rather old. 

So are the sun coding conventions but I still basically use the "Sun" config in Eclipse...

> There is a google one as well they include that works well for us. 

Yeah, with code-assert I included both checks (with some suppressions).

Maybe if I'm rolling my own I'll just do the same thing here.

> Yes that is extra work and checkstyle verioning never tells you when it will hard break.

Ugh.

dbwiddis:
> 0 It seems like a lot of work

Closing in favor of #1803, which was a lot more work :-)

PR #1808, opened 2021-12-01
Restrict imports with maven enforcer

dbwiddis:
Redo of #1802.

WIth the new version I don't have to specify static separately, so the configuration is pretty streamlined.  I like that the checkstyle plugin can be very strict on the modular part, allowing this to be a little more lenient.  But using enforcer has the advantage of working with mvn test (or most other maven commands), and not requiring a separate "checkstyle:check" call.  

Set as draft because it's using a snapshot dependency, but wanted to stick this up here for @mprins  and @hazendaz  review.  Will wait to merge until it's a released version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,7 @@
             <url>github:ssh://oshi.github.io/oshi/</url>
         </site>
         <snapshotRepository>
-            <id>ossrh</id>
+            <id>ossrh-snapshot</id>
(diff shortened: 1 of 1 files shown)


hazendaz (approved):
LGTM, interesting move on that project to 'develop' line that I would not have otherwise even noticed.  I like how this enforces this project to use only what it needs and nothing more.

Issue #1819, opened 2021-12-30
Revamping README and site pages


TurboFiero:
Forgive my ignorance but it seems there isnt even a guide on how to use this?

Claims to be compatible with everything, alright then, show us? 

launching with java does literally nothing?

dbwiddis:
As a library, OSHI is not intended to be run as a stand-alone application. It provides a pure Java interface that you can incorporate into your own application.

There are specific instructions in the README document (also the project's github homepage) showing:
1. Usage: https://github.com/oshi/oshi/blob/master/README.md#usage
2. Getting output from your own machine: https://github.com/oshi/oshi/blob/master/README.md#output

In addition, several common implementation examples are implemented as proof-of-concept in the `oshi-demo` artifact here: https://github.com/oshi/oshi/tree/master/oshi-demo/src/main/java/oshi/demo

If these links do not provide enough information for you to use the library, please be more specific in what you're trying to do and we can give a more specific answer.

dblock:
Maybe mention https://github.com/rememberber/MooInfo as a demo very early in the README, or another example @TurboFiero please contribute to the README if it's unclear. 

I suggest closing this because if the README isn't  "a guide on how to use this", I don't know what is.

dbwiddis:
> I suggest closing this because if the README isn't "a guide on how to use this", I don't know what is.

I think I'll move the SystemInfoTest snippet from "Output" to the "Usage" (how to use) section of the README. The whole output section could probably be trimmed to just show the screencaps.

I'll leave this open for now as a reminder to myself.

adrian-kong:
> I think I'll move the SystemInfoTest snippet from "Output" to the "Usage" (how to use) section of the README. The whole output section could probably be trimmed to just show the screencaps.

oh yea maybe a mkdocs/docsify or something documentation site generator would be really cool maybe on like readthedocs.io
or just github pages ? and we can move some of the readme to there to keep everything nice and clean

that would be something i could contribute to but looks like someone else already took the name oshi https://oshi.readthedocs.io/en/latest/

dbwiddis:
For most github projects, the README with markdown is sufficient. Certainly it could be trimmed up with some organization.

That other project looks abandoned and most companies have procedures to claim abandoned names. (Our github oshi org was pre-owned.)

We do have github pages but it just mirrors the README: https://oshi.github.io/oshi/README.html (but could be customized).

adrian-kong:
[...]

PR #1832, opened 2022-01-04
Small refactor to the site

adrian-kong:
For #1819 
- Fixed the broken images (?)
- Created separate pages for huge chunks of information


adrian-kong:
there wasn't much to add/change because oshi is already so simple to use by reading the javadocs but i just thought separating out the information would be better to read instead of a huge chunk on README that needs to scroll through for new users maybe

mprins (approved):
personally I'm not too keen on all those capitalised filenames, but others may think differently

adrian-kong:
> personally I'm not too keen on all those capitalised filenames, but others may think differently

same i just followed the README.html format also maybe the markdown folder could also use some refactoring with html files and png files

adrian-kong:
also oops it shows 2 commits on the logs i have no clue what I just did
did i just merge an already merged PR

dbwiddis:
One is a "merge commit" which is actually empty. The other was the actual commit.  You should be using "squash and merge" which combines all the individual commits into the one PR merge commit and keeps the log history cleaner.

Note you only changed the README for the site, which would need to actually be republished (it happens as part of a release) so nothing has changed yet.  The main README on the github home page is the same.  Also typically I've been copy/pasting the main readme to the SITE readme, so maintaining two separate versions adds a lot of overhead to releases.

I still think we might want to re-organize the main page to match, as I'm really not sure how much traffic the github pages output actually sees. 

Regarding the all caps filenames, that has been a consistent format at least on the main site homepage for key documents needed by developers, e.g., README, UPGRADING, CHANGELOG.   If we're reorganizing content we can probably point to a single source (e.g., under the site tree) referenced from the main page.

Also the "projects using OSHI" from the main page README was not copied into the site version, but it would be nice to stick all those users/usages onto their own page somewhere.

Bottom line: thanks for the PR but I think there's a bit more to do :)

Issue #927, opened 2019-08-02
Javadoc @author tag is unnecessary and possibly harmful


dbwiddis:
We should [stop using the `@author` tag](https://www.databasesandlife.com/dont-use-the-author-javadoc-tag/).  It's [not included as part of the generated javadoc](https://www.vojtechruzicka.com/stop-using-javadoc-author-tag/), and we haven't bothered updating it when making large changes to the code so in [most cases it's inaccurate](https://implementsblog.com/2018/02/16/dont-use-the-author-javadoc-tag/).  Should we ever wish to relicense under [Apache, it's required to remove them](https://blog.mkowalski.net/2016/12/29/still-using-author-please-dont/).

Git history tracks all our contributions.

No need to mass edit files to remove the tag if that's the only change, but if you're making other changes to a file, kill the `@author` tag while you're at it.  Leaving this issue open as a reminder until the number of edits is small enough for a final cleanup.

cilki:
What are your thoughts on the `@since` tag?

dbwiddis:
It's not harmful, but I don't think it's really needed either.

spyhunter99:
mm i dont think the author tag is enforced. it's everywhere in asf repos

dbwiddis:
Well, enforced or not, we don't need it and the multiple linked articles above explain why :)

hazendaz:
Just my 2 cents.

I agree, use of 'at author' is bad for quite a number of reasons.  Many open source projects have been removing it.  There is a point where you need to blame the code for an issue not the original author(s).  Plus git provies blame of the commit directly which still is nicer that calling a person out on something they did at some point in time.  "Blame the code not the author".  We have long stopped using it at my day job for that very reason.  And I insist on repos I work on to not see it come in.

'at since' is ok but that only matters on a very public api and and needs to be consistently used otherwise it's a bit useless.  The history still can indicate when something came in and if users are using the latest which we hope they are, then it's use deminishes.  One prime example of it being noise in my opinion is a recent additionl to 'spotbugs-maven-plugin'.  The submitter used what would be thought to be next version release but I had to change the version number for a patch revision and as such I had to additionally change that commited code.  Further, prior new code was not using 'at since' at all.  So that non consistency gives some very real false impressions.

dbwiddis:
[...]

PR #1850, opened 2022-01-15
Skip license header rewrites on CI

dbwiddis:
The `license-maven-plugin` version 4.2 enabled a new feature to use the file's creation date and last edit date as a copyright range, which we have now configured.  However, this new feature:
 - is resource intensive, navigating the git history for each file
 - breaks on shallow clones, since git history doesn't identify the oldest commit per file properly

Our CI workflows using the `actions/checkout` GitHub Action are shallow clones (to improve performance), using a default `fetch-depth` of 1.  This results in all license headers being incorrectly rewritten (and then thrown away) during CI runs.

Even without rewriting, the multiple checks are unneded.  This PR skips the `license:format` goal for CI, and only executes `license:check` once on PRs (after fetching full history).

### DIFF
--- a/.github/workflows/aix.yaml
+++ b/.github/workflows/aix.yaml
@@ -35,4 +35,4 @@ jobs:
           git checkout master
           git reset --hard HEAD~2
           git pull upstream master
-          ./mvnw clean test -B -Djacoco.skip=true
+          ./mvnw clean test -B -Djacoco.skip=true -Dlicense.skip=true
--- a/.github/workflows/freebsd.yaml
+++ b/.github/workflows/freebsd.yaml
@@ -30,5 +30,5 @@ jobs:
             pkg install -y curl
             pkg install -y openjdk8
           run: |
-            ./mvnw test -B -Djacoco.skip=true
+            ./mvnw test -B -Djacoco.skip=true -Dlicense.skip=true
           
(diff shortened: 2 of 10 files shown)


hazendaz (approved):

hazendaz:
@dbwiddis Just for clarity on this new feature of license plugin, its different than the original check of inception date on project in that the starting date will vary per file based on its creation instead?  Definiteily can see that as being heavy on processing and this is definitely an improvement either way as we didn't need to check on so many CI's either way.

dbwiddis:
> Just for clarity on this new feature of license plugin, its different than the original check of inception date on project in that the starting date will vary per file based on its creation instead? 

Correct, it navigates the git history (using `jgit`'s tree traversal) and uses the date of the first commit for the initial copyright date. 

There's also an option to use the name/email of the first committer for the copyright, but I kind of like hand-waving it and saying "look at the commit history, everyone's commits are their own work".

hazendaz:
@dbwiddis Thanks and agreed.  I remove author's from files as I see them.  It further causes us humans to blame the name instead of the code ;)

dbwiddis:
@hazendaz see #927 ;)

Also:
> Unknown lifecycle phase ".skip=true".   

Seems I need quotes. Should have tested the maven command....

hazendaz:
ah...yah 'for the windows bit' it needs the quotes.  Always a fun gotcha!

PR #1877, opened 2022-01-29
Replace tidy plugin with sortpom

dbwiddis:


hazendaz:
LIttle noisy the first time and I suspect you probably feel a few of these were odd (at least I did with the <tag> one).  I've known about this plugin for a few years but it used to run in a completely custom way it made up rather than using the approved override for maven which they did switch too (it always could do that but wasn't 100%).  Where I have used it, I had to do as you did and customize a few things to make it more appealing in L&F.  But overall its well maintained compared to mojohaus one and grew on me over time :)  I have a use case where I automated its usage in a PR fashion for projects not being well formatted and opted for that over the other as it was better at fixing bad pom layouts.

PR #1879, opened 2022-01-29
[pom] Update bind-api to jakarta 2.3.3

hazendaz:


hazendaz:
forgot to apply dependency name change to properties, squashed and repushed.

PR #1882, opened 2022-01-30
[DRAFT] Re-tooling the profiling

hazendaz:
Default activation in same pom can be with the activeByDefault or specific jdk in usage but not both.  To better setup usage here to split jdk 8 vs 11, this has been retooled.  To understand what is activated, on current master run 'mvn help:active-profiles', then try the very first commit to see that the 'default' profile is no longer activated.  This is normal maven behaviour.

- Commit 1 makes animal sniffer only run when activated by jdk 8 usage.  See 1.
- Commit 2 adds maven compiler release properties picking up off of maven.compiler target options for jdk 11+.  Additionally, modern '8' is used to indicate java 8 instead of '1.8'.
- Commit 3 is required to occur in removing default profile that won't run now but begs a larger discussion.  See 2.

1) With jdk 11+ as we are now with build, it is entirely unnecessary to even build jdk 8 but rather rely on perfect cross compilation.  We can drop animal sniffer entirely this way by simply dropping this profile and then removing any jdk 8 build usage unless there is some specific reason not to.

2) Since modular build currently is using java11 profile, the default profile is not run.  That means demo / dist are not produced during modular build.  I don't currently see why that case is necessary.  This however means that in order to actually allow that still, jdk 11+ build cannot build the modules either.  This clearly becomes an issue with how maven handles modules in that it won't reliably allow splitting them on profiles.  Further, this is bad practice to have profiles build other modules for the exact reason that some releases will just skip delivery.

The last point is why this is 'draft' for now.  I feel after a few hours getting here, its safe to pass this over to @dbwiddis to think about overall.  I think there is most likely something in modular path causing some issue and we need to further address that before this can happen so all modules are built at all times as maven wants.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,10 +82,10 @@
     <properties>
         <copyright>2010 - 2022</copyright>
         <maven.build.timestamp.format>yyyy-MM-dd HH:mm:ss</maven.build.timestamp.format>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
-        <maven.compiler.testSource>1.8</maven.compiler.testSource>
-        <maven.compiler.testTarget>1.8</maven.compiler.testTarget>
+        <maven.compiler.source>8</maven.compiler.source>
+        <maven.compiler.target>8</maven.compiler.target>
+        <maven.compiler.testSource>8</maven.compiler.testSource>
+        <maven.compiler.testTarget>8</maven.compiler.testTarget>
         <maven.min-version>3.3.9</maven.min-version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
@@ -1136,6 +1136,16 @@
                     </plugin>
                 </plugins>
             </build>
+        </profile>
+        <profile>
+            <id>jdk11on</id>
+            <activation>
+                <jdk>[11,)</jdk>
+            </activation>
+            <properties>
+                <maven.compiler.release>${maven.compiler.target}</maven.compiler.release>
+                <maven.compiler.testRelease>${maven.compiler.testTarget}</maven.compiler.testRelease>
+            </properties>
         </profile>
     </profiles>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -44,7 +44,7 @@
 
     <modules>
         <module>oshi-core</module>
-        <!-- other modules added by default profile -->
+        <!-- other modules added by jdk8 profile -->
     </modules>
 
     <scm>
@@ -854,17 +854,6 @@
     </reporting>
 
     <profiles>
-        <!-- Default module configuration for java8 build -->
-        <profile>
-            <id>default</id>
-            <activation>
-                <activeByDefault>true</activeByDefault>
-            </activation>
-            <modules>
-                <module>oshi-demo</module>
-                <module>oshi-dist</module>
-            </modules>
-        </profile>
         <!-- Custom configuration for Java11 build -->
         <profile>
             <id>java11</id>
@@ -1094,6 +1083,10 @@
             <activation>
                 <jdk>1.8</jdk>
             </activation>
+            <modules>
+                <module>oshi-demo</module>
+                <module>oshi-dist</module>
+            </modules>
             <properties>
                 <!-- Animal Sniffer for jdk 8 -->
                 <animal-sniffer-maven-plugin.version>1.20</animal-sniffer-maven-plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -90,8 +90,6 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <signature.artifact>java18</signature.artifact>
-        <signature.version>1.0</signature.version>
         <!-- Base directory, overwrite in submodules -->
         <main.basedir>${project.basedir}</main.basedir>
         <!-- Dependency versions -->
@@ -150,7 +148,6 @@
         <!-- Misc. -->
         <sonar-maven-plugin.version>3.9.1.2184</sonar-maven-plugin.version>
         <wagon-maven-plugin.version>2.0.2</wagon-maven-plugin.version>
-        <animal-sniffer-maven-plugin.version>1.20</animal-sniffer-maven-plugin.version>
         <dependency-check-maven.version>6.5.3</dependency-check-maven.version>
         <puppycrawl.checkstyle.version>9.2.1</puppycrawl.checkstyle.version>
         <m2e.lifecycle-mapping.version>1.0.0</m2e.lifecycle-mapping.version>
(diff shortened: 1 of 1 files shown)


hazendaz:
note I have commit ready to push on top here that will just build all modules all the time as well as moves java11 down under jdk9on for clarity.

dbwiddis:
> I feel after a few hours getting here, its safe to pass this over to @dbwiddis

I spent a few hours probably treading this same ground before, and ending up where I did with #1864. 

> Default activation in same pom can be with the activeByDefault or specific jdk in usage but not both. 

This was the main blocker to what I wanted to do, e.g., a jdk8 activation for animal sniffer.

> * Commit 1 makes animal sniffer only run when activated by jdk 8 usage.  See 1.

Which disables the active by default.  This fine for this narrow use case, but I personally have JDK17 and want to run most testing/releasing on JDK 8.  Detecting my  Java version produces the wrong result here.

> * Commit 2 adds maven compiler release properties picking up off of maven.compiler target options for jdk 11+.  Additionally, modern '8' is used to indicate java 8 instead of '1.8'.

This is cosmetic but fine.

> * Commit 3 is required to occur in removing default profile that won't run now but begs a larger discussion.  See 2.
> 
> 1. With jdk 11+ as we are now with build, it is entirely unnecessary to even build jdk 8 but rather rely on perfect cross compilation.  We can drop animal sniffer entirely this way by simply dropping this profile and then removing any jdk 8 build usage unless there is some specific reason not to.

I'm not sure I follow this.   I'm committed to supporting jdk8 through its EOL.  I've read enough about incompatibilities of modular builds and MRJARs to stick to a vanilla release through May 2026.  As a library it's better to support the lowest common denominator.  (Heck, only a few months ago I released a JDK6 version of OSHI 3.x.). Unless and until we need features of a higher version, I see no strong reason to raise the minimum level of support.

> 2. Since modular build currently is using java11 profile, the default profile is not run.  That means demo / dist are not produced during modular build.  

This is intentional.  The primary release/artifact/usage is the jdk8 build.  The modular java11 build is provided as a convenience for those who want a module descriptor to use with jlink, etc.  Most users, even those on jdk11 or jdk17 can just use the regular version with its automatic module name.

Honestly, we probably don't even need to release the demo artifact, and perhaps we shouldn't.  And all that the dist does is provide zip and tar.gz versions for anyone who doesn't use the Central Repository and wants to download and extract.  I expect the number of people using it is near zero.

> I don't currently see why that case is necessary.  This however means that in order to actually allow that still, jdk 11+ build cannot build the modules either.  

It doesn't need to.  The JDK11/JPMS version is intended to be stand-alone.  Really, I'd have liked it as another module alongside oshi-core, but I couldn't figure out an easy way to share the same source tree (yes, I tried dynamic links).  There are probably maven-ish ways to copy directory trees over that could work in theory.  The current setup seems to work well enough, though.

> This clearly becomes an issue with how maven handles modules in that it won't reliably allow splitting them on profiles.  Further, this is bad practice to have profiles build other modules for the exact reason that some releases will just skip delivery.

Which is why we might only use modules to do specific things not associated with the regular release.

> I think there is most likely something in modular path causing some issue and we need to further address that before this can happen so all modules are built at all times as maven wants.

I'd love if the jdk11 build happened as an extra module, which didn't break our tests on jdk 8.  I'm sure that's possible with some maven plugins, I just haven't spent the time to hunt it down because the current setup works.

I think using JDK version as an activation for profiles is the wrong approach, though.  I think the pom should be optimized for a jdk8 build, and have a few bells and whistles to enable an on-demand-with-slightly-different-settings jdk11 release.

pom.xml
```diff
@@ -82,16 +82,14 @@
...
         <copyright>2010 - 2022</copyright>
         <maven.build.timestamp.format>yyyy-MM-dd HH:mm:ss</maven.build.timestamp.format>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
-        <maven.compiler.testSource>1.8</maven.compiler.testSource>
-        <maven.compiler.testTarget>1.8</maven.compiler.testTarget>
+        <maven.compiler.source>8</maven.compiler.source>
+        <maven.compiler.target>8</maven.compiler.target>
+        <maven.compiler.testSource>8</maven.compiler.testSource>
+        <maven.compiler.testTarget>8</maven.compiler.testTarget>
```
dbwiddis, line 88: These are fine.

pom.xml
```diff
@@ -82,16 +82,14 @@
...
         <maven.min-version>3.3.9</maven.min-version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <signature.artifact>java18</signature.artifact>
-        <signature.version>1.0</signature.version>
```
dbwiddis, line 94: What is the benefit of moving properties to a profile? They are harmless in the main POM and easier to maintain if they are all together in one spot. We have the JNA and SLF4J versions in the parent pom too, even though they are only used by child modules.

pom.xml
```diff
@@ -1132,5 +1078,67 @@
...
+            <properties>
+                <!-- Animal Sniffer for jdk 8 -->
+                <animal-sniffer-maven-plugin.version>1.20</animal-sniffer-maven-plugin.version>
+
+                <!-- Signature Settings -->
+                <signature.artifact>java18</signature.artifact>
+                <signature.version>1.0</signature.version>
+            </properties>
+            <build>
+                <pluginManagement>
```
dbwiddis, line 1099: what is the need or benefit for pluginManagement in a profile?  If we're changing the profile to "opt in" instead of "opt out" why can't it just be all in one section?

pom.xml
```diff
@@ -1132,5 +1078,67 @@
...
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+        <profile>
+            <id>jdk11on</id>
+            <activation>
+                <jdk>[11,)</jdk>
```
dbwiddis, line 1136: I don't like this at al.

I want a compiler target of 8 when I build on my local machine which runs JDK17, unless I tell it otherwise.

pom.xml
dbwiddis: Also, will dependabot maintain versions in profiles? I don’t think so… by removing the main execution from the default POM it hides these.  Not that they will ever change, of course.  They may as well be hard coded without properties.

hazendaz:
Thanks for looking this completely over.  I believe with this info it's possibly to adjust and get all you mentioned here and possible integrate jdk 11 modular part in without mvn11.  I'll look more at this tonight.

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>

dbwiddis:
> Thanks for looking this completely over. I believe with this info it's possibly to adjust and get all you mentioned here and possible integrate jdk 11 modular part in without mvn11. I'll look more at this tonight.

My thoughts:
 - get rid of java11 profile
 - make a new module 
   - with a pom overriding the appropriate properties and using the other dependencies
   - with the two module-info.[java|test] files in src/main/java and src/test/java
   - an empty oshi directory that's .gitignore'd
 - as part of the build, recursively copy from oshi-core/src/[main|test]/java/oshi to oshi-core-java11 with appropriate cleanup either after, or before copying so it always gets the latest copy
   - this is the only bit I couldn't find. I did play with [`copy-rename-maven-plugin`](http://coderplus.github.io/copy-rename-maven-plugin/) but it only copies one file at a time.
   - I think it can be done with the [`antrun`](https://stackoverflow.com/questions/586202/best-practices-for-copying-files-with-maven) plugin as described [here](https://stackoverflow.com/a/694175/1161484)

pom.xml
dbwiddis: OK, disregard this complaint. I see it's necessary. :)

dbwiddis:
> I feel after a few hours getting here, its safe to pass this over to @dbwiddis

Since you passed it over to me, I feel safe incorporating the parts that are needed and that I like.  I've committed a subset of these changes to PR #1887 in https://github.com/oshi/oshi/pull/1887/commits/f4e25f968e08900606bba0639c1f63902177d373

I'll go ahead and close this PR as I fully understand all the proposed changes.

Suggested changes I implemented:
1. Changed the maven compiler source to 8 as proposed
2. Moved animal sniffer inside a jdk8-activated profile along with the unchanging signature properties
3. Created a jdk9+-activated profile for maven.compiler.(test)release property now that I grok its purpose

Suggested changes I didn't implement (yet) and what I'm doing.
1. Keeping oshi-demo and oshi-dist modules in all (jdk8) builds.  This probably doesn't matter, as the intention is to always run builds with JDK11+, with the exception of the CI testing on JDK8, and neither of these modules has tests.  But since they aren't strictly limited by JDK I decided to leave them alone.
2. Keeping all properties together at the top of the POM.  This probably doesn't matter but it's preference.
3. Did not use plugin management.

PR #1884, opened 2022-01-30
[skip-ci] Fix trailing spaces and make sure files end in proper end of line marker

hazendaz:


hazendaz:
cool skip-ci skips most everything.

PR #1887, opened 2022-01-31
Configure java11 build in a new module

dbwiddis:
This could be a much simpler approach to a modular build.   

It works on my local machine, copying the sources for mvn test and running modular tests.

I'm sure I'm missing something.   For example it didn't complain about animal-sniffer even before I tried to suppress it... and suppressing it doesn't seem to be working either (perhaps deleting the id did that).  Maybe I can do a build target for 11 but keep the "signature" that's being sniffed for java 8, without any problems...

~This leaves all the files in the directories, but I think I could use `maven-clean-plugin` to go through at the end of the build and delete them.~ The clean plugin now cleans up the directories too.

Only thing left is that this fails on Java 8 because of the target, so needs some sort of jdk activation just on that final module.  I'm sure @hazendaz can tell me how to fix that. :)


dbwiddis:
OK, I think i have taken a crash course in Maven tonight and learned a whole lot!

This is almost there!
 - Added a jdk8 profile to suppress essentially the entire lifecycle for the java11 module when jdk is 1.8
 - Fine tuned the cleanup of the temporary build files
 - Suppressed animal-sniffer, even though it's ok to run it, just keep the old signature.
 - Suppressed license-header since it would rewrite every single copied file...

Still to do:
 - Scrutinizer is complaining about lack of manifest/empty jar so more lifecycle suppression needed
 - Figured out a big flaw in the mvnw11 script.  It passes tests even when they fail :)
   - Need to just delete that and fix all the workflows I altered to use it

hazendaz:
@dbwiddis We probably need to figure out a zoom call :)  Keep this up along with other.  I like the draft ideas but think there is a fundamenal issue here we are all trying to learn.  I took a look at jackson and junit as I knew they were both doing something with module path and I think based on that both jna and this project has it wrong.

dbwiddis:
OK, I've removed "draft" and marked this ready for review.  Here's why I'd like to go ahead and merge this:

1. It accomplishes the same thing I've been doing manually for months with the `java11` branch and what was intended with the (broken) `mvnw11` hack I did earlier, but better.  
2. It is, at a minimum, better than the current master branch, even if it's not the final solution.

I'm sure it can be improved on, but I think it's a better starting point for future discussions of other options being discussed in the other two open (at the time of writing this) PRs.

I'll briefly summarize my thoughts on the other options here; not as a final decision, but why I'm leaning against them:
 - #1882 is very reliant on the JDK being used to build; given that I do most of the releasing, and I run JDK17, but want to release JDK8 artifacts, I don't want the build to compile to a higher target based on my JDK
   - I understand if I'm using a lower JDK to build I can't do stuff the higher JDK can. This is intended "failure"
   - But I should always be able to "dumb down" the build without having to install a lower JDK and redefine `JAVA_HOME`.
 - #1888 is the MRJAR approach.  
   - I strongly considered this early on because it's relatively easy to implement (as the PR shows) and I thought it could also bring us the benefit of implementing better/more efficient internal implementations.  
   - Unfortunately there are only a few improvements possible 
     - the process listing stuff doesn't get all the info OSHI does so still requires more effort and may end up less efficient
     - there are improvements for file reading "stream" APIs but they are largely programmer convenience and I'm not sure actually improve performance
   - The MRJAR is reliant on the user's JDK for determining the API behavior.  This is fine when you're doing more optimal code, but in the use case of "adding a module descriptor to enforce API privacy" it imposes restrictions on the API simply based on the user's JDK, and I really don't like that
 - I've looked into Moditect (mentioned in #1888) which is basically a "throw a modular descriptor into an existing JAR".  While in theory this could work:
   - We would still have to change the tooling to release two artifacts, one with a modular descriptor (that strongly enforces API visibility limits if the user has JDK 9+) and one without (the status quo that doesn't care about the user's JDK)
   - I haven't yet figured out how to adequately test a modular build produced with moditect.  And yes, I've found errors in our modular setup by testing it.
   - I think the best use case for moditect is in adding modular descriptors to non-modular dependencies, but we don't have those to deal with.

In sum... this PR permits a single release process, lets me create a jdk8 build using jdk17 that a user using jdk17 won't be forced into modules using, does not require special maintenance and cherry picking of a separate branch, continues publishing the set of artifacts we've been advertising, and works with our existing test setup (by basically skipping everything and not failing if on jdk8).  The only real "restriction" is that if we want to publish stuff to Central (snapshots or releases) we have to use JDK11+.  Not a concern.

Again, some points:
 - I know it's not perfect but it's better than what we've had
 - I really appreciate the work in #1882 because it got me moving in this direction and I referred to it frequently

dbwiddis:
OK, @hazendaz, I think I've tweaked this all I'm going to and would appreciate your review.  

I went the long way around to eventually get back to something close to what you had in #1882 and probably could have built on that one instead of starting fresh, but what's done is done.  And I really learned a metric ton about Maven along the way so it was a good exercise.

What this PR does:
1. Removes the `mvnw11` script and reverts the workflows using it back to their original state
  a. Added JDK11 to the PR workflow, though, so it does the modular tests 
2. Updates the RELEASING doc specifying minimum java version
3. Updates the parent pom:
  a. Moves animal-sniffer execution into a jdk8-activated profile as per #1882 commit 1
    - This is probably unneeded to restrict, it's ok to run it on higher JDKs, keeping the same target
    - but no need to execute if we don't have to with cross-compilation.  
    - I left the properties in the main list but moved the signature configuration into hard coded attributes
  b. Adds compiler release properties in a jdk11-activated profile and changes compiler source/target/etc. from 1.8 to 8 as per #1882 commit 2 
    - I didn't initially understand what it was doing!
  c. Created new (Maven) module for java11 release and moved that module and the dist module into the jdk11-activated profile as per the direction you were going with #1882 commit 3, but the way I think it should be:
    - We only need oshi-dist with the deploy goal which we can/should require jdk11 for
    - I could have moved oshi-demo there too, but decided to leave it as part of the default build.  Really at this stage we don't need it, but it needs to be a pom module for IDEs to pick it up easily.  I decided if we ever do put something releasable in that module it would belong alongside oshi-core with jdk-8 support.
  d. Added new oshi-core-java11 module
    - Moved the module-info files into their proper places
    - The java11 pom does exactly 3 things (in addition to an updated dependency list and maven compiler target):
      - Uses maven-resource-plugin to copy over source and resource files from oshi-core during the "generate-*" goals.
      - Uses maven-clean-plugin to delete the resources after the test goal when we're done with them
        - NOTE: we could/should move the cleanup up to "process-sources" and "process-test-sources" goals, except at present some of our tests (FileUtil) have hard dependency on the actual source code files so they have to remain through test.  I'll fix that hack (from my early days) with a file write to a temporary location later.
      - Skips the license plugin goal which assumed all the files were brand new with no git history.
  e. Added the java11 module as a dependency to oshi-dist.  This update was simple if we skip oshi-dist on jdk8.

hazendaz:
A journey indeed.  Lot here.  I'm logging off now but will take a look tomorrow.  Did ready through all you wrote.  All sounds good!

dbwiddis:
I'm glad you went to bed... I knew I was abusing the "resource" plugin to copy "source" files and kept digging.  Turns out build-helper-maven-plugin has an `add-source` (and `add-test-source`) goal perfectly suited to this task.   I could also have added the resources with build-helper, but there's a much more concise way of doing it with resource plugin.

So no copying, no need to clean up the copies.  I still had to tweak the FileUtilTest until it gets fixed properly, but it's all working now, and I think I'm doing it "the maven way" now.  Or at least a lot less hacky.

I actually think I'm really done messing with this now.

Issue #1238, opened 2020-05-22
OS processes Shmem RedHat (RHEL 7.6)


battousai-dev:
hi, working with qualification of RHEL platform.
How do you get the ShMem from oshi for redhat? I'm getting values of Shmem from sigar. 
Here's the sample data from sigar:

[...]

Issue #535, opened 2018-06-25
Add child thread cpu usage values to OSProcess


mar1ged:
I copied this from the Sysinfo example:

`100d * (p.getKernelTime() + p.getUserTime()) / p.getUpTime()`

I created a process that keeps several cores busy, htop displays this correctly as `390%` (please see the top of the screenshot).
But it seems like I only get the time values of a single thread (bottom of the screenshot), so the accumulated percentage is only `5.4%`.

![image](https://user-images.githubusercontent.com/16045549/41834287-fd07bc6a-7853-11e8-819f-f4c6a7bc5a21.png)

dbwiddis:
The values returned by OSHI are cumulative for the life of the process.  If I recall correctly, htop gives you a snapshot.   Try calculating the kernel+user time differences over a short time period and divide by the elapsed time.

OSHI gets its values from `/proc/<pid>/stat` so if there's an error in parsing, let me know...

dbwiddis:
Continuing from the above, OSHI's Kernel and user time values come from fields 14 and 15.

>               (14) utime  %lu
>                         Amount of time that this process has been scheduled
>                         in user mode, measured in clock ticks (divide by
>                         sysconf(_SC_CLK_TCK)).  This includes guest time,
>                         guest_time (time spent running a virtual CPU, see
>                         below), so that applications that are not aware of
>                         the guest time field do not lose that time from
>                         their calculations.
> 
>               (15) stime  %lu
>                         Amount of time that this process has been scheduled
>                         in kernel mode, measured in clock ticks (divide by
>                         sysconf(_SC_CLK_TCK)).

It appears you want the rollup for all child processes which would be the next two fields:

>               (16) cutime  %ld
>                         Amount of time that this process's waited-for chil‐
>                         dren have been scheduled in user mode, measured in
>                         clock ticks (divide by sysconf(_SC_CLK_TCK)).  (See
>                         also times(2).)  This includes guest time,
>                         cguest_time (time spent running a virtual CPU, see
>                         below).
> 
>               (17) cstime  %ld
>                         Amount of time that this process's waited-for chil‐
>                         dren have been scheduled in kernel mode, measured in
>                         clock ticks (divide by sysconf(_SC_CLK_TCK)).

OSHI does provide the parent PID, so it would be an easy exercise to just roll up the sum of user/system time for the children.

dbwiddis:
[...]

PR #1888, opened 2022-01-31
Move module-info directly into the build

hazendaz:
This is part of the puzzle.  This puts modular-info.java / modular-info.test directly into the build.  If a runtime using it is java 8, it doesn't get used.  If the runtime is java 11+, it does.


hazendaz:
@dbwiddis This in combination with 'https://github.com/moditect/moditect' is likely what is needed to pull this off without having to create extra modules and to completely do the entire job.  I think the upstream on JNA probably needs the same.  This change by itself would not matter what java version is used to release.  It will only pick up the modular information in the build if java 11 or higher.

The only thing I see with the mvn11 (java11 profile) is trying to set binary at java 11.  But in reality there are so few libraries that have forced that and it doesn't really matter.  If you want binary at java 11, sure fine, but what about 100s of other libraries in a potential project that don't.  Its never mattered.  I wouldn't doubt some are java 4 compiled still and all still ok.

What I do know is the IDEs such as Eclipse that I use daily already throws everything in modular path as it is, regardless if the jar was or not.  It likely does this by adding automatic module naming on the fly when using higher jdks.

hazendaz:
Signing off for night, but think this along with dropping java11 profile entirely probably would in fact give you all you were trying to do.  It would not pick up 'jpms' modules from jna but I feel like that ist just another case of compiled with java 11 targetting jdk 11 binary that doesn't quite mean much unless physically recoded to new java constructs or has something that is otherwise missing.  I could be 100% off base here but have also never really seen any direct modular usage in my day job.  No one seems to know how to do it nor seems to care and most still prefer java 8.

dbwiddis:
This is the MRJAR approach. I considered this very early on -- it was really the simplest approach -- but after reading about some of the drawbacks, decided against it.  

The biggest concern for me is similar to our discussion on the other issue -- I don't think the JDK should automatically enforce new behavior.  Just because I'm running on JDK17 doesn't mean OSHI should force me to use modules.  Unfortunately I think with the approach in this PR, it will force me to use modules even if I don't want to, just because I've got a newer JDK.  So if I'm directly using one of OSHI's driver classes to access information not exposed in the official API, I'll get different behavior.

Yeah, I already hear the objection about implementation details/API/etc. that modules are supposed to enforce.  I'm reluctant to "enforce" that because I really don't see a good reason to do so.

I primarily want OSHI to remain primarily JDK8 non-module compatible.  I still don't think there's much of a use case for modules and am primarily providing the added artifact solely to support those who want to use those features, jlink, etc.

Here's an article from the Gradle team that helped sway me against MRJARs early on.  Interested in your thoughts on it: https://blog.gradle.org/mrjars

I think eventually ... another year or two... this may be the right approach, but I honestly still think Java hasn't figured all this stuff out. I've seen MRJAR issues as recently as JDK16.

mvnw11
```diff
@@ -32,8 +28,6 @@ echo "> ./mvnw \"$@\" -P java11"
 ./mvnw "$@" -P java11
```
dbwiddis, line 28: For the sake of review, I'll pick on something you didn't even change.  Test failures in this call aren't propagated to the script result, so the `mvn test` can fail gloriously but this script will happily exit with no problems.

oshi-core/src/test/resources/META-INF/versions/11/module-info.test
```diff
@@ -22,4 +22,4 @@
 --add-reads
   com.github.oshi=org.hamcrest
 --add-reads
-  com.github.oshi=org.junit.jupiter.api
\ No newline at end of file
+  com.github.oshi=org.junit.jupiter.api
```
dbwiddis, line 25: I'm curious why the newline at end of file is a requirement.

dbwiddis (commented):
You asked for a review so I'm giving a review.   `mvnw11` is broken and should be totally eliminated.   MRJAR has some use cases but I don't think "adding a module descriptor" is one that we should leverage.

dbwiddis:
With a full night's sleep and full dose of caffeine, I'll try to rewrite what I said above but more briefly:
1. A MRJAR is fine for users with java 8 only, or users with java 9+ who are only using API functions.
2. A MRJAR as proposed here would force any user with JDK9+ into having oshi on the module path along with the API restrictions that come with modules
3. There may be some users who use util classes/driver functions.  See #535 and #1238.  While those are non-API they are useful. 
4. I'm fine with the idea of an MRJAR for all variants of an already fully modular setup so newer versions can take advantage of newer features under the hood based on the user's runtime.   But during this transition period while supporting Java 8 I'm a pretty hard no on MRJAR.

PR #1889, opened 2022-01-31
Improved calculation of AIX disk reads and writes

### LINKED ISSUE #1876: AIX Disk reads is not monotonic
AIX only provides total transfers for disk IO.  We *do* have data on reads/writes in blocks, so OSHI currently uses this to estimate the reads/writes.  From `AIXHWDiskStore`:
```java
// we only have total transfers so estimate read/write ratio from blocks
long blks = stat.rblks + stat.wblks;
this.reads = stat.xfers;
if (blks > 0L) {
    this.writes = stat.xfers * stat.wblks / blks;
    this.reads -= this.writes;
}
this.readBytes = stat.rblks * stat.bsize;
this.writeBytes = stat.wblks * stat.bsize;
```
The problem with this approach is that the integer division for `this.writes` biases writes slightly lower and reads slightly higher, and it is possible for `this.reads` to decrement by 1:
```
Failures (1):
  JUnit Jupiter:DisksTest:testDisks()
    MethodSource [className = 'oshi.hardware.DisksTest', methodName = 'testDisks', methodParameterTypes = '']
    => java.lang.AssertionError: Number of reads from the disk has not been updated
Expected: is a value equal to or greater than <130684L>
     but: <130683L> was less than <130684L>
       org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
       oshi.hardware.DisksTest.testDisks(DisksTest.java:87)
```
The estimation should be updated to cast `blks` to `double` and use `Math.round()` to eliminate the bias.

siddhantdixit:
Fixes #1876

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ##### Bug fixes / Improvements
 * [#1869](https://github.com/oshi/oshi/pull/1869): Ignore mount headers on AIX filesystem - [@dbwiddis](https://github.com/dbwiddis).
+* [#1889](https://github.com/oshi/oshi/pull/1889): Improved calculation of AIX disk reads and writes - [@siddhantdixit](https://github.com/siddhantdixit).
 
 ##### Artifact Removal
 * [#1868](https://github.com/oshi/oshi/pull/1868): Remove shaded jar artifact - [@dbwiddis](https://github.com/dbwiddis).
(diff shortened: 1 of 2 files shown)


dbwiddis (changes requested):
Looks good.  Two tiny tweaks please:
 - Checkstyle is complaining: 'typecast' is not followed by whitespace. [WhitespaceAfter]
 - Please add a change log entry (and change the issue title to match) something along the lines of "Improved calculation of AIX disk reads and writes."

dbwiddis (approved):

PR #1902, opened 2022-02-05
Add spotless-maven-plugin

dbwiddis:
Easy way to get users to apply correct formatting regardless of how they edit files.


hazendaz:
@dbwiddis Nice!  Had not heard of this one, looks like it covers a lot of ground that was either WIP or a struggle with formatter-maven-plugin.  I feel like I need to add this one to my main processing and see how it goes to cover other file types.  Really interested in fact it handles kotlin and groovy.

dbwiddis:
Figured you might like it! I encountered this submitting a PR to junit.  I like how (similar to sortpom and the license plugin) you can just do a single maven command line to fix all the formatting stuff I've normally had to do in Eclipse.

PR #1903, opened 2022-02-07
Reorganize README

dbwiddis:


dbwiddis:
@adrian-kong This is completely independent of what you did for #1832 but those changes were mostly just for the site and I'm trying to update the page I think more people land on.

Very interested in whether you think this reorg could replace what you did, or if there are things you did in #1832 that are a better option than my choices.   I think we both agree that "output" needs to be moved.  I wanted to keep usage on the main page but higher up and all in one block.  I think the site.xml needs to be updated for this new arrangement, though.

PR #1918, opened 2022-02-13
[actions] Ignore jakarta xml bind 3.0.0+

hazendaz:


dbwiddis:
I've already told dependabot to ignore this major version so 3.x is not a problem for us, but this will save us from 4.x.  

Although it's a dependency on the coveralls plugin which the (idle) owner finally gave permissions to someone else to release, which will include the fix transitively and we can remove it then.

Also you can delete the trajano wagon git as it's now been removed from our pom.

hazendaz:
@dbwiddis Yup on trajano, saw that after I pushed, I have that in a separate mixed commit I'll extract and remove.  On dependabot, yes it remembers you closed and it ignored until new version is rasied but on forks (well old forks), dependabot will continue to open unless the same is done.  This is the suggested way.  I'll bring in the other removal shortly.  

On site, I think it is safe to bring in what I have since you have deploy skipped now.  Or at least most of what I have.  I had settings setup slightly differently and configured to use git under scm per maven instructions.  It was just super slow given how they checkout one file after another which doesn't quite make sense.  Based on information posted around, I think they are trying to get scm plubish to work but it is unclear how that works in with release plugin.  Probably a maven 4 thing.  Most of the rest I think you picked up what I had already so its feasible to run locally, just takes forever :(  The gh action for that works different so it feels best to keep that one in rotation.

dbwiddis:
OK, for site, I will assign you #1922 and you can make it "work" even if not used.

PR #1924, opened 2022-02-14
[actions] Skip license plugin on site generation as unnecessary

hazendaz:


hazendaz:
@dbwiddis I believe this is what is needed to address license issue here.

dbwiddis:
This is only for the site; the issue I added is for `mvn release:perform`.  I could add the skip there, but it seems there's a better programmatic approach.

hazendaz:
@dbwiddis ah ok, so no issue with site then, I'll close this.  I know what is needed on release.  Its due to shallow clones in recent release changes.

hazendaz:
actually this wouldn't hurt, might skip a cycle, need to confirm looking at the output.  Still no reason to acutally run it if it does here.  Checking now.

hazendaz:
I do believe we want to skip here after looking.  It isn't a huge hit or anything but better not to waste time.

dbwiddis (approved):
Yeah, it doesn't hurt anything here. Should also add it to our snapshot deployment workflow if it isn't there already.

hazendaz:
Its on snapshot to sonatype already :)  It is in a number of places but probably worth checking them all to be sure given its probably the slowest plugin in use.

PR #1925, opened 2022-02-14
[site] Rework site so it works from local

hazendaz:
Note: Added maven scm publish plugin as it appears maven is going towards that being way to publish.  It does not help the slow checkout issue but deploy is skipped right now outside 'site' branch so this does no harm either.  Will come back to this as they refine process at a future date.

I am not sure if this config will affect the site.yaml file.  We would need to confirm that.  It also is forcing some of underlying items on site plugin so scm used is consistent.  

wagon-ssh is deprecated and removed in version 2 as I have here for scm.  This works through unwrapping (ie scm contains git which contains ssh).  Ultimately this makes it match the other distribution setup.  The site will work this way on release if wanted but long time to process.  Using scm plublish is separate if you were running site manually and then using that to deploy it.  To keep this separated from the current settings.xml, new entry to align to it.

If nothing more than awareness, this is where I had gotten which I think is better but this would have no impact now that it is now used.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -234,8 +234,8 @@
                     <version>${maven-scm-publish-plugin.version}</version>
                     <configuration>
                         <checkoutDirectory>${user.home}/maven-sites/${project.name}</checkoutDirectory>
-                        <!-- no need for site:stage, use target/site -->
-                        <content>${project.reporting.outputDirectory}</content>
+                        <!-- need site:stage to collect multimodule project -->
+                        <content>${project.output.directory}/staging/site</content>
                         <scmBranch>gh-pages</scmBranch>
                         <tryUpdate>true</tryUpdate>
                     </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -235,6 +235,7 @@
                     <configuration>
                         <checkoutDirectory>${user.home}/maven-sites/oshi</checkoutDirectory>
                         <scmBranch>gh-pages</scmBranch>
+                        <skipDeletedFiles>true</skipDeletedFiles>
                         <tryUpdate>true</tryUpdate>
                     </configuration>
                 </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -145,7 +145,6 @@
         <versions-maven-plugin.version>2.10.0</versions-maven-plugin.version>
         <!-- Misc. -->
         <sonar-maven-plugin.version>3.9.1.2184</sonar-maven-plugin.version>
-        <wagon-maven-plugin.version>2.0.2</wagon-maven-plugin.version>
         <animal-sniffer-maven-plugin.version>1.21</animal-sniffer-maven-plugin.version>
         <dependency-check-maven.version>7.0.4</dependency-check-maven.version>
         <puppycrawl.checkstyle.version>10.1</puppycrawl.checkstyle.version>
@@ -616,12 +615,6 @@
                     <groupId>org.sonarsource.scanner.maven</groupId>
                     <artifactId>sonar-maven-plugin</artifactId>
                     <version>${sonar-maven-plugin.version}</version>
-                </plugin>
-                <!-- Wagon -->
-                <plugin>
-                    <groupId>org.codehaus.mojo</groupId>
-                    <artifactId>wagon-maven-plugin</artifactId>
-                    <version>${wagon-maven-plugin.version}</version>
                 </plugin>
                 <plugin>
                     <groupId>org.owasp</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -235,7 +235,7 @@
                     <configuration>
                         <checkoutDirectory>${user.home}/maven-sites/${project.name}</checkoutDirectory>
                         <!-- need site:stage to collect multimodule project -->
-                        <content>${project.output.directory}/staging/site</content>
+                        <content>${project.build.directory}/staging/site</content>
                         <scmBranch>gh-pages</scmBranch>
                         <tryUpdate>true</tryUpdate>
                     </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -234,8 +234,6 @@
                     <version>${maven-scm-publish-plugin.version}</version>
                     <configuration>
                         <checkoutDirectory>${user.home}/maven-sites/oshi</checkoutDirectory>
-                        <!-- need site:stage to collect multimodule project -->
-                        <content>${project.build.directory}/staging</content>
                         <scmBranch>gh-pages</scmBranch>
                         <tryUpdate>true</tryUpdate>
                     </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -233,9 +233,9 @@
                     <artifactId>maven-scm-publish-plugin</artifactId>
                     <version>${maven-scm-publish-plugin.version}</version>
                     <configuration>
-                        <checkoutDirectory>${user.home}/maven-sites/${project.name}</checkoutDirectory>
+                        <checkoutDirectory>${user.home}/maven-sites/oshi</checkoutDirectory>
                         <!-- need site:stage to collect multimodule project -->
-                        <content>${project.build.directory}/staging/site</content>
+                        <content>${project.build.directory}/staging</content>
                         <scmBranch>gh-pages</scmBranch>
                         <tryUpdate>true</tryUpdate>
                     </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -47,8 +47,8 @@
     </modules>
 
     <scm>
-        <connection>scm:git:git@github.com:oshi/oshi.git</connection>
-        <developerConnection>scm:git:git@github.com:oshi/oshi.git</developerConnection>
+        <connection>scm:git:ssh://git@github.com/oshi/oshi.git</connection>
+        <developerConnection>scm:git:ssh://git@github.com:oshi/oshi.git</developerConnection>
         <tag>HEAD</tag>
         <url>https://github.com/oshi/oshi.git</url>
     </scm>
@@ -74,7 +74,7 @@
         <site>
             <id>gh-pages</id>
             <name>OSHI GitHub Pages</name>
-            <url>scp://oshi.github.io/oshi/</url>
+            <url>scm:git:ssh://git@github.com/oshi/oshi.git</url>
         </site>
     </distributionManagement>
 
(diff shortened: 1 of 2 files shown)


.mvn/settings.xml
```diff
@@ -16,6 +16,13 @@
         <server>
             <id>gh-pages</id>
         </server>
+        <server>
+            <id>gh-pages-scm</id>
```
dbwiddis, line 20: Can you please add some notes to the `Releasing` doc about these necessary `settings.xml` entries?

pom.xml
```diff
@@ -231,6 +248,14 @@
                         </dependency>
                     </dependencies>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-scm-publish-plugin</artifactId>
```
dbwiddis, line 253: +1 on this plugin.  I had gotten around to this config at some point in my experimenting and think it's the way to go. The only thing I couldn't figure out was tying it to the lifecycle... but that's OK as I think we want it separate from the lifecycle anyway.

pom.xml
```diff
@@ -221,8 +223,23 @@
                     <dependencies>
                         <dependency>
                             <groupId>org.apache.maven.wagon</groupId>
-                            <artifactId>wagon-ssh</artifactId>
-                            <version>${wagon-ssh.version}</version>
+                            <artifactId>wagon-scm</artifactId>
```
dbwiddis, line 226: Is `wagon-scm` necessary for the scm publish plugin to do its work?  I don't see it documented at all on that plugin site. 

The dependencies below are all dependencies for `wagon-scm` so I am not sure if they are needed here.  They can also be put as extensions instead of dependencies if they are needed.

dbwiddis (approved):
I'm not sure you need all these dependencies, but if you're sure you do, LGTM.

pom.xml
hazendaz: extesion based on maven documentation was for maven 2.  maven 3 is in site directly.  Its very confusing as it seemed to be the other way some time ago.  Wagon-ssh is obsolete (in fact deleted now in recent release).  The other items in site (or should say based off other junk potentially in my main setup) resulted in scm not being same level so that just ensures it isn't messed up on resolution.  Do we need it all?  Possibly not, some of that issue my be unrelated and worth further review / cleanup.

.mvn/settings.xml
hazendaz: sure - logging off to watch rest of the game but will take a further look early in the week.

pom.xml
dbwiddis: Looking at the plugin [pom](https://github.com/apache/maven-scm-publish-plugin/blob/ed970672a2878b6601c5a4771aa60bbf6352c041/pom.xml#L116-L125) it does include at least the api and scm-provider-gitexe as dependencies so those are probably fine to keep here at a higher level.  

I do not see wagon-scm or the plexus dependency so I think they can be deleted.

dbwiddis: Dug around in the deploy plugin. It does not use any of the scm/etc. stuff from POM.  It generates/stages the site and then has its own internal git implementation.   So I think we are safe with none of these dependencies, or explicitly specifying the two which are transitively imported anyway.  I actually think keeping `wagon-scm` might potentially cause problems as a competing implementation.

But I also know you've been testing all of this and trust your final judgment!

.mvn/settings.xml
dbwiddis: > sure - logging off to watch rest of the game but will take a further look early in the week.

I notice you didn't say *which* week. 😁 

Happy to finish this for you if you can just quickly summarize in a reply why the differently-named gh_pages config.   And I'll probably remove the dependencies that I don't think are needed.

hazendaz: sorry saw this earlier today and realized I forgot about it.  this can certainly be cleaned up and I know how now to address scm publish.  In fact, you probably could find that if you look at maven-parent github project.  That was updated to not use the site deploy during release but rather publish and configured to use it directly from the site already built rather than checkout which was causing this to be slow.  I was going to look more at that this coming weekend to see how the scm works (wrote it down so I don't forget) but if you have time, try seeing what they did that might finish this off.  See here https://github.com/apache/maven-parent/blob/master/maven-skins/pom.xml

dbwiddis: I might take a peek but I'm pretty slammed this week/month as well.  If this is still open in April I'll deal with it. :)

dbwiddis: OK, peeked. Config looks pretty simple.  It's a matter of testing and documenting the appropriate mvn commands (apparently no longer staging) to deploy manually.  All of which seems more complicated than just pushing to the site branch upstream.  And I don't know what mvn commands are needed to make that work... what we have now works and I'd be making work for myself breaking it.

So, yeah.  Leaving this for you to work, probably after you've got it working elsewhere.

hazendaz:
@dbwiddis Coming back to this now.  I've got some setup locally that I'm going to test to see how fast it pushes.  I've yet to test the release plugin cycle as that is far more complicated to do but believe what I have locally is all that is needed and potentially can be cleaned up more than what I've done.  I'll push that up here after I've run a test to see how fast it performs.

basically expecting this.

- release plugin at site-deploy phase will skip site deploy which is taking forever.
- release plugin at site-deploy will run scm publish instead using the site it built already at that point.
- user space is used for the scm checkout used copying what maven does as 'maven-sites/${project.name'.  It will attempt updates upon that to speed it up and is setup to use the site from reporting property location.  With release I expect that to be under the checkout when its running.  For testing though, using mvn site, mvn scm-publish:scm-publish.

If my testing shows it runs far faster than it did on site actions, then I will push that up here and we can review again.  I'm doing this off the site branch so the site doesn't change.

.mvn/settings.xml
hazendaz: I think provided what I'm doing works out, the 'gh-pages-scm' will be removed later entirely.  site-deploy will be disabled so that won't be used.  The scm publish intead uses 'gh-pages' directly as a branch from the pom.

pom.xml
hazendaz: wagon-scm and others will likely go away.  Since site won't be used to deploy, those are likey not necessary.  will look at that further after testing to streamline the setup.

dbwiddis:
> I'm doing this off the site branch so the site doesn't change.

Pushing site branch is what triggers the GHA to publish the site...

hazendaz:
@dbwiddis I believe this does it.  What I'm not 100% on is during release I belive it runs site:stage.  This I believe is simply using target/site.  So we may what further adjustment.  It is however significantly faster than the normal cycles up through site deploy given that last one is now replaced.  I guess what would need checked is something known specifically that should change on next release to confirm it does.  As for local release, should just be the normal release now without any skipping.

dbwiddis (approved):
Sounds good. Let me know what I should look for and/or do on the next release.

hazendaz:
Making a few adjustments...
- Dropping wagon maven plugin as I don't think that was ever used
- Adjusted the `<content>` tag to use the staging site as release plugin would create and needed after further review.  Only a flag single pom project would work with site directly.

As I ended up running from this branch after initial test on 'site' branch, the current commit on gh-pages should be the entire site distribution as we would normally expect.  Please spot check and see if that is in fact ok.

How I ran this locally when just pushing site
- mvn site site:stage scm-publish:publish-scm -DskipTests

On the original PR here before I force pushed, I had that gh-pages-scm.  That purpose was that the gh-pages branch was in settings.xml.  Now that publish is used, that is unnecessary so that is now left as gh-pages.  While settings.xml still has that entry for gh-pages, I'm pretty sure its now unused.  The site entry in pom is used by publish plugin.

During a release, I expect the release plugin to run the same site cycle above.  While it will try to run site:deploy which I believe was our issue, that will now be skipped.  So a normal mvn release:clean release:prepare release:perform should do it.

I did have issues otherwise running tests on latest code.  Seems like something fragile in recent master but noted that the site branch run on GHA skipped tests so I did the same.

You may want to test out this branch as noted here and check gh-pages to see that is ok as noted.  As to when to release next, do anytime.  If all goes well it will be uneventual and finally fix this issue.

Site branch is also still viable anytime to push the site for any snapshot.  

This does also still seem to suffer at least from the checkout issues we saw but since scm stores that on disk and tries to simply update later, it is far faster than we have seen.  This does feel somewhat simplified overall.  Hopefully this will serve us well for a while until maven gets this figured out.  Notes on various places indicate some hope that eventually this just works via single commit without running excessive actions.  Sure commit is single, but it does appear to run single actions to each file.

Time it took to run 'mvn site site:stage scm-publish:publish-scm -DskipTests', 15 minutes

note: you may need this 'git config core.longpaths true' or do that as --global if it doesn't matter for all repos.  This assumes windows 260 limitation.


:( publish wants to delete some site files but it fails due to length.  I've not figured out how to get around that.  Deleting gh-pages and starting over would do it but not clear on the issue itself.

dbwiddis:
> As I ended up running from this branch after initial test on 'site' branch, the current commit on gh-pages should be the entire site distribution as we would normally expect. Please spot check and see if that is in fact ok.

The javadocs weren't published as part of the site.  I'm pretty sure they are with the GHA.

> Time it took to run 'mvn site site:stage scm-publish:publish-scm -DskipTests', 15 minutes

Given only ~30 minutes for it to run from the GA, but only about 30 seconds of my time to execute 4 git commands and let some cloud server do the work, I'm content keeping the way I'm doing it.  It's fine to have the config available in the pom, though, although I'll have to refer back to this PR for the commands. 😁

hazendaz:
ok added skipDelete to the mix as that was causing the failure for me to push.  All items are updated now.  To clear anything that should be deleted, that should just be a cycle of delete everything, push, then run this to restore.  So a temporarily cleanup action.  I doubt the GHA push does that bit give what it wanted to delete and technically ran the identical command there other than the push is done via publish.

So release should be used next without anything special, it should release site property.  Overall time is less than 10 minutes on site at least when site was ran this way.

hazendaz:
I'm going to push the same here to 'site' branch which means just equivalent master up to this point to confirm that still works.

hazendaz:
ok 'site' branch is good.  So only thing left now is to merge in and try a release...

dbwiddis:
Ok, looks good.  I want to make one tweak to the index to add a javadoc link to the menu bar, then I’ll merge. 

No substantive changes to release so I’ll wait for the next bug fix.

dbwiddis:
Just released 6.1.6.  It did not do anything with the site.  

So I rebased site branch to master, backed up a commit, and pushed upstream.

PR #1932, opened 2022-02-14
Revert "Bump maven-compiler-plugin from 3.9.0 to 3.10.0"

dbwiddis:
Reverts oshi/oshi#1929

A change to the maven-compiler-plugin adds missing package-info files, but it seems to not work on Windows in conjunction with mvn-bnd-plugin.  Reverting until the next version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,7 +101,7 @@
         <hamcrest.version>2.2</hamcrest.version>
         <!-- Compile versions -->
         <maven-clean-plugin.version>3.1.0</maven-clean-plugin.version>
-        <maven-compiler-plugin.version>3.10.0</maven-compiler-plugin.version>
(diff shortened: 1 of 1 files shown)


hazendaz:
also seeing it broken elsewhere as well.  seems problematic the logic added.  worse on windows for sure.  I had some branches scan clean on PR but once merged, other actions had issues.

dbwiddis:
Filed https://issues.apache.org/jira/browse/MCOMPILER-485

hazendaz:
Great detail, I would not have caught the issue.  After seeing that it makes sense.  I believe they want a small sample now to quickly write into the compiler plugin as an IT test to get it fixed and prevent future issues.

dbwiddis:
Yeah, it was fun to debug.  I think I've given them about as much as I can to solve it, although I'm 99% sure I the fix is `string.replace(File.separator, '/');` (or some variation thereof).

Issue #1936, opened 2022-02-18
Issues follow #Wrong CPU load Data ? #1870


Jiangyongqiang:
@dbwiddis   Hello, Daniel, I use the latest version of 6.1.1 to test the CPU utilization. The utilization rate will be greater than the actual utilization rate within about five minutes after startup, and exceed 100%. It will return to normal after five minutes. Why?
![企业微信截图_20220218095903](https://user-images.githubusercontent.com/13825585/154603233-db9c5c5a-2a36-4915-b7ee-73ff4c37507e.png)

dbwiddis:
What does the task manager show over the same time period?

Jiangyongqiang:
@dbwiddis 
I've used the snapshot version before. This version is accurate
https://oss.sonatype.org/content/repositories/snapshots/com/github/oshi/oshi-core/6.1.1-SNAPSHOT/oshi-core-6.1.1-20220129.232529-24.jar

![333](https://user-images.githubusercontent.com/13825585/154613304-6bc51269-7cc7-49a3-af88-f1b5a366cbc8.png)

dbwiddis:
Can you print out the ticks each second so I can see what’s happening with them?

Jiangyongqiang:
> Can you print out the ticks each second so I can see what’s happening with them?
![444](https://user-images.githubusercontent.com/13825585/154624165-57a6e85a-3afe-4707-a29d-f23871c2779a.png)

dbwiddis:
OK, I'm unable to reproduce this on my machine.  There's clearly a problem in the data, however, as the system ticks are decreasing which should not be possible.

Can you please fork the project and execute from the fork, and gather some data on the internal calculations?  The ones that are important here are:
- The baseticks, privticks, and procticks arrays populated in the WindowsCentralProcessor constructor (lines 114-116, so print out the value after that loop completes)
- Those same values when queried from the counters at lines 350-352 (those are lists that you may need to create arrays from to print)
- The intermediate idlepercent and syspercent (per processor) on lines 402-403

dbwiddis:
Some other thoughts. The way the code works is:
- it immediately catches a snapshot of the capacity counter ticks and the base (timer snapshot) when the CentralProcessor is instantiated, and saves them.
- for all future ticks it calculates the difference in counters (both process/system and the "base") which produces a "percentage" over that time period.
- The percentage can then be multiplied by elapsed time to determine the overall tick calculation
[...]

PR #1937, opened 2022-02-19
Refactor processor utility calculations to reduce duplication

dbwiddis:
Initial attempt to simplify calculations for #1936


Jiangyongqiang:
OK, let me try

PR #1958, opened 2022-03-03
Include subdirectories in spotless configuration

### LINKED ISSUE #1957: spotless misconfigured (minor issues - missing all possible files)
I tried to use the spotless here as shown.  *.xml for example is incorrect.  That only will touch the root directory.  The <includes> is maven based type of includes.  Each that is *.xxx will need to be **/*.xxx to be accurate.  I didn't have time now to address as I'm working this at work but wanted to note it so its not forgotten.  Otherwise spotless worked out really well from what I can see.

dbwiddis:
Fixes #1957 (and lots of trailing whitespace / EOF issues)


dbwiddis:
Unrelated, it's failing license check, thinking WIndowsCentralProcessor should be 2022 and not 2021-2022.  I'm thinking there's a bug in the license check plugin...

hazendaz (approved):
LFTM.  I'm sure it cycles through more than it should but that was kind of what I saw after adjusting it.  Far more files than I expected in my use case too :)

hazendaz:
java 8 is complaining that its missing a header.

PR #1977, opened 2022-03-15
KstatUtil.java: test for UnsatisfiedLinkError for KStat2

jimklimov:
KStat2.java says callers should "test for UnsatisfiedLinkError" since "Kstat2 is available in Solaris 11.4 and later."
Running an oshi consumer on an older Solaris, or in its illumos siblings, crashes due to JNA, e.g.:
````
Exception in thread "main" java.lang.UnsatisfiedLinkError: Unable to load library 'kstat2':
ld.so.1: java: fatal: libkstat2.so: open failed: No such file or directory
Native library (sunos-x86-64/libkstat2.so) not found in resource path ([file:/export/home/abuild/jenkins-swarm/swarm-client-3.31.jar])
...
        Suppressed: java.lang.UnsatisfiedLinkError: ld.so.1: java: fatal: libkstat2.so: open failed: No such file or directory
                at com.sun.jna.Native.open(Native Method)
                at com.sun.jna.NativeLibrary.loadLibrary(NativeLibrary.java:191)
                ... 16 more
````

### DIFF
--- a/oshi-core/src/main/java/oshi/util/platform/unix/solaris/KstatUtil.java
+++ b/oshi-core/src/main/java/oshi/util/platform/unix/solaris/KstatUtil.java
@@ -305,7 +305,14 @@ public final class KstatUtil {
      */
     public static Object[] queryKstat2(String mapStr, String... names) {
         Object[] result = new Object[names.length];
-        Kstat2MatcherList matchers = new Kstat2MatcherList();
+        Kstat2MatcherList matchers = null;
+        try {
+            matchers = new Kstat2MatcherList();
+        } catch (java.lang.UnsatisfiedLinkError e) {
+            LOG.debug("Is this Solaris 11.4+? Failed to get stats on {} for names {}: {}", mapStr, Arrays.toString(names), e.getMessage());
+            return result;
+        }
+
         KstatUtil.CHAIN.lock();
         try {
             matchers.addMatcher(Kstat2.KSTAT2_M_STRING, mapStr);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by dbwiddis** [comment]
Thanks for this submission.  

However, I believe I have already addressed this for OSHI in #1966 (in SNAPSHOT but not yet released) which exposes a `HAS_KSTAT2` global `boolean` that can be used.   It's still up to the user to check this before using the Kstat2 methods, but in this case I think allowing the exception is preferable.

**Comment by dbwiddis** [comment]
Given the code in #1966, I think a better approach for the two methods you've indicated would be to throw a more specific exception alerting the developer of their error, e.g., 
```java
public static Object[] queryKstat2(String mapStr, String... names) {
    if (!SolarisOperatingSystem.HAS_KSTAT2) {
        throw new UnsupportedOperationException("Kstat2 requires Solaris 11.4+. Use SolarisOperatingSystem#HAS_KSTAT2 to test this.");
    }
}
```

jimklimov:
Frankly, I got here because I had a program (jenkins swarm client) that
worked before bug out now. After all in this case, it is OSHI internals
using methods inappropriate for the platform - why propagate the exception
further when callers might not be aware of it?

As far as that swarm client is concerned, it just calls
`SystemInfo().getOperatingSystem().getProcess(oldPid)` to check if an
earlier copy is running... what could possibly go wrong? :) (per trace,
getOS fails there)

dbwiddis:
> Frankly, I got here because I had a program (jenkins swarm client) that worked before bug out now. After all in this case, it is OSHI internals using methods inappropriate for the platform

The earlier check for Kstat2 presence attempted to parse operating system major and minor version to detect 11.4 by parsing `uname -rv`.  I'm not sure which version of Solaris/Illumos gave you the bug, but I'm suspecting that parsing failed.  (I'm curious what the output is for that on the system giving you trouble.)

But the bottom line is, that bug has already been fixed in #1966.  It's not in a released version yet, which is why you are seeing it; but if you were to run 6.1.5-SNAPSHOT it would hopefully not occur.

I'll update this PR per my suggestion and merge it.

jimklimov:
FWIW (for more patterns), OpenIndiana using a "vanilla" illumos-gate
replies e.g.

````
5.11 illumos-bf41048984
````

and OmniOS CE says
````
5.11 omnios-r151036-a13510b579
````

dbwiddis:
Given that I'm no longer checking that, we're good in either case.  The new code actually attempts to load the library so doesn't even care what version it is.

I did just cut a release for 6.1.5 that fixes this, should be on Central in a few hours.

PR #1978, opened 2022-03-17
Update all CI machines to JDK11

dbwiddis:
- Installed JDK11 on all CI machines and updated workflows.
~- Removed JDK activated profiles~
~- Removed animal-sniffer~

Is it possible to simplify the duplicate configurations in java11 module for javadoc plugin? (It needs source path overridden in both `<build>` and `<reporting>` sections.)

### DIFF
--- a/.github/workflows/aix.yaml
+++ b/.github/workflows/aix.yaml
@@ -15,13 +15,13 @@ on:
       - '**.yaml'
 
 jobs:
-  # SSH into AIX server and run test on AIX 7.1
+  # SSH into AIX server and run test on AIX 7.2
   # Configured to pull latest from oshi master branch
   # To test on a PR, log into the aix server and create a new branch
   testaix:
     if: github.repository_owner == 'oshi'
     runs-on: ubuntu-latest
-    name: Test JDK 8, aix
+    name: Test JDK 11, aix
     steps:
     - name: Test in AIX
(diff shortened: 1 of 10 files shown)


dbwiddis:
Scrutinizer is failing on the sonar:sonar goal on the module-info.java file... can't figure out how to tell it to ignore that.

dbwiddis:
Got that excluded and hit new error: with javax/xml/bind/JAXBException.  From [here](https://community.sonarsource.com/t/sonarqube-error-javax-xml-bind-jaxbexception/25027/3) it appears SonarQube 6.5 (which Scrutinizer uses) only supports jdk 8.

hazendaz:
> Got that excluded and hit new error: with javax/xml/bind/JAXBException. From [here](https://community.sonarsource.com/t/sonarqube-error-javax-xml-bind-jaxbexception/25027/3) it appears SonarQube 6.5 (which Scrutinizer uses) only supports jdk 8.

Why not use sonar directly?  Example.  Not sure if that tool provides anything else but 6.x is really, really old.  https://github.com/mybatis/mybatis-3/actions/workflows/sonar.yaml

dbwiddis:
We do use sonar directly.  I don’t think we have it configured on PR workflows though.

dbwiddis:
And Sonar doesn't do PR workflows for the free/open source level.

PR #2010, opened 2022-04-13
Replace Finalizers with Cleaners for PDH Query Handles

dbwiddis:
Bring over a portion of https://github.com/java-native-access/jna/pull/1402 for handles associated with PDH Queries

This will help test whether that upstream fix is a good solution to #2009


dbwiddis:
Closing this PR as the JNA code it was based on has been significantly revamped, and will soon be merged... we will gain the benefits in JNA version update.

Issue #2019, opened 2022-04-23
In the docker container, oshi gets cpu information will be blocked


conghuhu:
![image](https://user-images.githubusercontent.com/56248584/164884012-c453680d-d62e-4b09-93b3-2db8ed58de88.png)
My springboot application is deployed in a docker container. Calling the cpu api as shown in the figure below cannot obtain information, and it will be blocked all the time, as shown in the above figure.
![image](https://user-images.githubusercontent.com/56248584/164884055-47fc42d0-9b83-47a6-881a-548b947f9849.png)

After testing, it is possible to obtain memory information and obtain jvm information, and hope to get a reply.

dbwiddis:
Which operating system is the container running on?

Does the container have access to procfs or have an alternate in-container location for the equivalent of the `/proc` directory?

conghuhu:
Docker is deployed on centos, and the base image of the container is openjdk:8-jre-alpine.
The project is a springboot, so should I change the base image in the dockerfile to centos or ubuntu?
Thank you again.


| |
丛国庆
|
|
***@***.***
|
On 4/23/2022 15:09，Daniel ***@***.***> wrote：

Which operating system is the container running on?

Does the container have access to procfs or have an alternate in-container location for the equivalent of the /proc directory?

dbwiddis:
Can you see the contents of `/proc/stat` from inside the container?

dbwiddis:
There is not enough information in your post for me to diagnose the problem.
 - The top image appears to show a synchronization thread triggering every 5 seconds, and two log messages from an attempt to get CPU information, with only the first of the two log messages shown
 - This implies the CPU information thread is terminating somewhere in the code snippet shown
 - The only possibility inside the thread would be a hang/delay/error executing `getSystemCpuLoadTicks()` 
   - The code for this on Linux simply reads the contents of `/proc/stat` and parses it
 - The other possibility is that the thread is being terminated from outside by the application, possibly due to the internal 1000 ms delay in your code

Suggestion: don't use code with an internal delay.  Just return the raw tick values to your springboot application and do the usage calculations there.

dbwiddis:
Heh @conghuhu, haven't heard back from you so I'm going to assume this is resolved and close this issue.

I researched enough to know that springboot does include default timeouts and it's not waiting for you do delay as long as you do before responding to the query.  As mentioned in the previous comment, you should immediately return the raw tick values, and your base application should compare them to an older returned set to do the calculation.

[...]

PR #2046, opened 2022-05-26
Added getSystemCpuLoad/getProcessorCpuLoad convenience methods

Osiris-Team:
Small but useful util methods to retrieve the current CPU usage easily.

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -173,7 +173,7 @@ public interface CentralProcessor {
      *
      * @param delay
      *          Milliseconds to wait.
-     * @return Value between 0 and 1 (100%) that represents the cpu usage in the provided time period.
+     * @return value between 0 and 1 (100%) that represents the cpu usage in the provided time period.
      */
     default double getSystemCpuLoad(long delay) {
         long start = System.nanoTime();
@@ -191,7 +191,7 @@ public interface CentralProcessor {
      *
      * @param delay
      *          Milliseconds to wait.
-     * @return Value between 0 and 1 (100%) that represents the cpu usage in the provided time period.
+     * @return array of CPU load between 0 and 1 (100%) for each logical processor, for the provided time period.
      */
     default double[] getProcessorCpuLoad(long delay){
         long start = System.nanoTime();
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -166,14 +166,14 @@ public interface CentralProcessor {
     double[] getSystemLoadAverage(int nelem);
 
     /**
-     * This is a convenience method which collects an initial set
-     * of ticks using {@link #getSystemCpuLoadTicks()} and passes
-     * that result to {@link #getSystemCpuLoadBetweenTicks(long[])}
-     * after the specified delay.
+     * This is a convenience method which collects an initial set of ticks using
+     * {@link #getSystemCpuLoadTicks()} and passes that result to
+     * {@link #getSystemCpuLoadBetweenTicks(long[])} after the specified delay.
      *
      * @param delay
-     *          Milliseconds to wait.
-     * @return value between 0 and 1 (100%) that represents the cpu usage in the provided time period.
+     *            Milliseconds to wait.
+     * @return value between 0 and 1 (100%) that represents the cpu usage in the
+     *         provided time period.
      */
     default double getSystemCpuLoad(long delay) {
         long start = System.nanoTime();
@@ -187,14 +187,14 @@ public interface CentralProcessor {
     }
 
     /**
-     * This is a convenience method which collects an initial set
-     * of ticks using {@link #getProcessorCpuLoadTicks()} and passes
-     * that result to {@link #getProcessorCpuLoadBetweenTicks(long[][])}
-     * after the specified delay.
+     * This is a convenience method which collects an initial set of ticks using
+     * {@link #getProcessorCpuLoadTicks()} and passes that result to
+     * {@link #getProcessorCpuLoadBetweenTicks(long[][])} after the specified delay.
      *
      * @param delay
-     *          Milliseconds to wait.
-     * @return array of CPU load between 0 and 1 (100%) for each logical processor, for the provided time period.
+     *            Milliseconds to wait.
+     * @return array of CPU load between 0 and 1 (100%) for each logical processor,
+     *         for the provided time period.
      */
     default double[] getProcessorCpuLoad(long delay) {
         long start = System.nanoTime();
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -166,23 +166,30 @@ public interface CentralProcessor {
     double[] getSystemLoadAverage(int nelem);
 
     /**
-     * This method waits one second and then returns
-     * the processor usage for that second.
+     * This method waits one second and then returns the overall processor load average for that second.
+     * Note that the returned value may be above 1 (100%), which means the processor is being overloaded.
+     * Also keep in mind, that the average load of all logical processors is getting returned.
+     *
      * @return Value between 0 and 1 (100%) that represents the average processor usage of the next second.
-     * @see #getProcessorCpuLoadBetweenTicks(long[][])
+     *         Note that the returned value may be above 1, which means the processor is being overloaded.
+     * @see #getSystemLoadAverage(int)
      */
-    default double getCurrentUsage() throws InterruptedException {
-        return getCurrentUsage(1000);
+    default double getOverallCurrentSystemLoadAverage() throws InterruptedException {
+        return getOverallCurrentSystemLoadAverage(1000);
     }
 
     /**
-     * This method waits the provided amount of milliseconds and then returns
-     * the processor usage for that time period.
-     * @param ms Milliseconds to wait.
+     * This method waits the provided amount of milliseconds and then returns the overall processor load average for that time period.
+     * Note that the returned value may be above 1 (100%), which means the processor is being overloaded.
+     * Also keep in mind, that the average load of all logical processors is getting returned.
+     *
+     * @param ms
+     *          Milliseconds to wait.
      * @return Value between 0 and 1 (100%) that represents the average processor usage of the provided time period.
-     * @see #getProcessorCpuLoadBetweenTicks(long[][])
+     *         Note that the returned value may be above 1, which means the processor is being overloaded.
+     * @see #getSystemLoadAverage(int)
      */
-    default double getCurrentUsage(long ms) throws InterruptedException {
+    default double getOverallCurrentSystemLoadAverage(long ms) throws InterruptedException {
         long[][] oldTicks = getProcessorCpuLoadTicks();
         Thread.sleep(ms);
         double[] arr = getProcessorCpuLoadBetweenTicks(oldTicks);
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -180,7 +180,7 @@ public interface CentralProcessor {
         long[] oldTicks = getSystemCpuLoadTicks();
         long toWait = delay - (System.nanoTime() - start) / 1_000_000;
         // protect against IllegalArgumentException
-        if(toWait > 0L) {
+        if (toWait > 0L) {
             Util.sleep(delay);
         }
         return getSystemCpuLoadBetweenTicks(oldTicks);
@@ -201,7 +201,7 @@ public interface CentralProcessor {
         long[][] oldTicks = getProcessorCpuLoadTicks();
         long toWait = delay - (System.nanoTime() - start) / 1_000_000;
         // protect against IllegalArgumentException
-        if(toWait > 0L) {
+        if (toWait > 0L) {
             Util.sleep(delay);
         }
         return getProcessorCpuLoadBetweenTicks(oldTicks);
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -196,7 +196,7 @@ public interface CentralProcessor {
      *          Milliseconds to wait.
      * @return array of CPU load between 0 and 1 (100%) for each logical processor, for the provided time period.
      */
-    default double[] getProcessorCpuLoad(long delay){
+    default double[] getProcessorCpuLoad(long delay) {
         long start = System.nanoTime();
         long[][] oldTicks = getProcessorCpuLoadTicks();
         long toWait = delay - (System.nanoTime() - start) / 1_000_000;
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -166,36 +166,19 @@ public interface CentralProcessor {
     double[] getSystemLoadAverage(int nelem);
 
     /**
-     * This method waits one second and then returns the overall processor load average for that second.
-     * Note that the returned value may be above 1 (100%), which means the processor is being overloaded.
-     * Also keep in mind, that the average load of all logical processors is getting returned.
+     * This method waits the provided amount of milliseconds
+     * and then returns the cpu usage for that time period.
      *
-     * @return Value between 0 and 1 (100%) that represents the average processor usage of the next second.
-     *         Note that the returned value may be above 1, which means the processor is being overloaded.
-     * @see #getSystemLoadAverage(int)
-     */
-    default double getOverallCurrentSystemLoadAverage() throws InterruptedException {
-        return getOverallCurrentSystemLoadAverage(1000);
-    }
-
-    /**
-     * This method waits the provided amount of milliseconds and then returns the overall processor load average for that time period.
-     * Note that the returned value may be above 1 (100%), which means the processor is being overloaded.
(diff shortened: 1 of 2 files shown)


dbwiddis (changes requested):
Thanks for your contribution.  A few problems, though.

I'm not sure I really like having a method in the API that doesn't return immediately.  See #2019 for unexpected behavior when there is a delay in a value.  I might be convinced of this, but I don't see anything here that a user can't do, and would probably prefer this method were put in a "Util" class rather than here.

I am not sure "load average" is the correct term to use here. It has a very specific meaning in the Unix-based implementations and is associated with longer-term averages.  The "CPU Load between ticks" methods are closer to this.

Why add up individual processor ticks instead of just collecting total system ticks (and multiplying by number of processors)?

Osiris-Team:
> See https://github.com/oshi/oshi/issues/2019 for unexpected behavior...

The method throws an interrupted exception, which should already be enough to let the developer know that the method waits. That issue seems more like an edge case instead of a real problem.

> I might be convinced of this, but I don't see anything here that a user can't do...

Yeah, but isn't that exactly the point of a library? To provide useful methods to the developer and make their lives a tiny bit easier.

> ...and would probably prefer this method were put in a "Util" class rather than here.

Yeah can be done, but it felt right putting those methods there. I personally don't use util classes of libraries that often, and it wouldn't be the first thing that comes to my mind to search for.

> [It has a very specific meaning ](https://www.howtogeek.com/194642/understanding-the-load-average-on-linux-and-other-unix-like-systems/)in the Unix-based implementations...

Oh yes, I see, didn't know about that, and looks like I misunderstood the getSystemLoadAverage() method. I am trying to retrieve the overall CPU usage percentage, so it could be named `getCurrentUsage()` for example?

> Why add up individual processor ticks instead of just collecting total system ticks (and multiplying by number of processors)?

I have no clue what that means.

Just added a new commit fixing a few mentioned things.

dbwiddis:
> The method throws an interrupted exception, which should already be enough to let the developer know that the method waits. That issue seems more like an edge case instead of a real problem.

Yes and no.  It may appear as an edge case but it is a fundamental departure from the existing API pattern that I think deserves extra scrutiny and consideration from other users.

> Yeah, but isn't that exactly the point of a library? To provide useful methods to the developer and make their lives a tiny bit easier.

Documenting how to do this in user code would be equally as valuable. The `oshi-demo` project exists to provide things like this.  I can envision an example which creates a daemon thread to collect the appropriate information for users that can be used as a template.

> > Why add up individual processor ticks instead of just collecting total system ticks (and multiplying by number of processors)?
> 
> I have no clue what that means.

And this is something that needs to be resolved before merging this PR.

There are multiple ways of measuring CPU load, and this PR proposes one that doesn't really align exactly with the existing measures.  This is fine if you're doing it in your own app, but putting it in a cross-platform app requires enough documentation to understand just what value is being returned.

OSHI returns a "load average" on Unix platforms, as cited above.  This is an OS-provided metric on all systems but Windows.

OSHI provides tick counters to enable users to count per-processor or system-wide CPU usage (CPU Load) over some time periods.  These provide a lot of detail into the types of usage (user/kernel).

OSHI provides a "convenience method" to calculate, based on the tick values, the percent usage per-processor or for the whole system given a previously stored set of ticks.  It is envisioned that any user who wants to collect long term averages will save such ticks themselves and use them.

This PR proposes an additional convenience method on top of the existing convenience method, choosing one of many possible interpretations of CPU Load.  I am very hesitant to add something to the API that is not a well-known or well-understood standard way of measuring CPU load.

The implementation you propose adds up the CPU usage load from each of the individual processors.  This does *not* correspond to existing metrics in existing OS displays and will potentially lead to user confusion. I can almost guarantee a "Why does the load average not match the task manager?" question.    It can (and usually will) exceed 100% and there is zero explanation of why, or why that 's a useful metric.

The calculation "roughly" correlates to the "Load Average" metric on unix-based machines, except:
 - load average actually measures something completely different.
 - load average is typically counted at 1/5/15 minute intervals, not 1 second, which seems less useful.  By the time you've measured it, it's changed.  I really see very little value in it unless it's being measured in a separate thread reporting values. 
   - in which case such thread should be just collecting the ticks and doing the same calculation.
 - the javadocs for your proposal don't really outline what the statistic means or how it differs from other metrics available already.

In sum:
 - If we implement this, the name should more closely align with the underlying methods (e.g., "Cpu Load") but not including the terminology "Load Average".
 - The javadocs should very clearly specify expected values and their meaning/interpretation.  This seems to be an invented metric that does not correlate to an existing standard load measurement on any existing OS.  It must be clearly documented to be understood.
 - I really don't think this is a good idea. I am willing to wait to be convinced otherwise and am happy for other maintainers to chime in with their opinions so I am not the mean guy saying no.   But inventing a new, non-standard metric requires a higher bar, in my opinion.

dbwiddis:
> > But inventing a new, non-standard metric requires a higher bar, in my opinion.
> 
> Oh, well that wasn't my intention either. I just wanted to measure the CPU usage. Didn't know this was such a complicated topic. 

Yeah, different operating systems measure different things, and then sometimes aren't consistent with each other so it does get a bit complex.  And as a cross-platform project, OSHI must try to stay as close as possible to well-defined definitions of these terms.

> Did a bit more of digging and the load average you were talking (UNIX) about can go over 100%, which means overload, which means there were processes waiting, aka being blocked by other processes currently using the CPU. 

Load average over 1 doesn't mean overload unless you're on a single processor machine.  On a 4-core machine, a load average of 4 is fully loaded.  A load average of 1 means only 25% load.

> This concept doesn't exist in Windows I think. There's just 100% in TaskManager and that's it, which is kinda sad, to be honest. 

Windows is even worse as starting with Windows 8 they changed the way they calculated the Task manager value!

And you are correct, Windows does not measure load average in the background.  It is possible to start a daemon thread in the background to collect the performance statistic to calculate it the same way as Unix machines.  You'd record the queue at periodic intervals and then average them over the 1, 5, and 15 minute intervals to match.  

> So lets stick to the Unix variant.

We already have the operating-system provided "load average" on unix-based systems that give values over 1, 5, and 15 minute averages.  Based on how much load changes second-to-second, this is a far more accurate and useful measure than a one-second poll.

Further, the one-second values can be really inaccurate due to roundoff.  On Windows, the tick granularity is 1/64 of a second, so each processor can be inaccurate by up to 1/32 of a second.  On an 8-logical-core windows machine that means your ticks measuring usage for one second can be off by 25% high or low!

> > This does not correspond to existing metrics in existing OS displays...
> 
> If I understood it correctly my calculation returns a value between 0 and 1 (100%), or above 1 if the cpu is overloaded. 

Not quite and perhaps this is the misunderstanding.  Let me review what OSHI provides and its meaning:

**Load Average** 

This is the OS-provided 1-, 5-, and 15-minute averages on Unix-based machines.  (It is not provided on Windows.).  It is normally expressed as a numeric value, not as a percentage.   It represents "number of processors fully loaded" so can range from 0 up to the number of logical cores on the machine.  An 8-core machine at 75% load will have a load average of 6.0.   

Load Average is always measured at the "system" level.

**Cpu Load**

This is a percentage measure, normally between 0% and 100%.  OSHI does not provide this directly; it simply supplies "tick counters" -- number of milliseconds the processor has spent in various states (Idle, User, System, etc.) and the usage calculation is (non-Idle ticks / all ticks).  For convenience, OSHI will calculate the usage between "now" and a previous set of ticks provided as an argument.  SO if you are polling the ticks every second you can store the last set, then when you update you grab the new ticks, calculate the change, and save the newest ones.

Windows Task Manager doesn't display this load, however. It displays a different metric "amount of work completed compared to the nominal workload of the processor".  So for modern processors which can vary their frequency for power savings or performance boost, the "percent of time" measure is inaccurate.  A processor running at half-frequency may be busy twice the amount of time as one at nominal frequency but they complete the same amount of work.  Or a processor can raise its frequency ("turbo boost", etc.) and do the work of 2 processors for a second.  Scale that up to a full machine and it can exceed 100% -- except in Windows 8, users got confused at >100% values so they made it even worse in 8.1, capping it at 100%.

CPU Load is calculated for both the whole system (0-100%) and for individual processors (0-100% per processor).  And this is where your implementation is very confusing.  You calculate the individual processor load and add it up, which makes it correlate somewhat with the load average, but calculates it in a completely different way so it is not appropriate to call it that.

>  So it returns the same as `getSystemLoadAverage()`, but instead of an average of 1min it gets the average for 1 second (and of all logical processors). 

It's not the same. It's roughly correlated, but as mentioned before, highly inaccurate over a short time period.

But it doesn't match any existing displays.  Load Average values are well-understood for those who use them, and they would have very little use for a highly variable 1-second average, especially one they have to wait a second to even get.

> So the stricly correct name would be `getOverallCurrentSystemLoadAverage()`. So why do you want to change the name? Or did I miss something?

I think any use of the terminology "load average" that does not match the decades-old interpretation on Unix-based machines is inaccurate. 

You are essentially returning the result of `getSystemCpuLoadBetweenTicks(oldTicks)` except:
 - you are collecting both sets of ticks yourself and delaying the response
 - you are essentially multiplying that value by the number of processors, giving confusing results (instead of returning a single value 0-100% you are adding up multiple values 0-100% and not really explaining why)

What is your ultimate goal here?  Are you just trying to add convenience in tick calculation?  Then you should match something OSHI already returns.  Are you trying to replicate "Load Average" on Windows?  Then we should do our best to match the actual calculation with a daemon thread.

Osiris-Team:
I am on Windows. My ultimate goal at the start was to replicate the CPU usage (0-100%) the TaskManager shows.

dbwiddis:
Sorry about that... have been traveling for the last few days and trying to squeeze in these reviews in at the end of long days.  You're correct that you are dividing at the end... in which case why don't you just get the system ticks up front instead of the processor ticks?  On most operating systems that will be a more efficient code path and should generate an identical result.  Possibly a more accurate one in some cases.

So here's what would make me happy:
 - Use `getSystemCpuLoadTicks()` and `getSystemCpuLoadBetweenTicks(long[] oldTicks)` instead of your processor tick average.
 - Use a name similar to the above, e.g., `someverbSystemCpuLoad(long delay)`.  Document that it's a convenience method for the above two calls with the specified delay.
   - The reason I say "some verb" is that "get" implies a getter fetching a field, and is not expected to have a delayed return.  Possibly "calculate" or "query".
   - I'm not strongly opposed to using "get" here to match the "between ticks" getter, as long as the argument is clearly named/labeled to indicate the delay.
 - Do not do the no-arg one that chooses an arbitrary time (1 second).  
   - For users who convert OSHI output into JSON using the Jackson object mapper, a no-arg method starting with "get" will be assumed to be a getter for a field and included in the JSON output. A delay here is not acceptable.  This could be mitigated with another verb than get
   - I think 1 second is too short for useful output, and it's not really asking too much of the user to specify the delay.

If you want an accurate Windows CPU Load calculation to match the unix-based values, let's make that a new issue/PR.

dbwiddis:
Also, consider whether you need to throw the interrupted exception.  OSHI has a `Util.sleep()` method that catches that exception and returns immediately, which is probably what we want.

Osiris-Team:
> Sorry about that... have been traveling for the last few days and trying to squeeze in these reviews in at the end of long days.

Haha ok, hope u have a nice vacation!
I implemented the changes u requested, tell me if I missed something.

I'll open an issue for the other feature.

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
```diff
@@ -165,6 +165,22 @@ public interface CentralProcessor {
...
+     *
+     * @param delay
+     *          Milliseconds to wait.
+     * @return Value between 0 and 1 (100%) that represents the cpu usage in the provided time period.
+     * @see #getSystemCpuLoadTicks()
+     * @see #getSystemCpuLoadBetweenTicks(long[])
+     */
+    default double getSystemCpuLoad(long delay) {
+        long[] oldTicks = getSystemCpuLoadTicks();
+        Util.sleep(delay);
```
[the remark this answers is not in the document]
dbwiddis: @sonatype-lift ignore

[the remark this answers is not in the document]
Osiris-Team: Oh I guess bc of Util.sleep?
I thought the logger would ultimately use a lock or synchronized method when writing to the output stream?

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
```diff
@@ -165,6 +165,22 @@ public interface CentralProcessor {
      */
     double[] getSystemLoadAverage(int nelem);
 
+    /**
+     * This method waits the provided amount of milliseconds
+     * and then returns the cpu usage for that time period.
```
dbwiddis, line 170: While this is sort-of okay, I'd prefer something more explicit, such as:
> `This is a convenience method which collects an initial set of ticks using {@link #getSystemCpuLoadTicks()} and passes that result to {@link getSystemCpuLoadBetweenTicks(long[])} after the specified delay.`

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
dbwiddis, line 180: A suggestion here.   The `getSystemCpuLoadTicks()` call takes a few milliseconds so the overall time period is going to be slightly more than the specified delay.   For example, if it takes 25ms to fetch the ticks, timestamped somewhere near the end of that period, you might get the following sequence:
 - Initial get call. 20ms later, ticks collected
 - 5 ms later, method returns
 - 1000ms delay
 - Final get call.  20ms later, ticks collected
 - 5 ms later, method returns

Total ticks collected end up being 1025 (5ms at tail end of initial call, 20ms at beginning of final call).

This can be corrected using the following sequence:
```java
long start = System.nanoTime();
long[] oldTicks = getSystemCpuLoadTicks();
long toWait = (System.nanoTime() - start) / 1_000_000;
// protect against IllegalArgumentException
if (toWait > 0L) {
    Util.sleep(toWait);
}
```

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
```diff
@@ -165,6 +165,22 @@ public interface CentralProcessor {
...
+     *
+     * @param delay
+     *          Milliseconds to wait.
+     * @return Value between 0 and 1 (100%) that represents the cpu usage in the provided time period.
+     * @see #getSystemCpuLoadTicks()
+     * @see #getSystemCpuLoadBetweenTicks(long[])
+     */
+    default double getSystemCpuLoad(long delay) {
```
dbwiddis, line 178: For completeness, do you also want to have a `double[] getProcessorCpuLoad(long delay)` method?  (Yeah, you had that before, but this one would return and not do the averaging.)

dbwiddis (approved):
OK, looking good.  A few suggestions inline.

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
```diff
@@ -165,6 +165,42 @@ public interface CentralProcessor {
...
+     *
+     * @param delay
+     *          Milliseconds to wait.
+     * @return Value between 0 and 1 (100%) that represents the cpu usage in the provided time period.
+     */
+    default double[] getProcessorCpuLoad(long delay){
+        long start = System.nanoTime();
+        long[][] oldTicks = getProcessorCpuLoadTicks();
+        long toWait = delay - (System.nanoTime() - start) / 1_000_000;
+        if(toWait > 0L) Util.sleep(delay); // protect against IllegalArgumentException
```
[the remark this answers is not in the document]
dbwiddis: @sonatype-lift ignore

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
```diff
@@ -165,6 +165,42 @@ public interface CentralProcessor {
...
+     *
+     * @param delay
+     *          Milliseconds to wait.
+     * @return value between 0 and 1 (100%) that represents the cpu usage in the provided time period.
+     */
+    default double getSystemCpuLoad(long delay) {
+        long start = System.nanoTime();
+        long[] oldTicks = getSystemCpuLoadTicks();
+        long toWait = delay - (System.nanoTime() - start) / 1_000_000;
+        if(toWait > 0L) Util.sleep(delay); // protect against IllegalArgumentException
```
dbwiddis, line 182: (Style) Please use curly braces and indent the result of this conditional.  Checkstyle will probably already be telling you this.

oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
```diff
@@ -165,6 +165,42 @@ public interface CentralProcessor {
...
+     *
+     * @param delay
+     *          Milliseconds to wait.
+     * @return array of CPU load between 0 and 1 (100%) for each logical processor, for the provided time period.
+     */
+    default double[] getProcessorCpuLoad(long delay){
+        long start = System.nanoTime();
+        long[][] oldTicks = getProcessorCpuLoadTicks();
+        long toWait = delay - (System.nanoTime() - start) / 1_000_000;
+        if(toWait > 0L) Util.sleep(delay); // protect against IllegalArgumentException
```
dbwiddis, line 200: (Style) same comment here.   Also a space between if and the parentheses.

dbwiddis (approved):
OK, looking good.  Almost ready to merge!.  Two style fixes please (I expect checkstyle will fail with those, see the tests).   Also:
 - please add an entry in `CHANGELOG.md` under the new features section
 - please run `mvn spotless:apply` before your final commit

dbwiddis:
`if(toWait > 0L)` should be `if (toWait > 0L)`.  Running spotless should fix this.  (Use maven if you have it or the maven wrapper otherwise.)

Osiris-Team:
Im having troubles with maven (IDE tells me its not installed, even though it is wtf) at the moment, so I can't run it.

dbwiddis:
> Im having troubles with maven (IDE tells me its not installed, even though it is wtf) at the moment, so I can't run it.

You should be able to use the maven wrapper (mvnw or mvnw.bat) on the command line.

Issue #2097, opened 2022-06-30
Prefer character classes to alternators in regex


dbwiddis:
### First Timers Only

I'm marking this issue for [first-timers-only](https://www.firsttimersonly.com/).  That means that I will only accept a Pull Request (PR) for this one from someone who's never contributed to an open source project on GitHub before.

The main goal of this task is to learn the process for submitting a PR -- the GitHub equivalent of writing a "Hello World!" program.  If you already know the PR workflow, please leave this issue for someone else!

This one is easy (but don't make that statement make you feel bad if you have a hard time with it, there's more to contributing to open source than changing lines of code, especially if it's your first time). I'll hold your hand through this if you need me to. 😄 

### The change you are making 

In the `FreeBsdCentralProcessor` class, a regex uses an alternation (`|`) to identify line terminators.
https://github.com/oshi/oshi/blob/43e5e6eaf968281457b25ecbec8dcf322265a98a/oshi-core/src/main/java/oshi/hardware/platform/unix/freebsd/FreeBsdCentralProcessor.java#L152

Character classes (`[]`) are more efficient.

### What to do 

Replace the regex in the above line with `"[\\n\\r]"`.

### How to submit your change to the repository

Here are the steps to get a PR merged here.
- Read the [Contributing](https://github.com/oshi/oshi/blob/master/src/site/markdown/Contributing.md) file for reference.  We'll follow those steps, and I'll walk you through anything you don't understand.
- Fork the project to your own GitHub account, make a local clone, and create a branch for your submission.
- Make the changes described above.
- Commit your changes, push to to your GitHub fork, and create a Pull Request
- Respond to the usually minor comments on code review
- Celebrate! :tada:

PR #2107, opened 2022-07-03
Fixed Issue #2097, by updating regEx

varun8338:


dbwiddis (changes requested):
Thanks for the fix. Your original change worked.

Sonatype suggested using Guava's `Splitter` but we do not have that dependency.  In the data we are parsing, the "surprising" behavior of Java's `split()` is OK. 

So can you roll back to your own first commit?

varun8338:
Creating new pull request as, still familiarizing with GIT :)

dbwiddis:
You could have done:
```
git pull
git reset --hard HEAD~1
git push -f
```
:)

varun8338:
While running **mvn test** in  cmd, it is failing to build for oshi-core-java11. Can you help me with this, attaching SS for your reference. 

image

dbwiddis:
That's a known bug in `junit-platform-maven-plugin`.   You need to do `mvn clean test` for it to work.

dbwiddis:
(Or just `mvn clean` after your previous `mvn test` run.... bottom line, multiple `test` without `clean` on the JPMS project fails.)

varun8338:
Just a suggestion, can we update the **Contributing.md** file, by replacing **mvn test** to **mvn clean test**? it might be confusing for first timers,

dbwiddis:
Will do, although once the bug is fixed, "test" will work again ;)

varun8338:
I have raised another pull request, [](https://github.com/oshi/oshi/pull/2108)

PR #2116, opened 2022-07-14
Add a workflow to use the vmactions/openbsd-vm image for testing OpenBSD

mprins:


mprins:
I've opened an issue for the VM as it is missing some required filesets for Java: https://github.com/vmactions/openbsd-builder/issues/1

.github/workflows/openbsdvm.yaml
```diff
@@ -0,0 +1,30 @@
+name: OpenBSD VM
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+      - openbsd-vm-workflow
```
dbwiddis, line 8: This can be removed now, right?

.github/workflows/openbsdvm.yaml
```diff
@@ -0,0 +1,30 @@
...
+    paths-ignore:
+      - '**.md'
+      - '**.yml'
+
+
```
dbwiddis, line 13: wouldn't have commented if this was the only thing, but let's lose one of the two newlines. I'll let you pick which one. :-)

dbwiddis (changes requested):
LTGM!   A few minor changes, and go ahead and remove (or replace with this content) the other OpenBSD test that's currently running on my Synology diskstation VM.

PR #2129, opened 2022-07-19
Jmx demo

SalvadorRomo:
This is just demo for JMX hope you can find it useful Regards!

### DIFF
--- a/oshi-demo/src/main/java/oshi/demo/jmx/ContextRegistrationPlatform.java
+++ b/oshi-demo/src/main/java/oshi/demo/jmx/ContextRegistrationPlatform.java
@@ -31,13 +31,17 @@ import javax.management.MalformedObjectNameException;
 import javax.management.NotCompliantMBeanException;
 import javax.management.MBeanRegistrationException;
 import java.beans.IntrospectionException;
+
 public class ContextRegistrationPlatform {
     private StrategyRegistrationPlatformMBeans strategyRegistrationContext;
+
     public void setStrategyRegistrationContext(StrategyRegistrationPlatformMBeans platformMBeans) {
         this.strategyRegistrationContext = platformMBeans;
     }
-    public void registerMBeans(SystemInfo sisInfo, MBeanServer mBeanServer) throws MalformedObjectNameException,
-        NotCompliantMBeanException, InstanceAlreadyExistsException, MBeanRegistrationException, IntrospectionException, javax.management.IntrospectionException {
+
+    public void registerMBeans(SystemInfo sisInfo, MBeanServer mBeanServer)
+            throws MalformedObjectNameException, NotCompliantMBeanException, InstanceAlreadyExistsException,
+            MBeanRegistrationException, IntrospectionException, javax.management.IntrospectionException {
         this.strategyRegistrationContext.registerMBeans(sisInfo, mBeanServer);
     }
 }
--- a/oshi-demo/src/main/java/oshi/demo/jmx/ContextRegistrationPlatform.java
+++ b/oshi-demo/src/main/java/oshi/demo/jmx/ContextRegistrationPlatform.java
@@ -1,0 +1,43 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2022 The OSHI Project Contributors: https://github.com/oshi/oshi/graphs/contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package oshi.demo.jmx;
+
+import oshi.SystemInfo;
+import oshi.demo.jmx.api.StrategyRegistrationPlatformMBeans;
+import javax.management.MBeanServer;
+import javax.management.InstanceAlreadyExistsException;
+import javax.management.MalformedObjectNameException;
+import javax.management.NotCompliantMBeanException;
+import javax.management.MBeanRegistrationException;
+import java.beans.IntrospectionException;
+public class ContextRegistrationPlatform {
+    private StrategyRegistrationPlatformMBeans strategyRegistrationContext;
+    public void setStrategyRegistrationContext(StrategyRegistrationPlatformMBeans platformMBeans) {
+        this.strategyRegistrationContext = platformMBeans;
+    }
+    public void registerMBeans(SystemInfo sisInfo, MBeanServer mBeanServer) throws MalformedObjectNameException,
+        NotCompliantMBeanException, InstanceAlreadyExistsException, MBeanRegistrationException, IntrospectionException, javax.management.IntrospectionException {
+        this.strategyRegistrationContext.registerMBeans(sisInfo, mBeanServer);
+    }
+}
--- a/oshi-demo/src/main/java/oshi/demo/jmx/JMXOshiAgentServer.java
+++ b/oshi-demo/src/main/java/oshi/demo/jmx/JMXOshiAgentServer.java
@@ -1,0 +1,350 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2022 The OSHI Project Contributors: https://github.com/oshi/oshi/graphs/contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package oshi.demo.jmx;
+import oshi.SystemInfo;
+import oshi.demo.jmx.api.JMXOshiAgent;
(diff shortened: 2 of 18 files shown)


dbwiddis:
Looks like Sonatype and the CI is giving you some stuff to work with, so I'll hold off on further review.  

But can you rebase onto master and make sure you keep the final solaris GHA file as it currently is on the master branch?

SalvadorRomo:
ready!! This last commit now has the changes  without conflict  87858b8

oshi-demo/src/main/java/oshi/demo/jmx/ContextRegistrationPlatform.java
```diff
@@ -0,0 +1,43 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package oshi.demo.jmx;
+
+import oshi.SystemInfo;
+import oshi.demo.jmx.api.StrategyRegistrationPlatformMBeans;
+
+import javax.management.*;
```
dbwiddis, line 29: Don't use star imports. :)

oshi-demo/src/main/java/oshi/demo/jmx/JMXOshiAgentServer.java
```diff
@@ -0,0 +1,338 @@
...
+    private Map properties;
+
+    private MBeanServer server;
+    private JMXServiceURL address;
+    private JMXConnectorServer cntorServer;
+
+    private ContextRegistrationPlatform contextRegistrationPlatform;
+
+    private static JMXOshiAgentServer jmxOshiAgentServer;
+    private static final PlatformEnum CURRENT_PLATFORM = PlatformEnum.getValue(Platform.getOSType());
```
dbwiddis, line 61: `SystemInfo.getCurrentPlatform()` already gets this, and you only use it once so it's probably fine to use it inline there.

oshi-demo/src/main/java/oshi/demo/jmx/JMXOshiAgentServer.java
```diff
@@ -0,0 +1,338 @@
...
+    private void initilizeMBeans() throws MalformedObjectNameException, NotCompliantMBeanException,
+            InstanceAlreadyExistsException, MBeanRegistrationException {
+
+        StrategyRegistrationPlatformMBeans platformMBeans = null;
+        SystemInfo si = new SystemInfo();
+
+        switch (CURRENT_PLATFORM) {
+        case WINDOWS:
+            platformMBeans = new WindowsStrategyRegistrattionPlatform();
+            contextRegistrationPlatform.setStrategyRegistrationContext(platformMBeans);
```
dbwiddis, line 116: You need a default case here, and a protection against NPE on the line below.

oshi-demo/src/main/java/oshi/demo/jmx/mbeans/BaseBoard.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    @Override
+    public String getSerialNumber() {
+        return this.getSerialNumber();
+    }
+
+    @Override
+    public String getManufacterer() {
+        return this.getManufacterer();
```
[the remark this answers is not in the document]
dbwiddis: See comment above by Sonatype... do you need `baseboard.getManufacturer()`?

oshi-demo/src/main/java/oshi/demo/jmx/mbeans/BaseBoardMBean.java
```diff
@@ -0,0 +1,32 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package oshi.demo.jmx.mbeans;
+
+public interface BaseBoardMBean {
```
dbwiddis, line 26: I'm wondering if rather than manually writing these MBeans you can use Jackson's `ObjectMapper` to accomplish the same purpose.  Otherwise that's a ton of boilerplate code.

dbwiddis (changes requested):

oshi-demo/src/main/java/oshi/demo/jmx/mbeans/BaseBoardMBean.java
SalvadorRomo: ohh  got it, will work over it, to use Jacksons ObjectMapper

oshi-demo/src/main/java/oshi/demo/jmx/JMXOshiAgentServer.java
SalvadorRomo: got it, and  in the baseboard a made mistake he, yes what i needed was baseboard.getManufacturer()

oshi-demo/src/main/java/oshi/demo/jmx/ContextRegistrationPlatform.java
SalvadorRomo: ohh got it

oshi-demo/src/main/java/oshi/demo/jmx/mbeans/BaseBoardMBean.java
SalvadorRomo: Hi Already chnaged all the others comments,  the only one pending is this one.

dbwiddis: Sorry, busy today.  

No specific suggestion here.  I'm just thinking from the experience of the oshi-json artifact that maintaining a whole "mirror" of the API is a lot of work.  I was hoping ObjectMapper could directly communicate with JMX without coding up the beans.

SalvadorRomo: I see some problems with those approach, but here are some solutions. 

1 - we can convert the oshi objects to Json, and create  DynamicMBeaans, this mbean will accept an oshi  json representation and the Dynamic MBean  will just set up that mbean with the json provided, but the problem here is that we would need to always keep updating that instance of DynamicMBeans since it wont hold any actual instance 

2- we can deal with reflection, we can pass the actual object and execute methods by reflection and sets the fields of this DynamicMBean by iterating over each field of each object and that way we can call also the update method on each oshi object  to keep the obj updateded

3 or we can make an empty interface eg BaseboardMBean  and that interface will extend the Baseboard interface,  that way any implementation from windows , mac , linux could implement the BaseBoardMBean and refactor the code to just offering a method for retrive the the MBean representation of whatever object , this new method could be only available at the oshi package level so anyone can't use it

oshi-demo/src/main/java/oshi/demo/jmx/ContextRegistrationPlatform.java
```diff
@@ -0,0 +1,48 @@
...
+import oshi.SystemInfo;
+import oshi.demo.jmx.api.StrategyRegistrationPlatformMBeans;
+import javax.management.MBeanServer;
+import javax.management.InstanceAlreadyExistsException;
+import javax.management.MalformedObjectNameException;
+import javax.management.NotCompliantMBeanException;
+import javax.management.MBeanRegistrationException;
+
+
+
```
dbwiddis, line 35: Lots of whitespace here.  Please import OSHI's [formatting conventions from here](https://github.com/oshi/oshi/blob/master/config/OSHIJavaFormatConventions.xml) into your IDE.

Alternately, run `mvn spotless:apply" and it should fix these for you.

oshi-demo/src/main/java/oshi/demo/jmx/mbeans/BaseBoardMBean.java
dbwiddis: Which option do you recommend?

OSHI already has interfaces that define what gets returned, which is essentially how Jackson converts them to its object.  The DynamicMBean approach sounds to me like a reasonable option, but you have a lot more experience here so I'd like to hear your thoughts.

.github/workflows/solaris.yaml
```diff
@@ -41,19 +40,19 @@ jobs:
     runs-on: ubuntu-latest
     name: Test JDK 11, Solaris SPARC
     steps:
-    - name: Test in Solaris SPARC
```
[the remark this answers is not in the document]
SalvadorRomo: Hi So sorry for this, I want to apologize because of this mess, first time being a contributor, thanks a lot for this steps, thanks to you I now more stuff about git!! Will fix my commit with this valuable information

oshi-demo/src/main/java/oshi/demo/jmx/mbeans/BaseBoardMBean.java
SalvadorRomo: well, rethinking more about what you had proposed, Im thinking in using [ASM](https://asm.ow2.io/) with a maven steps in which we can grab all the interfaces on the corresponding classpath,  parse those clases into JSON with object mapper and later create the MBeans dinamycaally , so when the strategy for regsitering MBeans is executed, can have those MBenas  ref, will work on that

SalvadorRomo: beacuese sadly for JMX we need interface with the suffix MBena and the DynamicMBean would have been a good option but we would need to mantain some enums to specify the operation on each interface and  that would end up in the same problem

SalvadorRomo: Well at the end I went with the approach of DynamicMBean and reflection, so it resloves the overhead of creating repetitive  code. And for giving the users the capability to know what properties are in each MBena I offer an Interface for giving them the list of props they can retrive for each MBEan

dbwiddis (approved):
This looks great!  Any way we can write up some short docs on how to use it?

oshi-demo/src/main/java/oshi/demo/jmx/mbeans/BaseBoardMBean.java
SalvadorRomo: I'm completely in for this effort about creating these servers relying on OSHI and from my perspective that I worked in System Managment systems, it is really hard to find these Open Source efforts well mantained and documented. Actually we struggle a lot with [sblim-sfcb](http://sblim.sourceforge.net/wiki/index.php/Sfcb) beacuse either you need to pay for support at the different Linux distos or you need to have  c developers. and thks to you and all this comunity I came up with this awsome  library that helped the businnes at that time. And with Converting this into  complete server , the posibility are endlees, from either sending threshold event and more.   And Regarding this effort, for overcoming the need to construct these interfaces I just implemented and approach with reflection, that way we can have the DynamicMBeams always in sync with watever change is made to the oshi interfaces

SalvadorRomo:
and regarding the doc will work on an md! hope I can get it done by the end of the day! Regards!!

SalvadorRomo:
done!

dbwiddis (commented):
This looks really good, thanks!  Codacy doesn't like some things that Spotless could probably fix. :)

I think this could be improved just a little bit with actual code samples (unless those are actually completed in the "demo" directory, in which case link to them!)  For example at the top, you have "all you need to do is..." but you could actually code a small class to do that and display it inline.   You can keep the "as an alternative" paragraph as is.

Similarly it looks like you have most of the code for a client already typed... can you wrap it in minimal class headers so that it works?  Just the basics to make it work, with the "if..." paragraphs following advising how to extend and adapt the code.  Again, if the "demo" folder has these complete classes then just link to them, no need to repeat the code.

Alternately just have some small classes that actually run the minimal code included in the project

I note we sill have the Baseboard mbean. Is that still needed? And I'm not sure I understand the WindowsStrategyRegistration file, which also looks Baseboard-specific.

SalvadorRomo:
Great!! Will work on those observations.

 And regarding the `WindowsStrategyRegistration`, My idea was to have an strategy for each platform so it could specialize the registration of each MBean into each platform, expample: cuztomize the MBean domain to have an specific name as  `oshi-windows:` or add more stuff specific to platform at the MBean registration. And in that class I just wrote only one Mbean registration that is the Baseboard, but what I had in mind is that strategy had all the MBeans being initialized as CPU and more, but just for simplicity reason I just declare the baseboard one. 

And Regarding the Baseboard MBean, it is not a MBean at all. Is a Proxy class that is extending `DynamicMBean`, so whenever we want to get some attribute from the MBean we can just override the `getAttribute ` and  call the specific getter method on the actual oshi object by reflection. But At the end we still need that proxy beacuse is the only way we can register this mbean into the MBean server beacause of the DyniamicMBean extension applying the MBean convention.

SalvadorRomo:
Another note i forgot to mention about `WindowsStrategyRegistration ` ,  If this server in the near future can make it to the core API,  we can  take advantge of going directly into the internal API to just get specific information regarding OS and Hadware without the need to use a `SystemInfo`  or `HadwareAbstractLyer` object , or another option is to aslk clients that they need specify these object and get rid of this class and make a generic one because, already those object has reference to each platform.

Issue #2141, opened 2022-07-28
Add GitHub Actions on multiple native libraries and architectures


dbwiddis:
Add a new GHA workflow to test:
 - architectures armv6, armv7, aarch64, s390x, or ppc64le
 - alpine linux which uses MUSL libc

https://github.com/marketplace/actions/run-on-architecture

dbwiddis:
Closing this issue as our attempts to add this (see #2151) didn't work, likely due to the emulation capabilities.

PR #2151, opened 2022-08-02
Added alpine CI

adrian-kong:
Adding alpine distro with armv6, armv7, aarch64, s390x, or ppc64le architectures via run-on-arch composite action - #2141

### DIFF
--- a/.github/workflows/alpine.yaml
+++ b/.github/workflows/alpine.yaml
@@ -25,12 +25,11 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: '8'
-      - name: Test on Alpine
-        id: test-alpine
+      - name: Test on Ubuntu
         uses: uraimo/run-on-arch-action@v2
         with:
           arch: ${{ matrix.arch }}
-          distro: alpine_latest
+          distro: ubuntu22.04
           install: |
             apk update
             apk add maven musl
--- a/.github/workflows/alpine.yaml
+++ b/.github/workflows/alpine.yaml
@@ -31,8 +31,8 @@ jobs:
         with:
           arch: ${{ matrix.arch }}
           distro: alpine_latest
-          run: |
-            mvn test -B -Djacoco.skip=true
+          shell: /bin/bash
+          run: mvn test -B -Djacoco.skip=true
           install: |
             apk update
             apk add maven musl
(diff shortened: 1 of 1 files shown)


dbwiddis:
Thanks for the submission!  Looks like it's not working... perhaps this GHA isn't ready for prime time yet.

dbwiddis:
These are all failing on not finding the JDK:
> Executing process failed: java.io.IOException: Cannot run program "/usr/lib/jvm/java-17-openjdk/bin/java" (in directory "/home/runner/work/oshi/oshi/oshi-core"): error=0, Failed to exec spawn helper: pid: 125, exit value: 1

I think the issue here is that the java executable available in the Ubuntu host is not accessible inside the QEMU vm.

Try adding java (`openjdk`?) to the `apk add maven musl` line.

dbwiddis:
(looks like you had that in c6016298f27fd333af9e7541db1afe5b795ac2e5 but then backed it out?)

adrian-kong:
Huh interestingly, some of them runs for hours whereas the others

```
  This command is not for general use and should only be run as the result of a call to
  ProcessBuilder.start() or Runtime.exec() in a java application
```

Maybe failing to run a command inside the docker?

adrian-kong:
@dbwiddis 

Managed to get 5 failures on [aarch64](https://github.com/oshi/oshi/actions/runs/3564655261/jobs/5988880376), [ppc64le](https://github.com/oshi/oshi/actions/runs/3564655261/jobs/5988880651) !

dbwiddis:
> Managed to get tests to run with 5 failures 

Woohoo!   I knew it was choking in the tests but now we have a stack trace!   It's the `ExecutingCommand.runNative()` that doesn't seem to be working.

adrian-kong:
Wow drains 5+ hours of running armv7 and armv6

.github/workflows/alpine.yaml
```diff
@@ -25,12 +25,11 @@ jobs:
...
-      - name: Test on Alpine
-        id: test-alpine
+      - name: Test on Ubuntu
         uses: uraimo/run-on-arch-action@v2
         with:
           arch: ${{ matrix.arch }}
-          distro: alpine_latest
+          distro: ubuntu22.04
           install: |
             apk update
```
mprins, line 34: Ubuntu uses `apt`

adrian-kong: oops

Issue #1533, opened 2021-02-19
Android Compatiblity


muneebqadir:
Hi, I couldn't find anything on android compatibility so just decided to ask here.
Sorry if I overlooked it but I found it was working with ARM Linux systems

mprins:
I don't know, did you try? 
Oshi might work, but it relies on https://github.com/java-native-access/jna for native code binding which might fail, part of the problem is android does not run java bytecode but dalvik bytecode and binding to native libs might fail and oshi might be using language features not available in your version of android.

may be of interest: https://stackoverflow.com/questions/15765212/android-jna-library

dbwiddis:
JNA is supposed to be Android compatible.   Right now if you tried it in OSHI, you'd get the "Operating System not supported" exception because we haven't added a switch for JNA's `Platform.ANDROID`.  I made a brief attempt to spin up an Android VM to try to test this but didn't get very far. I'm not sure how well an emulator will test this, and it's not been a priority for me to learn how to spin up an emulator just to test it.

It is possible that simply adding that as an option that feeds into OSHI's Linux code will "just work".

It's also possible that some things may break.

Bottom line:  if the Community wants Android support, the community needs to do this testing.

Start by adding `Platform.ANDROID` to the existing `SystemInfo` case statements leading to the Linux branch, and see what works.

sbotev5:
Hi and great to see this being discussed!

First of all, thanks for this amazing library.

I wanted to briefly touch on this library being used on Android devices (as part of the functionality of an Android app).

To start off, I want to ask would this library (assume setup hiccups resolved) provide additional stuff on top of Android SDK programmatic calls already available? 

From a Google search [result](https://www.viralandroid.com/2017/02/how-to-get-android-hardware-and-system-info-programmatically.html) I see the following are available:


`        button.setOnClickListener(new View.OnClickListener() {
            @Override
            public void onClick(View view) {
                textView.setText(
                        "SERIAL: " + Build.SERIAL + "\n" +
                        "MODEL: " + Build.MODEL + "\n" +
                        "ID: " + Build.ID + "\n" +
                        "Manufacture: " + Build.MANUFACTURER + "\n" +
                        "Brand: " + Build.BRAND + "\n" +
                        "Type: " + Build.TYPE + "\n" +
                        "User: " + Build.USER + "\n" +
                        "BASE: " + Build.VERSION_CODES.BASE + "\n" +
[...]

PR #2197, opened 2022-09-27
Added support for Android OS

milan-fabian:
As Android is based on Linux, reusing LinuxOperatingSystem and LinuxHardwareAbstractionLayer works for the most important parts (CPU, RAM, system description etc.).

Tested on Android 7, 9 and 11.

Resolves https://github.com/oshi/oshi/issues/1533.


dbwiddis:
This is awesome, thanks!

I'm curious what "doesn't" work... is it the Udev-based stuff?

milan-fabian:
Listing stuff currently produces no visible error / exception, it just returns no data for some of the queries.

For example `HardwareAbstractionLayer.getMemory().getPhysicalMemory()` and `HardwareAbstractionLayer.getDisplays()` return empty list on all test devices, `HardwareAbstractionLayer.getNetworkIFs()` return empty list on non-rooted Android 11 (probably because Android permissions).

I would like to have a closer look at some of the Android specifics in the future, but right now I am happy that it doesn't crash and that it can detect the CPU correctly :)

PR #2198, opened 2022-09-28
Add Processor Cache Information

### LINKED ISSUE #2190: CPU Cache
Hey,
is it possible to find out the L1, L2 and L3 cache of a processor via the API?

dbwiddis:
Fixes #2190

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/platform/unix/aix/AixCentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/unix/aix/AixCentralProcessor.java
@@ -128,7 +128,7 @@ final class AixCentralProcessor extends AbstractCentralProcessor {
         int lpPerPp = lcpus / physProcs;
         // Get L2 caches, one per physical processor
         List<ProcessorCache> caches = new ArrayList<>();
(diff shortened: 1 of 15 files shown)


oshi-core/src/main/java/oshi/hardware/platform/unix/aix/AixCentralProcessor.java
```diff
@@ -121,15 +123,48 @@ protected Pair<List<LogicalProcessor>, List<PhysicalProcessor>> initProcessorCou
         if (lcpus < 1) {
             lcpus = 1;
         }
+        int lpPerPp = lcpus / physProcs;
```
mprins, line 126: it seems that `lpPerPp` can become 0... eg. 1/2, 1/3... and then cause a divide by zero on line 132.

see #2228
explainer https://stackoverflow.com/questions/4685450/int-division-why-is-the-result-of-1-3-0

dbwiddis: > explainer https://stackoverflow.com/questions/4685450/int-division-why-is-the-result-of-1-3-0

Better explainer:  https://xkcd.com/2295/

PR #2203, opened 2022-10-01
NPE potentially caused by empty directories

adrian-kong:
Think iterating
```
for (File f : dir.listFiles((f, name) -> name.toLowerCase().endsWith(".conf"))) {
```
potentially causes unhandled null pointer errors in

`LinuxOperatingSystem` and `MacOperatingSystem` which wraps `Arrays.asList(null)`

And included some tiny refactoring (maybe should have split out in a different PR but hard to spot a lot)

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
@@ -272,8 +272,8 @@ public final class LinuxHWDiskStore extends AbstractHWDiskStore {
         }
         // Iterate the list and make the partitions unmodifiable
         for (HWDiskStore hwds : result) {
-            ((LinuxHWDiskStore) hwds).partitionList = Collections.unmodifiableList(hwds.getPartitions().stream()
-                    .sorted(Comparator.comparing(HWPartition::getName)).collect(Collectors.toList()));
+            ((LinuxHWDiskStore) hwds).partitionList = hwds.getPartitions().stream()
+                .sorted(Comparator.comparing(HWPartition::getName)).collect(Collectors.toUnmodifiableList());
         }
         return result;
     }
@@ -312,11 +312,11 @@ public final class LinuxHWDiskStore extends AbstractHWDiskStore {
     }
 
     private static String getPartitionNameForDmDevice(String vgName, String lvName) {
-        return new StringBuilder().append(DEV_LOCATION).append(vgName).append('/').append(lvName).toString();
+        return DEV_LOCATION + vgName + '/' + lvName;
     }
 
     private static String getMountPointOfDmDevice(String vgName, String lvName) {
-        return new StringBuilder().append(DEV_MAPPER).append(vgName).append('-').append(lvName).toString();
+        return DEV_MAPPER + vgName + '-' + lvName;
     }
 
     private static String getDependentNamesFromHoldersDirectory(String sysPath) {
--- a/oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
@@ -27,6 +27,9 @@ import static oshi.software.os.OSService.State.RUNNING;
 import static oshi.software.os.OSService.State.STOPPED;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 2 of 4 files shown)


oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
```diff
@@ -272,8 +272,8 @@ private static List<HWDiskStore> getDisks(LinuxHWDiskStore storeToUpdate) {
         }
         // Iterate the list and make the partitions unmodifiable
         for (HWDiskStore hwds : result) {
-            ((LinuxHWDiskStore) hwds).partitionList = Collections.unmodifiableList(hwds.getPartitions().stream()
```
adrian-kong, line 275: hmmm CI throwing errors not sure why, this is introduced since [10](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/stream/Collectors.html#toUnmodifiableList())

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -266,18 +270,19 @@ public List<OSService> getServices() {
         }
         // Get Directories for stopped services
         ArrayList<File> files = new ArrayList<>();
-        File dir = new File(SYSTEM_LIBRARY_LAUNCH_AGENTS);
-        if (dir.exists() && dir.isDirectory()) {
-            files.addAll(Arrays.asList(dir.listFiles((f, name) -> name.toLowerCase().endsWith(".plist"))));
-        } else {
+
+        try (Stream<Path> file = Files.list(Path.of(SYSTEM_LIBRARY_LAUNCH_AGENTS))) {
```
adrian-kong, line 274: as well as `Paths.of` since [11](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Path.html#of(java.lang.String,java.lang.String...))...

adrian-kong (commented):
weird ci errors, also having trouble compiling tests locally 
```
cannot access oshi.jna.platform.unix.AixLibc.AixLwpsInfo
```

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
dbwiddis: >  this is introduced since [10](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/stream/Collectors.html#toUnmodifiableList())

We still compile with compatibility for 8, at least until 31 Mar 2025.

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
dbwiddis: [Same answer as above.](https://endoflife.date/java)

dbwiddis:
> Think iterating
> 
> ```
> for (File f : dir.listFiles((f, name) -> name.toLowerCase().endsWith(".conf"))) {
> ```
> 
> potentially causes unhandled null pointer errors in
> 
> `LinuxOperatingSystem` and `MacOperatingSystem` which wraps `Arrays.asList(null)`

Based on the subject line "caused by empty directories" I'm not so sure it's an issue.  Per the javadoc for [`File.listFiles()`](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#listFiles()):
> The array will be empty if the directory is empty. 

The documented null returns:
> Returns null if this abstract pathname does not denote a directory, or if an I/O error occurs.

There's already an `isDirectory()` check and an `exists()` check should prevent I/O error.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxHWDiskStore.java
```diff
@@ -312,11 +312,11 @@ private static void computeDiskStats(LinuxHWDiskStore store, String devstat) {
     }
 
     private static String getPartitionNameForDmDevice(String vgName, String lvName) {
-        return new StringBuilder().append(DEV_LOCATION).append(vgName).append('/').append(lvName).toString();
+        return DEV_LOCATION + vgName + '/' + lvName;
```
dbwiddis, line 315: This compiles to the same thing.  I agree your version is more readable and don't mind the change as long as we're touching the file for other reasons...

dbwiddis (changes requested):
Not sure a fix is needed here.  Need to fix the JDK 8 compatibility issues before I'll review. :)

adrian-kong:
not needed

PR #2204, opened 2022-10-01
Improve performance using parallel streams for processes and threads

### LINKED ISSUE #2196: Improve processing speed with parallel streams
While doing some benchmarking on the FFM project I realized there are many places in OSHI that can benefit from simply adding `parallel()` to existing streams, particularly in the getProcess implementations.

This is a relatively simple code change with a potentially big payoff.

adrian-kong:
Not sure if this was done, 

Adds parallel on every stream call 
Fixes #2196 

@dbwiddis I'm not actually sure when parallel should be used according to
https://www.baeldung.com/java-when-to-use-parallel-stream
so I have never actually used this method, is it necessary to use `.parallel` on every stream call?

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
@@ -154,7 +154,7 @@ public abstract class AbstractNetworkIF implements NetworkIF {
         List<NetworkInterface> interfaces = getAllNetworkInterfaces();
 
         return includeLocalInterfaces ? interfaces
-                : getAllNetworkInterfaces().stream().filter(networkInterface1 -> !isLocalInterface(networkInterface1))
+                : getAllNetworkInterfaces().stream().parallel().filter(networkInterface1 -> !isLocalInterface(networkInterface1))
                         .collect(Collectors.toList());
     }
 
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
@@ -155,8 +155,8 @@ public abstract class AbstractNetworkIF implements NetworkIF {
         List<NetworkInterface> interfaces = getAllNetworkInterfaces();
 
         return includeLocalInterfaces ? interfaces
-                : getAllNetworkInterfaces().stream().parallel()
-                        .filter(ni -> !isLocalInterface(ni)).collect(Collectors.toList());
+                : getAllNetworkInterfaces().stream().parallel().filter(ni -> !isLocalInterface(ni))
+                        .collect(Collectors.toList());
     }
 
     /**
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
@@ -35,7 +35,6 @@ import java.util.Collections;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Properties;
-import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
@@ -35,6 +35,7 @@ import java.util.Collections;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Properties;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
@@ -155,7 +156,7 @@ public abstract class AbstractNetworkIF implements NetworkIF {
 
         return includeLocalInterfaces ? interfaces
                 : getAllNetworkInterfaces().stream().parallel()
-                        .filter(networkInterface1 -> !isLocalInterface(networkInterface1)).collect(Collectors.toList());
+                        .filter(Predicate.not(AbstractNetworkIF::isLocalInterface)).collect(Collectors.toList());
     }
 
     /**
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
@@ -154,8 +154,8 @@ public abstract class AbstractNetworkIF implements NetworkIF {
         List<NetworkInterface> interfaces = getAllNetworkInterfaces();
 
         return includeLocalInterfaces ? interfaces
-                : getAllNetworkInterfaces().stream().parallel().filter(networkInterface1 -> !isLocalInterface(networkInterface1))
-                        .collect(Collectors.toList());
+                : getAllNetworkInterfaces().stream().parallel()
+                        .filter(networkInterface1 -> !isLocalInterface(networkInterface1)).collect(Collectors.toList());
     }
 
     /**
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
@@ -1,7 +1,7 @@
 /*
  * MIT License
  *
- * Copyright (c) 2020-2021 The OSHI Project Contributors: https://github.com/oshi/oshi/graphs/contributors
+ * Copyright (c) 2020-2022 The OSHI Project Contributors: https://github.com/oshi/oshi/graphs/contributors
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
@@ -156,7 +156,7 @@ public abstract class AbstractNetworkIF implements NetworkIF {
 
         return includeLocalInterfaces ? interfaces
                 : getAllNetworkInterfaces().stream().parallel()
-                        .filter(Predicate.not(AbstractNetworkIF::isLocalInterface)).collect(Collectors.toList());
+                        .filter(ni -> !isLocalInterface(ni)).collect(Collectors.toList());
     }
 
     /**
--- a/oshi-core/src/main/java/oshi/software/os/OSThread.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSThread.java
@@ -25,10 +25,25 @@ package oshi.software.os;
 
 import oshi.software.os.OSProcess.State;
 
+import java.util.function.Predicate;
+
 /**
  * Represents a Thread/Task on the operating system.
  */
 public interface OSThread {
+
+    /**
+     * Constants which may be used to filter Thread lists in
+     */
+    final class ThreadFiltering {
+        private ThreadFiltering() {
+        }
+
+        /**
+         * Exclude processes with {@link State#INVALID} process state.
+         */
+        public static final Predicate<OSThread> VALID_THREAD = p -> !p.getState().equals(State.INVALID);
+    }
 
     /**
      * The thread id. The meaning of this value is OS-dependent.
--- a/oshi-core/src/main/java/oshi/software/os/OSThread.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSThread.java
@@ -33,7 +33,7 @@ import java.util.function.Predicate;
 public interface OSThread {
 
     /**
-     * Constants which may be used to filter Thread lists in
+     * Constants which may be used to filter Thread lists
      */
     final class ThreadFiltering {
         private ThreadFiltering() {
--- a/oshi-core/src/main/java/oshi/software/os/OSThread.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSThread.java
@@ -1,7 +1,7 @@
 /*
  * MIT License
  *
- * Copyright (c) 2020-2021 The OSHI Project Contributors: https://github.com/oshi/oshi/graphs/contributors
+ * Copyright (c) 2020-2022 The OSHI Project Contributors: https://github.com/oshi/oshi/graphs/contributors
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
@@ -247,7 +247,7 @@ public class LinuxOSProcess extends AbstractOSProcess {
 
     @Override
     public List<OSThread> getThreadDetails() {
-        return ProcessStat.getThreadIds(getProcessID()).stream().map(id -> new LinuxOSThread(getProcessID(), id))
+        return ProcessStat.getThreadIds(getProcessID()).stream().parallel().map(id -> new LinuxOSThread(getProcessID(), id))
                 .collect(Collectors.toList());
     }
 
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
@@ -247,8 +247,8 @@ public class LinuxOSProcess extends AbstractOSProcess {
 
     @Override
     public List<OSThread> getThreadDetails() {
-        return ProcessStat.getThreadIds(getProcessID()).stream().parallel().map(id -> new LinuxOSThread(getProcessID(), id))
-                .collect(Collectors.toList());
+        return ProcessStat.getThreadIds(getProcessID()).stream().parallel()
+                .map(id -> new LinuxOSThread(getProcessID(), id)).collect(Collectors.toList());
     }
 
     @Override
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
@@ -24,6 +24,7 @@
 package oshi.software.os.linux;
 
 import static oshi.software.os.OSProcess.State.INVALID;
+import static oshi.software.os.OSThread.ThreadFiltering.VALID_THREAD;
 import static oshi.util.Memoizer.memoize;
 
 import java.io.File;
@@ -248,7 +249,7 @@ public class LinuxOSProcess extends AbstractOSProcess {
     @Override
     public List<OSThread> getThreadDetails() {
         return ProcessStat.getThreadIds(getProcessID()).stream().parallel()
-                .map(id -> new LinuxOSThread(getProcessID(), id)).collect(Collectors.toList());
+                .map(id -> new LinuxOSThread(getProcessID(), id)).filter(VALID_THREAD).collect(Collectors.toList());
     }
 
     @Override
--- a/oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
@@ -30,6 +30,7 @@ import static oshi.software.os.OSProcess.State.SLEEPING;
 import static oshi.software.os.OSProcess.State.STOPPED;
 import static oshi.software.os.OSProcess.State.WAITING;
 import static oshi.software.os.OSProcess.State.ZOMBIE;
+import static oshi.software.os.OSThread.ThreadFiltering.VALID_THREAD;
 import static oshi.util.Memoizer.defaultExpiration;
 import static oshi.util.Memoizer.memoize;
 
@@ -38,7 +39,10 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.*;
+import java.util.List;
+import java.util.Collections;
+import java.util.Arrays;
+import java.util.Map;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -298,7 +302,7 @@ public class AixOSProcess extends AbstractOSProcess {
 
         return Arrays.stream(numericFiles).parallel()
                 .map(lwpidFile -> new AixOSThread(getProcessID(), ParseUtil.parseIntOrDefault(lwpidFile.getName(), 0)))
-                .filter(thread -> thread.getState() != INVALID).collect(Collectors.toList());
+                .filter(VALID_THREAD).collect(Collectors.toList());
     }
 
     @Override
--- a/oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
@@ -38,10 +38,9 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.slf4j.Logger;
@@ -290,24 +289,16 @@ public class AixOSProcess extends AbstractOSProcess {
 
     @Override
     public List<OSThread> getThreadDetails() {
-        List<OSThread> threads = new ArrayList<>();
-
         // Get process files in proc
         File directory = new File(String.format("/proc/%d/lwp", getProcessID()));
         File[] numericFiles = directory.listFiles(file -> Constants.DIGITS.matcher(file.getName()).matches());
         if (numericFiles == null) {
-            return threads;
+            return Collections.emptyList();
         }
 
-        // Iterate files
-        for (File lwpidFile : numericFiles) {
-            int lwpidNum = ParseUtil.parseIntOrDefault(lwpidFile.getName(), 0);
-            OSThread thread = new AixOSThread(getProcessID(), lwpidNum);
-            if (thread.getState() != INVALID) {
-                threads.add(thread);
-            }
-        }
-        return threads;
+        return Arrays.stream(numericFiles).parallel()
+            .map(lwpidFile -> new AixOSThread(getProcessID(), ParseUtil.parseIntOrDefault(lwpidFile.getName(), 0)))
+            .filter(thread -> thread.getState() != INVALID).collect(Collectors.toList());
     }
 
     @Override
--- a/oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
@@ -297,8 +297,8 @@ public class AixOSProcess extends AbstractOSProcess {
         }
 
         return Arrays.stream(numericFiles).parallel()
-            .map(lwpidFile -> new AixOSThread(getProcessID(), ParseUtil.parseIntOrDefault(lwpidFile.getName(), 0)))
-            .filter(thread -> thread.getState() != INVALID).collect(Collectors.toList());
+                .map(lwpidFile -> new AixOSThread(getProcessID(), ParseUtil.parseIntOrDefault(lwpidFile.getName(), 0)))
+                .filter(thread -> thread.getState() != INVALID).collect(Collectors.toList());
     }
 
     @Override
--- a/oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
@@ -30,6 +30,7 @@ import static oshi.software.os.OSProcess.State.SLEEPING;
 import static oshi.software.os.OSProcess.State.STOPPED;
 import static oshi.software.os.OSProcess.State.WAITING;
 import static oshi.software.os.OSProcess.State.ZOMBIE;
+import static oshi.software.os.OSThread.ThreadFiltering.VALID_THREAD;
 import static oshi.util.Memoizer.memoize;
 
 import java.util.ArrayList;
@@ -337,10 +338,11 @@ public class FreeBsdOSProcess extends AbstractOSProcess {
         if (getProcessID() >= 0) {
             psCommand += " -p " + getProcessID();
         }
-        Predicate<Map<PsThreadColumns, String>> columnPRI = threadMap -> threadMap.containsKey(PsThreadColumns.PRI);
+        Predicate<Map<PsThreadColumns, String>> hasColumnsPri = threadMap -> threadMap.containsKey(PsThreadColumns.PRI);
         return ExecutingCommand.runNative(psCommand).stream().skip(1).parallel()
-                .map(thread -> ParseUtil.stringToEnumMap(PsThreadColumns.class, thread.trim(), ' ')).filter(columnPRI)
-                .map(threadMap -> new FreeBsdOSThread(getProcessID(), threadMap)).collect(Collectors.toList());
+                .map(thread -> ParseUtil.stringToEnumMap(PsThreadColumns.class, thread.trim(), ' '))
+                .filter(hasColumnsPri).map(threadMap -> new FreeBsdOSThread(getProcessID(), threadMap))
+                .filter(VALID_THREAD).collect(Collectors.toList());
     }
 
     @Override
--- a/oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
@@ -332,25 +332,15 @@ public class FreeBsdOSProcess extends AbstractOSProcess {
 
     @Override
     public List<OSThread> getThreadDetails() {
-        List<OSThread> threads = new ArrayList<>();
         String psCommand = "ps -awwxo " + PS_THREAD_COLUMNS + " -H";
         if (getProcessID() >= 0) {
             psCommand += " -p " + getProcessID();
         }
-        List<String> threadList = ExecutingCommand.runNative(psCommand);
-        if (threadList.size() > 1) {
-            // remove header row
-            threadList.remove(0);
-            // Fill list
-            for (String thread : threadList) {
-                Map<PsThreadColumns, String> threadMap = ParseUtil.stringToEnumMap(PsThreadColumns.class, thread.trim(),
-                        ' ');
-                if (threadMap.containsKey(PsThreadColumns.PRI)) {
-                    threads.add(new FreeBsdOSThread(getProcessID(), threadMap));
-                }
-            }
-        }
-        return threads;
+        return ExecutingCommand.runNative(psCommand).stream().skip(1).parallel()
+            .map(thread -> ParseUtil.stringToEnumMap(PsThreadColumns.class, thread.trim(), ' '))
+            .filter(threadMap -> threadMap.containsKey(PsThreadColumns.PRI))
+            .map(threadMap -> new FreeBsdOSThread(getProcessID(), threadMap))
+            .collect(Collectors.toList());
     }
 
     @Override
--- a/oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
@@ -37,6 +37,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
@@ -336,9 +337,10 @@ public class FreeBsdOSProcess extends AbstractOSProcess {
         if (getProcessID() >= 0) {
             psCommand += " -p " + getProcessID();
         }
+        Predicate<Map<PsThreadColumns, String>> columnPRI = threadMap -> threadMap.containsKey(PsThreadColumns.PRI);
         return ExecutingCommand.runNative(psCommand).stream().skip(1).parallel()
                 .map(thread -> ParseUtil.stringToEnumMap(PsThreadColumns.class, thread.trim(), ' '))
-                .filter(threadMap -> threadMap.containsKey(PsThreadColumns.PRI))
+                .filter(columnPRI)
                 .map(threadMap -> new FreeBsdOSThread(getProcessID(), threadMap)).collect(Collectors.toList());
     }
 
--- a/oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
@@ -337,10 +337,9 @@ public class FreeBsdOSProcess extends AbstractOSProcess {
             psCommand += " -p " + getProcessID();
         }
         return ExecutingCommand.runNative(psCommand).stream().skip(1).parallel()
-            .map(thread -> ParseUtil.stringToEnumMap(PsThreadColumns.class, thread.trim(), ' '))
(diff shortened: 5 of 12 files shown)


oshi-core/src/main/java/oshi/driver/mac/ThreadInfo.java
```diff
@@ -59,8 +59,8 @@ public static List<ThreadStats> queryTaskThreads(int pid) {
         List<ThreadStats> taskThreads = new ArrayList<>();
         // Only way to get thread info without root permissions
         // Using the M switch gives all threads with no possibility to filter
-        List<String> psThread = ExecutingCommand.runNative("ps -awwxM").stream().filter(s -> s.contains(pidStr))
-                .collect(Collectors.toList());
+        List<String> psThread = ExecutingCommand.runNative("ps -awwxM").stream().parallel()
+                .filter(s -> s.contains(pidStr)).collect(Collectors.toList());
```
dbwiddis, line 63: Probably not needed here but doesn't hurt, we're just filtering a list of strings and you'd really have to benchmark this to see if the parallel thread overhead beats the savings from simultaneously doing hundreds of `String.contains()` checks (which are pretty fast).

oshi-core/src/main/java/oshi/driver/unix/freebsd/disk/GeomPartList.java
```diff
@@ -133,7 +133,7 @@ public static Map<String, List<HWPartition>> queryPartitions() {
             }
             // Process last diskstore
             if (!partList.isEmpty()) {
-                partList = partList.stream().sorted(Comparator.comparing(HWPartition::getName))
+                partList = partList.stream().parallel().sorted(Comparator.comparing(HWPartition::getName))
```
dbwiddis, line 136: Doesn't actually do anything here, as the sorting list needs the whole stream anyway.

oshi-core/src/main/java/oshi/driver/windows/DeviceTree.java
```diff
@@ -145,8 +146,8 @@ public static Quintet<Set<Integer>, Map<Integer, Integer>, Map<Integer, String>,
             }
         }
         // Look for output without parents, these are top of tree
-        Set<Integer> controllerDevices = deviceIdMap.keySet().stream().filter(k -> !parentMap.containsKey(k))
-                .collect(Collectors.toSet());
+        Set<Integer> controllerDevices = deviceIdMap.keySet().stream().parallel()
+                .filter(Predicate.not(parentMap::containsKey)).collect(Collectors.toSet());
```
dbwiddis, line 150: This processes a filter in parallel but given it's just a hashmap lookup, probably no performance benefit (particularly for a small number of devices... 2 or 3 usually.)

oshi-core/src/main/java/oshi/driver/windows/wmi/Win32Process.java
```diff
@@ -72,7 +72,7 @@ public static WmiResult<CommandLineProperty> queryCommandLines(Set<Integer> pids
         String sb = WIN32_PROCESS;
         if (pidsToQuery != null) {
             sb += " WHERE ProcessID="
-                    + pidsToQuery.stream().map(String::valueOf).collect(Collectors.joining(" OR PROCESSID="));
+                    + pidsToQuery.stream().parallel().map(String::valueOf).collect(Collectors.joining(" OR PROCESSID="));
```
dbwiddis, line 75: Any performance benefit here is dwarfed by the resulting WMI call overhead.  Probably doesn't hurt, but I doubt it's a noticeable help.

oshi-core/src/main/java/oshi/driver/windows/wmi/Win32Process.java
```diff
@@ -89,7 +89,7 @@ public static WmiResult<ProcessXPProperty> queryProcesses(Collection<Integer> pi
         String sb = WIN32_PROCESS;
         if (pids != null) {
             sb += " WHERE ProcessID="
-                    + pids.stream().map(String::valueOf).collect(Collectors.joining(" OR PROCESSID="));
+                    + pids.stream().parallel().map(String::valueOf).collect(Collectors.joining(" OR PROCESSID="));
```
dbwiddis, line 92: Same as above. I'm not sure whether a random sort order of the process ID filters would make any difference on query speed.  Probably not needed.

oshi-core/src/main/java/oshi/hardware/common/AbstractCentralProcessor.java
```diff
@@ -87,9 +87,9 @@ protected AbstractCentralProcessor() {
             Set<Integer> pkgCoreKeys = this.logicalProcessors.stream()
                     .map(p -> (p.getPhysicalPackageNumber() << 16) + p.getPhysicalProcessorNumber())
                     .collect(Collectors.toSet());
-            List<PhysicalProcessor> physProcs = pkgCoreKeys.stream().sorted()
-                    .map(k -> new PhysicalProcessor(k >> 16, k & 0xffff)).collect(Collectors.toList());
-            this.physicalProcessors = Collections.unmodifiableList(physProcs);
+            this.physicalProcessors = pkgCoreKeys.stream().parallel().sorted()
+                .map(k -> new PhysicalProcessor(k >> 16, k & 0xffff))
```
dbwiddis, line 91: Not needed, sorted gets everything back into a single thread where order matters.

oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
```diff
@@ -154,7 +154,7 @@ protected static List<NetworkInterface> getNetworkInterfaces(boolean includeLoca
         List<NetworkInterface> interfaces = getAllNetworkInterfaces();
 
         return includeLocalInterfaces ? interfaces
-                : getAllNetworkInterfaces().stream().filter(networkInterface1 -> !isLocalInterface(networkInterface1))
+                : getAllNetworkInterfaces().stream().parallel().filter(networkInterface1 -> !isLocalInterface(networkInterface1))
```
dbwiddis, line 157: This is a good one.  See #1606 which noted the `isLocalInterface()` check consumed 1/3 of OSHI's CPU usage in profiling testing.  PR #1610 cut that in half by removing redundancy, but parallelizing it will shorten it even more!

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
```diff
@@ -191,7 +191,7 @@ protected Pair<List<LogicalProcessor>, List<PhysicalProcessor>> initProcessorCou
         // Sort
         logProcs.sort(Comparator.comparingInt(LogicalProcessor::getProcessorNumber));
 
-        List<PhysicalProcessor> physProcs = coreEfficiencyMap.entrySet().stream().sorted(Map.Entry.comparingByKey())
+        List<PhysicalProcessor> physProcs = coreEfficiencyMap.entrySet().stream().parallel().sorted(Map.Entry.comparingByKey())
```
dbwiddis, line 194: No need to parallel before sorting (see above)

oshi-core/src/main/java/oshi/hardware/platform/mac/MacCentralProcessor.java
```diff
@@ -143,7 +143,7 @@ protected Pair<List<LogicalProcessor>, List<PhysicalProcessor>> initProcessorCou
             pkgCoreKeys.add((pkgId << 16) + coreId);
         }
         Map<Integer, String> compatMap = queryCompatibleStrings();
-        List<PhysicalProcessor> physProcs = pkgCoreKeys.stream().sorted().map(k -> {
+        List<PhysicalProcessor> physProcs = pkgCoreKeys.stream().parallel().sorted().map(k -> {
```
dbwiddis, line 146: No parallel+sort :)

oshi-core/src/main/java/oshi/hardware/platform/mac/MacCentralProcessor.java
```diff
@@ -182,9 +182,9 @@ public long[] querySystemCpuLoadTicks() {
     public long[] queryCurrentFreq() {
         if (isArmCpu) {
             Map<Integer, Long> physFreqMap = new HashMap<>();
-            getPhysicalProcessors().stream().forEach(p -> physFreqMap.put(p.getPhysicalProcessorNumber(),
+            getPhysicalProcessors().stream().parallel().forEach(p -> physFreqMap.put(p.getPhysicalProcessorNumber(),
```
dbwiddis, line 185: Probably doesn't hurt but we're just parallelizing inserting into a map, and a relatively small number of physical processors (4 to 8 on most modern machines) so likely not needed.

oshi-core/src/main/java/oshi/hardware/platform/mac/MacCentralProcessor.java
```diff
@@ -182,9 +182,9 @@ public long[] querySystemCpuLoadTicks() {
...
                     p.getEfficiency() > 0 ? performanceCoreFrequency : efficiencyCoreFrequency));
-            return getLogicalProcessors().stream().map(LogicalProcessor::getPhysicalProcessorNumber)
+            return getLogicalProcessors().stream().parallel().map(LogicalProcessor::getPhysicalProcessorNumber)
```
dbwiddis, line 187: NOPE. The `getLogicalProcessors()` is in a particular order and the output of this method has to match.

oshi-core/src/main/java/oshi/hardware/platform/mac/MacCentralProcessor.java
```diff
@@ -302,7 +302,7 @@ private static Map<Integer, String> queryCompatibleStrings() {
     // populated physical processors
     private boolean isArmCpu() {
         // M1 / M2 chips will have an efficiency > 0
-        return getPhysicalProcessors().stream().map(PhysicalProcessor::getEfficiency).anyMatch(e -> e > 0);
+        return getPhysicalProcessors().stream().parallel().map(PhysicalProcessor::getEfficiency).anyMatch(e -> e > 0);
```
dbwiddis, line 305: The small number of processors means this probably doesn't gain anything.

oshi-core/src/main/java/oshi/hardware/platform/unix/UnixDisplay.java
```diff
@@ -53,6 +53,6 @@ public final class UnixDisplay extends AbstractDisplay {
      * @return An array of Display objects representing monitors, etc.
      */
     public static List<Display> getDisplays() {
-        return Xrandr.getEdidArrays().stream().map(UnixDisplay::new).collect(Collectors.toList());
+        return Xrandr.getEdidArrays().stream().parallel().map(UnixDisplay::new).collect(Collectors.toList());
```
dbwiddis, line 56: This could in theory improve performance but at the cost of giving users a random order of their displays each time they ran the method.  I'd vote no.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsUsbDevice.java
```diff
@@ -133,7 +133,7 @@ private static WindowsUsbDevice queryDeviceAndChildren(Integer device, Map<Integ
             }
         }
         // Iterate the parent map looking for children
-        Set<Integer> childDeviceSet = parentMap.entrySet().stream().filter(e -> e.getValue().equals(device))
+        Set<Integer> childDeviceSet = parentMap.entrySet().stream().parallel().filter(e -> e.getValue().equals(device))
```
dbwiddis, line 136: Filtering a map entryset in parallel could slightly improve performance but this would result in a random order of USB device listing each time someone ran it, so probably not worth it.

oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
```diff
@@ -103,7 +103,7 @@ private int queryPlatformBitness() {
 
     @Override
     public List<OSProcess> getProcesses(Predicate<OSProcess> filter, Comparator<OSProcess> sort, int limit) {
-        return queryAllProcesses().stream().filter(filter == null ? ALL_PROCESSES : filter)
+        return queryAllProcesses().stream().parallel().distinct().filter(filter == null ? ALL_PROCESSES : filter)
```
dbwiddis, line 106: put distinct before parallel or you lose the benefits.

But in this case you shouldn't use it as there's a limit at the end and random order (possible with no sorting option) would lead to nondeterministic results.

oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
```diff
@@ -116,11 +116,11 @@ public List<OSProcess> getChildProcesses(int parentPid, Predicate<OSProcess> fil
         // Get this pid and its children
         List<OSProcess> childProcs = queryChildProcesses(parentPid);
         // Extract the parent from the list
-        OSProcess parent = childProcs.stream().filter(p -> p.getProcessID() == parentPid).findAny().orElse(null);
+        OSProcess parent = childProcs.stream().parallel().filter(p -> p.getProcessID() == parentPid).findAny().orElse(null);
```
dbwiddis, line 119: In theory could improve performance, finding any in parallel... vs. finding any in a single thread... but may end up doing extra work too.  Would have to benchmark this.  It's just an object getter though, and a relatively small number (child processes are usually in the single digits) so I'd say no here.

oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
```diff
@@ -116,11 +116,11 @@ public List<OSProcess> getChildProcesses(int parentPid, Predicate<OSProcess> fil
...
         // Get the parent's start time
         long parentStartTime = parent == null ? 0 : parent.getStartTime();
         // Get children after parent
-        return queryChildProcesses(parentPid).stream().filter(filter == null ? ALL_PROCESSES : filter)
+        return queryChildProcesses(parentPid).stream().parallel().filter(filter == null ? ALL_PROCESSES : filter)
```
dbwiddis, line 123: Probably very little value in the parallel filters, and concern about nondeterministic results with a limit at the end.

oshi-core/src/main/java/oshi/software/common/AbstractOperatingSystem.java
```diff
@@ -134,11 +134,11 @@ public List<OSProcess> getDescendantProcesses(int parentPid, Predicate<OSProcess
         // Get this pid and its descendants
         List<OSProcess> descendantProcs = queryDescendantProcesses(parentPid);
         // Extract the parent from the list
-        OSProcess parent = descendantProcs.stream().filter(p -> p.getProcessID() == parentPid).findAny().orElse(null);
+        OSProcess parent = descendantProcs.stream().parallel().filter(p -> p.getProcessID() == parentPid).findAny().orElse(null);
```
dbwiddis, line 137: Similar to child process answer above. Just parallelizing a getter of a relatively small number, and finding any... single thread likely to stop more quickly when it finds one than trying to coordinate multiple threads which may each run until they find one and then have to talk to each other.....

oshi-core/src/main/java/oshi/software/os/OperatingSystem.java
```diff
@@ -213,7 +213,7 @@ default List<OSProcess> getProcesses() {
      *         specified process ids if it is running
      */
     default List<OSProcess> getProcesses(Collection<Integer> pids) {
-        return pids.stream().map(this::getProcess).filter(Objects::nonNull).filter(ProcessFiltering.VALID_PROCESS)
+        return pids.stream().parallel().map(this::getProcess).filter(Objects::nonNull).filter(ProcessFiltering.VALID_PROCESS)
```
dbwiddis, line 216: Probably doesn't hurt, but also would give nondeterministic ordering of results when a user would expect the results in the same order of a list (if a list was the input).  So no.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -247,7 +247,7 @@ public long getBytesWritten() {
 
     @Override
     public List<OSThread> getThreadDetails() {
-        return ProcessStat.getThreadIds(getProcessID()).stream().map(id -> new LinuxOSThread(getProcessID(), id))
+        return ProcessStat.getThreadIds(getProcessID()).stream().parallel().map(id -> new LinuxOSThread(getProcessID(), id))
```
dbwiddis, line 250: YES.  This is the primary use case envisioned by #2196.   the new thread call is heavy so this will help.   Not sure why we aren't sorting these, though...

oshi-core/src/main/java/oshi/software/os/mac/MacFileSystem.java
```diff
@@ -204,7 +204,7 @@ private static List<OSFileStore> getFileStoreMatching(String nameToMatch, boolea
                     }
 
                     StringBuilder options = new StringBuilder((MNT_RDONLY & flags) == 0 ? "rw" : "ro");
-                    String moreOptions = OPTIONS_MAP.entrySet().stream().filter(e -> (e.getKey() & flags) > 0)
+                    String moreOptions = OPTIONS_MAP.entrySet().stream().parallel().filter(e -> (e.getKey() & flags) > 0)
```
dbwiddis, line 207: too few options to gain much benefit here.

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -199,14 +199,14 @@ public OSProcess getProcess(int pid) {
     public List<OSProcess> queryChildProcesses(int parentPid) {
         List<OSProcess> allProcs = queryAllProcesses();
         Set<Integer> descendantPids = getChildrenOrDescendants(allProcs, parentPid, false);
-        return allProcs.stream().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
+        return allProcs.stream().parallel().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
```
dbwiddis, line 202: Very little benefit if any.

oshi-core/src/main/java/oshi/software/os/mac/MacOperatingSystem.java
```diff
@@ -199,14 +199,14 @@ public OSProcess getProcess(int pid) {
...
     }
 
     @Override
     public List<OSProcess> queryDescendantProcesses(int parentPid) {
         List<OSProcess> allProcs = queryAllProcesses();
         Set<Integer> descendantPids = getChildrenOrDescendants(allProcs, parentPid, true);
-        return allProcs.stream().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
+        return allProcs.stream().parallel().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
```
dbwiddis, line 209: very little benefit if any.

oshi-core/src/main/java/oshi/software/os/unix/aix/AixOperatingSystem.java
```diff
@@ -130,14 +130,14 @@ public List<OSProcess> queryAllProcesses() {
     public List<OSProcess> queryChildProcesses(int parentPid) {
         List<OSProcess> allProcs = queryAllProcesses();
         Set<Integer> descendantPids = getChildrenOrDescendants(allProcs, parentPid, false);
-        return allProcs.stream().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
+        return allProcs.stream().parallel().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
```
dbwiddis, line 133: very little benefit if any

oshi-core/src/main/java/oshi/software/os/unix/aix/AixOperatingSystem.java
```diff
@@ -130,14 +130,14 @@ public List<OSProcess> queryAllProcesses() {
...
     }
 
     @Override
     public List<OSProcess> queryDescendantProcesses(int parentPid) {
         List<OSProcess> allProcs = queryAllProcesses();
         Set<Integer> descendantPids = getChildrenOrDescendants(allProcs, parentPid, true);
-        return allProcs.stream().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
+        return allProcs.stream().parallel().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
```
dbwiddis, line 140: very little benefit if any

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOperatingSystem.java
```diff
@@ -127,14 +127,14 @@ public List<OSProcess> queryAllProcesses() {
     public List<OSProcess> queryChildProcesses(int parentPid) {
         List<OSProcess> allProcs = queryAllProcesses();
         Set<Integer> descendantPids = getChildrenOrDescendants(allProcs, parentPid, false);
-        return allProcs.stream().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
+        return allProcs.stream().parallel().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
```
dbwiddis, line 130: very little benefit if any (repeating myself, all of these are parallelizing set.contains.... which is O(1) fast)

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOperatingSystem.java
```diff
@@ -127,14 +127,14 @@ public List<OSProcess> queryAllProcesses() {
...
     }
 
     @Override
     public List<OSProcess> queryDescendantProcesses(int parentPid) {
         List<OSProcess> allProcs = queryAllProcesses();
         Set<Integer> descendantPids = getChildrenOrDescendants(allProcs, parentPid, true);
-        return allProcs.stream().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
+        return allProcs.stream().parallel().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
```
dbwiddis, line 137: ditto

oshi-core/src/main/java/oshi/software/os/unix/openbsd/OpenBsdOperatingSystem.java
```diff
@@ -124,14 +124,14 @@ public List<OSProcess> queryAllProcesses() {
     public List<OSProcess> queryChildProcesses(int parentPid) {
         List<OSProcess> allProcs = queryAllProcesses();
         Set<Integer> descendantPids = getChildrenOrDescendants(allProcs, parentPid, false);
-        return allProcs.stream().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
+        return allProcs.stream().parallel().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
```
dbwiddis, line 127: more set contains in parallel

oshi-core/src/main/java/oshi/software/os/unix/openbsd/OpenBsdOperatingSystem.java
```diff
@@ -124,14 +124,14 @@ public List<OSProcess> queryAllProcesses() {
...
     }
 
     @Override
     public List<OSProcess> queryDescendantProcesses(int parentPid) {
         List<OSProcess> allProcs = queryAllProcesses();
         Set<Integer> descendantPids = getChildrenOrDescendants(allProcs, parentPid, true);
-        return allProcs.stream().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
+        return allProcs.stream().parallel().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
```
dbwiddis, line 134: I think I'll stop commenting on everyone of these :)

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOperatingSystem.java
```diff
@@ -144,14 +144,14 @@ public List<OSProcess> queryAllProcesses() {
     public List<OSProcess> queryChildProcesses(int parentPid) {
         List<OSProcess> allProcs = queryAllProcessesFromPrStat();
         Set<Integer> descendantPids = getChildrenOrDescendants(allProcs, parentPid, false);
-        return allProcs.stream().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
+        return allProcs.stream().parallel().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
```
dbwiddis, line 147: but there's only 2 more

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOperatingSystem.java
```diff
@@ -144,14 +144,14 @@ public List<OSProcess> queryAllProcesses() {
...
     }
 
     @Override
     public List<OSProcess> queryDescendantProcesses(int parentPid) {
         List<OSProcess> allProcs = queryAllProcessesFromPrStat();
         Set<Integer> descendantPids = getChildrenOrDescendants(allProcs, parentPid, true);
-        return allProcs.stream().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
+        return allProcs.stream().parallel().filter(p -> descendantPids.contains(p.getProcessID())).collect(Collectors.toList());
```
dbwiddis, line 154: and down to 1

oshi-core/src/main/java/oshi/software/os/windows/WindowsFileSystem.java
```diff
@@ -204,7 +204,7 @@ static ArrayList<OSFileStore> getLocalVolumes(String volumeToMatch) {
                     strFsType = Native.toString(fstype);
 
                     StringBuilder options = new StringBuilder((FILE_READ_ONLY_VOLUME & flags) == 0 ? "rw" : "ro");
-                    String moreOptions = OPTIONS_MAP.entrySet().stream().filter(e -> (e.getKey() & flags) > 0)
+                    String moreOptions = OPTIONS_MAP.entrySet().stream().parallel().filter(e -> (e.getKey() & flags) > 0)
```
dbwiddis, line 207: too few options to gain benefit here.

dbwiddis (changes requested):
> Not sure if this was done,
> 
> Adds parallel on every stream call #2196
> 
> @dbwiddis I'm not actually sure when parallel should be used according to https://www.baeldung.com/java-when-to-use-parallel-stream so I have never actually used this method, is it necessary to use `.parallel` on every stream call?

No, it's not necessary, or even desired to use it on every stream.  From a performance perspective (the main reason for #2196) we want to balance the additional overhead of spawning/managing the streams, vs. the faster time from the parallel processing.  And we don't want to use it in any place that order maters.

I've looked at every change and the vast majority aren't needed. Some are good.  But you missed the primary reason for #2196 by only looking at existing streams. 

For the below:
 - keep the one for network local interface check
 - (probably) keep the one for thread details, but there should probably also be ones for the other OS implementations

For things not included:
 - please focus on methods that instantiate (a lot of) new OSProcess or new OSThread objects.   Look at every getProcess() implementation on every OS.  Some of them are in lists which could be turned into streams and parallelized.
   - also look into the `getThreadDetails()` implementations for the same reason. You found one... look at the others.

oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
```diff
@@ -38,10 +38,9 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
dbwiddis, line 41: No star imports.

oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
```diff
@@ -290,24 +289,16 @@ public long getAffinityMask() {
...
-            int lwpidNum = ParseUtil.parseIntOrDefault(lwpidFile.getName(), 0);
-            OSThread thread = new AixOSThread(getProcessID(), lwpidNum);
-            if (thread.getState() != INVALID) {
-                threads.add(thread);
-            }
-        }
-        return threads;
+        return Arrays.stream(numericFiles).parallel()
+            .map(lwpidFile -> new AixOSThread(getProcessID(), ParseUtil.parseIntOrDefault(lwpidFile.getName(), 0)))
+            .filter(thread -> thread.getState() != INVALID).collect(Collectors.toList());
```
dbwiddis: Nice job with the stream.  

We're re-using this invalid thread state check in multiple classes, so it would make sense to make it a constant on either the AbstractOSProcess parent or make it part of the API on OSProcess, similar to `OperatingSystem.VALID_PROCESS`.   Probably don't need a whole ThreadSorting and ThreadFiltering set of classes, though. WDYT?

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
```diff
@@ -332,25 +332,15 @@ private int queryBitness() {
...
-                        ' ');
-                if (threadMap.containsKey(PsThreadColumns.PRI)) {
-                    threads.add(new FreeBsdOSThread(getProcessID(), threadMap));
-                }
-            }
-        }
-        return threads;
+        return ExecutingCommand.runNative(psCommand).stream().skip(1).parallel()
+            .map(thread -> ParseUtil.stringToEnumMap(PsThreadColumns.class, thread.trim(), ' '))
+            .filter(threadMap -> threadMap.containsKey(PsThreadColumns.PRI))
```
dbwiddis: This one's local to the class, but the purpose here is to make sure we get all the columns so extracting this predicate to a constant would improve readability.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOSProcess.java
```diff
@@ -38,10 +38,9 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
dbwiddis, line 41: No `*` :)

dbwiddis (changes requested):
Great job, I love the improved readability and the performance will be good too.

Will approve with removal of star imports, but consider making the repeated valid thread check a constnat.

dbwiddis:
failing license check CI is due to copyright change date. Run mvn license:format to fix.

oshi-core/src/main/java/oshi/software/os/unix/aix/AixOSProcess.java
adrian-kong: sounds good! I copied the format of how we have in `OperatingSystem` placed the constants in interface `OSThread` so it would kinda look the same, (but that way, seems like we have to keep the classes in there just like `OperatingSystem` tho...)

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -247,8 +247,8 @@ public long getBytesWritten() {
 
     @Override
     public List<OSThread> getThreadDetails() {
-        return ProcessStat.getThreadIds(getProcessID()).stream().map(id -> new LinuxOSThread(getProcessID(), id))
-                .collect(Collectors.toList());
+        return ProcessStat.getThreadIds(getProcessID()).stream().parallel()
+                .map(id -> new LinuxOSThread(getProcessID(), id)).collect(Collectors.toList());
```
dbwiddis, line 251: We can probably add that valid thread filter here.

oshi-core/src/main/java/oshi/software/os/OSThread.java
```diff
@@ -25,11 +25,26 @@
...
 
+import java.util.function.Predicate;
+
 /**
  * Represents a Thread/Task on the operating system.
  */
 public interface OSThread {
 
+    /**
+     * Constants which may be used to filter Thread lists in
```
dbwiddis, line 36: in ... ? :-)

oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
```diff
@@ -154,8 +154,8 @@ protected static List<NetworkInterface> getNetworkInterfaces(boolean includeLoca
         List<NetworkInterface> interfaces = getAllNetworkInterfaces();
 
         return includeLocalInterfaces ? interfaces
-                : getAllNetworkInterfaces().stream().filter(networkInterface1 -> !isLocalInterface(networkInterface1))
-                        .collect(Collectors.toList());
+                : getAllNetworkInterfaces().stream().parallel()
+                        .filter(networkInterface1 -> !isLocalInterface(networkInterface1)).collect(Collectors.toList());
```
dbwiddis, line 158: Huh, that `networkInterface1` snuck in at some point, probably due to a refactoring. I like shorter names in lambdas, though, why not [`ni`](https://www.youtube.com/watch?v=zIV4poUZAQo)?

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
```diff
@@ -332,25 +333,14 @@ private int queryBitness() {
...
-                if (threadMap.containsKey(PsThreadColumns.PRI)) {
-                    threads.add(new FreeBsdOSThread(getProcessID(), threadMap));
-                }
-            }
-        }
-        return threads;
+        Predicate<Map<PsThreadColumns, String>> columnPRI = threadMap -> threadMap.containsKey(PsThreadColumns.PRI);
+        return ExecutingCommand.runNative(psCommand).stream().skip(1).parallel()
+                .map(thread -> ParseUtil.stringToEnumMap(PsThreadColumns.class, thread.trim(), ' ')).filter(columnPRI)
+                .map(threadMap -> new FreeBsdOSThread(getProcessID(), threadMap)).collect(Collectors.toList());
```
dbwiddis, line 343: Another place to insert valid thread.

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
```diff
@@ -332,25 +333,14 @@ private int queryBitness() {
...
-            for (String thread : threadList) {
-                Map<PsThreadColumns, String> threadMap = ParseUtil.stringToEnumMap(PsThreadColumns.class, thread.trim(),
-                        ' ');
-                if (threadMap.containsKey(PsThreadColumns.PRI)) {
-                    threads.add(new FreeBsdOSThread(getProcessID(), threadMap));
-                }
-            }
-        }
-        return threads;
+        Predicate<Map<PsThreadColumns, String>> columnPRI = threadMap -> threadMap.containsKey(PsThreadColumns.PRI);
```
dbwiddis, line 340: (Preference) let's name this for the reason it exists, e.g., `hasAllPSColumns`

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOperatingSystem.java
```diff
@@ -147,27 +148,17 @@ public OSProcess getProcess(int pid) {
...
-            }
-        }
-        return procs;
+
+        Predicate<Map<PsKeywords, String>> keywordARGS = psMap -> psMap.containsKey(PsKeywords.ARGS);
+        return ExecutingCommand.runNative(psCommand).stream().skip(1).parallel()
+                .map(proc -> ParseUtil.stringToEnumMap(PsKeywords.class, proc.trim(), ' ')).filter(keywordARGS)
+                .map(psMap -> new FreeBsdOSProcess(
+                        pid < 0 ? ParseUtil.parseIntOrDefault(psMap.get(PsKeywords.PID), 0) : pid, psMap))
+                .collect(Collectors.toList());
```
dbwiddis, line 161: Another case for valid process filter, and rename the hasall columns predicate.

oshi-core/src/main/java/oshi/software/os/unix/openbsd/OpenBsdOSProcess.java
```diff
@@ -349,26 +350,14 @@ public long getAffinityMask() {
...
-                    ' ');
-            if (threadMap.containsKey(PsThreadColumns.ARGS)) {
-                threads.add(new OpenBsdOSThread(getProcessID(), threadMap));
-            }
-        }
-        return threads;
+        Predicate<Map<PsThreadColumns, String>> columnARGS = threadMap -> threadMap.containsKey(PsThreadColumns.ARGS);
+        return ExecutingCommand.runNative(psCommand).stream().skip(1)
+                .map(thread -> ParseUtil.stringToEnumMap(PsThreadColumns.class, thread.trim(), ' ')).filter(columnARGS)
+                .map(threadMap -> new OpenBsdOSThread(getProcessID(), threadMap)).collect(Collectors.toList());
```
dbwiddis, line 360: +valid filter, +rename all columns

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
```diff
@@ -336,11 +337,11 @@ private List<OSProcess> processMapToList(Collection<Integer> pids) {
...
-        List<OSProcess> processList = new ArrayList<>();
-        for (Integer pid : mapKeys) {
-            processList.add(new WindowsOSProcess(pid, this, processMap, processWtsMap, threadMap));
-        }
-        return processList;
+        Map<Integer, ProcessPerformanceData.PerfCounterBlock> finalProcessMap = processMap;
+        Map<Integer, ThreadPerformanceData.PerfCounterBlock> finalThreadMap = threadMap;
+        return mapKeys.stream().parallel()
+                .map(pid -> new WindowsOSProcess(pid, this, finalProcessMap, processWtsMap, finalThreadMap))
+                .collect(Collectors.toList());
```
dbwiddis, line 344: +valid filter, +rename all columns

oshi-core/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
```diff
@@ -336,11 +337,11 @@ private List<OSProcess> processMapToList(Collection<Integer> pids) {
         Set<Integer> mapKeys = new HashSet<>(processWtsMap.keySet());
         mapKeys.retainAll(processMap.keySet());
 
-        List<OSProcess> processList = new ArrayList<>();
-        for (Integer pid : mapKeys) {
-            processList.add(new WindowsOSProcess(pid, this, processMap, processWtsMap, threadMap));
-        }
-        return processList;
+        Map<Integer, ProcessPerformanceData.PerfCounterBlock> finalProcessMap = processMap;
+        Map<Integer, ThreadPerformanceData.PerfCounterBlock> finalThreadMap = threadMap;
```
dbwiddis, line 341: if we're naming them final, why not make them actually `final`? ;)

dbwiddis (approved):
This looks really good, thanks for the updates!   I've made a few more suggestions if you're willing to iterate one last time, otherwise this is probably GTG.

dbwiddis:
Oh, and also a changelog entry would be nice!

oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
adrian-kong: hmmm how about just `.filter(Predicate.not(AbstractNetworkIF::isLocalInterface))`

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
adrian-kong: would `hasColumnsPri` suffice? im not sure what `AllPS` means i.e. the other ones would be `hasKeywordsArgs` etc...

oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java
adrian-kong: nevermind it seems has compilation errors
```
Error:  /home/runner/work/oshi/oshi/oshi-core/src/main/java/oshi/hardware/common/AbstractNetworkIF.java:[159,42] cannot find symbol

```

dbwiddis:
Spotless should help with the unused imports :)

adrian-kong:
oops

dbwiddis:
> oops EDIT: spotless didnt catch the unused import 😢

Sounds like a PR to fix spotless config is in order ;)

dbwiddis (approved):
LGTM.   Let me know if you're ready to merge this or still working.

adrian-kong:
> LGTM. Let me know if you're ready to merge this or still working.

yep its all good to go!

PR #2207, opened 2022-10-03
Bump jackson-databind to 2.14.0-rc1

dbwiddis:
Wouldn't normally bump to a RC but this is getting flagged for security.   We aren't vulnerable as we haven't opted in to the setting, but I figure why not update to make the warning go away? 

Curious if there's a way to tell Renovate to do this update without changing configs...

https://security.snyk.io/vuln/SNYK-JAVA-COMFASTERXMLJACKSONCORE-3038426

### DIFF
--- a/oshi-demo/pom.xml
+++ b/oshi-demo/pom.xml
@@ -14,7 +14,7 @@
     <name>oshi-demo</name>
 
     <properties>
-        <jackson.version>2.13.4</jackson.version>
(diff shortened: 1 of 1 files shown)


hazendaz:
guessing configs would need changed to allow non final releases to show up.

dbwiddis:
Yeah, but I don't want that in normal config, just if it's a CVE.  Or to be able to easily manually trigger it from the dashboard using an @command or something. Oh well...

Issue #2206, opened 2022-10-02
Allow suppressing sysctl error for known error conditions


dbwiddis:
Too many of these in test logs:
> Warning:  2022-10-02 07:11:40.966 [WARN] MacOSProcess - Failed sysctl call for process arguments (kern.procargs2), process 10323 may not exist. Error code: 22

Also with the cache implementation in #2198 we'll be intentionally checking for some platform-specific sysctls that may not be there.

We have a boolean in the FileUtil classes optionally suppressing log messages.  We should add the same to the sysctls.

See also #2149 on Linux with essentially the same issue.

pavangole:
Hello @dbwiddis . I want to fix this issue. As this is my first time I don't know much about sysctls. Will you help me? Additionally, will you provide a location where to write the code and where to take references? Thank you.

dbwiddis:
Hey @pavangole ... the one most wanted doesn't use the util class. You can find the `MacOSProcess` class in the `oshi.software.os.mac` package. There's a `sysctl` call there inside the `queryArgsAndEnvironment()` method.   Make that log warning conditional on a global config parameter similar to what was done for Linux proc (look at `LinuxOSProcesss` and the `LOG_PROCFS_WARNING` constant and its use.

Also, if you like ... the `SyscltUtil` class is in `oshi.util.platform.mac`.   I actually already did the change for the one with the `int` default, which you can use as an example for the others.  This is optional.

pavangole:
What I understand is that I have to create a global config variable if it is set then we don't have to give log messages. Please correct me If I am wrong.

dbwiddis:
@pavangole that's correct.  See how LOG_PROCFS_WARNING is implemented and implement something similar for MAC_SYSCTL_WARNING (or another name you choose).

pavangole:
I tried to follow the contributing guidelines but it is giving errors. I tried mvn clean test command and it is showing following output.

![image](https://user-images.githubusercontent.com/64621994/195361983-69d1b8f8-aaaa-4eb8-9d22-85f583114ebf.png)

dbwiddis:
Further up it should tell you what failed.

pavangole:
This is the error.

![image](https://user-images.githubusercontent.com/64621994/195393088-30c7d63d-56b8-4639-bec5-2be68443eaaa.png)

dbwiddis:
Further up it should tell you exactly which test failed.

pavangole:
[...]

PR #2212, opened 2022-10-14
Suppress MacSysctl Warning Logs

pavangole:


oshi-core/src/main/java/oshi/software/os/mac/MacOSProcess.java
```diff
@@ -53,10 +54,13 @@ public class MacOSProcess extends AbstractOSProcess {
...
     private static final int ARGMAX = SysctlUtil.sysctl("kern.argmax", 0);
 
+    private static final boolean LOG_MAC_SYSCTL_WARNING = GlobalConfig.get(GlobalConfig.OSHI_OS_MAC_SYSCTL_LOGWARNING,
+        false);
+
     // 64-bit flag
     private static final int P_LP64 = 0x4;
     /*
-     * macOS States:
+     * macOS States:    7
```
dbwiddis, line 63: What's this 7 for?

oshi-core/src/main/resources/oshi.properties
```diff
@@ -116,12 +116,17 @@ oshi.os.windows.perfproc.disabled=
...
-# On Linux, most process metrics are read from the proc pseudo-filesystem. 
+# On Linux, most process metrics are read from the proc pseudo-filesystem.
 # When operating without elevated permissions, this results in frequent error
 # messages for failures to read the process environment files. Set this to true
 # to receive these warnings.
 oshi.os.linux.procfs.logwarning=false
 
+#Macos gives log warning if SYSCTL calls fails
+#In order to avoid warning make this property true
+
```
dbwiddis, line 127: Since the reasoning for this is the same as Linux (no elevated permissions) can we merge the two paragraphs above?  For example (feel free to use different wording) insert a comment right below line 119, `# On macOS, process environment is read via sysctl`. And then you can put the two settings next to each other.

dbwiddis (commented):
Code looks great with a few document requests.  Can you also add a change log entry?

oshi-core/src/main/java/oshi/software/os/mac/MacOSProcess.java
pavangole: I think by mistake it gets typed.

oshi-core/src/main/resources/oshi.properties
pavangole: Ya thanks for the suggesstion.

dbwiddis:
Thanks for your contribution!  Fantastic job.  I'm off for the night but other than the minor comment changes should be good to merge tomorrow.

pavangole:
Yes

pavangole:
Resolves [#2206
](https://github.com/oshi/oshi/issues/2206)

Issue #2215, opened 2022-10-18
Add max file descriptors per process


dbwiddis:
See commentary [here](https://stackoverflow.com/a/74101019/1161484).

OSHI currently returns the systemwide limit in `getMaxFileDescriptors()`.  On some OS's there is a per-process limit, for which we could add a `getMaxFileDescriptorsPerProcess()`.
 - on Linux, `/proc/sys/fs/nr-open` or `/proc/self/limits`.
 - on macOS and FreeBSD, sysctl `kern.maxfilesperproc`
 - on Windows, try [`_getmaxstdio`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/getmaxstdio?view=msvc-170) or hardcode 16 * 1024 * 1024 = 16,777,216.
 - on OpenBSD, try [`getdtablesize()`](https://man.openbsd.org/getdtablesize.3) or `getrlimit()` with `RLIMIT_NOFILE`.
 - on Solaris, try `getrlimit()` or [these options](https://www.thegeekdiary.com/solaris-how-to-set-limit-on-the-maximum-number-of-open-files-per-process/)
 - on AIX try `/etc/security/limits`, if it's not there assume unlimited

gitseti:
When I get to it, I would gladly contribute a PR to this. :) 

For now, just some thoughts / questions:
- Would it make sense to also offer a method to get the per-process soft limit for open files, sth. like `getMaxFileDescriptorsSoftLimitPerProcess`?
- On Linux `/proc/sys/fs/nr-open` does return the hard limit for open files afaik, but not the soft limit. Using `getrlimit()` would return both limits. Also another option would be to read out `/proc/{current-pid}/limits`.
- Afaik, macOS, FreeBSD and OpenBSD also support the `getrlimit()` POSIX function, so it might be worth to think about using the same implementation to get the limits. Or are there any concerns about using JNA calls instead of reading out the files?

dbwiddis:
Thanks for volunteering to add this new capability. 

> Would it make sense to also offer a method to get the per-process soft limit for open files

I'll admit I'm not an expert in these metrics, so I'll defer to your choice here.  I'm all for adding as much information as is generally needed, and if there are different hard and soft limits that are easily available, that's great... just make sure the javadocs are clear.  I'm not sure if the per-process limit differs across different process, in which case:
 - we may want a systemwide "default" or report values for "current process"
 - we may want something on the OSProcess object to get the value per-process

[...]

PR #2218, opened 2022-10-23
Add system-wide per-process file descriptors limits

gitseti:
Context: #2215


oshi-core/src/main/java/oshi/software/os/unix/aix/AixFileSystem.java
```diff
@@ -169,4 +170,18 @@ public long getOpenFileDescriptors() {
     public long getMaxFileDescriptors() {
         return ParseUtil.parseLongOrDefault(ExecutingCommand.getFirstAnswer("ulimit -n"), 0L);
     }
+
+    @Override
+    public long getMaxFileDescriptorsPerProcess() {
+        final List<String> lines = FileUtil.readFile("/etc/security/limits");
+        for (final String line : lines) {
+            if (line.startsWith("nofiles(descriptors)")) {
+                final String[] split = line.split("\\w+");
```
dbwiddis, line 179: I'm not sure if this is the best delimiter choice, and compiling regexes other than a single character is less performant.  There are some pre-compiled regexes in `ParseUtil` like `ParseUtil.whitespaces` or `ParseUtil.notDigits` that you can use, or this may be a good candidate for `ParseUtil.getLastInt()`.

Do you need to `trim()` the string before splitting?  [This example](https://sites.ualberta.ca/dept/chemeng/AIX-43/share/man/info/C/a_doc_lib/files/aixfiles/limits.htm) (from AIX 4) shows indented values per-user, although I haven't found an example file for AIX 7 which we should use the format for.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisFileSystem.java
```diff
@@ -20,10 +20,7 @@
 import oshi.annotation.concurrent.ThreadSafe;
 import oshi.software.common.AbstractFileSystem;
 import oshi.software.os.OSFileStore;
-import oshi.util.ExecutingCommand;
-import oshi.util.FileSystemUtil;
-import oshi.util.Memoizer;
-import oshi.util.ParseUtil;
+import oshi.util.*;
```
dbwiddis, line 23: No star imports, please. I'm sure this has failed checkstyle somewhere in the CI tests below. :)

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisFileSystem.java
```diff
@@ -182,6 +179,20 @@ public long getMaxFileDescriptors() {
         return 0L;
     }
 
+    @Override
+    public long getMaxFileDescriptorsPerProcess() {
+        final List<String> lines = FileUtil.readFile("/etc/system");
+        for (final String line : lines) {
+            if (line.matches("set.+rlim_fd_cur=\\d+")) {
```
dbwiddis, line 186: Please compile this regex pattern as a static variable, although this could also use a combination of `startsWith()`, `contains()`, and `getLastInt()`.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisFileSystem.java
```diff
@@ -182,6 +179,20 @@ public long getMaxFileDescriptors() {
...
+                final String[] split = line.split("set.+rlim_fd_cur=");
```
dbwiddis, line 187: If you use a capturing regex above you can just extract the `group()` and not worry about the split and other checking.

oshi-core/src/main/java/oshi/util/platform/linux/ProcPath.java
```diff
@@ -43,6 +43,7 @@ public final class ProcPath {
     public static final String SELF_STAT = PROC + "/self/stat";
     public static final String STAT = PROC + "/stat";
     public static final String SYS_FS_FILE_NR = PROC + "/sys/fs/file-nr";
+    public static final String SYS_FS_FILE_MAX = PROC + "/proc/sys/fs/file-max";
```
dbwiddis, line 46: you've duplicated `PROC` here.

oshi-core/src/test/java/oshi/software/os/FileSystemTest.java
```diff
@@ -33,6 +33,8 @@ void testFileSystem() throws IOException {
                 is(true));
         assertThat("File system max open file descriptors should be 0 or higher",
                 filesystem.getMaxFileDescriptors() >= 0L, is(true));
+        assertThat("File system max open file descriptors per process should be 0 or higher",
+                filesystem.getMaxFileDescriptorsPerProcess() >= 0L, is(true));
```
dbwiddis, line 37: I realize you copied the one above, but both of these should be `greaterThanOrEqualTo(0)`

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # 6.3.1 (in progress)
 
-* Your contribution here!
+* [#2218](https://github.com/oshi/oshi/pull/2218): Added system-wide per-process file descriptor limits - [@gitseti](https://github.com/gitseti)
```
dbwiddis, line 3: Add a `##### New Features` header here. :)

oshi-core/src/main/java/oshi/software/os/FileSystem.java
```diff
@@ -61,4 +61,12 @@ public interface FileSystem {
      * @return The maximum number of file descriptors if available, 0 otherwise.
      */
     long getMaxFileDescriptors();
+
+    /**
+     * The maximum number of open file descriptors per process. This returns the system-wide limit. The actual limit of
```
dbwiddis, line 66: Using both "per-procsess" and "systemwide" may be confusing.  Can you slightly clarify this?  Perhaps the word "default" or is this the "hard" limit?

oshi-core/src/main/java/oshi/software/os/FileSystem.java
```diff
@@ -61,4 +61,12 @@ public interface FileSystem {
...
+     * the calling process may be lower.
+     *
+     * @return The maximum number of file descriptors of every process if available, 0 otherwise.
```
dbwiddis, line 69: Similar to the above comment, I'd say "each process, unless configured to a lower value" or similar.

oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -250,4 +255,13 @@ private static long getFileDescriptors(int index) {
         }
         return 0L;
     }
+
+    private static long getFileDescriptorsPerProcess() {
+        final String filename = ProcPath.SYS_FS_FILE_MAX;
+        final List<String> fileContent = FileUtil.readFile(filename);
+        if (!fileContent.isEmpty()) {
+            return ParseUtil.parseLongOrDefault(fileContent.get(0), 0L);
```
dbwiddis, line 263: We have a `getLongFromFile()` method on the fileUtil class that does exactly this logic.

oshi-core/src/main/java/oshi/software/os/windows/WindowsFileSystem.java
```diff
@@ -281,4 +281,9 @@ public long getOpenFileDescriptors() {
     public long getMaxFileDescriptors() {
         return MAX_WINDOWS_HANDLES;
     }
+
+    @Override
+    public long getMaxFileDescriptorsPerProcess() {
+        return MAX_WINDOWS_HANDLES;
```
dbwiddis, line 287: This is a smaller constant: 16 * 1024 * 1024 (64 bit) or 16 * 1024 * 512 (32 bit).

dbwiddis (changes requested):
Great start, and thanks for your contribution!   

Some improvements suggested inline.

oshi-core/src/main/java/oshi/software/os/unix/aix/AixFileSystem.java
gitseti: `ParseUtil.getLastLong()` seems to be a good fit. 👍🏻 

I think you are correct. [This 7.2 reference](https://www.ibm.com/docs/en/ssw_aix_72/filesreference/filesreference_pdf.pdf) looks exactly as the AIX 4 example I will add a `trim()` and `startsWith("nofiles")` check.

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisFileSystem.java
gitseti: Also this should be `rlim_fd_max`, not `rlim_fd_cur`.

oshi-core/src/main/java/oshi/software/os/unix/aix/AixFileSystem.java
dbwiddis: You'll also need to check if it matches the current user name...

oshi-core/src/main/java/oshi/software/os/windows/WindowsFileSystem.java
gitseti: Hmm, I thought that the per-process limit cannot be higher than the value returned by `getMaxFileDescriptors()`. So, wouldn't it make sense to set it like that, because there exists no per-process file limit?

dbwiddis: Huh, nevermind. These are the same limit, for some reason I thought the systemwide was much larger.

I've seen a per-process of 512 when using the MS C runtime, but that probably doesn't apply here.

oshi-core/src/main/java/oshi/software/os/FileSystem.java
gitseti: I tried to add some clarification. Hope this fits better.

oshi-core/src/test/java/oshi/software/os/FileSystemTest.java
dbwiddis: Missed one :)

dbwiddis (approved):

dbwiddis:
I went ahead and approved this and will merge when CI finishes, as I know you're working on "part 2".  I think the AIX still needs to somehow check the per-user limit, and it requires admin privileges (so I have no way to even test) so I'm ok leaving it is as is for now.

PR #2225, opened 2022-10-26
Add process specific open file descriptor limits

### LINKED ISSUE #2215: Add max file descriptors per process

gitseti:
This resolves #2215 as it implements the second needed part.

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/OSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSProcess.java
@@ -249,44 +249,24 @@ public interface OSProcess {
     long getOpenFiles();
 
     /**
-     * Gets the soft limit for open file handles (or network connections) that belong to the process.
+     * Gets the soft limit for open file handles (or network connections) of the given process.
      * <p>
-     * On Windows this returns the maximum number of open files for the process. The result is the same as
-     * {@link #getHardOpenFileLimit()} ()}.
+     * Retrieving the soft limit for processes other than the calling process is only supported on Linux, FreeBsd and
+     * Solaris.
      *
-     * @return the soft open file limit for the process.
+     * @return the soft open file limit for the process if available or -1 if not supported.
      */
     long getSoftOpenFileLimit();
 
     /**
-     * Gets the soft limit for open file handles (or network connections) that belong to the given process.
-     *
-     * This function is only supported on Linux, FreeBsd and Solaris.
-     *
-     * @param proc the process to retrieve the soft open file limit for.
-     * @return the soft open file limit for the process if available or -1 if not supported.
-     */
-    long getSoftOpenFileLimit(OSProcess proc);
-
-    /**
-     * Gets the hard limit for open file handles (or network connections) that belong to the process.
+     * Gets the hard limit for open file handles (or network connections) that belong to the given process.
      * <p>
-     * On Windows this returns the maximum number of open files for the process. The result is the same as
-     * {@link #getSoftOpenFileLimit()}.
+     * Retrieving the hard limit for processes other than the calling process is only supported on Linux, FreeBsd and
+     * Solaris.
      *
-     * @return the hard open file limit for the process.
+     * @return the soft open file limit for the process if available or -1 if not supported.
      */
     long getHardOpenFileLimit();
-
-    /**
-     * Gets the hard limit for open file handles (or network connections) that belong to the given process.
-     *
-     * This function is only supported on Linux, FreeBsd and Solaris.
-     *
-     * @param proc the process to retrieve the soft open file limit for.
-     * @return the soft open file limit for the process if available or -1 if not supported.
-     */
-    long getHardOpenFileLimit(OSProcess proc);
 
     /**
      * Gets cumulative CPU usage of this process.
@@ -311,7 +291,6 @@ public interface OSProcess {
      *
      * @param proc An {@link OSProcess} object containing statistics for this same process collected at a prior point in
      *             time. May be null.
-     *
      * @return If the prior snapshot is for the same process at a prior point in time, the proportion of elapsed up time
      *         between the current process snapshot and the previous one that the process was executing in kernel or
      *         user mode. Returns cumulative load otherwise.
--- a/oshi-core/src/main/java/oshi/software/os/OSProcess.java
+++ b/oshi-core/src/main/java/oshi/software/os/OSProcess.java
@@ -249,6 +249,46 @@ public interface OSProcess {
     long getOpenFiles();
 
     /**
+     * Gets the soft limit for open file handles (or network connections) that belong to the process.
+     * <p>
+     * On Windows this returns the maximum number of open files for the process. The result is the same as
+     * {@link #getHardOpenFileLimit()} ()}.
+     *
+     * @return the soft open file limit for the process.
+     */
+    long getSoftOpenFileLimit();
+
+    /**
+     * Gets the soft limit for open file handles (or network connections) that belong to the given process.
+     *
+     * This function is only supported on Linux, FreeBsd and Solaris.
+     *
+     * @param proc the process to retrieve the soft open file limit for.
+     * @return the soft open file limit for the process if available or -1 if not supported.
+     */
+    long getSoftOpenFileLimit(OSProcess proc);
+
+    /**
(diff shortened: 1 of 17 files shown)


oshi-core/src/main/java/oshi/software/os/OSProcess.java
```diff
@@ -248,6 +248,46 @@ public interface OSProcess {
...
+
+    /**
+     * Gets the soft limit for open file handles (or network connections) that belong to the given process.
+     *
+     * This function is only supported on Linux, FreeBsd and Solaris.
+     *
+     * @param proc the process to retrieve the soft open file limit for.
+     * @return the soft open file limit for the process if available or -1 if not supported.
+     */
+    long getSoftOpenFileLimit(OSProcess proc);
```
dbwiddis, line 269: I'm really confused here.  Why do we have two methods with the same name?  I don't see us calling one OSProcess object using a completely different OSProcess object as an argument, when we could just call this method from that other object.   Why do they return different results?

gitseti: Oh, I somehow assumed that OSProcess is always the calling process. 🙈 
I am myself a bit confused now where the method for getting the calling process soft/hard limits should go to, because having it in OSProcess suggests that it will return the limit for the OSProcess, but the `getrlimits()` call returns the limit of the calling process. 

We could merge the methods by checking `OSProcess.getProcessId() == ProcessHandle.current().pid` and then doing the system call, else we try to fallback on the filesystem method to retrieve the limits of another process.

gitseti: So to get the limits of the current process a user would have to call `OSProcess.getSoftOpenFileLimit(ProcessHandle.current().pid())`.

gitseti: So for OSs which support getting open file limits from other processes the merged method would look like:
```java 
// Example Linux
if (getProcessID() == ProcessHandle.current().pid()) {
            final Resource.Rlimit rlimit = new Resource.Rlimit();
            LinuxLibc.INSTANCE.getrlimit(LinuxLibc.RLIMIT_NOFILE, rlimit);
            return rlimit.rlim_cur;
        } else {
            return getProcessOpenFileLimit(getProcessID(), 1);
        }
´´´

And for OSs which do not support to get the limits of other processes it would look like:
```java
// Example for Mac
public long getSoftOpenFileLimit() {
        if (getProcessID() == ProcessHandle.current().pid()) {
            final Resource.Rlimit rlimit = new Resource.Rlimit();
            SystemB.INSTANCE.getrlimit(MAC_RLIMIT_NOFILE, rlimit);
            return rlimit.rlim_cur;
        } else {
            return -1; // Not supported
        }
    }
``` 

What do you think?

dbwiddis: > having it in OSProcess suggests that it will return the limit for the OSProcess referred process, but the getrlimits() call returns the limit of the calling process.

Ah, yes, I recall that discussion now.

> We could merge the methods by checking `OSProcess.getProcessId() == ProcessHandle.current().pid`

A few notes:
1. I like the idea of merging them into one method and having different behavior based on whether it's the calling process.   The user knows the current PID so they could rely on `get...Limit()` when called from the matching `OSProcess` object.  So `os.getProcess(os.getCurrentProcess()).getHardOpenFileLimit()` would reliably return a value.   We could return the other-process value (or -1 for "unsupported") if called from other processes.
2. `ProcessHandle` is JDK9 so you can't use it.  However, the appropriate `OperatingSystem` object is usually passed to the corresponding `OSProcess` constructor so it has access to `os.getProcessId()` for purposes like this.  For example, `WindowsOSProcess.getCwd()` has `if (getProcessID() == this.os.getProcessId()) {... `

dbwiddis: > What do you think?

LGTM.   Nit: use -1L.   I'm not sure it matters as the compiler probably is smart enough to figure that constant out, but I like to be explicit with my types!

Make sure the javadocs are really clear about this, though!  I'd like the 'not supported' to only apply to "not the calling process".

gitseti: Cool. 
Getting the OperatingSystem instance is a bit messy for the parts where the processes can be retrived via a static method. I worked around that for now by using `new SystemInfo().getOperationSystem()`. Not sure though if that is fine. You can have a look in [`a948bc8` (#2225)](https://github.com/oshi/oshi/pull/2225/commits/a948bc8a8839bfe3cec1045664c2687d83c87a90)

Another option would be to just pass the processId of the calling process to the OSProcess implementations.

dbwiddis: Creating a new OS object is probably a little heavy.  And we can always make the static methods non-static. 

As for the pid vs object, keeping an already instantiated object seems easier than creating (a bunch of) ints, particularly if most of them will never be used.

gitseti: > Creating a new OS object is probably a little heavy. And we can always make the static methods non-static.

Oh, I somehow though that these methods are public and changing them to non-static would result in a breaking change. But that's actually not the case. I make them non-static then 👍🏻 

> As for the pid vs object, keeping an already instantiated object seems easier than creating (a bunch of) ints, particularly if most of them will never be used.

 I agree 👍🏻

gitseti: So the only thing left to do, so that this can leave the draft state is the calling of the Windows `_getmaxstdio` method. 
Can you point me to some documentation or already used code on how to implement this JNA call?

Currently, there are just placeholder methods for WindowsOSProcess:
```java
@Override
    public long getSoftOpenFileLimit() {
        // TODO how to execute _getmaxstdio
        return new SystemInfo().getOperatingSystem().getFileSystem().getMaxFileDescriptorsPerProcess();
    }

    @Override
    public long getHardOpenFileLimit() {
        // FIXME how to execute _getmaxstdio
        return new SystemInfo().getOperatingSystem().getFileSystem().getMaxFileDescriptorsPerProcess();
    }
```

dbwiddis:
Hey @gitseti don't want to rush this PR, but based on a bug report in AIX I'm probably going to try to push out a patch release this weekend.  Do you want me to try to wait for you to finish this implementation or is it close enough to release it as-is with some more tweaks for the next version?

gitseti:
Hey @dbwiddis, imo, it should be ready for review now.

The only thing left would be to have a solution for Windows as I'm not sure how to do the `_getmaxstdio` JNA call. 
Currently, we just use the `getMaxFileDescriptorsPerProcess` returned limits for `getSoftOpenFileLimit` and `getHardOpenFileLimit`. We can also deliver the more precise limits for Windows in a later patch. 
WDYT?

dbwiddis:
Sure, we can leave those implementations for later. 

But can you remove the FIXME and TODO notes? Just create a new issue. :)

dbwiddis:
We can table the windows `_getmaxstdio` question. I think it actually only applies to the Windows C runtime so would be inaccurate for other processes.    I replaced your call navigating the API to get to a constant with a direct fetching of the constant (which happens to be in a class in the same package).  Will merge shortly.

Thanks again for your contribution!

PR #2229, opened 2022-10-29
Fix division by zero on AIX with fewer logProcs than physProcs

### LINKED ISSUE #2228: 6.3.0 throws exception on AIX from AixCentralProcessor.java:113
Hi,

After upgrading oshi-core from 6.2.2 to 6.3.0 I started getting errors when running on AIX:

```
java.lang.ArithmeticException: divide by zero
        at oshi.hardware.platform.unix.aix.AixCentralProcessor.initProcessorCounts(AixCentralProcessor.java:113)
        at oshi.hardware.common.AbstractCentralProcessor.<init>(AbstractCentralProcessor.java:65)
        at oshi.hardware.platform.unix.aix.AixCentralProcessor.<init>(AixCentralProcessor.java:38)
        at oshi.hardware.platform.unix.aix.AixHardwareAbstractionLayer.createProcessor(AixHardwareAbstractionLayer.java:55)
        at oshi.util.Memoizer$1.get(Memoizer.java:61)
        at oshi.hardware.common.AbstractHardwareAbstractionLayer.getProcessor(AbstractHardwareAbstractionLayer.java:48)
        at sysmon.plugins.os_base.BaseProcessorExtension.getMetrics(BaseProcessorExtension.java:84)
```

Call in my code is from [BaseProcessorExtension.java:84](https://bitbucket.org/mnellemann/sysmon/src/63d2fbd7800c217284a67583e3085fa1110f1ab1/plugins/os-base/src/main/java/sysmon/plugins/os_base/BaseProcessorExtension.java#lines-84), which works as expected with 6.2.2. 

Code is tested on a Power9 LPAR (VM/Guest) running AIX 7200-05-04-2220.
```System configuration: type=Dedicated mode=Capped smt=8 lcpu=8 mem=4096MB ```

Works fine on Linux - have not tested on any other OS.

Have I missed something important in the release notes?

dbwiddis:
Fixes #2228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ##### Bug fixes / Improvements
 * [#2224](https://github.com/oshi/oshi/pull/2224): Detect Windows Server 2022 in older JDKs - [@dbwiddis](https://github.com/dbwiddis).
+* [#2229](https://github.com/oshi/oshi/pull/2229): Fix division by zero on AIX with fewer logical processors than physical processors - [@dbwiddis](https://github.com/dbwiddis).
 
 # 6.3.0 (2022-10-16)
 
--- a/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java
@@ -64,8 +64,11 @@ public interface CentralProcessor {
 
     /**
      * Returns an {@code UnmodifiableList} of the CPU's logical processors. The list will be sorted in order of
-     * increasing NUMA node number, and then processor number. This order is consistent with other methods providing
-     * per-processor results.
+     * increasing NUMA node number, and then processor number. This order is (usually) consistent with other methods
+     * providing per-processor results.
+     * <p>
+     * On some operating systems with variable numbers of logical processors, the size of this array could change and
+     * may not align with other per-processor methods.
      *
      * @return An {@code UnmodifiabeList} of logical processors.
      */
@@ -83,8 +86,9 @@ public interface CentralProcessor {
      * Makes a best-effort attempt to identify the CPU's processor caches. For hybrid processors, both performance and
      * efficiency core caches are shown. Only one instance of per-core caches is shown.
      * <p>
-     * Values are unreliable in a VM. Callers should conduct sanity checking of the returned objects. Not all values are
-     * available on all operating systems or architectures.
+     * Values are unreliable in virtual machines and rely in information made available by the VM or hypervisor. Callers
+     * should conduct sanity checking of the returned objects. Not all values are available on all operating systems or
+     * architectures.
      * <p>
      * Not available on Solaris.
      *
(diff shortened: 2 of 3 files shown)


mnellemann:
Hi @dbwiddis.

Thank you for looking into this. The case is closed, but I would like to provide my view in the physical processors, so writing my thoughts here:


I have no experience with Raptor or other Power based systems, but let me try to explain how this works on IBM Power systems.  I might not be 100% accurate on this - feel free to correct me if I have gotten something wrong :)

Besides very old systems and a few speciel models, all Power servers are shipped with the PowerVM hypervisor in firmware. There is no way to run “bare-metal” - the hypervisor is always present. On top of PowerVM you install any number of AIX, Linux or IBM i logical partitions (LPAR’s or what everyone else calls a VM or guest). 

The Power servers exists in configurations from 1,2,3 and 4 processor sockets up to 16 processor sockets in a single system. The socket is populated with a Power processor (latest gen. is Power10) which could have from 4 to 16 processor cores each. So a very small system could have 1-socket w. 4 Power cores and the largest could have 16 sockets with 240 cores.

Each of the Power cores are able to execute independant hardware threads, similar to Hyperthreading (SMT-2) on Intel. The Power cores are able to execute in SMT-2, SMT-4 or SMT-8 mode independently for each LPAR/VM running. Each of these hardware threads will show up as a logical processor / vCPU in the LPAR/VM. So the large system with 240 Power10 cores would be able to provide up to 1920 logical processors (in SMT-8) assigned to any number of LPAR’s / VM’s.

To my understanding the number of physical processors (or cores) is not really relevant for any LPAR/VM, and I don’t think oshi would have to take this into account. It is true that you can see this number for within an LPAR with the ‘lparstat’ command (and possible in other ways). If you could run bare-metal then ‘logical processors’ > cores > sockets. Metrics on the physical system, I/O adapters and virtualisation is provided through the HMC (Hardware Management Console) which is where you could look at the overall system allocation and resources.

When you configure an LPAR you can specify its processor mode as ‘dedicated’ in which case you state a minimum, allocated and maximum number of cores assigned to this specific LPAR. You can not over-subscribe the number of allocated processors. The min and max values are the range within you are able to dynamically change the allocation value within having to reboot the LPAR. So if you allocate 2 cores (in SMT-8) this would show up as 2*8 = 16 logical processors in your AIX or Linux LPAR.

You can also configure an LPAR with a processor mode as ‘shared’ and things get’s a little more complicated here.  You still allocate processor cores and have the min and max which works in the same way and will show up in AIX or Linux as logical processors. The extra dimension is entitlements. As an example, you can allocate 2 cores but only entitle 1.5 cores, which results in 2*8 = 16 logical processors being available (and visible) to AIX or Linux but only 12 of the 16 cores being guarantied (entitled) to the LPAR. If the 0.5 processor not entitled is not in use elsewhere, then PowerVM will allow the LPAR to use it. PowerVM will look at a ‘weight’ which you can set for each LPAR to determine who get’s extra resources first in case of constraints. When you run in the type of setup you can see from ‘lparstat’ that you might be using 100% of you logical processors, but 200% of you entitlements. In this case you could have given the LPAR eg. 4 cores, but one 2.0 as entitlement.  You could also allocate 1 cores but only entitle 0.25 cores or even less, which is called micro-partitioning - in this case you could end up using 100% entitlements but only 25% if the visible logical processors. There is also a min and max for the entitlements.


/Mark

Issue #2309, opened 2023-02-12
FreeBSD CPU counts not accurate on FreeBSD 13.1


decketron:
cpu count on FreeBSD is calculated from output of 
  sysctl kern.sched.topology_spec        
And parsing the resulting <groups> XML.
Comment states:
	* If there is no level 3, then all logical processors are physical
	
I have a Dell XPS 8500:
  kern.ostype: FreeBSD
  kern.osrelease: 13.1-RELEASE-p2
  kern.osrevision: 199506
  kern.version: FreeBSD 13.1-RELEASE-p2 n245412-484f039b1d0 TRUENAS
  hw.model: Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz
  kern.smp.topology: 0
  kern.smp.cores: 4
  kern.smp.threads_per_core: 2
  kern.smp.cpus: 8
  hw.ncpu: 8  
	# sysctl kern.sched.topology_spec
	kern.sched.topology_spec: ```

```
<groups>
	 <group level="1" cache-level="3">
	  <cpu count="8" mask="ff,0,0,0">0, 1, 2, 3, 4, 5, 6, 7</cpu>
	  <children>
	   <group level="2" cache-level="2">
		<cpu count="2" mask="3,0,0,0">0, 1</cpu>
		<flags><flag name="THREAD">THREAD group</flag><flag name="SMT">SMT group</flag></flags>
	   </group>
	   <group level="2" cache-level="2">
		<cpu count="2" mask="c,0,0,0">2, 3</cpu>
		<flags><flag name="THREAD">THREAD group</flag><flag name="SMT">SMT group</flag></flags>
	   </group>
	   <group level="2" cache-level="2">
		<cpu count="2" mask="30,0,0,0">4, 5</cpu>
		<flags><flag name="THREAD">THREAD group</flag><flag name="SMT">SMT group</flag></flags>
	   </group>
	   <group level="2" cache-level="2">
		<cpu count="2" mask="c0,0,0,0">6, 7</cpu>
		<flags><flag name="THREAD">THREAD group</flag><flag name="SMT">SMT group</flag></flags>
	   </group>
	  </children>
	 </group>
	</groups>  

```  
  
And a Dell OptiPlex 7020:
  kern.ostype: FreeBSD
  kern.osrelease: 13.1-RELEASE-p2
  kern.osrevision: 199506
  kern.version: FreeBSD 13.1-RELEASE-p2 n245412-484f039b1d0 TRUENAS
[...]

PR #2318, opened 2023-02-26
adjust fix in regex for #2309

decketron:
I don't fully understand the meaning of the values of the CSV exposed by FreeBSD.
However this fix allows the test to pass by inspecting the first element of the list,
rather than expecting a scalar value.
I think this allows the test to be expanded to cover the other values in the list, should there
meanings and associated requirements become clear.

Also not sure what is the best way to run all tests locally before submitting a PR?
thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ##### Bug fixes / Improvements
 * [#2292](https://github.com/oshi/oshi/pull/2292): Update to JNA 5.13.0 - [@dbwiddis](https://github.com/dbwiddis).
+* [#2309](https://github.com/oshi/oshi/pull/2309): Fix FreeBSD CPU count parsing - [@decketron](https://github.com/decketron).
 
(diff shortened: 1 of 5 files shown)


dbwiddis:
You should be able to run `./mvnw test` to run the tests locally on any platform.

I'm not sure about that meaning either, as the [man page](https://man.freebsd.org/cgi/man.cgi?query=smp&apropos=0&sektion=4&manpath=FreeBSD+14.0-CURRENT&arch=default&format=html) still has mask values of `0xff` and such.

I did find [this thread](https://forums.freebsd.org/threads/number-of-cpus-and-cores.41299/) which shows the updated format but doesn't explain what it means.  My suspicion is that it's for support for up to 256 processors as each mask would be limited to 64 bits.

dbwiddis:
Anyway, your fix looks good, can I ask you to add a changelog entry before I merge it (forgot to do so on the last submission).

Issue #2326, opened 2023-03-15
Relax slf4j OSGi requirement to include slf4j 1


Artur-:
Many existing libraries are still using slf4j. When you are deploying to OSGi, it would be preferrable if you can deploy only slf4j 1 even though oshi is using slf4j 2. To my understanding they are API compatible.

e.g. Jetty has defined a range of `[1.7,3.0)`  https://github.com/eclipse/jetty.project/blob/e6de81084d8f69007102bd3a6a12ad67555e3eb6/pom.xml#L22

dbwiddis:
Hey @Artur- took a quick look at implementing this so I can cut a new release, but I have no experience with OSGi.  It looks like a line needs to be added somewhere in here, but I'm not sure what.
https://github.com/oshi/oshi/blob/b8527282d949d5c2ad7655b59d24f87c0a83210b/pom.xml#L312-L322

If this is a feature you want in the next release, I really need you or another OSGi user to do this fix.

hazendaz:
There has been a lot of chatter on slf4j around osgi.  Bet you can find help from there.  Latest release of 2.x got some osgi automated fixes so probably worth checking that pom.  No experience either here just noting possible place to check.

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

mshabarov:
@dbwiddis thanks for fast reply!
I've made a patch here https://github.com/oshi/oshi/pull/2329. With this patch I have no more "missing requirement" errors in our OSGi environment, that I had with oshi-core 6.4.0. Could you please review.

dbwiddis:
> @dbwiddis thanks for fast reply! I've made a patch here #2329. With this patch I have no more "missing requirement" errors in our OSGi environment, that I had with oshi-core 6.4.0. Could you please review.

Reviewed.  Inclined to approve as is based on your report that "it works" but do have one comment.

Will try to get a 6.4.1 release out this weekend.

PR #2329, opened 2023-03-17
Allow using SLF4J 1 in OSGi containers

mshabarov:
Extends the SLF4J version range for OSGi environments, so SLF4J 1.7 for instance doesn't raise a conflicts with latest oshi-core.

Fixes https://github.com/oshi/oshi/issues/2326

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,6 +95,7 @@
         <!-- Users of the Spring Boot Starter Parent should include this property in their POM -->
         <jna.version>5.13.0</jna.version>
         <slf4j.version>2.0.6</slf4j.version>
+        <osgi.slf4j.import.packages>org.slf4j;version="[1.7,3.0)"</osgi.slf4j.import.packages>
         <junit.version>5.9.2</junit.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -316,6 +317,7 @@
                     <version>${bnd-maven-plugin.version}</version>
                     <configuration>
                         <bnd><![CDATA[Export-Package: oshi.*;-noimport:=true;-split-package:=merge-first
+                        Import-Package: ${osgi.slf4j.import.packages}, *
```
dbwiddis, line 320: I had assumed it was similar to this!  Does the `-noimport:=true` line above have any relevance here?

mshabarov: This directive is for "Export-Package" only as I understand it, from the javadocs: `The package must not be substitutably imported.`
Quick search led me to this article https://bnd.bndtools.org/heads/export_package.html. So I'm not sure this is needed in "Import-Package".

dbwiddis:
Thanks for this PR.  I'll cut a new release this weekend.

PR #2343, opened 2023-03-31
Tried refactoring the code smells that I found while going through this project. #first-timers-only

Faizal17:
Refactored the code and also performed the suggestions provided by sonatype-lift. Haven't changed one suggestion as I wasn't sure if we can modify that or not as it was using the same logic priorly too.

### DIFF
--- a/oshi-core/src/main/java/oshi/driver/linux/Who.java
+++ b/oshi-core/src/main/java/oshi/driver/linux/Who.java
@@ -6,7 +6,7 @@ package oshi.driver.linux;
 
 import static oshi.jna.platform.unix.CLibrary.LOGIN_PROCESS;
 import static oshi.jna.platform.unix.CLibrary.USER_PROCESS;
-import static oshi.util.Util.isSessionInvalid;
+import static oshi.util.Util.isSessionValid;
 
 import java.nio.charset.Charset;
 import java.util.ArrayList;
@@ -50,7 +50,7 @@ public final class Who {
                     String host = ParseUtil.parseUtAddrV6toIP(ut.ut_addr_v6);
                     long loginTime = ut.ut_tv.tv_sec * 1000L + ut.ut_tv.tv_usec / 1000L;
                     // Sanity check. If errors, default to who command line
-                    if (isSessionInvalid(user, device, loginTime)) {
+                    if (!isSessionValid(user, device, loginTime)) {
                         return oshi.driver.unix.Who.queryWho();
                     }
                     whoList.add(new OSSession(user, device, loginTime, host));
--- a/oshi-core/src/main/java/oshi/driver/linux/Who.java
+++ b/oshi-core/src/main/java/oshi/driver/linux/Who.java
@@ -1,11 +1,12 @@
 /*
- * Copyright 2020-2022 The OSHI Project Contributors
+ * Copyright 2020-2023 The OSHI Project Contributors
  * SPDX-License-Identifier: MIT
  */
 package oshi.driver.linux;
 
 import static oshi.jna.platform.unix.CLibrary.LOGIN_PROCESS;
 import static oshi.jna.platform.unix.CLibrary.USER_PROCESS;
+import static oshi.util.Util.isSessionInvalid;
 
 import java.nio.charset.Charset;
 import java.util.ArrayList;
@@ -49,7 +50,7 @@ public final class Who {
                     String host = ParseUtil.parseUtAddrV6toIP(ut.ut_addr_v6);
                     long loginTime = ut.ut_tv.tv_sec * 1000L + ut.ut_tv.tv_usec / 1000L;
                     // Sanity check. If errors, default to who command line
-                    if (user.isEmpty() || device.isEmpty() || loginTime < 0 || loginTime > System.currentTimeMillis()) {
+                    if (isSessionInvalid(user, device, loginTime)) {
                         return oshi.driver.unix.Who.queryWho();
                     }
                     whoList.add(new OSSession(user, device, loginTime, host));
--- a/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
@@ -7,13 +7,10 @@ package oshi.hardware.platform.unix.solaris;
 import static oshi.util.Memoizer.memoize;
 import static oshi.util.ParseUtil.getValueOrUnknown;
 
-import java.util.Arrays;
 import java.util.EnumMap;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.function.Supplier;
-import java.util.stream.Collectors;
 
 import oshi.annotation.concurrent.Immutable;
 import oshi.hardware.Baseboard;
@@ -79,22 +76,6 @@ final class SolarisComputerSystem extends AbstractComputerSystem {
     }
 
     private static SmbiosStrings readSmbios() {
-        String biosVendor = null;
-        String biosVersion = null;
-        String biosDate = null;
-
-        String manufacturer = null;
-        String model = null;
-        String serialNumber = null;
-        String uuid = null;
-
-        String boardManufacturer = null;
-        String boardModel = null;
-        String boardVersion = null;
-        String boardSerialNumber = null;
-
-        Map<String, String> biosStrings = new HashMap<>();
-
         // $ smbios
         // ID SIZE TYPE
         // 0 87 SMB_TYPE_BIOS (BIOS Information)
@@ -131,15 +112,7 @@ final class SolarisComputerSystem extends AbstractComputerSystem {
         // ID SIZE TYPE
         // 3 .... <snip> ...
 
-        final String vendorMarker = "Vendor:";
-        final String biosDateMarker = "Release Date:";
-        final String biosVersionMarker = "VersionString:";
-
-        final String manufacturerMarker = "Manufacturer:";
-        final String productMarker = "Product:";
         final String serialNumMarker = "Serial Number:";
-        final String uuidMarker = "UUID:";
-        final String versionMarker = "Version:";
 
         SmbType smbTypeId = null;
 
@@ -241,29 +214,5 @@ final class SolarisComputerSystem extends AbstractComputerSystem {
             this.boardVersion = getValueOrUnknown(smbTypeBaseboardStrings, versionMarker);
             this.boardSerialNumber = getValueOrUnknown(smbTypeBaseboardStrings, serialNumMarker);
         }
-    }
-
-    /**
-     * Generate a map of strings parsing the BIOS Strings
-     *
-     * @param checkLines The lines received by running the bios command
-     * @param smbTypeId  The smbTypeID fetched
-     * @param param      contains the biosStrings data to be fetched for
-     * @return Map of strings based on the param and smbTypeId passed in the argument.
-     */
-    private static Map<String, String> parseBIOSStrings(List<String> checkLines, int smbTypeId, String... param) {
-//        Map<String, String> biosStrings = Arrays.stream(param).filter(checkLine::contains)
-//                .collect(Collectors.toMap(p -> p, p -> checkLine.split(p)[1].trim()));
-//        Map<String, String> biosStrings = checkLines.stream()
-//            .filter(line -> Arrays.stream(param).anyMatch(line::startsWith))
-//            .collect(Collectors.toMap(
-//                s -> s.substring(0, s.indexOf(":")).trim(),
-//                s -> s.substring(s.indexOf(":") + 1).trim()));
-        Map<String, String> biosStrings = Arrays.stream(param)
-                .filter(key -> checkLines.stream().anyMatch(line -> line.startsWith(key)))
-                .collect(Collectors.toMap(s -> s.substring(0, s.indexOf(":")).trim(),
-                        s -> s.substring(s.indexOf(":") + 1).trim(), (oldValue, newValue) -> newValue));
-        biosStrings.put("smbTypeId", Integer.toString(smbTypeId));
-        return biosStrings;
     }
 }
--- a/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
@@ -7,9 +7,11 @@ package oshi.hardware.platform.unix.solaris;
 import static oshi.util.Memoizer.memoize;
 import static oshi.util.ParseUtil.getValueOrUnknown;
 
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 
 import oshi.annotation.concurrent.Immutable;
 import oshi.hardware.Baseboard;
@@ -251,13 +253,9 @@ final class SolarisComputerSystem extends AbstractComputerSystem {
      * @return True if the user of device is empty or the login time is lesser than zero or greater than current time.
      */
     private static Map<String, String> parseBIOSStrings(String checkLine, int smbTypeId, String... param) {
-        Map<String, String> biosStrings = new HashMap<>();
+        Map<String, String> biosStrings = Arrays.stream(param).filter(checkLine::contains)
+                .collect(Collectors.toMap(p -> p, p -> checkLine.split(p)[1].trim()));
         biosStrings.put("smbTypeId", Integer.toString(smbTypeId));
-        for (int i = 0; i < param.length; i++) {
-            if (checkLine.contains(param[i])) {
-                biosStrings.put(param[i], checkLine.split(param[i])[1].trim());
-            }
-        }
         return biosStrings;
     }
 }
--- a/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
@@ -245,12 +245,12 @@ final class SolarisComputerSystem extends AbstractComputerSystem {
     }
 
     /**
-     * Tests if session of a user logged in a device is valid or not.
+     * Generate a map of strings parsing the BIOS Strings
      *
      * @param checkLine The line received by running the bios command
      * @param smbTypeId The smbTypeID fetched
      * @param param     contains the biosStrings data to be fetched for
-     * @return True if the user of device is empty or the login time is lesser than zero or greater than current time.
+     * @return Map of strings based on the param and smbTypeId passed in the argument.
      */
     private static Map<String, String> parseBIOSStrings(String checkLine, int smbTypeId, String... param) {
         Map<String, String> biosStrings = Arrays.stream(param).filter(checkLine::contains)
--- a/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
@@ -27,7 +27,7 @@ import oshi.util.Util;
  */
 @Immutable
 final class SolarisComputerSystem extends AbstractComputerSystem {
-    private enum SmbType {
+    public enum SmbType {
         /**
          * BIOS
          */
--- a/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
@@ -5,7 +5,6 @@
 package oshi.hardware.platform.unix.solaris;
 
 import static oshi.util.Memoizer.memoize;
-import static oshi.util.Util.parseBIOSStrings;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -262,5 +261,24 @@ final class SolarisComputerSystem extends AbstractComputerSystem {
                 this.boardSerialNumber = Constants.UNKNOWN;
             }
         }
+    }
+
+    /**
+     * Tests if session of a user logged in a device is valid or not.
+     *
+     * @param checkLine The line received by running the bios command
+     * @param smbTypeId The smbTypeID fetched
+     * @param param     contains the biosStrings data to be fetched for
+     * @return True if the user of device is empty or the login time is lesser than zero or greater than current time.
+     */
+    private static Map<String, String> parseBIOSStrings(String checkLine, Integer smbTypeId, String... param) {
+        Map<String, String> biosStrings = new HashMap<>();
+        biosStrings.put("smbTypeId", smbTypeId.toString());
+        for (int i = 0; i < param.length; i++) {
+            if (checkLine.contains(param[i])) {
+                biosStrings.put(param[i], checkLine.split(param[i])[1].trim());
+            }
+        }
+        return biosStrings;
     }
 }
--- a/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
@@ -271,9 +271,9 @@ final class SolarisComputerSystem extends AbstractComputerSystem {
      * @param param     contains the biosStrings data to be fetched for
      * @return True if the user of device is empty or the login time is lesser than zero or greater than current time.
      */
-    private static Map<String, String> parseBIOSStrings(String checkLine, Integer smbTypeId, String... param) {
+    private static Map<String, String> parseBIOSStrings(String checkLine, int smbTypeId, String... param) {
         Map<String, String> biosStrings = new HashMap<>();
-        biosStrings.put("smbTypeId", smbTypeId.toString());
+        biosStrings.put("smbTypeId", Integer.toString(smbTypeId));
         for (int i = 0; i < param.length; i++) {
             if (checkLine.contains(param[i])) {
                 biosStrings.put(param[i], checkLine.split(param[i])[1].trim());
--- a/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
@@ -240,7 +240,7 @@ final class SolarisComputerSystem extends AbstractComputerSystem {
                 this.boardManufacturer = Constants.UNKNOWN;
                 this.boardModel = Constants.UNKNOWN;
                 this.boardSerialNumber = Constants.UNKNOWN;
-            } else if (biosStrings.get("smbTypeId") == "2") {
+            } else if ("2".equals(biosStrings.get("smbTypeId"))) {
                 this.manufacturer = Constants.UNKNOWN;
                 this.model = Constants.UNKNOWN;
                 this.serialNumber = Constants.UNKNOWN;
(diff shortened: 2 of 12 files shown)


oshi-core/src/main/java/oshi/util/Util.java
```diff
@@ -83,4 +86,35 @@ public static void freeMemory(Pointer p) {
...
+     * @param smbTypeId The smbTypeID fetched
+     * @param param     contains the biosStrings data to be fetched for
+     * @return True if the user of device is empty or the login time is lesser than zero or greater than current time.
+     */
+    public static Map<String, String> parseBIOSStrings(String checkLine, Integer smbTypeId, String... param) {
+        Map<String, String> biosStrings = new HashMap<>();
+        biosStrings.put("smbTypeId", smbTypeId.toString());
+        for (int i = 0; i < param.length; i++) {
+            if (checkLine.contains(param[i])) {
+                biosStrings.put(param[i], checkLine.split(param[i])[1].trim());
```
dbwiddis, line 115: These multiple array indices impact readability.  I wonder if you could stream the array, filter by the contains, and then collect to a map?
```java
public static Map<String, String> parseBIOSStrings(String checkLine, int smbTypeId, String... param) {
    Map<String, String> biosStrings = Arrays.stream(param)
        .filter(checkLine::contains)
        .collect(Collectors.toMap(p -> p, p -> checkLine.split(p)[1].trim()));
    biosStrings.put("smbTypeId", Integer.toString(smbTypeId));
    return biosStrings;
}
```

dbwiddis (changes requested):
Thanks for your contribution!   I really like the improved readability.

I've made a few additional suggestions.

dbwiddis:
> Haven't changed one suggestion as I wasn't sure if we can modify that or not as it was using the same logic priorly too.

Yeah, that's a conversion of a 64-bit number in 1601 epoch 100-nanosecond ticks to Java's milliseconds-since-1970 epoch.  It's ok to leave as is.  [Here's the actual math it's doing.](https://github.com/java-native-access/jna/blob/e96f30192e9455e7cc4117cce06fc3fa80bead55/contrib/platform/src/com/sun/jna/platform/win32/WinBase.java#L869-L874)

oshi-core/src/main/java/oshi/driver/linux/Who.java
```diff
@@ -49,7 +50,7 @@ public static synchronized List<OSSession> queryUtxent() {
                     String host = ParseUtil.parseUtAddrV6toIP(ut.ut_addr_v6);
                     long loginTime = ut.ut_tv.tv_sec * 1000L + ut.ut_tv.tv_usec / 1000L;
                     // Sanity check. If errors, default to who command line
-                    if (user.isEmpty() || device.isEmpty() || loginTime < 0 || loginTime > System.currentTimeMillis()) {
+                    if (isSessionInvalid(user, device, loginTime)) {
```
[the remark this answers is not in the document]
Faizal17: thanks!

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
```diff
@@ -217,22 +201,66 @@ private static final class SmbiosStrings {
...
+            final String biosDateMarker = "Release Date:";
+            final String biosVersionMarker = "VersionString:";
+
+            final String manufacturerMarker = "Manufacturer:";
+            final String productMarker = "Product:";
+            final String serialNumMarker = "Serial Number:";
+            final String uuidMarker = "UUID:";
+            final String versionMarker = "Version:";
+
+            this.biosVendor = !biosStrings.containsKey(vendorMarker) || Util.isBlank(biosStrings.get(vendorMarker))
```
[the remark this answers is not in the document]
Faizal17: Yes, thanks for the feedback. I have updated all the assignments using this method.

oshi-core/src/main/java/oshi/util/Util.java
```diff
@@ -83,4 +86,35 @@ public static void freeMemory(Pointer p) {
...
+
+    /**
+     * Tests if session of a user logged in a device is valid or not.
+     *
+     * @param user      The user logged in
+     * @param device    The device used by user
+     * @param loginTime The login time of the user
+     * @return True if the user of device is empty or the login time is lesser than zero or greater than current time.
+     */
+    public static boolean isSessionInvalid(String user, String device, Long loginTime) {
```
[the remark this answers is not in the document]
Faizal17: Yes, I have modified the method and its implementation to isSessionValid.

oshi-core/src/main/java/oshi/util/Util.java
```diff
@@ -83,4 +86,35 @@ public static void freeMemory(Pointer p) {
...
+
+    /**
+     * Tests if session of a user logged in a device is valid or not.
+     *
+     * @param checkLine The line received by running the bios command
+     * @param smbTypeId The smbTypeID fetched
+     * @param param     contains the biosStrings data to be fetched for
+     * @return True if the user of device is empty or the login time is lesser than zero or greater than current time.
+     */
+    public static Map<String, String> parseBIOSStrings(String checkLine, Integer smbTypeId, String... param) {
```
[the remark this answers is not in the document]
Faizal17: Thank you, I have moved the method to `SolarisComputerSystem` class

oshi-core/src/main/java/oshi/util/Util.java
```diff
@@ -83,4 +86,35 @@ public static void freeMemory(Pointer p) {
...
+        Map<String, String> biosStrings = new HashMap<>();
+        biosStrings.put("smbTypeId", smbTypeId.toString());
```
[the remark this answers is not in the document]
Faizal17: Thankyou, I updated the data type from Integer to int.

Faizal17:
> Thanks for your contribution! I really like the improved readability.
> 
> I've made a few additional suggestions.

Thank you! I have worked on the suggestions and pushed the updated code accordingly.

rodrodmendez:
pull oshi

dbwiddis (changes requested):
Code changes look great!

Can you please add test cases for the new `ParseUtil` and `Util` methods you've created?

oshi-core/src/main/java/oshi/util/Util.java
dbwiddis: OK, looking at this again, I'm not sure this is right.  While the new code matches the functionality of your initial code, `checkLine` is just a String and we are iterating over all of them, but this method is creating a new map every time after iterating over all the params (so it would only have, at most, one entry for the matching param, plus the id entry).

That seems like a lot of overhead vs. checking each line in an if/else.

I think the approach is valid but instead of passing the single `checkLine` strings you need to pass the whole list from `ExecutingCommand.runNative("smbios")` (or separate the list when you hit the 3 type delimiters and pass each of the 3 sub-lists).  That kind of makes the stream approach not ideal any more as you will be checking all lines against all params.

Faizal17: If we pass the whole list from executing `ExecutingCommand.runNative("smbios")`, Won't we also need to consider a couple of conditions as below?

```
if (checkLine.contains("SMB_TYPE_") && (smbTypeId = getSmbType(checkLine)) == Integer.MAX_VALUE) {
    // If we get past what we need, stop iterating
    break;
} 
// Based on the smbTypeID we are processing for
switch (smbTypeId) { ... }
```
Should I revert back the code to the previously implemented if-else logic?

dbwiddis: I think there's room for a happy medium somewhere. 

The original list has sections.  My thoughts:
1. Create 3 HashMaps, one for each of the SMB_TYPE_x we care about.  (Maybe an array of them?)
2. Collect the command line results and iterate them
3. When you see the `SMB_TYPE_BIOS` marker, make the "biosMap" the current map you're dealing with.
4. Add _all_ the colon-delimited fields you see into the map.
5. When you get to `SMB_TYPE_SYSTEM` switch to the "systemMap" and repeat.  (Maybe change array index?)
6. Same as above with `SMB_TYPE_BASEBOARD`.

Then you have 3 hashmaps and you can just look up from the keys.

Faizal17:
> Code changes look great!
> 
> Can you please add test cases for the new `ParseUtil` and `Util` methods you've created?

Sure. I have added the test cases. Not sure why the code analysis failed. Tried to look the exact reason, but unable to figure it out!

oshi-core/src/main/java/oshi/util/Util.java
Faizal17: I kind of understand what you are trying to explain here. I was thinking if there is any other alternative way. I am thinking of one possible solution. Not sure if it works or not.

1. Suppose we have a list named biosStringsList.
2. While iterating through command line results, we will add each line to the list.
3. At the end, we have a single list having all the required values.
4. We can call the `parseBIOSStrings` with this list, the last smbTypeId value encountered and all the params (including all 3 SMB type parameters)

In this case, the `parseBIOSStrings` function would be like:

```
private static Map<String, String> parseBIOSStrings(List<String> checkLines, int smbTypeId, String... param) {
  Map<String, String> biosStrings = Arrays.stream(param)
    .filter(key -> checkLines.stream().anyMatch(line-> line.startsWith(key)))
    .collect(Collectors.toMap(
      s -> s.substring(0, s.indexOf(":")).trim(),
      s -> s.substring(s.indexOf(":") + 1).trim(),
      (oldValue, newValue) -> newValue));
  biosStrings.put("smbTypeId", Integer.toString(smbTypeId));
  return biosStrings;
}
```

dbwiddis: >  I am thinking of one possible solution. Not sure if it works or not.

It would work if all the strings were unique.

But there are duplicated fields.  Both the System Information and Baseboard have Manufacturer, Product, Version, and Serial Number fields. See the sample output in the comments in the class.  (It's from a VM so they happen to match, but I'm suspecting they are different on a real machine, that I don't have sudo access to test.)

Plus there are more sections than the ones we copy that probably duplicate this as well.

So somehow there needs to be a notion, as we iterate through, of the "last header field we saw" where we only really care about 0, 1, and 2.

dbwiddis: YOu could do 3 command lines `smbios -i 0`, `smbios -i 1`, and `smbios -i 2` but my theory is running the command once is probably faster.

Faizal17: yes, I got your point. So this would definitely fail when there are duplicate fields. As per your suggestion, I worked on having 3 different maps, please verify if now I'm on the same page or still missing something.

```
Map<Integer, List<String>> smbTypesBIOSStringsMap = new HashMap<>();
smbTypesBIOSStringsMap.put(0, new ArrayList<>());
smbTypesBIOSStringsMap.put(1, new ArrayList<>());
smbTypesBIOSStringsMap.put(2, new ArrayList<>());

for (final String checkLine : ExecutingCommand.runNative("smbios")) {
  // Change the smbTypeId when hitting a new header
  if (checkLine.contains("SMB_TYPE_") && (smbTypeId = getSmbType(checkLine)) == Integer.MAX_VALUE) {
      // If we get past what we need, stop iterating
      break;
  }
  // Based on the smbTypeID we are processing for
  smbTypesBIOSStringsMap.get(smbTypeId).add(checkLine);
}
```

And based on the accumulated strings, we can get three different maps.

```
Map<String, String> smbTypeBIOSStrings = parseBIOSStrings(smbTypesBIOSStringsMap.get(0), 0, vendorMarker, biosVersionMarker, biosDateMarker);
Map<String, String> smbTypeSystemStrings = parseBIOSStrings(smbTypesBIOSStringsMap.get(1), 1, manufacturerMarker, productMarker, serialNumMarker, uuidMarker);
Map<String, String> smbTypeBaseboardStrings = parseBIOSStrings(smbTypesBIOSStringsMap.get(2), 2, manufacturerMarker, productMarker, versionMarker, serialNumMarker);
```

And the method parseBIOSStrings would be:

```
private static Map<String, String> parseBIOSStrings(List<String> checkLines, int smbTypeId, String... param) {
  Map<String, String> biosStrings = Arrays.stream(param)
    .filter(key -> checkLines.stream().anyMatch(line-> line.startsWith(key)))
    .collect(Collectors.toMap(
      s -> s.substring(0, s.indexOf(":")).trim(),
      s -> s.substring(s.indexOf(":") + 1).trim(),
      (oldValue, newValue) -> newValue));
  biosStrings.put("smbTypeId", Integer.toString(smbTypeId));
  return biosStrings;
}
```

Is the logic valid and will this work now for all the cases?

dbwiddis: Logic looks great from my perspective!

One improvement: instead of 0, 1, and 2, create an Enum of the SMB_TYPE_X values and use those as a key in an enumMap.  Essentially the same thing but more readable when youre "getting".  See the various PS-type maps in the linux driver package.

dbwiddis: Looking at it again...

Logic works, but you end up iterating multiple times and duplicating the filtering logic with nested streams.  Basically your logic is:
 - iterate the whole list and copy over into smaller lists
 - For each param, iterate the sub-list, filtering by startsWith, and then adding the one thing you get to a map
 - Later for each param, pull out the one thing from the map that you just put in it.

I think it's better if you:
 - Have an enumMap (by `SMB_TYPE_*` key) of maps
 - Iterate the entire list once, starting with a "null" value for the id
   - If the line is a `SMB_TYPE` header, update the id to the right enum value (or null)
   - Otherwise, If the id is not null, parse the string and put it into the map, ignoring whether or not it's one of the params.  Just collect everything.
 - After all that's done, fetch the appropriate values that you need from the different maps using the getOrUnknown() util.

Faizal17: Yes, there is indeed duplication. I have modified the code as per the feedback given. The logic is as below!

1. An Enum having 3 values
```
private enum SMB_TYPE_X {
  BIOS,
  System,
  Baseboard
}
```
2. Creating an enumMap and assigning an empty hashmap as a value to it.
```
SMB_TYPE_X smbTypeId = null;
EnumMap<SMB_TYPE_X, Map<String, String>> smbTypesBIOSStringsMap = new EnumMap<>(SMB_TYPE_X.class);
smbTypesBIOSStringsMap.put(SMB_TYPE_X.BIOS, new HashMap<>());
smbTypesBIOSStringsMap.put(SMB_TYPE_X.System, new HashMap<>());
smbTypesBIOSStringsMap.put(SMB_TYPE_X.Baseboard, new HashMap<>());
```
3. Looping on all the statements and storing them in the respective type map separated as key values (colon delimited). And returning an object of SmbiosStrings from the map generated.
```
for (final String checkLine : ExecutingCommand.runNative("smbios")) {
    if (checkLine.contains("SMB_TYPE_") && (smbTypeId = getSmbType(checkLine)) == null) {
        break;
    }
    String key = checkLine.substring(0, checkLine.indexOf(":")).trim();
    String val = checkLine.substring(checkLine.indexOf(":") + 1).trim();
    smbTypesBIOSStringsMap.get(smbTypeId).put(key, val);
}

// If we get to end and haven't assigned, use fallback
if (!smbTypesBIOSStringsMap.get(SMB_TYPE_X.System).containsKey(serialNumMarker)
        || Util.isBlank(smbTypesBIOSStringsMap.get(SMB_TYPE_X.System).get(serialNumMarker))) {
    smbTypesBIOSStringsMap.get(SMB_TYPE_X.System).put(serialNumMarker, readSerialNumber());
}
return new SmbiosStrings(smbTypesBIOSStringsMap.get(SMB_TYPE_X.BIOS), smbTypesBIOSStringsMap.get(SMB_TYPE_X.System), smbTypesBIOSStringsMap.get(SMB_TYPE_X.Baseboard));
```

Is there any other update to be done? or am I getting it right?

dbwiddis: Sorry if the TYPE_X was clear... was meaning to suggest to fill in the X with the values. Follow Java style for the naming.  `SmbType` for the class name, and all caps enum values `SMB_TYPE_BIOS`, `SMB_TYPE_SYSTEM`, `SMB_TYPE_BASEBOARD`.  These values are ideal as they exactly match the log output we're already testing against!

The `break` here is problematic if we initialize the enum with null.
```java
    if (checkLine.contains("SMB_TYPE_") && (smbTypeId = getSmbType(checkLine)) == null) {
        break;
    }
```
We'd end up not reading anything at all if we start out as null.  The original code using an integer value set it to max integer after we'd iterated.   So I guess having null for both initial and "a value signaling we're done" isn't the best thing to do.   

Ways to solve this:
1. Just change the `break` to a `continue`.  We'll iterate a bunch of strings and do nothing, but with branch prediction the few extra nanoseconds probably won't matter.
2. Use a boolean to indicate whether we've hit the first one or not, and test it. I don't like this option that much.
3. Use a fourth enum value to represent SmbTypes that mark the "we're done" point.  I prefer this to option 2, but that ends up with an uninitialized enumMap value (that we'll never use) which seems a bit code-smelly to me.   

What do you think?

Faizal17: Sorry for the variable name. I will keep that in mind from now onwards. 

Yes, the second option is not too preferable in my opinion too. 

I am not sure I understood when you said `We'd end up not reading anything at all if we start out as null.` If we start with null, it would validate the second condition only if the line contains `SMB_TYPE_`. And if it contains, it would first assign the value to smbTypeId using the getSmbType method and then it will check if it's null or not. If it is a valid line, its value would be from one of the enums, else null. In either case, code would work. But, I agree that the initial condition and termination condition having as same value (null) is not an ideal approach.

Although the third option you specified can be a valid one, we can create a default type enum and use that as a starting value. I also had one more idea. Can we create an enum of key value-based as I defined earlier? and make a map of an integer as a key instead of an enum. That can also survive the previous logic of getting smbTypeID as 0,1,2 and INT_MAX.+

1. The enum would look something like this:

```
private enum SmbType{
  SMB_TYPE_BIOS(0),
  SMB_TYPE_SYSTEM(1),
  SMB_TYPE_BASEBOARD(2),;

  private final int smbTypeId;

  SmbType(int smbTypeId) {
      this.smbTypeId = smbTypeId;
  }
  public Integer getIndex() {
      return this.smbTypeId;
  }
}
```

2. The hashmap would be defined as an integer with the key instead of an enum!

```
int smbTypeId = -1;
Map<Integer, List<String>> smbTypesMap= new HashMap<>();

Integer smbTypeBIOSIndex = SmbType.SMB_TYPE_BIOS.getIndex();
Integer smbTypeSystemIndex = SmbType.SMB_TYPE_SYSTEM.getIndex();
Integer smbTypeBaseboardIndex = SmbType.SMB_TYPE_BASEBOARD.getIndex()

smbTypesMap.put(smbTypeBIOSIndex , new ArrayList<>());
smbTypesMap.put(smbTypeSystemIndex, new ArrayList<>());
smbTypesMap.put(smbTypeBaseboardIndex , new ArrayList<>());
```

3. Now, when we will loop, we will insert based on the integer id, because that's what would be the key of the map or values of the enum in other words.

```
for (final String checkLine : ExecutingCommand.runNative("smbios")) {
  if (checkLine.contains("SMB_TYPE_") && (smbTypeId = getSmbType(checkLine)) == Integer.MAX_VALUE) {
      break;
  }
  String key = checkLine.substring(0, checkLine.indexOf(":")).trim();
  String val = checkLine.substring(checkLine.indexOf(":") + 1).trim();
  smbTypesMap.get(smbTypeId).put(key, val);
}
if (!smbTypesMap.get(smbTypeSystemIndex).containsKey(serialNumMarker)
        || Util.isBlank(smbTypesMap.get(smbTypeSystemIndex).get(serialNumMarker))) {
    smbTypesMap.get(smbTypeSystemIndex).put(serialNumMarker, readSerialNumber());
}
return new SmbiosStrings(smbTypesMap.get(smbTypeBIOSIndex), smbTypesMap.get(smbTypeSystemIndex), smbTypesMap.get(smbTypeBaseboardIndex));
```

The functionality of the code remains the same and it also passes all the logic (If I'm not wrong, lol). And in addition, as we are not using constants as 0, 1, or 2, the code has better readability.

What are your thoughts on this approach?

dbwiddis: > I am not sure I understood when you said `We'd end up not reading anything at all if we start out as null`.

Nevermind!  I made a mistake.  I was thinking that check evaluated on every iteration and we'd break on the first line, but the initial `checkLine.contains("SMB_TYPE_")` conditional makes sure I only break on those lines and when the test fails.  So we're good and you can ignore that entire second part of the comment.

Faizal17: thanks! so should I push the code?

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
```diff
@@ -73,6 +93,8 @@ private static SmbiosStrings readSmbios() {
         String boardVersion = null;
         String boardSerialNumber = null;
 
+        Map<String, String> biosStrings = new HashMap<>();
```
dbwiddis, line 96: As Sonatype says, this is unused. :)

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
```diff
@@ -119,70 +141,48 @@ private static SmbiosStrings readSmbios() {
...
-        int smbTypeId = -1;
-        // Only works with root permissions but it's all we've got
+        SmbType smbTypeId = null;
+
+        EnumMap<SmbType, Map<String, String>> smbTypesMap = new EnumMap<>(SmbType.class);
+        smbTypesMap.put(SmbType.SMB_TYPE_BIOS, new HashMap<>());
+        smbTypesMap.put(SmbType.SMB_TYPE_SYSTEM, new HashMap<>());
+        smbTypesMap.put(SmbType.SMB_TYPE_BASEBOARD, new HashMap<>());
+
+//        // Only works with root permissions but it's all we've got
```
dbwiddis, line 151: stray double comment, let's put it back to original.

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
```diff
@@ -119,70 +141,48 @@ private static SmbiosStrings readSmbios() {
...
-                } else if (checkLine.contains(versionMarker)) {
-                    boardVersion = checkLine.split(versionMarker)[1].trim();
-                } else if (checkLine.contains(serialNumMarker)) {
-                    boardSerialNumber = checkLine.split(serialNumMarker)[1].trim();
-                }
-                break;
-            default:
-                break;
+            if (smbTypeId != null) {
+                String key = checkLine.substring(0, checkLine.indexOf(":")).trim();
```
dbwiddis, line 160: I think you need to test whether `indexOf(":")` is non-negative.

Bonus: you can capture the value in a variable and re-use it on these two lines.

Also I did look up whether string vs. character is better for `indexOf()`. The answer is it's a compiler intrinsic and both are replaced with identical code at compile time so it doesn't matter.

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
```diff
@@ -119,70 +141,48 @@ private static SmbiosStrings readSmbios() {
...
+                smbTypesMap.get(smbTypeId).put(key, val);
             }
         }
+
         // If we get to end and haven't assigned, use fallback
-        if (Util.isBlank(serialNumber)) {
-            serialNumber = readSerialNumber();
+        if (!smbTypesMap.get(SmbType.SMB_TYPE_SYSTEM).containsKey(serialNumMarker)
+                || Util.isBlank(smbTypesMap.get(SmbType.SMB_TYPE_SYSTEM).get(serialNumMarker))) {
+            smbTypesMap.get(SmbType.SMB_TYPE_SYSTEM).put(serialNumMarker, readSerialNumber());
```
dbwiddis, line 169: Can we do this `get(SMB_TYPE_SYSTEM)` just once to avoid this repetition?  Also why not make those enums static imports?

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
```diff
@@ -119,70 +141,48 @@ private static SmbiosStrings readSmbios() {
...
-                boardManufacturer, boardModel, boardVersion, boardSerialNumber);
+        return new SmbiosStrings(smbTypesMap.get(SmbType.SMB_TYPE_BIOS), smbTypesMap.get(SmbType.SMB_TYPE_SYSTEM),
+                smbTypesMap.get(SmbType.SMB_TYPE_BASEBOARD));
     }
 
-    private static int getSmbType(String checkLine) {
+    private static SmbType getSmbType(String checkLine) {
         if (checkLine.contains("SMB_TYPE_BIOS")) {
-            return 0; // BIOS
+            return SmbType.SMB_TYPE_BIOS; // BIOS
```
dbwiddis, line 177: This works and is probably just fine, although one could iterate the enum values and do the contains match in the loop (possibly even in a method inside the enum class).

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
```diff
@@ -217,22 +217,53 @@ private static final class SmbiosStrings {
...
-            this.serialNumber = Util.isBlank(serialNumber) ? Constants.UNKNOWN : serialNumber;
-            this.uuid = Util.isBlank(uuid) ? Constants.UNKNOWN : uuid;
-
-            this.boardManufacturer = Util.isBlank(boardManufacturer) ? Constants.UNKNOWN : boardManufacturer;
-            this.boardModel = Util.isBlank(boardModel) ? Constants.UNKNOWN : boardModel;
-            this.boardVersion = Util.isBlank(boardVersion) ? Constants.UNKNOWN : boardVersion;
-            this.boardSerialNumber = Util.isBlank(boardSerialNumber) ? Constants.UNKNOWN : boardSerialNumber;
+        private SmbiosStrings(Map<String, String> smbTypeBIOSStrings, Map<String, String> smbTypeSystemStrings,
+                Map<String, String> smbTypeBaseboardStrings) {
+            final String vendorMarker = "Vendor:";
```
dbwiddis, line 222: I think we're leaving off the ending colon in our substring parsing?  I may be wrong, though.

dbwiddis (changes requested):
Pretty sure you need to verify the string contains `:` before the substring splitting, otherwise looks good and a few suggestions.

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
Faizal17: yes, we are leaving it. I'll update the code.

oshi-core/src/main/java/oshi/hardware/platform/unix/solaris/SolarisComputerSystem.java
```diff
@@ -217,22 +193,29 @@ private static final class SmbiosStrings {
...
-
-            this.boardManufacturer = Util.isBlank(boardManufacturer) ? Constants.UNKNOWN : boardManufacturer;
-            this.boardModel = Util.isBlank(boardModel) ? Constants.UNKNOWN : boardModel;
-            this.boardVersion = Util.isBlank(boardVersion) ? Constants.UNKNOWN : boardVersion;
-            this.boardSerialNumber = Util.isBlank(boardSerialNumber) ? Constants.UNKNOWN : boardSerialNumber;
+        private SmbiosStrings(Map<String, String> smbTypeBIOSStrings, Map<String, String> smbTypeSystemStrings,
+                Map<String, String> smbTypeBaseboardStrings) {
+            final String vendorMarker = "Vendor";
+            final String biosDateMarker = "Release Date";
+            final String biosVersionMarker = "VersionString";
```
dbwiddis, line 200: This isn't your error, but a bug I found testing your code. Should be a space between Version and String to match the sample output.

Other than that LGTM and ready to merge!

Faizal17: I have updated it. Thank you :)

dbwiddis (approved):

dbwiddis:
Thanks for your contribution!  Very much appreciated!

PR #2347, opened 2023-04-07
Tried some refactoring at architecture level

Faizal17:


oshi-core/src/main/java/oshi/software/common/AbstractOSThread.java
```diff
@@ -20,12 +20,23 @@ public abstract class AbstractOSThread implements OSThread {
 
     private final Supplier<Double> cumulativeCpuLoad = memoize(this::queryCumulativeCpuLoad, defaultExpiration());
 
+    public void setThreadId(int threadId) {
```
dbwiddis, line 23: Good idea to move the threadId to the superclass with a setter, but these fields are set on instantiation in the constructor and don't change during operation, so this setter is probab better `protected`. 

Also, I see you've implemented inheritance for the 4 Unix OS's and Windows, but don't see Linux or macOS.

dbwiddis (commented):
Quick review of the threadId changes.

Could you explain what you're doing on the MBeans? It's not obvious.

PR #2366, opened 2023-05-11
[GHA] Switch from 'dblock' to 'oshi' for coverity

hazendaz:


hazendaz:
@dbwiddis Noticed on a separate repo I have that I had coverity broken.  Took a look to fix as looking into jdk 17 usage with it.  Realized I failed to add COVERITY_TOKEN in that case.  Looked here as I know this has long been still stating dblock from before.  Realized on coverity dashboard I just needed to refresh my repos, that showed OSHI so I added that, got the new token and replaced it.  This PR then brings that in.  Going to merge and see that it works now under proper project naming.

dbwiddis:
Pretty sure I looked at this way back when, and couldn't change the dblock on the account without setting up a whole new account and losing history.

hazendaz:
I know there was a reason we had issues, couldn't recall what but I had switched Waffle over back in 2020.  The old will stay out there.  Does that really matter that much on switch over from that perspective?

dbwiddis:
Nah, now that the pace of changes is tiny I don't think there is as much value in the history. Go for it if you can fix things.

hazendaz:
Sent you an invite as 'owner' as appears that part is needed to see.  Other project I was not setup on states it submitted but haven't seen an update yet :(

hazendaz:
cool thanks!

dbwiddis:
Hmm.

Still no build at [new location](https://scan.coverity.com/projects/oshi-oshi).  Looks like it's waiting for us to submit a build.

[Old location](https://scan.coverity.com/projects/dblock-oshi) shows it was still doing builds as of May 9 (2 days ago) so I'm not sure what was broken?

hazendaz:
Seemed like first run they have some manual process to confirm owners before allowing it.  Good to know it requires another trigger after.  Was not 100% sure last night as oshi one had a different look than another I was working on.  But in both cases appeared waiting on that manual step.

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>

dbwiddis:
> Seemed like first run they have some manual process to confirm owners before allowing it. Good to know it requires another trigger after. Was not 100% sure last night as oshi one had a different look than another I was working on. But in both cases appeared waiting on that manual step. 

Might be something where the "scan" works better if there's a more detailed "build" file or something?

Issue #2373, opened 2023-05-19
Plugin Validation Failing


dbwiddis:
For tests:
```
[WARNING] 
[WARNING] Plugin validation issues were detected in 2 plugin(s)
[WARNING] 
[WARNING]  * org.jacoco:jacoco-maven-plugin:0.8.10
[WARNING]   Declared at location(s):
[WARNING]    * com.github.oshi:oshi-parent:6.4.3-SNAPSHOT (pom.xml) @ line 736
[WARNING]   Used in module(s):
[WARNING]    * com.github.oshi:oshi-parent:6.4.3-SNAPSHOT (pom.xml)
[WARNING]    * com.github.oshi:oshi-core:6.4.3-SNAPSHOT (oshi-core/pom.xml)
[WARNING]   Plugin issue(s):
[WARNING]    * Plugin is a Maven 2.x plugin, which will be not supported in Maven 4.x
[WARNING]    * Plugin mixes multiple Maven versions: [3.0, 2.0.2]
[WARNING]    * Plugin should declare these Maven artifacts in `provided` scope: [org.apache.maven:maven-repository-metadata:3.0, org.apache.maven:maven-artifact:3.0]
[WARNING]    * Plugin depends on plexus-container-default, which is EOL
[WARNING] 
[WARNING]  * pl.project13.maven:git-commit-id-plugin:4.9.10
[WARNING]   Declared at location(s):
[WARNING]    * com.github.oshi:oshi-parent:6.4.3-SNAPSHOT (pom.xml) @ line 598
[WARNING]   Used in module(s):
[WARNING]    * com.github.oshi:oshi-parent:6.4.3-SNAPSHOT (pom.xml)
[WARNING]    * com.github.oshi:oshi-core:6.4.3-SNAPSHOT (oshi-core/pom.xml)
[WARNING]   Plugin issue(s):
[WARNING]    * Plugin should declare these Maven artifacts in `provided` scope: [org.apache.maven:maven-settings-builder:3.0, org.apache.maven:maven-repository-metadata:3.0, org.apache.maven:maven-artifact:3.0, org.apache.maven:maven-plugin-api:3.0, org.apache.maven:maven-settings:3.0, org.apache.maven:maven-aether-provider:3.0, org.apache.maven:maven-model:3.0, org.apache.maven:maven-core:3.0, org.apache.maven:maven-model-builder:3.0]
[WARNING] 
[WARNING] 
[...]

PR #2378, opened 2023-05-21
Update openbsd.yaml to install jdk11

mprins:
The default jdk for maven is 8, but some maven plugins no longer work, the jdk version is specific for the OBSD release, ic. 7.3 ; there is no meta package like eg. "jdk-11"

see. https://github.com/oshi/oshi/issues/2373#issuecomment-1556191016

### DIFF
--- a/.github/workflows/openbsd.yaml
+++ b/.github/workflows/openbsd.yaml
@@ -22,8 +22,11 @@ jobs:
         with:
           usesh: false
           mem: 2048
+          # note default jdk for maven is 8, but some maven plugins no longer work
+          # the version is specific for the OBSD release, ic.7.3 ; there is no meta package like eg. "jdk-11"
           prepare: |
             pkg_add curl
+            pkg_add jdk-11.0.18.10.1p0v0
             pkg_add maven
           run: |
             mvn clean test -B -Djacoco.skip=true


mprins:
not sure the GH action is using the current obsd release; it seems to be using the previous 7.2

there is a PR for 7.3: https://github.com/vmactions/openbsd-vm/pull/9

hazendaz:
Thanks!  I was playing around with trying to use jdk-11 or jdk-11.0 and didn't want to do full string but guess that is what it took.  Assuming this is all worked out ;)

dbwiddis:
Doesn't seem to have worked:
```
  pkg_add jdk-11.0.18.10.1p0v0
  pkg_add maven
  /bin/bash /Users/runner/work/_actions/vmactions/openbsd-vm/v0/run.sh execSSH
  Config file: openbsd-7.2.conf
  Pseudo-terminal will not be allocated because stdin is not a terminal.
  quirks-6.42 signed on 2023-04-06T19:16:59Z
  quirks-6.42 signed on 2023-04-06T19:16:59Z
  Can't find jdk-11.0.18.10.1p0v0
  Couldn't install jdk-11.0.18.10.1p0v0
```
Looking [here](https://cdn.openbsd.org/pub/OpenBSD/snapshots/packages/amd64/) should it be `jdk-11.0.19.7.1p0v0` (or `jdk-17.0.7.7.1p0v0`)?

dbwiddis:
I tried `pkg_add jdk%11` and successfully got `--- +jdk-11.0.16.8.1v0 -------------------`

dbwiddis:
> The special form ‘pkgname%branch’ can be used to restrict matches to a branch matching the [pkgpath(7)](https://man.openbsd.org/pkgpath.7).

So as long as there is exactly one JDK 11 package on the jdk/11 branch this should work

PR #2379, opened 2023-05-21
Use branch matching for OpenBSD JDK 11 lookup

dbwiddis:
The special `package%branch` syntax appears to work to get the correct JDK 11.
`pkg_add jdk%11` installs `jdk-11.0.16.8.1v0`


mprins (approved):
It's new to me, but it's in the man age.  
Learned something after using obsd for more than a decade 

Jdk packages aren't updated very often for a stable release as there's a new stable every 6 months.

dbwiddis:
Happened to catch similar syntax on Stack Overflow for a different question and figured I'd give it a shot!

Issue #2389, opened 2023-06-01
java.lang.UnsatisfiedLinkError: Error looking up function 'gettid': java: undefined symbol: gettid


silencewood:
oshi version is 6.4.2
jna version is 5.13.0
cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 6.5 (Santiago)

Exception in thread "main" java.lang.UnsatisfiedLinkError: Error looking up function 'gettid': java: undefined symbol: gettid
        at com.sun.jna.Function.<init>(Function.java:252)
        at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:620)
        at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:596)
        at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:582)
        at com.sun.jna.Library$Handler.invoke(Library.java:248)
        at com.sun.proxy.$Proxy1.gettid(Unknown Source)
        at oshi.software.os.linux.LinuxOperatingSystem.getThreadId(LinuxOperatingSystem.java:265)

dbwiddis:
OSHI is looking for the [`gettid()`](https://man7.org/linux/man-pages/man2/gettid.2.html) function in your C library and not finding it.

This function is in glibc 2.3 and later.  It was added to musl libc in 1.2.2.

What is your libc version?

dbwiddis:
> Red Hat Enterprise Linux Server release 6.5 (Santiago)

RHEL 6.5 was released in 2013; the latest 6.x is 6.10.  Looks like it's in "extended support" phase with EOL in 2024.

RHEL 6 uses an older version of glibc that doesn't have `gettid()`.

TLDR: OSHI's thread code above is not compatible with very old versions of RHEL/glibc.  

If you'd like to contribute a PR to fetch the information another way to support your old OS, it'd be welcome.

silencewood:
Yes,is libc version problem

[root@test1 bin]# cd /bin
[root@test1 bin]# ldd ls
        linux-vdso.so.1 =>  (0x00007fffb0dff000)
        libselinux.so.1 => /lib64/libselinux.so.1 (0x0000003536c00000)
        librt.so.1 => /lib64/librt.so.1 (0x0000003535c00000)
        libcap.so.2 => /lib64/libcap.so.2 (0x0000003539000000)
        libacl.so.1 => /lib64/libacl.so.1 (0x0000003540000000)
[...]

PR #2393, opened 2023-06-02
fix undefined symbol: gettid #2389

silencewood:
#2389 
gettid
Glibc does not provide a wrapper for this system call; call it using syscall(2).

       #define _GNU_SOURCE
       #include <unistd.h>
       #include <sys/syscall.h>
       #include <sys/types.h>

       int
       main(int argc, char *argv[])
       {
           pid_t tid;

           tid = syscall(SYS_gettid);
       }

### DIFF
--- a/oshi-core/src/main/java/oshi/jna/platform/linux/LinuxLibc.java
+++ b/oshi-core/src/main/java/oshi/jna/platform/linux/LinuxLibc.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2022 The OSHI Project Contributors
+ * Copyright 2019-2023 The OSHI Project Contributors
  * SPDX-License-Identifier: MIT
  */
 package oshi.jna.platform.linux;
@@ -20,6 +20,11 @@ import oshi.jna.platform.unix.CLibrary;
 public interface LinuxLibc extends LibC, CLibrary {
 
     LinuxLibc INSTANCE = Native.load("c", LinuxLibc.class);
+
+    /**
+     * SYS_gettid Defined in /usr/include/asm/unistd_32.h or /usr/include/asm/unistd_64.h
+     */
+    NativeLong SYS_GETTID = new NativeLong(Platform.is64Bit() ? (Platform.isARM() ? 178 : 186) : 224);
 
     /**
      * Return type for getutxent()
@@ -77,20 +82,13 @@ public interface LinuxLibc extends LibC, CLibrary {
     int gettid();
 
     /**
-     * SYS_gettid
-     * Defined in /usr/include/asm/unistd_32.h or /usr/include/asm/unistd_64.h
-     */
-    int SYS_GETTID = Platform.is64Bit() ? (Platform.isARM() ? 178 : 186) : 224;
-
-    /**
-     * syscall()  performs the system call whose assembly language interface has the specified number with the specified
+     * syscall() performs the system call whose assembly language interface has the specified number with the specified
      * arguments.
      *
      * @param number sys call number
-     * @param args sys call arguments
-     * @return The return value is defined by the system call being invoked.  In general, a 0 return value indicates success.   A  -1  return  value  indicates  an
-     *        error, and an error code is stored in errno.
+     * @param args   sys call arguments
+     * @return The return value is defined by the system call being invoked. In general, a 0 return value indicates
+     *         success. A -1 return value indicates an error, and an error code is stored in errno.
      */
-    NativeLong syscall(int number, Object... args);
-
+    NativeLong syscall(NativeLong number, Object... args);
 }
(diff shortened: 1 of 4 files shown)


dbwiddis (changes requested):
Thanks for your contribution!  Comments inline.  Also please add a change log entry.

oshi-core/src/main/java/oshi/jna/platform/linux/LinuxLibc.java
```diff
@@ -67,10 +68,20 @@ class Ut_Tv extends Structure {
     LinuxUtmpx getutxent();
 
     /**
-     * Returns the caller's thread ID (TID). In a single-threaded process, the thread ID is equal to the process ID. In
-     * a multithreaded process, all threads have the same PID, but each one has a unique TID.
+     * SYS_gettid
+     * Defined in /usr/include/asm/unistd_32.h or /usr/include/asm/unistd_64.h
+     */
+    int SYS_GETTID = Platform.is64Bit() ? 186 : 224;
```
dbwiddis, line 74: So it looks like there are 3 options for this, so you'll need another conditional based on `Platform.isARM()`:
 - x86_64: 0xba = 186
 - x86 or arm: 0xe0 = 224
 - arm64: 0xb2 = 178

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -262,7 +262,7 @@ public int getProcessCount() {
 
     @Override
     public int getThreadId() {
-        return LinuxLibc.INSTANCE.gettid();
```
dbwiddis, line 265: One other rather simple fallback which should be cross-platform:  just read `/proc/thread-self` and return the value of `Pid` (which is the tid in the proc/pid/task tree)

dbwiddis:
I played around with some microbenchmarking, with the syscall method vs. resolving the dynamic link of `/proc/thread-self`:
```java
syscall = LinuxLibc.INSTANCE.syscall(code).intValue();
// vs.
proctask = ParseUtil.parseIntOrDefault(Files.readSymbolicLink(link).getFileName().toString(), 0);
```
Not much difference between them, and pretty sure the proc approach is more portable.
```
  Method     TID      Nanos
 syscall   27693   10589.64
proctask   27693   12773.40
```

dbwiddis:
Another test on an x86 system including the gettid native:
```
  Method     TID      Nanos
  gettid  955738     762.50
 syscall  955738    1836.52
proctask  955738    2543.79
```

oshi-core/src/main/java/oshi/jna/platform/linux/LinuxLibc.java
```diff
@@ -73,4 +75,22 @@ class Ut_Tv extends Structure {
...
+    /**
+     * syscall()  performs the system call whose assembly language interface has the specified number with the specified
+     * arguments.
+     *
+     * @param number sys call number
+     * @param args sys call arguments
+     * @return The return value is defined by the system call being invoked.  In general, a 0 return value indicates success.   A  -1  return  value  indicates  an
+     *        error, and an error code is stored in errno.
+     */
+    NativeLong syscall(int number, Object... args);
```
dbwiddis, line 94: The type of `number` needs to be `NativeLong` as well.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -262,7 +274,7 @@ public int getProcessCount() {
...
+        return HAS_GETTID ? LinuxLibc.INSTANCE.gettid() : LinuxLibc.INSTANCE.syscall(LinuxLibc.SYS_GETTID).intValue();
```
dbwiddis, line 277: Can we do one more fallback check here?  Collect this return value and if it's 0, then use  the `/proc/thread-self` symbolic link name.

dbwiddis (changes requested):

dbwiddis (approved):

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
silencewood: Under the redhat linux not have /proc/thread-self
ls /proc/thread-self
ls: cannot access /proc/thread-self: No such file or directory

dbwiddis: Well the syscall should work there so it doesn't get to that third fallback.  That one is necessary on Alpine Linux though.

dbwiddis:
Thanks for your contribution!  I made a few tweaks and merged it. 

Let me know if there are any other changes needed, I'll probably get a release out tomorrow.

dbwiddis:
> * x86_64: 0xba = 186
> * x86 or arm: 0xe0 = 224
> * arm64: 0xb2 = 178

Digging into this again from primary headers and not a summary page, I think this is slightly off.
 - arm64 is 224
 - x86 is 224
 - x86_64 is 186
 - generic (fallback for non-intel/non-arm) is 178

I'll submit a quick fix for this.

PR #2443, opened 2023-09-03
Include IPConnections on Mac that listen on both IPv4 and IPv6 protocols

rieck0:


oshi-core/src/test/java/oshi/software/os/InternetProtocolStatsTest.java
```diff
@@ -119,16 +119,16 @@ void testGetConnections() {
         for (IPConnection conn : ipStats.getConnections()) {
             assertThat("Protocol name is not null or empty", conn.getType(),
                     allOf(is(not(emptyString())), is(notNullValue())));
-            if (conn.getType().contains("6")) {
-                assertThat("Local address array size should be 0, or 16", conn.getLocalAddress().length,
-                        anyOf(is(0), is(16)));
-                assertThat("Foreign address rray size should be 0, or 16", conn.getForeignAddress().length,
-                        anyOf(is(0), is(16)));
-            } else {
+            if (conn.getType().contains("4")) {
```
dbwiddis, line 122: It seems the "46" case would trigger this case. Not sure why it's not failing.  It should probably have its own "46" case here.

dbwiddis (changes requested):
Code changes look good.  Please have a look at the tests and make sure the "tcp46" and "udp46" cases are adequately handled.

Also please add a change log entry!

oshi-core/src/test/java/oshi/software/os/InternetProtocolStatsTest.java
rieck0: It's not failing because for connection types tcp4, udp4, tcp46 and udp46 the ipv4 address is stored in the local address and foreign address properties of the IPConnection instance. The conditional:  

if (conn.getType().contains("4")) 

is meant to handle tcp4, udp4, tcp46 and udp46 types.

dbwiddis: Got it, so if it's 46 it will met the "4" requirement even if it would also meet the "6" requirement.

dbwiddis (approved):

dbwiddis:
Merged! 🎉 

Thanks for your contribution!

PR #2446, opened 2023-09-05
Fix parsing Loongson CPU names

Glavo:
On some platforms, the case of cpuinfo entry names does not match the names hardcoded in oshi.
For example, `/proc/cpuinfo` on linux loongarch64 looks like this:

```
system type             : generic-loongson-machine

processor               : 0
package                 : 0
core                    : 1
global_id               : 0
CPU Family              : Loongson-64bit
Model Name              : Loongson-3A6000
CPU Revision            : 0x00
FPU Revision            : 0x00
CPU MHz                 : 2500.00
BogoMIPS                : 5000.00
TLB Entries             : 2112
Address Sizes           : 48 bits physical, 48 bits virtual
ISA                     : loongarch32 loongarch64
Features                : cpucfg lam ual fpu lsx lasx crc32 complex crypto lvz lbt_x86 lbt_arm lbt_mips
Hardware Watchpoint     : yes, iwatch count: 8, dwatch count: 4
```
This causes oshi to be unable to obtain some CPU information.

This PR fixes the issue.

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
@@ -81,13 +81,13 @@ final class LinuxCentralProcessor extends AbstractCentralProcessor {
                 }
                 continue;
             }
-            switch (splitLine[0]) {
+            switch (splitLine[0].toLowerCase()) {
             case "vendor_id":
-            case "CPU implementer":
+            case "cpu implementer":
                 cpuVendor = splitLine[1];
                 break;
             case "model name":
-            case "Processor": // some ARM chips
+            case "processor": // some ARM chips
                 // May be a processor number. Check for a space.
                 if (splitLine[1].indexOf(' ') > 0) {
                     cpuName = splitLine[1];
@@ -105,26 +105,26 @@ final class LinuxCentralProcessor extends AbstractCentralProcessor {
             case "stepping":
                 cpuStepping = splitLine[1];
                 break;
-            case "CPU variant":
+            case "cpu variant":
                 if (!armStepping.toString().startsWith("r")) {
                     // CPU variant format always starts with 0x
                     int rev = ParseUtil.parseLastInt(splitLine[1], 0);
                     armStepping.insert(0, "r" + rev);
                 }
                 break;
-            case "CPU revision":
+            case "cpu revision":
                 if (!armStepping.toString().contains("p")) {
                     armStepping.append('p').append(splitLine[1]);
                 }
                 break;
             case "model":
-            case "CPU part":
+            case "cpu part":
                 cpuModel = splitLine[1];
                 break;
             case "cpu family":
                 cpuFamily = splitLine[1];
                 break;
-            case "cpu MHz":
+            case "cpu mhz":
                 cpuFreq = ParseUtil.parseHertz(splitLine[1]);
                 break;
             default:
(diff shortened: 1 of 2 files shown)


Glavo:
Well, it looks like I still need a small patch to get it working.
I wish to make a modification like this:

```diff
- // May be a processor number. Check for a space.
- if (splitLine[1].indexOf(' ') > 0) {
-    cpuName = splitLine[1];
- }
+ // Ignore processor number
+ try {
+    Long.parseLong(splitLine[1].trim());
+ } catch (NumberFormatException ignored) {
+        cpuName = splitLine[1];
+ }
```

Is this a safe modification?

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
```diff
@@ -81,13 +81,13 @@ protected ProcessorIdentifier queryProcessorId() {
                 }
                 continue;
             }
-            switch (splitLine[0]) {
+            switch (splitLine[0].toLowerCase()) {
```
mprins, line 84: Make sure this is Locale independent:
```suggestion
            switch (splitLine[0].toLowerCase(Locale.ROOT)) {
```

mprins (changes requested):

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
Glavo: I know, but I see toLowerCase() being used elsewhere, so I'm going to open a new PR to clean up all use cases of toLowerCase()/toUpperCase().

dbwiddis:
> Is this a safe modification?

Depends on your definition of safe. :)

I think it results in the same output, but I personally dislike intentionally using exceptions as process flow.  Is there a specific case where the previous code isn't a sufficient check?

Glavo:
> Is there a specific case where the previous code isn't a sufficient check?

The cpuinfo I gave above is a special case.  The model name is `Loongson-3A6000`, which contains no spaces.

> I think it results in the same output, but I personally dislike intentionally using exceptions as process flow. 

How about using regular expressions to check this?

dbwiddis:
> The model name is `Loongson-3A6000`, which contains no spaces.

Ah, yes, I missed that.  <insert annoyance at Longsoon for altering the standard output>

> How about using regular expressions to check this?

Ugh.  I dislike regex too. :)

I had a look at the JDK source code to see if they did it any better, but it's pretty much character-by-character iteration, multiply by 10, move on throw an exception on bad character.  I'm sure it's pretty efficient but annoying.

Given this is not a "hot path" parsing, I'd probably prefer regex here, more from a philosophical POV than code.

dbwiddis:
Hey @Glavo, no rush, but were you going to implement the regex check you proposed?

Glavo:
> Hey @Glavo, no rush, but were you going to implement the regex check you proposed?

Sorry I didn't reply in time because I'm busy with other work recently.

I've updated this PR to implement it.

dbwiddis (approved):
LGTM, thanks for fixing this!

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxCentralProcessor.java
dbwiddis: Opened https://github.com/oshi/oshi/issues/2451 to track this.

Issue #2455, opened 2023-09-20
JDK 21 is GA!


dbwiddis:
JDK 21 (LTS) is GA.

1. Some of our CI is starting to fail because it's based on -ea.  It should be updated to 22-ea
2. We should update all our test matrices for LTS and LTS-1 (17 and 21 and excluding 11) but open to other maintainer ideas.  Yes, this is a blatant request for input.
3. A year ago when the foreign functions became more accessible, I had an idea that a whole new library was the answer.  Sadly, I've not had the time and only one other contributor has helped out at the oshi-ffm project, so the plan needs reworking. The impacts of this are:
   - I think we should bring the FFI / FFM changes into OSHI via MRJAR, where we can keep the same API but introduce significant performance improvements under the hood in "hot path" cases (primarily process-related code).  I don't think we're ever going to get totally free from JNA, at least in the near future.
   - I'm not sure how to continue to support java 8 under this new approach.  I'm leaning toward a major version bump to 7 and simply maintaining 6.x with bugfixes for java 8.
   - I'm not in a hurry to do any of this, but wanted to at least get my thoughts out there for discussion.

mprins:
Many large projects are a little slow on adopting new a LTS as a baseline.
Looking at GeoServer we are still at Java 11 as a minimum in our development branch, a MRJAR could fix that though

dbwiddis:
> Many large projects are a little slow on adopting new a LTS as a baseline.

Great input, thanks. We also support multiple operating systems which lag behind OpenJDK by some time. Plus, 21 is only 2 years after 17 while there had been a 3-year expectation previously.  We definitely need to keep 11 around just to test MRJAR code < 17 once we go that route, so I guess a blanket policy doesn't need to apply.

I'll take another look at JDK8 with modular MRJAR as well.  When I first looked into it a few years ago I saw it was possible but there were some gotchas.  I'm not sure we need to do anything special to "support" JDK8; I'd still make the "MRJAR" distribution a breaking change/major version bump, but that would open the door to performance improvements on newer JDKs.

dbwiddis:
Closing as I've addressed CI for now.

I'm still pondering how to release FFI-based code.

PR #2456, opened 2023-09-20
Use Java 21 release

mprins:
Java 21 is now released, no more ea, this should be enough to fix the build

### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        java: [11, 17, 20, 21-ea]
+        java: [11, 17, 20, 21]
         os: [ubuntu-latest]
       fail-fast: false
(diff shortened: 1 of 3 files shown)


.github/workflows/linux.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        java: [11, 17, 20, 21-ea]
+        java: [11, 17, 20, 21]
```
dbwiddis, line 20: Can we drop 11 and 20 and add 22-ea?

See #2455, where I beat you by a minute or two. ;)

dbwiddis: Also, this should probably be done for Windows and macOS....

mprins: I'm on mobile, so things don't move as fast...
Java 11 being a current lts,  I'd keep that, no qualms on dropping 20

Issue #2452, opened 2023-09-19
InternetProtocolStats tests broken for dual connections


dbwiddis:
In #2443, [the logic was changed for testing IP stats](https://github.com/oshi/oshi/pull/2443/files#diff-0a53ffe8ba6bb980763391279f42be5744135f718344f8e8a7b329527b0308d4).  It did look a bit suspicious, but tests passed on mac.

But OpenBSD tests are now failing, see https://github.com/oshi/oshi/actions/workflows/openbsd.yaml
```
  Error:  Failures (1):
  Error:    JUnit Jupiter:InternetProtocolStatsTest:testGetConnections()
  Error:      MethodSource [className = 'oshi.software.os.InternetProtocolStatsTest', methodName = 'testGetConnections', methodParameterTypes = '']
  Error:      => java.lang.AssertionError: Local address array size should be 0, or 16
  Error:  Expected: (is <0> or is <16>)
  Error:       but: was <4>
```

The issue here is that there is no "4" in the OpenBSD output, it's either `tcp` or `tcp6`.  

```diff
- if (conn.getType().contains("6")) {
-     assertThat("Local address array size should be 0, or 16", conn.getLocalAddress().length,
-             anyOf(is(0), is(16)));
-     assertThat("Foreign address rray size should be 0, or 16", conn.getForeignAddress().length,
-             anyOf(is(0), is(16)));
- } else {
+ if (conn.getType().contains("4")) {
      assertThat("Local address array size should be 0, or 4", conn.getLocalAddress().length,
              anyOf(is(0), is(4)));
      assertThat("Foreign address rray size should be 0, or 4", conn.getForeignAddress().length,
              anyOf(is(0), is(4)));
+ } else {
+     assertThat("Local address array size should be 0, or 16", conn.getLocalAddress().length,
+             anyOf(is(0), is(16)));
+     assertThat("Foreign address rray size should be 0, or 16", conn.getForeignAddress().length,
+             anyOf(is(0), is(16)));
```
@rieck0 can you update the tests please?  I'd suggest testing if "46" else if "6" else (assume "4")

dbwiddis:
> Fwiw, On obsd it's either ipv4 or ipv6, never together

Right, but the `type` OSHI reports for ipv4 is "tcp" which falls into the "else" branch with the test changes.

PR #2457, opened 2023-09-21
Fix failing unit test on OpenBSD that broke due to InternetProtocolStats changes on Mac

rieck0:
Fixes [https://github.com/oshi/oshi/issues/2452](https://github.com/oshi/oshi/issues/2452)


dbwiddis:
Thanks for the fix!

PR #2466, opened 2023-10-07
defined locale for String#format method

### LINKED ISSUE #2451: Audit and fix locale- and encoding- specific conversions
Coming from https://github.com/oshi/oshi/pull/2446#discussion_r1315664949

 - [x] Search for uppercase / lowercase conversions and make sure they use `Locale.ROOT`.
 - [x] Search for String format and make sure it uses `Locale.ROOT`.
 - [x] Search for String instantiation (`new String(byte[])`) that does not specify an encoding.
 - [x] Add CI checking such as checkstyle, sonar, or forbidden APIs that automates these checks in the future

Taz03:
#2451


dbwiddis (approved):
Thanks again!

PR #2473, opened 2023-10-09
update changes.

Palak4620:


oshi-core/src/main/java/oshi/annotation/concurrent/SuppressForbidden.java
dbwiddis, line 1: 1. Just put this in `oshi.annotation`, not the concurrent subpackage.
2. You'll need to update the import statements where this is used in `oshi.demo`.

pom.xml
```diff
@@ -161,6 +161,25 @@
...
+                    </manifest>
+                    <manifestEntries>
+                        <!-- Add other manifest entries if needed -->
+                    </manifestEntries>
+                </archive>
+                <signaturesFiles>
+                    <signaturesFile><your-signatures-file></signaturesFile>
+                </signaturesFiles>
+            </configuration>
+        </plugin>
```
dbwiddis, line 182: This is the wrong plugin to add.  Go to the `forbiddenapis` plugin here:
https://github.com/oshi/oshi/blob/2a37177fe9c9c5c96f6711b62d7b8c97814f0d09/pom.xml#L746-L770

You'll be adding the `<signatureFiles>` section inside configuration.  See sample POM entry [here](https://github.com/policeman-tools/forbidden-apis/wiki/MavenUsage#maven-usage-instructions).  Format of that file (you can put it in the `config` directory under root) is [here](https://github.com/policeman-tools/forbidden-apis/wiki/SignaturesSyntax#syntax-of-forbidden-signatures).

dbwiddis (changes requested):

oshi-core/src/main/java/oshi/annotation/concurrent/SuppressForbidden.java
Palak4620: okay, I will do that.

dbwiddis:
Superceced by #2474

PR #2477, opened 2023-10-16
forbid parse apis

### LINKED ISSUE #2472: Add forbidden API for number parsing methods
Add a custom Forbidden API check for `Integer.parse()`, `Long.parse()` that suggests using the `ParseUtil` equivalent.

Taz03:
1/2 of #2472


oshi-core/src/main/java/oshi/driver/windows/registry/HkeyPerformanceDataUtil.java
```diff
@@ -272,7 +273,7 @@ private static Map<String, Integer> mapCounterIndicesFromRegistry() {
             String[] counterText = Advapi32Util.registryGetStringArray(WinReg.HKEY_LOCAL_MACHINE, HKEY_PERFORMANCE_TEXT,
                     COUNTER);
             for (int i = 1; i < counterText.length; i += 2) {
-                indexMap.putIfAbsent(counterText[i], Integer.parseInt(counterText[i - 1]));
+                indexMap.putIfAbsent(counterText[i], ParseUtil.parseLastInt(counterText[i - 1], 0));
```
dbwiddis, line 276: I think we can leave this one as-is and just add a `@SuppressForbidden` annotation.  We want to throw an exception here (it's handled on line 282) rather than quietly stick a 0 in the map which would possibly lead to other bugs when that 0 was used.

config/signatures.txt
```diff
@@ -0,0 +1,5 @@
+java.lang.Integer#parseInt(java.lang.String) @ Use oshi.util.ParseUtil#parseLastInt(java.lang.String,int)
```
dbwiddis, line 1: Just sticking this in `config` without mentioning it's for forbidden APIs is confusing.  I'd suggest either including "forbidden" in the file name or putting this file in a "forbidden" subdirectory.

Also all three of these should probably use the `parseFooOrDefault()` as the proposed suggestion (or at least the primary one) rather than the "last" one.

oshi-core/src/main/java/oshi/driver/linux/proc/UpTime.java
```diff
@@ -30,7 +31,7 @@ public static double getSystemUptimeSeconds() {
                 // No space, error
                 return 0d;
             }
-            return Double.parseDouble(uptime.substring(0, spaceIndex));
+            return ParseUtil.parseLastDouble(uptime.substring(0, spaceIndex), 0);
```
dbwiddis, line 34: Since you're doing the default here you don't need to catch the `NumberFormatException` below.  But should use `parseDoubleOrDefault()`.

Also, the compiler probably handles it but I like to include the `d` when defining (double) 0.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -467,6 +467,6 @@ private long getProcessOpenFileLimit(long processId, int index) {
 
         // Split all non-Digits away -> ["", "{soft-limit}, "{hard-limit}"]
         final String[] split = maxOpenFilesLine.get().split("\\D+");
-        return Long.parseLong(split[index]);
+        return ParseUtil.parseLastLong(split[index], 0);
```
dbwiddis, line 470: Looks like we use -1 as the default in the previous lines.

Also use the ..`orDefault()` version.

oshi-core/src/main/java/oshi/software/os/unix/freebsd/FreeBsdOSProcess.java
```diff
@@ -453,6 +453,6 @@ private long getProcessOpenFileLimit(long processId, int index) {
 
         // Split all non-Digits away -> ["", "{soft-limit}, "{hard-limit}"]
         final String[] split = maxOpenFilesLine.get().split("\\D+");
-        return Long.parseLong(split[index]);
+        return ParseUtil.parseLastLong(split[index], 0);
```
dbwiddis, line 456: Same comments as above (orDefault, -1)

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOSProcess.java
```diff
@@ -430,6 +430,6 @@ private long getProcessOpenFileLimit(final long processId, final int index) {
 
         // Split all non-Digits away -> ["", "{soft-limit}, "{hard-limit}"]
         final String[] split = nofilesLine.get().split("\\D+");
-        return Long.parseLong(split[index]);
+        return ParseUtil.parseLastLong(split[index], 0);
```
dbwiddis, line 433: Same comments as above (orDefault, -1)

oshi-core/src/main/java/oshi/util/FileUtil.java
```diff
@@ -300,7 +300,7 @@ public static int getIntFromFile(String filename) {
                 if (LOG.isTraceEnabled()) {
                     LOG.trace(READ_LOG, read.get(0));
                 }
-                return Integer.parseInt(read.get(0));
+                return ParseUtil.parseLastInt(read.get(0), 0);
```
dbwiddis, line 303: Use ... orDefault

oshi-core/src/main/java/oshi/util/ParseUtil.java
```diff
@@ -38,6 +39,7 @@
  * String parsing utility.
  */
 @ThreadSafe
+@SuppressForbidden(reason = "Require parse methods to parse in utility class")
```
dbwiddis, line 42: Can probably narrow this to only use the annotations on the methods that do the parsing rather than the whole class.

dbwiddis (changes requested):
Thanks for the contribution!  A few changes please!

oshi-core/src/main/java/oshi/util/ParseUtil.java
Taz03: There are many methods around 10 so ig it's better to skip the whole class, this way we can prevent duplication as well and since it is parse util class new methods will likely use parse APIs as well

dbwiddis: Fair 'nuff.

oshi-core/src/main/java/oshi/driver/windows/registry/HkeyPerformanceDataUtil.java
Taz03: Done

config/signatures.txt
Taz03: Done

oshi-core/src/main/java/oshi/driver/linux/proc/UpTime.java
Taz03: Done

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
Taz03: Done

oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisOSProcess.java
Taz03: Done

oshi-core/src/main/java/oshi/util/FileUtil.java
Taz03: Done

Taz03:
u didn't answered my question in the issue asking it here again

Do u want to forbid all the parse methods ex {Integer#parse(String), Integer#parse(String, int), Integer#parse(CharSequence, int,  int), same for other classes} or just the one I did?

dbwiddis (approved):
LGTM!  Let me know if you want to iterate on other parse methods here or in a new PR.

dbwiddis:
> u didn't answered my question in the issue asking it here again
> 
> Do u want to forbid all the parse methods ex {Integer#parse(String), Integer#parse(String, int), Integer#parse(CharSequence, int, int), same for other classes} or just the one I did?

I initially wanted to make this easier as a hactoberfest issue, but probably good to do all of the parse methods, and decode as well (I think it just delegates to the appropriate parse with radix method)

Taz03:
> Let me know if you want to iterate on other parse methods here or in a new PR.

lets do it in a separate pr if that's ok with u

PR #2481, opened 2023-10-18
forbid parse apis

### LINKED ISSUE #2472: Add forbidden API for number parsing methods

Taz03:
fixes #2472

### DIFF
--- a/config/forbidden-apis.txt
+++ b/config/forbidden-apis.txt
@@ -4,8 +4,8 @@ java.lang.Integer#parseInt(java.lang.String)
 java.lang.Integer#parseInt(java.lang.String,int)
 java.lang.Integer#parseInt(java.lang.CharSequence,int,int,int)
 
-java.lang.Long#parseLong(java.lang.String) @ Use oshi.util.ParseUtil#parseLongOrDefault(java.lang.String,int)
+java.lang.Long#parseLong(java.lang.String)
 java.lang.Long#parseLong(java.lang.String,int)
 java.lang.Long#parseLong(java.lang.CharSequence,int,int,int)
 
-java.lang.Double#parseDouble(java.lang.String) @ Use oshi.util.ParseUtil#parseDoubleOrDefault(java.lang.String,int)
+java.lang.Double#parseDouble(java.lang.String)
--- a/config/forbidden-apis.txt
+++ b/config/forbidden-apis.txt
@@ -1,4 +1,6 @@
-java.lang.Integer#parseInt(java.lang.String) @ Use oshi.util.ParseUtil#parseIntOrDefault(java.lang.String,int)
+@defaultMessage Use existing oshi.util.ParseUtil methods or create a new one
+
+java.lang.Integer#parseInt(java.lang.String)
 java.lang.Integer#parseInt(java.lang.String,int)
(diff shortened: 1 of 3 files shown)


oshi-core/src/main/java/oshi/hardware/platform/unix/freebsd/FreeBsdCentralProcessor.java
```diff
@@ -207,15 +207,16 @@ private static List<LogicalProcessor> parseTopology() {
 
                     // Regex guarantees parsing digits so we won't get a
                     // NumberFormatException
+                    long parsedVal = ParseUtil.hexStringToLong(firstVal, 0);
```
dbwiddis, line 210: Interesting.... the comments indicate the regex guarantees the parsing won't fail.  But at the same time it's more readable with the hex string parsing. But then when you look at that method it's a lot of manual manipulation when I think the actual parsing-with-radix would handle it!

Anyway, probably ok here.

oshi-core/src/main/java/oshi/util/EdidUtil.java
```diff
@@ -32,6 +33,7 @@ private EdidUtil() {
      * @param edid The EDID byte array
      * @return The manufacturer ID
      */
+    @SuppressForbidden(reason = "using base 2")
```
dbwiddis, line 36:
```suggestion
    @SuppressForbidden(reason = "customized base 2 parsing not in Util class")
```
I think we need to catch a NumberFormatException here?  Do you agree?

dbwiddis (commented):
Initial pass...

oshi-core/src/main/java/oshi/util/EdidUtil.java
Taz03: if we're doing it in other similar places then, yes. Also its out of the scope of this pr I can open a new pr if u want to catch the exception

config/forbidden-apis.txt
```diff
@@ -1,5 +1,9 @@
 java.lang.Integer#parseInt(java.lang.String) @ Use oshi.util.ParseUtil#parseIntOrDefault(java.lang.String,int)
+java.lang.Integer#parseInt(java.lang.String,int)
+java.lang.Integer#parseInt(java.lang.CharSequence,int,int,int)
```
dbwiddis, line 3: See [docs](https://jenkins.thetaphi.de/job/Forbidden-APIs/javadoc/signatures-syntax.html) for syntax of multi-line reasons.
```suggestion
@defaultMessage Use existing oshi.util.ParseUtil methods or create a new one

java.lang.Integer#parseInt(java.lang.String)
java.lang.Integer#parseInt(java.lang.String,int)
java.lang.Integer#parseInt(java.lang.CharSequence,int,int,int)
```

dbwiddis (commented):
Just a suggestion for a single combined message for all these parse methods. 

Also were we going to add `Integer#decode(String)` and `Long#decode(String)`?

Taz03:
> Also were we going to add `Integer#decode(String)` and `Long#decode(String)`?

no?

PR #2487, opened 2023-10-22
Improve performance of thread details query for a single process

### LINKED ISSUE #2483: Getting thread counters for a single process is horribly slow
Coming from https://github.com/oshi/oshi/issues/2482#issuecomment-1773253125

`ThreadPerformanceData.buildThreadMapFromPerfCounters()` takes nearly 3 seconds to complete, even when searching for a single PID, because it gathers all the data first before filtering for process.  It delegates to `ThreadInformation.queryThreadCounters()` without sending the PID, eventually making this call:
https://github.com/oshi/oshi/blob/17c721122dabae6b15c9e3a5b0a90d507b6de10c/oshi-core/src/main/java/oshi/driver/windows/perfmon/ThreadInformation.java#L68-L69

We need to add a new method to `ThreadInformation` which takes a pid argument and changes the filtering criteria from "Not total" to "matches the process".

**"But why didn't we do this before?" you might ask.**

Because we still have a challenge here: the instance names (for which the filtering would have real benefit) for threads do not include the process ID.  They're names like
```
fooprogram/0
fooprogram/1
fooprogram/2
```
And if there are multiple processes with the same name then we get
```
fooprogram/0
fooprogram/1
fooprogram#1/0
fooprogram#1/1
fooprogram#1/2
fooprogram#1/3
fooprogram#2/0
fooprogram#2/1
fooprogram#2/2
```
But that's not always consistent, because if fooprogram#1 terminates, then #2 becomes #1 (or loses the suffix entirely if it's now the only process by that name).  I *think* we could still just query for the process name prefix, we'd just end up with counters for all the same-name processes to filter, but it'd still be better than every single process.

I wish I had a windows machine to develop on... :-|

dbwiddis:
Bypasses registry query for thread details (which can't be filtered) and directly queries performance monitors only for the requested process when getting process thread details.  

```
[INFO] Trying the old way:
[INFO]   Total time: 805ms
[INFO] Trying the new way:
[INFO]   Total time: 47ms
```

Fixes #2483

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ##### Bug fixes / Improvements
 * [#2460](https://github.com/oshi/oshi/pull/2460): Fix AIX tests for virtual/unused drives - [@dbwiddis](https://github.com/dbwiddis).
 * [#2480](https://github.com/oshi/oshi/pull/2480): Use sysfs as a backup for Linux power supply without udev - [@dbwiddis](https://github.com/dbwiddis).
+* [#2487](https://github.com/oshi/oshi/pull/2487): Improve performance of thread details query for a single process - [@dbwiddis](https://github.com/dbwiddis).
 
 # 6.4.0 (2022-12-02), 6.4.1 (2023-03-18), 6.4.2 (2023-05-02), 6.4.3 (2023-06-06), 6.4.4 (2023-07-01), 6.4.5 (2023-08-20), 6.4.6 (2023-09-24)
 
--- a/oshi-core/src/main/java/oshi/driver/windows/perfmon/PerfmonConstants.java
+++ b/oshi-core/src/main/java/oshi/driver/windows/perfmon/PerfmonConstants.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 The OSHI Project Contributors
+ * Copyright 2022-2023 The OSHI Project Contributors
  * SPDX-License-Identifier: MIT
  */
 package oshi.driver.windows.perfmon;
@@ -31,6 +31,7 @@ public final class PerfmonConstants {
     static final String WIN32_PERFPROC_PROCESS_WHERE_IDPROCESS_0 = "Win32_PerfRawData_PerfProc_Process WHERE IDProcess=0";
 
     static final String THREAD = "Thread";
+    static final String WIN32_PERF_RAW_DATA_PERF_PROC_THREAD = "Win32_PerfRawData_PerfProc_Thread";
     static final String WIN32_PERF_RAW_DATA_PERF_PROC_THREAD_WHERE_NOT_NAME_LIKE_TOTAL = "Win32_PerfRawData_PerfProc_Thread WHERE NOT Name LIKE \"%_Total\"";
 
     // For Vista- ... Older systems just have processor #
--- a/oshi-core/src/main/java/oshi/driver/windows/perfmon/PerfmonDisabled.java
+++ b/oshi-core/src/main/java/oshi/driver/windows/perfmon/PerfmonDisabled.java
@@ -24,9 +24,11 @@ public final class PerfmonDisabled {
 
     private static final Logger LOG = LoggerFactory.getLogger(PerfmonDisabled.class);
 
-    static final boolean PERF_OS_DISABLED = isDisabled(GlobalConfig.OSHI_OS_WINDOWS_PERFOS_DIABLED, "PerfOS");
-    static final boolean PERF_PROC_DISABLED = isDisabled(GlobalConfig.OSHI_OS_WINDOWS_PERFPROC_DIABLED, "PerfProc");
-    static final boolean PERF_DISK_DISABLED = isDisabled(GlobalConfig.OSHI_OS_WINDOWS_PERFDISK_DIABLED, "PerfDisk");
+    public static final boolean PERF_OS_DISABLED = isDisabled(GlobalConfig.OSHI_OS_WINDOWS_PERFOS_DIABLED, "PerfOS");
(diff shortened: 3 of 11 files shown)


tomcruise81:
This breaks retrieving `OSThread` data on my machine. It results in an empty list being returned as a result of no longer considering `buildThreadMapFromRegistry` when trying to `getThreadDetails`

dbwiddis:
Are your performance counters disabled?  

You can always just call build thread map from registry and parse the result....

tomcruise81:
I configure the following at app startup:
```java
GlobalConfig.set(GlobalConfig.OSHI_OS_WINDOWS_PERFPROC_DIABLED, false);
GlobalConfig.set(GlobalConfig.OSHI_OS_WINDOWS_PERFOS_DIABLED, false);
GlobalConfig.set(GlobalConfig.OSHI_OS_WINDOWS_CPU_UTILITY, true);
```

Beyond that, I'm not sure how else to ensure that they are NOT disabled.

Thanks.

dbwiddis:
> I configure the following at app startup
Those just suppress log messages that would otherwise appear at startup, informing you that they are disabled.  

Disabling the counters can improve performance, and many gaming programs may recommend you do this (or do it themselves on your behalf without being obvious about it).

You can check whether they are disabled in the registry at `HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\%servicename%\Performance` where `%servicename` is PerfProc for the counters.   Or just call OSHI's `PerfmonDisabled.isDisabled("PerfProc")`.

Digging through the code it looks like if you toggle on the setting `USE_PROCSTATE_SUSPENDED` for the processes, then it will build the thread map using the registry.  As seen above, this may be slow.

tomcruise81:
`PerfmonDisabled.PERF_PROC_DISABLED` returns false for me.

And nothing in `HKEY_LOCAL_MACHINE\System\CurrentControlSet\Services\PerfProc\Performance` is indicating anything being disabled:
```
[HKEY_LOCAL_MACHINE\SYSTEM\ControlSet001\Services\PerfProc\Performance]
"Close"="CloseSysProcessObject"
"Collect"="CollectSysProcessObjectData"
"Collect Timeout"=dword:00001f40
"Library"=hex(2):25,00,53,00,79,00,73,00,74,00,65,00,6d,00,52,00,6f,00,6f,00,\
  74,00,25,00,5c,00,53,00,79,00,73,00,74,00,65,00,6d,00,33,00,32,00,5c,00,70,\
  00,65,00,72,00,66,00,70,00,72,00,6f,00,63,00,2e,00,64,00,6c,00,6c,00,00,00
"Object List"="230 232 786 740 816 1408 1500 1548 1760"
"Open"="OpenSysProcessObject"
"Open Timeout"=dword:00002710
```

tomcruise81:
I forgot to include the `CurrentControlSet` registry entry:
```
[HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\PerfProc\Performance]
"Close"="CloseSysProcessObject"
"Collect"="CollectSysProcessObjectData"
"Collect Timeout"=dword:00001f40
"Library"=hex(2):25,00,53,00,79,00,73,00,74,00,65,00,6d,00,52,00,6f,00,6f,00,\
  74,00,25,00,5c,00,53,00,79,00,73,00,74,00,65,00,6d,00,33,00,32,00,5c,00,70,\
  00,65,00,72,00,66,00,70,00,72,00,6f,00,63,00,2e,00,64,00,6c,00,6c,00,00,00
"Object List"="230 232 786 740 816 1408 1500 1548 1760"
"Open"="OpenSysProcessObject"
"Open Timeout"=dword:00002710
```
Nothing popping out there.

dbwiddis:
> Nothing popping out there.

It would have been an entry "Disable Performance Counters" with a nonzero value.

So there's got to be some reason you're not getting those details.

Despite that, the workaround of setting the config for `USE_PROCSTATE_SUSPENDED` (`GlobalConfig.OSHI_OS_WINDOWS_PROCSTATE_SUSPENDED`) should populate the thread counter blocks when you call `updateAttributes()` on a process object.

tomcruise81:
Thanks for trying. Still no dice.

With the following, I still only get the thread details from the `buildThreadMapFromRegistry` call:

```java
GlobalConfig.set(GlobalConfig.OSHI_OS_WINDOWS_PERFPROC_DIABLED, false);
GlobalConfig.set(GlobalConfig.OSHI_OS_WINDOWS_PERFOS_DIABLED, false);
GlobalConfig.set(GlobalConfig.OSHI_OS_WINDOWS_CPU_UTILITY, true);
GlobalConfig.set(GlobalConfig.OSHI_OS_WINDOWS_PROCSTATE_SUSPENDED, true);

SystemInfo systemInformation = new SystemInfo();
OperatingSystem operatingSystem = systemInformation.getOperatingSystem();
OSProcess currentProcess = operatingSystem.getCurrentProcess();

currentProcess.updateAttributes();

List<OSThread> threadDetails = currentProcess.getThreadDetails();
if (threadDetails == null || threadDetails.isEmpty()) {
    Map<Integer, ThreadPerformanceData.PerfCounterBlock> threads = ThreadPerformanceData
            .buildThreadMapFromPerfCounters(
                    Collections.singleton(currentProcess.getProcessID()),
                    currentProcess.getName(),
                    -1);

    if (threads == null || threads.isEmpty()) {
        threads = ThreadPerformanceData
                .buildThreadMapFromRegistry(Arrays.asList(currentProcess.getProcessID()));
    }

    threadDetails = threads
            .entrySet()
            .parallelStream()
            .map(entry -> {
                return new WindowsOSThread(
                        currentProcess.getProcessID(),
                        entry.getKey(),
                        currentProcess.getName(),
                        entry.getValue());
            })
            .map(OSThread.class::cast)
            .collect(Collectors.toList());
}
```

PR #2524, opened 2023-12-04
Use settings properties for NVD API key

dbwiddis:
Puts the NVD API key in a property in the settings.xml allowing both site deployment and my local release process (with the key hardcoded in my local settings rather than an env variable) to work without any additional configuration.

### DIFF
--- a/.github/workflows/site.yaml
+++ b/.github/workflows/site.yaml
@@ -31,10 +31,11 @@ jobs:
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
       - name: Build site
-        run: ./mvnw site site:stage -DskipTests -DnvdApiKey=${{ secrets.NVD_API_KEY }}
+        run: ./mvnw site site:stage -Djacoco.skip=true -DskipTests -B --settings ./.mvn/settings.xml
         env:
           CI_DEPLOY_USERNAME: ${{ secrets.CI_DEPLOY_USERNAME }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NVD_API_KEY: ${{ secrets.NVD_API_KEY }}
       - name: Deploy Site to gh-pages
         uses: JamesIves/github-pages-deploy-action@v4.5.0
         with:
--- a/.github/workflows/sonatype.yaml
+++ b/.github/workflows/sonatype.yaml
@@ -16,6 +16,7 @@ jobs:
     env:
       CI_DEPLOY_USERNAME: ${{ secrets.CI_DEPLOY_USERNAME }}
       CI_DEPLOY_PASSWORD: ${{ secrets.CI_DEPLOY_PASSWORD }}
+      NVD_API_KEY: ${{ secrets.NVD_API_KEY }}
     steps:
       - uses: actions/checkout@v4
       - name: Set up JDK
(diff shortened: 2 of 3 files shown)


mprins (approved):

hazendaz:
@dbwiddis I say you asked somewhere for my full configuration.  On my local, I simply have a variation of this in my settings.xml

```
    <!-- For NVD API -->
    <server>
      <id>nvd</id>
      <password>...</password>
    </server>
```

The same I put in the project level that needs resolved.  I don't use the properties part.  I haven't actually ran one using site.xml on github yet but trying to get that squared away.  I thought I had done that but I had missed passing the settings.xml and nvd api key as I gained from this pull.  I did recheck my local as the remote I thought I had ran this on stated it was skipping that part.  My local was running fine.

In my pom, I only have the following configuration on this plugin

```
<configuration>
    <failOnError>false</failOnError>
    <nvdApiServerId>nvd</nvdApiServerId>
</configuration>
```

So I'm using the nvd api server id whereas you are using the key.

hazendaz:
I'll have to wait to confirm on the remote this all works on project I'm testing on.  Since I had ran it earlier seems it has some other defect where it just states it was already done even though it had not actually used nvd api at that time.  So I'll check it again tomorrow to be certain it works there too.

dbwiddis:
> So I'm using the nvd api server id whereas you are using the key.

I tried it that way and it didn't work on github actions, see this run: https://github.com/oshi/oshi/actions/runs/7082153920/job/19272494852#step:6:2178

I also previously reported that didn't work in 9.0.1 here: https://github.com/jeremylong/DependencyCheck/issues/6117#issuecomment-1826939658

I may be doing something stupid, of course.

hazendaz:
No worries, will confirm tomorrow.  This is one of those things we will do once then probably forget all about it.

dbwiddis:
I have already pushed this branch to site and it worked: https://github.com/oshi/oshi/actions/runs/7082722744

hazendaz:
Same thing that caught me -> Skipping the NVD API Update as it was completed within the last 30 minutes

dbwiddis:
> Same thing that caught me -> Skipping the NVD API Update as it was completed within the last 30 minutes

That's the second thru fifth one, the first one worked.

Line 2201 (oshi-core check)
```
[INFO] Checking for updates
[INFO] NVD API has 232,110 records in this update
[INFO] Downloaded 10,000/232,110 (4%)
```

Line 3425 (oshi-core aggregate report)
Line 4526 (oshi-demo report)
Line 5574 (oshi-core-java11 report)
Line 5949 (oshi-dist report)
```
[INFO] Checking for updates
[INFO] Skipping the NVD API Update as it was completed within the last 30 minutes
```
]

hazendaz:
I did double check my setup tonight, it ran on GHA using standard maven way without issues.  Likely was due to a combination of issues as to why it didn't work for you before.  Possibly the block of NVD for a while and any of those various versions (now at 9.0.3).  I'm happy with it now :)  And this way works too.

dbwiddis:
Cool.  I really think it's only used in the site publishing (maybe the sonar workflow since we don't specifically exclude it) which only happens normally on releases.... which I'll do as soon as JNA 5.14 is out...

dbwiddis:
Cut a relase today. Only took 2 retries to get it published.  API key is definitely being used but still got 403 response from API around 100K records on the first two tries.   Not going to spend any more time tweaking this but hopefully this new API gets sorted out by all concerned parties soon.

PR #2525, opened 2023-12-04
Update for solaris-vm@v1

Neilpang:
solaris-vm@v1 is ready

### DIFF
--- a/.github/workflows/unix.yaml
+++ b/.github/workflows/unix.yaml
@@ -61,9 +61,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Test on Solaris
         id: test-solaris
-        uses: vmactions/solaris-vm@v0
-        # remove this line when solaris-vm updated to v1
-        continue-on-error: true
+        uses: vmactions/solaris-vm@v1
         with:
           run: |
             wget https://download.bell-sw.com/java/11.0.15.1+2/bellsoft-jdk11.0.15.1+2-solaris-x64-lite.tar.gz -nv


dbwiddis:
Good news: test runs!
Bad news: test is failing based on a `MalformedInputException`.
 - log: https://github.com/oshi/oshi/actions/runs/7089320523/job/19293798675
 - previous issue on Windows that may be related: https://github.com/sormuras/junit-platform-maven-plugin/issues/95

dbwiddis:
> so, what does that mean? 

I don't know!  Previously when it happened on Windows we sourced it to assumptions related to UTF-8 by default on JDK 18+.   But this is happening on JDK 11.

> is that a bug of solaris-vm ?

I don't know and don't presently have the time to investigate that much. It's likely a charset-related issue and either OSHI or the VM needs some environment setting tweaked somewhere.

dbwiddis:
The error is occurring in the last line of this code. Note the `getProperty()` fetch is JDK18+ so isn't relevant here; we're using the default charset.
```java
      if (captureIO) {
        String encoding = System.getProperty("native.encoding"); // Populated on Java 18 and later
        Charset charset = encoding != null ? Charset.forName(encoding) : Charset.defaultCharset();
        try (Stream<String> stdoutput = Files.lines(outputPath, charset);
            Stream<String> erroutput = Files.lines(errorPath, charset)) {
```
The log shows it logging OSHI's usual output and ending early while printing services, could be a service name running with an unexpected character?

dbwiddis:
Fixed by #2527 removing the degree sign. CI passes on Solaris VM.

PR #2562, opened 2024-01-11
Allow opting out of udev usage via GlobalConfig

chadlwilson:
As discussed in https://github.com/oshi/oshi/discussions/2561#discussioncomment-8090621 adds a configuration option `oshi.os.linux.allowudev` (default: `true`) to actively disable use of udev to load more detailed device information.

- If we'd prefer to invert the property to be `_disabled` instead, I can do that too.
- Solaris `allowKstat2` uses a type of camel case format. I can follow that, but other properties don't use that (`logwarning`, `commandline` etc) so bit unsure which format to follow.
- I didn't add a specific test as it'd need a VM fork for the test due to the use of `static` (or some bytecode/classloader magic)

Validated that this does work for the specific use case in the discussion:

**After**
```
jvm 1    | 2024-01-11 12:50:14,683 INFO  [main] AgentStatusHttpd:66 - Agent status HTTP API server running on http://localhost:8152.
jvm 1    | 2024-01-11 12:50:16,428 WARN  [main] LinuxOperatingSystem:92 - Loading of udev not allowed by configuration. Some features may not work.
jvm 1    | 2024-01-11 12:50:16,830 INFO  [main] ThreadPoolTaskScheduler:166 - Initializing ExecutorService 'scheduler'
...
```

**Before** (in my somewhat broken frankenstein Alpine / glibc environment)
```
jvm 1    | 2024-01-11 12:51:59,916 INFO  [main] AgentStatusHttpd:66 - Agent status HTTP API server running on http://localhost:8152.
jvm 1    | #
jvm 1    | # A fatal error has been detected by the Java Runtime Environment:
...
```

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
@@ -46,6 +46,7 @@ import oshi.software.os.OSThread;
 import oshi.util.Constants;
 import oshi.util.ExecutingCommand;
 import oshi.util.FileUtil;
+import oshi.util.GlobalConfig;
 import oshi.util.ParseUtil;
 import oshi.util.platform.linux.ProcPath;
 import oshi.util.tuples.Pair;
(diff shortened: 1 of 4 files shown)


oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
```diff
@@ -79,12 +80,16 @@ public class LinuxOperatingSystem extends AbstractOperatingSystem {
...
+            if (GlobalConfig.get(GlobalConfig.OSHI_OS_LINUX_ALLOWUDEV, true)) {
+                try {
+                    @SuppressWarnings("unused")
+                    Udev lib = Udev.INSTANCE;
+                    hasUdev = true;
+                } catch (UnsatisfiedLinkError e) {
+                    LOG.warn("Did not find udev library in operating system. Some features may not work.");
+                }
+            } else {
+                LOG.warn("Loading of udev not allowed by configuration. Some features may not work.");
```
chadlwilson, line 92: "disallowed" maybe better? or bike-shedding?

dbwiddis: Agree with INFO and "Loading of udev disabled by configuration."

dbwiddis (approved):
LGTM, I'll tweak that log message and merge.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOperatingSystem.java
dbwiddis: Actually given the name, "not allowed" is fine.  Changed it to info.  Will merge when CI passes and cut a release before I go on vacation :)

chadlwilson:
Thanks so much, really appreciate the interaction, response and release! All working perfectly for [GoCD](https://github.com/gocd/gocd/pull/12443)

Issue #2574, opened 2024-02-01
arm linux cannot  get cpuName, use processorIdentifer


13276965576:
use arthas get cpuName is empty.
os info:ubuntu 20.04.5 LTS (Focal Fossa) build 5.15.0-1022-aws
oshi version :6.4.11

[arthas@1317369]$ ognl -c  31cefde0 '@com.yeastar.cloud.dm.cluster.common.helper.OshiHelper@getHardwareLayer().getProcessor().getProcessorIdentifier()'
@ProcessorIdentifier[
    OSHI_ARCHITECTURE_PROPERTIES=@String[oshi.architecture.properties],
    cpuVendor=@String[ARM],
    cpuName=@String[],
    cpuFamily=@String[8],
    cpuModel=@String[0xd0c],
    cpuStepping=@String[r3p1],
    processorID=@String[C1 D0 3F 41 00 00 00 00],
    cpuIdentifier=@String[ARM Family 8 Model 0xd0c Stepping r3p1],
    cpu64bit=@Boolean[false],
    cpuVendorFreq=@Long[2500000000],
    microArchictecture=@[oshi.util.Memoizer$1@770cb057],
]

13276965576:
the /proc/cpuinfo info not found model name，
root@b-1:~# cat /proc/cpuinfo | grep "model name" | uniq
root@b-1:~# cat /proc/cpuinfo 
processor	: 0
BogoMIPS	: 243.75
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x3
CPU part	: 0xd0c
CPU revision	: 1

processor	: 1
BogoMIPS	: 243.75
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x3
CPU part	: 0xd0c
CPU revision	: 1

processor	: 2
BogoMIPS	: 243.75
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x3
[...]

PR #2578, opened 2024-02-07
Use lscpu Model Name as backup for cpuName

13276965576:
fix the cpu names is blank on arm.see #2574


dbwiddis (approved):
Looks great!  Can you please add a change log entry?

PR #2585, opened 2024-02-24
Add GH `macos-14` to teh MacOS buildmatrix

mprins:
According to the blog posts this is now generally available for public repositories and utilises the M1 cpu.

- https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/
- https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/

Let's see how it compares to `flyci-macos-large-latest-m1`

### DIFF
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         # Baseline support, Current LTS, Current GA, Current EA
         java: [11, 17, 21]
-        os: [macos-11, macos-12, macos-13]
+        os: [macos-11, macos-12, macos-13, macos-14]
         include:
           - os: flyci-macos-large-latest-m1
             java: 21


dbwiddis (approved):
Yup, just found out about this myself and would have got around to this eventually!

dbwiddis:
> Let's see how it compares to flyci-macos-large-latest-m1: https://github.com/oshi/oshi/actions/runs/8030710570
>
> the macos-14 job appears to be a tad faster than the flyci-macos-large-latest-m1, 58s vs 1m34s even though GH provides 1 core less, but there is probably some overhead transferring caches and maven artifacts to flyci.

1. FlyCI runs on macOS 13.6 while GHA is macOS 14.2.1.  This in itself is a good reason to continue to use both.
2. Other than the number of logical processors, both VMs look similarly spec'd (same memory, same frequency)
3. Both runs showed near-100% CPU utilization.  The difference in logical processors may just be cosmetic for the container and not actually much different for the underlying architecture.

As for the run times, it actually took the GHA action longer to finish. Here's a side-by-side of timestamps on the runs you conducted:

| Stamp | mm:ss | GHA | FlyCI elapsed | GHA elapsed | 
| ----- | ----- | ----- | ----- | ----- | 
| Waiting for runner | 55:18 | 55:18 |  | | 
| Runner assigned | 55:47 | 57:22 | 29s | 124s | 
| start action | 55:53 | 57:26 | 6s | 4s | 
| java setup | 56:01 | 57:34 | 8s | 8s | 
| cache restored | 56:10 | 57:36 | 9s | 2s | 
| start tests | 57:23 | 57:47 | 73s | 11s | 
| end tests | 57:29 | 58:07 | 6s | 20s | 
| done | 57:45 | 58:35 | 16s | 28s | 
| total time | |  | 02:27 | 03:17 | 

 - FlyCI only took 29s to get assigned a runner, it took over 2 minutes for GHA
 - Cache took longer on FlyCI
 - Maven downloads took a lot longer on FlyCI
 - Tests took over 3x longer to run on GHA 
 - FlyCI finished 50 seconds before GHA did, but took 40 seconds longer internally

One data point is too small a sample... just ran it again.  GHA found a runner in 3 seconds and ran in 1:24, FlyCI took 3:37 to find a runner and finished in 1:46.   Ran again, FlyCI took 1:35 (shorter than previous 2 and comparable to first macos-14 run) after waiting 29s for a runner. GHA took 1:14.

Seems a 20s longer run time is consistent, but does appear associated with maven downloads; the actual running of tests goes faster on FlyCI.

However, GitHub's free plan only allows for 5 concurrent macOS jobs.  Running 3 jdks x 4 macos versions is 12 jobs, so the FlyCI job was finished before the macos-14 JDK 21 job even started. 

TLDR: 20s run time difference isn't a big deal, and may just be network bandwidth.  Waiting for a runner and dealing with the 5 concurrent run limit is a bigger constraint.

PR #2588, opened 2024-02-26
Fix parsing issues

cl728:


oshi-core/src/main/java/oshi/util/ParseUtil.java
```diff
@@ -660,7 +663,7 @@ public static long[] parseStringToLongArray(String s, int[] indices, int length,
         boolean numeric = true;
         boolean numberFound = false; // ignore nonnumeric at end
         boolean dashSeen = false; // to flag uuids as nonnumeric
-        while (--charIndex > 0 && parsedIndex >= 0) {
+        while (charIndex-- > 0 && parsedIndex >= 0) {
```
dbwiddis, line 666: The change in ordering here doesn't make a difference for this conditional, but breaks the `charAt()` call below.

dbwiddis (commented):
I think just adding the trim() should be sufficient.

oshi-core/src/main/java/oshi/util/ParseUtil.java
cl728: update: 
1. use `--charIndex >= 0` instead of `--charIndex > 0` in the while loop condition
2. add a test case

dbwiddis (approved):

dbwiddis:
Thanks for finding this problem and for the fix, @cl728 !

I cut a new release with this fix included.

PR #2605, opened 2024-03-19
Reduce CpuStat.getSystemCpuLoadTicks memory allocation pressure

chrisribble:
Add a new method, FileUtil.readLines, which accepts a number of lines to read and allocates an ArrayList which contains a number of elements corresponding to the number of lines to be read. If file is not readable, return Collections.emptyList() to avoid allocating a list.

Use this new method in CpuStat.getSystemCpuLoadTicks to avoid allocating a String for each line in /proc/stat, instead allocating a single String and ArrayList with a capacity of one.

This reduces allocation of char[]'s to back the String for each line by two orders of magnitude (from a few kB to a few tens of bytes) on each call.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 * Your contribution here!
 
+##### Bug fixes / Improvements
+* [#2605](https://github.com/oshi/oshi/pull/2605): Reduce CpuStat.getSystemCpuLoadticks memory allocation pressure - [@chrisribble](https://github.com/chrisribble).
+
 # 6.5.0 (2024-03-10)
 
 ##### New Features
--- a/oshi-core/src/main/java/oshi/driver/linux/proc/CpuStat.java
+++ b/oshi-core/src/main/java/oshi/driver/linux/proc/CpuStat.java
@@ -32,12 +32,12 @@ public final class CpuStat {
         // first line is overall user,nice,system,idle,iowait,irq, etc.
         // cpu 3357 0 4313 1362393 ...
         String tickStr;
-        List<String> procStat = FileUtil.readFile(ProcPath.STAT);
-        if (!procStat.isEmpty()) {
-            tickStr = procStat.get(0);
-        } else {
+        List<String> procStat = FileUtil.readLines(ProcPath.STAT, 1);
+        if (procStat.isEmpty()) {
             return ticks;
         }
+        tickStr = procStat.get(0);
+
         // Split the line. Note the first (0) element is "cpu" so remaining
         // elements are offset by 1 from the enum index
         String[] tickArr = ParseUtil.whitespaces.split(tickStr);
--- a/oshi-core/src/main/java/oshi/driver/linux/proc/CpuStat.java
+++ b/oshi-core/src/main/java/oshi/driver/linux/proc/CpuStat.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2022 The OSHI Project Contributors
+ * Copyright 2020-2024 The OSHI Project Contributors
  * SPDX-License-Identifier: MIT
  */
 package oshi.driver.linux.proc;
--- a/oshi-core/src/main/java/oshi/util/FileUtil.java
+++ b/oshi-core/src/main/java/oshi/util/FileUtil.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2022 The OSHI Project Contributors
+ * Copyright 2016-2024 The OSHI Project Contributors
  * SPDX-License-Identifier: MIT
  */
 package oshi.util;
@@ -91,11 +91,11 @@ public final class FileUtil {
     }
 
     /**
-     * Read count lines from a file. Intended primarily for Linux /proc filesystem to avoid recalculating file
-     * contents on iterative reads.
+     * Read count lines from a file. Intended primarily for Linux /proc filesystem to avoid recalculating file contents
+     * on iterative reads.
      *
      * @param filename The file to read
-     * @param count The number of lines to read
+     * @param count    The number of lines to read
      * @return A list of Strings representing the first count lines of the file, or an empty list if file could not be
      *         read or is empty
      */
@@ -104,11 +104,11 @@ public final class FileUtil {
     }
 
     /**
-     * Read count lines from a file. Intended primarily for Linux /proc filesystem to avoid recalculating file
-     * contents on iterative reads.
+     * Read count lines from a file. Intended primarily for Linux /proc filesystem to avoid recalculating file contents
+     * on iterative reads.
      *
-     * @param filename The file to read
-     * @param count The number of lines to read
+     * @param filename    The file to read
+     * @param count       The number of lines to read
      * @param reportError Whether to log errors reading the file
      * @return A list of Strings representing the first count lines of the file, or an empty list if file could not be
      *         read or is empty
--- a/oshi-core/src/main/java/oshi/util/FileUtil.java
+++ b/oshi-core/src/main/java/oshi/util/FileUtil.java
@@ -4,6 +4,7 @@
  */
 package oshi.util;
 
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -12,6 +13,7 @@ import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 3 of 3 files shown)


chrisribble:
On my system, `/proc/stat` contains 5927 bytes of data currently:
```
$ cat /proc/stat | wc --bytes
5927
```

However, the first line `/proc/stat` contains a mere 55 bytes:
```
$ cat /proc/stat | head -n 1 | wc --bytes
55
```

I noticed while profiling one of our applications with Java Flight Recorder and Java Mission Control that `CpuStat.getSystemCpuLoadTicks()` - which we invoke twice a second so that we have accurate data about recent CPU usage - was responsible for about 10% of all `byte[]` allocations under light load. This surprised me and motivated this PR.

@dbwiddis I tried to mirror the existing code style while keeping the size of this change as minimal as possible, but let me know if you'd like to see a different approach.

dbwiddis:
Hey @chrisribble thanks a ton for this PR!   Given that CPU ticks are a "hot path" type operation I'm more than happy to try to optimize here.

> I tried to mirror the existing code style

A lot of this file was written in ~2015 with JDK 6 compatibility in mind.  We've been JDK8 minimum for a while, and I wonder if you can't do even better here using [`Files.lines()`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#lines-java.nio.file.Path-java.nio.charset.Charset-) in a stream.  
> Read all lines from a file as a Stream. Unlike [readAllLines](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#readAllLines-java.nio.file.Path-java.nio.charset.Charset-), this method does not read all lines into a List, but instead populates lazily as the stream is consumed.

It may be no better than what you've done, but as long as we're making an improvement, would be nice to compare these options.

chrisribble:
> Hey @chrisribble thanks a ton for this PR! Given that CPU ticks are a "hot path" type operation I'm more than happy to try to optimize here.
> 
> > I tried to mirror the existing code style
> 
> A lot of this file was written in ~2015 with JDK 6 compatibility in mind. We've been JDK8 minimum for a while, and I wonder if you can't do even better here using [`Files.lines()`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#lines-java.nio.file.Path-java.nio.charset.Charset-) in a stream.
> 
> > Read all lines from a file as a Stream. Unlike [readAllLines](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#readAllLines-java.nio.file.Path-java.nio.charset.Charset-), this method does not read all lines into a List, but instead populates lazily as the stream is consumed.
> 
> It may be no better than what you've done, but as long as we're making an improvement, would be nice to compare these options.

@dbwiddis do you just want to explore updating the implementation here to use `Files.lines(file)` to implement `FileUtils.readLines` instead of the imperative implementation currently in the PR?

I wrote a simple JMH benchmark (and enabled the `gc` profiler) for:
1. The old implementation (`Files.readAllLines`)
2. The implementation in this PR (sized `ArrayList` and imperative iteration)
3. `Files.lines(file).limit(count).collect(Collectors.toList())`
4. `Files.lines(file).limit(count).collect(Collectors.toCollection(() -> new ArrayList<>(count)))`

Execution time results:
```
Benchmark                                                                                 Mode  Cnt      Score     Error   Units
OshiBenchmark.readFile                                                            avgt    5     14.070 ±   0.340   us/op
OshiBenchmark.imperative                                                        avgt    5     11.497 ±   0.477   us/op
OshiBenchmark.streamLimitToCollection                                  avgt    5     12.832 ±   0.487   us/op
OshiBenchmark.streamLimitToList                                            avgt    5     12.957 ±   1.022   us/op
```
Allocation rate results:
```
Benchmark                                                                                 Mode  Cnt      Score     Error   Units
OshiBenchmark.readFile:gc.alloc.rate.norm                             avgt    5  32728.020 ±   0.001    B/op
OshiBenchmark.imperative:gc.alloc.rate.norm                        avgt    5  25616.016 ±   0.001    B/op
OshiBenchmark.streamLimitToCollection:gc.alloc.rate.norm  avgt    5  26424.018 ±   0.001    B/op
OshiBenchmark.streamLimitToList:gc.alloc.rate.norm            avgt    5  26464.018 ±   0.002    B/op
```

So the implementation in the PR executes in the least time and allocates the least amount of memory (although the fact that 25k is allocated is definitely something I want to attack in another PR ... I think it's mostly to the default 8k buffer size used by the JRE). I'm happy to change it to use `Files.lines(file).limit(count).collect(Collectors.toList())` though, since it's a pretty tiny amount of extra memory, but I think I will have to handle `UncheckedIOException` if we do that.

Anyway, if you actually wanted to explore exposing a more functional style in the file access classes, I'm happy to discuss that as well. Were you wanting to return `Stream` instead of a `List` so that the caller can control how much data is processed?

One thing to consider if we return a `Stream<String>` is that the caller will be responsible for closing it (to avoid leaking file handles/channels until the `Stream` is GC'd). This also means that the caller may have to deal with `UncheckedIOException` that can be thrown by the underlying `FileChannelLinesSpliterator` when a terminal operator is invoked on the stream (and also potentially when the `Stream` is closed).

For example:
```java
	public static Stream<String> readLines(final String filename, final boolean reportError) {
		Path file = Paths.get(filename);
		if (Files.isReadable(file)) {
			if (LOG.isDebugEnabled()) {
				LOG.debug(READING_LOG, filename);
			}
			try {
				return Files.lines(file);
			} catch (IOException e) {
				if (reportError) {
					LOG.error("Error reading file {}. {}", filename, e.getMessage());
				} else {
					LOG.debug("Error reading file {}. {}", filename, e.getMessage());
				}
			}
		} else if (reportError) {
			LOG.warn("File not found or not readable: {}", filename);
		}
		return Stream.empty();
	}
```

chrisribble:
> > So the implementation in the PR executes in the least time and allocates the least amount of memory
> 
> Thanks for doing the benchmarks! I'm totally fine with merging this PR as-is for this benefit. Goes to prove the "write dumb code" theory that the JIT compiler knows just what to do with this "low level" code.
> 
> My thought on the stream APIs is to just get rid of this util class altogether; it existed in a time before those APIs existed (pretty sure `readAllAlines()` replaced something else). But rather than `Files.lines(file).limit(count).collect(Collectors.toList())` we'd replace individual "just read the first line" implementations with something like `Files.lines(file).limit(count).forEach(s -> doTheThingWith(s))`. Or some util abstraction that we can pass a function to.
> 
> Again, not really worth it unless it gains us benefit.
> 
> Can you add a change log entry and do a quick `mvn spotless:apply`? I'll merge this as-is and we can discuss any other improvements. I really do appreciate you helping improve performance here.

Sounds good. I agree that eliminating the utility class is probably better in the long-run.

I added the changelog entry (hopefully I did it right) and checked in the results of `mvn spotless:apply`.

Btw, I also experimented with specifying a buffer size of 100 bytes and that yields another nice speedup and reduces memory allocation rates as well:
```
Benchmark                                                       Mode  Cnt      Score    Error   Units
OshiBenchmark.readFile                                          avgt    5     13.712 ±  0.205   us/op
OshiBenchmark.readFile:gc.alloc.rate                            avgt    5   2276.259 ± 34.230  MB/sec
OshiBenchmark.readFile:gc.alloc.rate.norm                       avgt    5  32728.019 ±  0.001    B/op
OshiBenchmark.readFile:gc.count                                 avgt    5    158.000           counts
OshiBenchmark.readFile:gc.time                                  avgt    5    105.000               ms
OshiBenchmark.readLinesImperative                               avgt    5     11.581 ±  0.295   us/op
OshiBenchmark.readLinesImperative:gc.alloc.rate                 avgt    5   2108.791 ± 53.969  MB/sec
OshiBenchmark.readLinesImperative:gc.alloc.rate.norm            avgt    5  25608.016 ±  0.001    B/op
OshiBenchmark.readLinesImperative:gc.count                      avgt    5    147.000           counts
OshiBenchmark.readLinesImperative:gc.time                       avgt    5     96.000               ms
OshiBenchmark.readLinesImperativeBufferSize                     avgt    5      9.923 ±  0.482   us/op
OshiBenchmark.readLinesImperativeBufferSize:gc.alloc.rate       avgt    5    905.831 ± 44.118  MB/sec
OshiBenchmark.readLinesImperativeBufferSize:gc.alloc.rate.norm  avgt    5   9424.014 ±  0.001    B/op
OshiBenchmark.readLinesImperativeBufferSize:gc.count            avgt    5     63.000           counts
OshiBenchmark.readLinesImperativeBufferSize:gc.time             avgt    5     44.000               ms
OshiBenchmark.readLinesStreamToCollection                       avgt    5     13.072 ±  0.129   us/op
OshiBenchmark.readLinesStreamToCollection:gc.alloc.rate         avgt    5   1928.262 ± 18.985  MB/sec
OshiBenchmark.readLinesStreamToCollection:gc.alloc.rate.norm    avgt    5  26432.018 ±  0.001    B/op
OshiBenchmark.readLinesStreamToCollection:gc.count              avgt    5    135.000           counts
OshiBenchmark.readLinesStreamToCollection:gc.time               avgt    5     86.000               ms
OshiBenchmark.readLinesStreamToList                             avgt    5     12.868 ±  0.091   us/op
OshiBenchmark.readLinesStreamToList:gc.alloc.rate               avgt    5   1962.367 ± 13.867  MB/sec
OshiBenchmark.readLinesStreamToList:gc.alloc.rate.norm          avgt    5  26480.018 ±  0.001    B/op
OshiBenchmark.readLinesStreamToList:gc.count                    avgt    5    136.000           counts
OshiBenchmark.readLinesStreamToList:gc.time                     avgt    5     91.000               ms
```

I think that just makes the API more obtuse, though, since the caller (the place where you know roughly how much data you're going to read) would need to specify the buffer size. I'm happy to push those changes in a separate PR where we can discuss them?

dbwiddis:
> I'm happy to push those changes in a separate PR where we can discuss them or maybe think some more about a better way to expose "read M bytes from N lines in file foo".

How about opening an issue to discuss, so it's easier to search for later?

Also take a look at the [Auxv](https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/driver/linux/proc/Auxv.java) class (which reads `/proc/auxv` as binary).  Seems there's something similar we could do with "read until we hit `n` newlines" rather than a fixed number of bytes.

dbwiddis:
One other idea. Old school, but why can't we just go old school and use `Scanner`?  It has a 1KB buffer rather than 8K.  And we could just read in the values into a `long[]` while we scan.....

PR #2612, opened 2024-03-30
Use 1k buffer in FileUtils.readLines to reduce heap allocation pressure

chrisribble:


chrisribble:
JMH benchmark results with various buffer sizes:
```
Benchmark                                      Mode  Cnt      Score     Error   Units
OshiBenchmark.readLines1k                      avgt    5     10.945 ±   1.473   us/op
OshiBenchmark.readLines1k:gc.alloc.rate        avgt    5    985.187 ± 134.199  MB/sec
OshiBenchmark.readLines1k:gc.alloc.rate.norm   avgt    5  11296.015 ±   0.002    B/op
OshiBenchmark.readLines1k:gc.count             avgt    5     69.000            counts
OshiBenchmark.readLines1k:gc.time              avgt    5     32.000                ms
OshiBenchmark.readLines2k                      avgt    5     10.561 ±   0.053   us/op
OshiBenchmark.readLines2k:gc.alloc.rate        avgt    5   1204.201 ±   6.095  MB/sec
OshiBenchmark.readLines2k:gc.alloc.rate.norm   avgt    5  13336.015 ±   0.001    B/op
OshiBenchmark.readLines2k:gc.count             avgt    5     84.000            counts
OshiBenchmark.readLines2k:gc.time              avgt    5     53.000                ms
OshiBenchmark.readLines4k                      avgt    5     10.928 ±   0.203   us/op
OshiBenchmark.readLines4k:gc.alloc.rate        avgt    5   1521.972 ±  28.238  MB/sec
OshiBenchmark.readLines4k:gc.alloc.rate.norm   avgt    5  17440.015 ±   0.001    B/op
OshiBenchmark.readLines4k:gc.count             avgt    5    106.000            counts
OshiBenchmark.readLines4k:gc.time              avgt    5     69.000                ms
OshiBenchmark.readLines512                     avgt    5     10.273 ±   0.148   us/op
OshiBenchmark.readLines512:gc.alloc.rate       avgt    5    952.852 ±  13.679  MB/sec
OshiBenchmark.readLines512:gc.alloc.rate.norm  avgt    5  10264.014 ±   0.001    B/op
OshiBenchmark.readLines512:gc.count            avgt    5     67.000            counts
OshiBenchmark.readLines512:gc.time             avgt    5     48.000                ms
OshiBenchmark.readLines8k                      avgt    5     12.006 ±   0.063   us/op
OshiBenchmark.readLines8k:gc.alloc.rate        avgt    5   2034.124 ±  10.731  MB/sec
OshiBenchmark.readLines8k:gc.alloc.rate.norm   avgt    5  25608.017 ±   0.001    B/op
OshiBenchmark.readLines8k:gc.count             avgt    5    142.000            counts
OshiBenchmark.readLines8k:gc.time              avgt    5     95.000                ms
```

The gains from going below 1k appear to be very small, but the delta from the default 8k to 1k are big: about `25 kB -> 11 kB` per invocation.

oshi-core/src/main/java/oshi/util/FileUtil.java
```diff
@@ -119,7 +124,9 @@ public static List<String> readLines(String filename, int count, boolean reportE
             if (LOG.isDebugEnabled()) {
                 LOG.debug(READING_LOG, filename);
             }
-            try (BufferedReader reader = Files.newBufferedReader(file, StandardCharsets.UTF_8)) {
+            CharsetDecoder decoder = StandardCharsets.UTF_8.newDecoder();
```
dbwiddis, line 127: Are we sure the files we are reading will always be in UTF-8?

We don't currently read any files in Windows, but the util should be generic.  And while procsys files are going to be a defined (ASCII) format, we also read various `/etc/` and other files which in theory could use a different character set.

oshi-core/src/main/java/oshi/util/FileUtil.java
```diff
@@ -119,7 +124,9 @@ public static List<String> readLines(String filename, int count, boolean reportE
...
+            try (Reader isr = new InputStreamReader(Files.newInputStream(file), decoder);
+                    BufferedReader reader = new BufferedReader(isr, BUFFER_SIZE)) {
                 List<String> lines = new ArrayList<>(count);
```
dbwiddis, line 130: Since we're micro-optimizing, we really should handle the "read one line" case separately to avoid the List overhead. No need to handle that in this PR but something to consider!

dbwiddis (commented):
LGTM with one question

oshi-core/src/main/java/oshi/util/FileUtil.java
chrisribble: Yes, I think that it's a bit problematic to have generic utility classes that read text files without allowing the caller to specify the character set with which to interpret the bytes, but that's where we are now, since that's what `FileUtil.readFile(String)` did previously (assumed UTF-8).

The previous PR provides an implementation that allows for reading a specific number of lines as UTF-8 strings (following roughly how `Files.readAllLines(Path, Charset)` is implemented.

This PR expands the same thing that `Files.newBufferedReader(Path, Charset)` does, but with a 1k buffer.

I think it would potentially make sense to have different types of reader classes that understand how to process specific types of files (for example a class for binary `/proc` files and a class for text `/proc` files, and a class for each different types of file in `/etc`, etc).

oshi-core/src/main/java/oshi/util/FileUtil.java
chrisribble: I agree. However, the savings from avoiding the list allocation are small (< 100 bytes) compared to the savings from reducing the buffer size (14kB).

dbwiddis (approved):

Issue #2689, opened 2024-07-16
Seeing [ERROR] Error reading file /proc/2045181/cmdline. null in log


dbwiddis:
### Discussed in https://github.com/oshi/oshi/discussions/2685

<div type='discussions-op-text'>

<sup>Originally posted by **wolfs** July 10, 2024</sup>
I am seeing
```
[ERROR] Error reading file /proc/2045181/cmdline. null
[ERROR] Error reading file /proc/1975607/cmdline. null
[ERROR] Error reading file /proc/1962012/cmdline. null
```
in my logs. I think this is because I am calling `OSProcess.queryArguments()`, that uses `FileUtil.readAllBytes()` and the process file is already gone.

https://github.com/oshi/oshi/blob/7023f06b5606d07a7f6dbf6d730beeb05c4f1a9d/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java#L129-L132

I think the call should be changed to `FileUtil.readAllBytes(String.format(Locale.ROOT, ProcPath.PID_CMDLINE, getProcessID()))), false)`, so it doesn't print out on the error log when reading the file fails.

It seems like `OSProcess.queryCommandline()` already calls a method that does not log on error:
https://github.com/oshi/oshi/blob/7023f06b5606d07a7f6dbf6d730beeb05c4f1a9d/oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java#L118-L121</div>

dbwiddis:
Thanks for the report, @wolfs!

We'd welcome a PR to fix this!

wolfs:
I created https://github.com/oshi/oshi/pull/2692 for doing so.

PR #2692, opened 2024-07-17
Do not log errors for reading process arguments on Linux

wolfs:
When the process is gone, there shouldn't be any error logging in the command line.

Fixes https://github.com/oshi/oshi/issues/2689

### DIFF
--- a/oshi-core/src/main/java/oshi/util/FileUtil.java
+++ b/oshi-core/src/main/java/oshi/util/FileUtil.java
@@ -155,17 +155,6 @@ public final class FileUtil {
      * Read an entire file at one time. Intended primarily for Linux /proc filesystem to avoid recalculating file
      * contents on iterative reads.
      *
-     * @param filename The file to read
-     * @return A byte array representing the file
-     */
-    public static byte[] readAllBytes(String filename) {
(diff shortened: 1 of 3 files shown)


oshi-core/src/main/java/oshi/util/FileUtil.java
```diff
@@ -151,17 +151,6 @@ public static List<String> readLines(String filename, int count, boolean reportE
...
     }
 
-    /**
-     * Read an entire file at one time. Intended primarily for Linux /proc filesystem to avoid recalculating file
-     * contents on iterative reads.
-     *
-     * @param filename The file to read
-     * @return A byte array representing the file
-     */
-    public static byte[] readAllBytes(String filename) {
```
wolfs, line 161: 💭 This method was now unused, and I think it is better to remove it since you should always think whether or not you want to log errors.

dbwiddis: Yeah, this method was a relic from pre-Java8 days.  The JDK has an equivalent if we ever need it again.

dbwiddis (approved):
Thanks for this fix!

PR #2709, opened 2024-08-20
Replace FlyCI macOS runners with GitHub macOS runners

tsonev7:
This PR replaces the use of FlyCI macOS runners with GitHub ones since FlyCI macOS runners will be discontinued effective Sep 30, 2024. 

[Read more about the discontinuation of the FlyCI macOS runners](https://flyci.net/blog/flyci-discontinue-macos-runners?utm_source=site_link&utm_medium=github&utm_campaign=discontinue-runners).

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -16,7 +16,7 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
         java: [11, 17, 21]
         include:
-          - os: flyci-macos-large-latest-m1
+          - os: macos-latest
             java: 21
       fail-fast: false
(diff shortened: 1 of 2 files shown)


mprins (approved):
thanks, lgtm

.github/workflows/pr.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
         java: [11, 17, 21]
         include:
-          - os: flyci-macos-large-latest-m1
+          - os: macos-latest
```
dbwiddis, line 19: `macos-latest` is already included in the above matrix so this seems to duplicate it.

We might consider using `macos-13` since `-latest` is now aarch64 and `-13` is still x86...

PR #2711, opened 2024-08-20
Do not log error on macOS for hw.nperflevels

### LINKED ISSUE #2710: Oshi logs error on macOS for hw.nperflevels
On some CI macOS agents, oshi logs
```
Failed sysctl call: hw.nperflevels, Error code: 2
```

You can see this in [this build scan](https://ge.androidx.dev/s/273vfft7mit3a/console-log?page=1#L1) of the AndroidX project. The error doesn't have any effect on how oshi is working, only the information is not as accurate as possible.

Oshi should not log this error when obtaining CPU information about the machine at the error level or not at all.

wolfs:
when the systctl call fails.

Fixes #2710

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/platform/mac/MacCentralProcessor.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/mac/MacCentralProcessor.java
@@ -147,7 +147,7 @@ final class MacCentralProcessor extends AbstractCentralProcessor {
             pkgCoreKeys.add((pkgId << 16) + coreId);
         }
         Map<Integer, String> compatMap = queryCompatibleStrings();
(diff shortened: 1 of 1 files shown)


dbwiddis:
Thanks for the contribution.  You missed the latest release by an hour :)

wolfs:
Thank you for the quick turnaround!

> You missed the latest release by an hour :)

Pity! Any idea when the next release will go out?

dbwiddis:
> Any idea when the next release will go out?

By request, generally, or after a few weeks without a request.  For a bug fix I'd probably push it out sooner but this is just a logging annoyance so I may wait a week or two for any other changes.

PR #2722, opened 2024-09-06
Fix speed value for LinuxNetworkIF

Puppy4C:
The output result of  `/sys/class/net/<card>/speed` is in Mbps and needs to be converted to bps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # 6.6.4 (in progress)
 
 * Your contribution here!
+##### Bug fixes / Improvements
+* [#2722](https://github.com/oshi/oshi/pull/2722): Fix speed value for LinuxNetworkIF - [@Puppy4C](https://github.com/Puppy4C).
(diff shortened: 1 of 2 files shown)


mprins:
> The output result of /sys/class/net/<card>/speed is in Mbps 

yes

> and needs to be converted to bps

because: https://github.com/oshi/oshi/blob/0ae5c65613e7b41752b83b7bfec0c24914482c2a/oshi-core/src/main/java/oshi/hardware/NetworkIF.java#L273

but this will result in new/unexpected output in `#toString()`

Puppy4C:
Of course

dbwiddis:
Thanks. The irony here is that I should have known this as one of my more popular Stack Overflow answers [is this one](https://stackoverflow.com/questions/17872257/how-do-we-determine-whether-a-size-is-base-on-1000-or-1024-when-we-see-kb-or-mb/36992829#36992829). ;)

PR #2726, opened 2024-09-15
JNA 5.15.0

dbwiddis:
Update JNA dependency and remove (some) temporary code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * [#2722](https://github.com/oshi/oshi/pull/2722): Fix speed value for LinuxNetworkIF - [@Puppy4C](https://github.com/Puppy4C).
 * [#2724](https://github.com/oshi/oshi/pull/2724): Clarify IO bytes documentation on OSProcess - [@dbwiddis](https://github.com/dbwiddis).
 * [#2725](https://github.com/oshi/oshi/pull/2725): Reduce redundant logging on perf counter failures - [@dbwiddis](https://github.com/dbwiddis).
+* [#2726](https://github.com/oshi/oshi/pull/2726): JNA 5.15.0 - [@dbwiddis](https://github.com/dbwiddis).
 
 # 6.6.0 (2024-04-13), 6.6.1 (2024-05-26), 6.6.2 (2024-07-21), 6.6.3 (2024-08-20)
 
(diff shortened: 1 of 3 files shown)


jthalmai:
The removed method IsProcessorFeaturePresent() is still called in oshi.hardware.platform.windows.WindowsCentralProcessor.java

dbwiddis:
> The removed method IsProcessorFeaturePresent() is still called in oshi.hardware.platform.windows.WindowsCentralProcessor.java

Yep.  I tried to get rid of it but it wasn't trivial and wasn't worth the time to deal with.

The problem (if it is a problem) is that I implemented the processor features as an enum in OSHI, allowing me to iterate over `ProcessorFeature.values()`.  The JNA equivalent of the (native) enum is simply a declaration of integer constants, with the "name" only available by reflection or reading the source code.  Given the main purpose of this code is to display an actual (human readable) list of features, I decided keeping the enum was probably the best option, at least for now.

If you've got a better idea, I'm open to it.

jthalmai:
The problem is that the other class fails with a NoSuchMethodError. If you remove a method, you should also adapt the code that calls it.

dbwiddis:
> The problem is that the other class fails with a NoSuchMethodError. If you remove a method, you should also adapt the code that calls it.

But it's inherited.  The calling code in `WindowsCentralProcessor` calls `oshi.jna.platform.windows.Kernel32` which `extends com.sun.jna.platform.win32.Kernel32` where the method is declared.  I can see the code working correctly here (classpath): https://github.com/oshi/oshi/actions/runs/10875123481/job/30173174640#step:4:1135 and here (module path): https://github.com/oshi/oshi/actions/runs/10875123481/job/30173174640#step:4:1891

Can you be more clear how you're seeing this error?

While I agree it could be considered a "code smell" and I might want to fully qualify the superclass when it's called, I don't think it causes an exception.

jthalmai:
In our project the call 'new oshi.SystemInfo().getHardware().getProcessor().getLogicalProcessorCount();' fails with the error.

dbwiddis:
Do you have the latest version of JNA (specifically jna-project) as a dependency?  `NoSuchMethod` might occur if you're inheriting from an earlier version of JNA.   Use maven or gradle to show you which dependencies you are pulling in.  See https://github.com/oshi/oshi/blob/master/src/site/markdown/FAQ.md#how-do-i-resolve-jna-noclassdeffounderror-or-nosuchmethoderror-issues

jthalmai:
Thanks, that was it. I don't know why it wasn't updated the first time as well, but now we also have the latest JNA version.

dbwiddis:
Dependency resolution is sometimes a mystery, particularly on gradle. Maven is a bit easier to predict!

Issue #2791, opened 2025-01-13
The CPU temperature obtained on the Windows platform is always an incorrect value of 27.850000000000023


pandalxb:
Running the OHM software indirectly to obtain the temperature is a bad way. It is best to call DLL to obtain the temperature of sensors such as CPU on Windows Platform. for example https://github.com/profesorfalken/jSensors which the author has already solved it . I hope to fix this issue as soon as possible

pandalxb:
It's running on the Windows platform with the wrong value 27.850000000000023.

dbwiddis:
> It's running on the Windows platform with the wrong value 27.850000000000023.

This is documented in [the API](https://www.oshi.ooo/oshi-core-java11/apidocs/com.github.oshi/oshi/hardware/Sensors.html)
> Windows information is retrieved via Windows Management Instrumentation (WMI). Unfortunately, most hardware providers do not publish values to WMI. Oshi attempts to retrieve values from [Open Hardware Monitor](https://openhardwaremonitor.org/) if it is running, in preference to the Microsoft API, which may require elevated permissions and still may provide no results or unchanging results depending on the motherboard manufacturer.

> Running the OHM software indirectly to obtain the temperature is a bad way. It is best to call DLL to obtain the temperature of sensors such as CPU on Windows Platform.

I do not have WIndows hardware and am unable to do this myself.

>  I hope to fix this issue as soon as possible

I hope this is you volunteering to submit a PR!  It would be welcome.

Note that https://github.com/oshi/oshi/issues/2695 suggests using **LibreHardwareMonitor** in preference to OHM.

pandalxb:
> I hope this is you volunteering to submit a PR! It would be welcome.
> 
> Note that [#2695](https://github.com/oshi/oshi/issues/2695) suggests using **LibreHardwareMonitor** in preference to OHM.
Thanks for your encouragement and suggestion. I have submitted PR #2814 to solve the problem. This is my first time submitting a PR on GitHub, forgive me for my lack of experience and submitting many duplicate build failures PR. Fortunately, after testing multiple times on my own fork, I finally succeeded in building it.

pandalxb:
Looking forward to your reply.

PR #2793, opened 2025-01-14
Enable configurable prioritization of thermal zones

joerg1985:
This PR will add a priority to the thermal zones, in the past the last termal zone found was reported as CPU temp.
In my local system this resolves to the sensor on the M2 disk, this might be the reason for #2791 too.

The priority is defined in a local list and unmapped types are treated as lowest priority.
In case there are multiple sensor with the same priority the last one does win, like in the past.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 6.6.6 (in progress)
 
 * Your contribution here!
+* [#2793](https://github.com/oshi/oshi/pull/2793), Enable configurable prioritization of thermal zones - [@joerg1985](https://github.com/joerg1985).
 
 # 6.6.0 (2024-04-13), 6.6.1 (2024-05-26), 6.6.2 (2024-07-21), 6.6.3 (2024-08-20), 6.6.4 (2024-09-15), 6.6.5 (2024-09-16)
 
--- a/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxSensors.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxSensors.java
@@ -13,11 +13,15 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.function.ToIntFunction;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import oshi.annotation.concurrent.ThreadSafe;
 import oshi.hardware.common.AbstractSensors;
 import oshi.util.ExecutingCommand;
 import oshi.util.FileUtil;
+import oshi.util.GlobalConfig;
 import oshi.util.ParseUtil;
 import oshi.util.platform.linux.SysPath;
 
@@ -27,6 +31,14 @@ import oshi.util.platform.linux.SysPath;
 @ThreadSafe
 final class LinuxSensors extends AbstractSensors {
 
+    public static final String OSHI_THERMAL_ZONE_TYPE_PRIORITY = "oshi.os.linux.sensors.cpuTemperature.types";
+
+    private static final List<String> THERMAL_ZONE_TYPE_PRIORITY = Stream
+            .of(GlobalConfig.get(OSHI_THERMAL_ZONE_TYPE_PRIORITY, "").split(","))
+            .filter((s) -> !s.isEmpty())
+            .collect(Collectors.toList());
+
+    private static final String TYPE = "type";
     // Possible sensor types. See sysfs documentation for others, e.g. current
     private static final String TEMP = "temp";
     private static final String FAN = "fan";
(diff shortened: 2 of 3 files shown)


dbwiddis (commented):
Thanks for this PR!  

I think this is a great addition but I'd like to see it configurable.  Can you make that change?

Also please add a change log entry.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxSensors.java
```diff
@@ -27,6 +31,12 @@
 @ThreadSafe
 final class LinuxSensors extends AbstractSensors {
 
+    private static final List<String> THERMAL_ZONE_TYPE_PRIORITY = Arrays.asList(
+            "x86_pkg_temp", // lowest priorty
+            "cpu-thermal" // highest priorty
+    );
```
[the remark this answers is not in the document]
joerg1985: I have updated the PR, thanks for the fast response!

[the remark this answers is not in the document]
dbwiddis: Excellent! Looks great at first glance, will give a proper review tonight.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxSensors.java
```diff
@@ -84,27 +95,69 @@ private void populateSensorsMapFromHwmon() {
...
+                            if (index != -1) {
+                                return index;
+                            }
+
+                            break;
+                        }
+                    }
+
+                    return THERMAL_ZONE_TYPE_PRIORITY.size();
+                });
```
dbwiddis, line 115:
```suggestion
                files -> Arrays.stream(files)
                    .filter(file -> file.getName().equals(TYPE))
                    .findFirst()
                    .map(file -> {
                        String type = FileUtil.getStringFromFile(file.getPath()).trim();
                        return THERMAL_ZONE_TYPE_PRIORITY.indexOf(type);
                    })
                    .filter(index -> index >= 0)
                    .orElse(THERMAL_ZONE_TYPE_PRIORITY.size()));
```

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxSensors.java
```diff
@@ -84,27 +95,69 @@ private void populateSensorsMapFromHwmon() {
...
+        getSensorFilesFromPath(sensorPath, sensor, sensorFileFilter, (files) -> 0);
+    }
+
+    /**
+     * Find all sensor files in a specific path and adds them to the sensorsMap
+     *
+     * @param sensorPath       A string containing the sensor path
+     * @param sensor           A string containing the sensor
+     * @param sensorFileFilter A FileFilter for detecting valid sensor files
+     * @param priorizer        A callback to priorize between multiple sensors
```
dbwiddis, line 135:
```suggestion
     * @param prioritizer      A callback to prioritize between multiple sensors
```
Typo

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxSensors.java
```diff
@@ -84,27 +95,69 @@ private void populateSensorsMapFromHwmon() {
...
+     */
+    private void getSensorFilesFromPath(String sensorPath, String sensor, FileFilter sensorFileFilter, ToIntFunction<File[]> priorizer) {
+        Map<Integer, String> byPriority = new TreeMap<>();
```
dbwiddis, line 138: While this works, we know we only want to keep the lowest priority `path` at the end.  So why not just have a single `String` variable for the `path`, store its priority, and update it if the new priority is lower?

Something like 
```java
    String selectedPath = null;
    int lowestPriority = Integer.MAX_VALUE;

    for ( ... ) {
    // existing code to get matching files for each sensor path

        if (matchingFiles != null && matchingFiles.length > 0) {
            int priority = prioritizer.applyAsInt(matchingFiles);
            if (priority < lowestPriority) {
                lowestPriority = priority;
                selectedPath = path;
            }
        }
    }

    if (selectedPath != null) {
        this.sensorsMap.put(sensor, String.format(Locale.ROOT, "%s/%s", selectedPath, sensor));
    }
```

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # 6.6.6 (in progress)
 
 * Your contribution here!
+* [#2793](https://github.com/oshi/oshi/pull/2793), use a priorization in thermal zones
```
dbwiddis, line 4:
```suggestion
* [#2793](https://github.com/oshi/oshi/pull/2793): Enable configurable prioritization of thermal zones - [@joerg1985](https://github.com/joerg1985).
```

oshi-core/src/main/resources/oshi.properties
```diff
@@ -267,3 +267,7 @@ oshi.os.openbsd.filesystem.path.excludes=/tmp**,/dev
 
 oshi.os.solaris.filesystem.path.excludes=/system**,/tmp**,/dev,/dev/fd**
 oshi.os.solaris.filesystem.volume.excludes=rpool
+
+# List of types inside /sys/class/thermal/thermal_zone*/type to detect as CPU temperature.
+# The order of entries in the list does represent the priority, the first match will be used.
+oshi.os.linux.sensors.cpuTemperature.types=cpu-thermal,x86_pkg_temp
```
dbwiddis, line 273: Nit: please add a final line break to end the file.  Apparently we didn't include the `.properties` extension in our `pom.xml` spotless configuration :)

dbwiddis (commented):
Looking good. A few readability/simplification suggestions and I think you need to handle the empty string edge case better.

pandalxb:
could you solve the problem when it's running on Windows platform without OHM running that the CPU temperature is the same value 27.850000000000023 ??

joerg1985:
@dbwiddis i have updated the PR, with small deviations.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxSensors.java
```diff
@@ -27,6 +31,14 @@
 @ThreadSafe
 final class LinuxSensors extends AbstractSensors {
 
+    public static final String OSHI_THERMAL_ZONE_TYPE_PRIORITY = "oshi.os.linux.sensors.cpuTemperature.types";
+
+    private static final List<String> THERMAL_ZONE_TYPE_PRIORITY = Stream
+            .of(GlobalConfig.get(OSHI_THERMAL_ZONE_TYPE_PRIORITY, "").split(","))
+            .filter((s) -> !s.isEmpty())
+            .collect(Collectors.toList());
```
dbwiddis, line 39: Nice approach!

dbwiddis (approved):
Looks great!  Will merge as soon as CI finishes.

Issue #2803, opened 2025-02-06
EdidUtil reads bits in getManufacturerID()


JonathanTheDev:
I'm working on an application where I'm using the manufacturer id to help identify monitors. I got the right id 2 of my monitors, but one of my monitors reported GCM instead of GSM. I tested that GSM was the proper value by checking here: https://people.freedesktop.org/~imirkin/edid-decode/

Debugging the code that decodes the manufacturer id, I noticed that letter two and three are both missing a bit when reading the edid.
Heres the split up code for the letters:
First: `(char)(64 + Integer.parseInt(temp.substring(1, 6), 2))`
Second: `(char)(64 + Integer.parseInt(temp.substring(7, 11), 2))`
Third: `(char)(64 + Integer.parseInt(temp.substring(12, 16), 2))`

The second and third letter are beginning their substring one letter to far to the right. They should start at at 6 and 11 instead of 7 and 12.
Changing this fixes the problem and makes i report the right manufacturer id.

Kind regards,
Jonathan B. Jørgensen

dbwiddis:
Nice catch. I agree with your analysis.

Can you please submit a PR to fix this?  Might be useful to change the EdidUtilTest `EDID_MANUFID` value to what you're seeing in your actual EDID.

JonathanTheDev:
I will give it a try :) I've never submitted a PR before, but this seems like a nice easy way to start.

dbwiddis:
> I will give it a try :) I've never submitted a PR before, but this seems like a nice easy way to start.

That is exactly how I got started at this project.  See #23 :)

PR #2805, opened 2025-02-11
Fixes manufacturer ID parsing logic in getManufacturerID()

JonathanTheDev:
It used to miss a few bits on the second and third, see issue #2803. 

This pull request changes the substring in the method to include the right bits.


dbwiddis:
Thanks for this PR.  I’m on vacation with limited internet this week.  Looks good.  Can you add a changelog please?

PR #2811, opened 2025-02-13
Fetch CPU temperature, Fan speeds and CPU voltage by jLibreHardwareMonitor

pandalxb:
Fetch CPU temperature, Fan speeds and CPU voltage by directly digging into the library LibreHardwareMonitorLib.dll(.NET 4.7.2 and above) or OpenHardwareMonitorLib.dll(.NET 2.0) without applications running on Windows platform.
See issue #2791 
This pull request supports a new method to fetch sensors values in WindowsSensors.

### DIFF
--- a/config/import-control-test.xml
+++ b/config/import-control-test.xml
@@ -18,6 +18,12 @@
         pkg="com\.sun\.(jna|jna\..*)"
         regex="true" />
 
+    <!-- jlibrehardwaremonitor -->
+    <allow pkg="io.github.pandalxb.jlibrehardwaremonitor" />
+
+    <!-- apache -->
+    <allow pkg="org.apache" />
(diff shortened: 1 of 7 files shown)


hazendaz:
@pandalxb Please don't spam us.  You can test on your own fork.  If what you are doing is not ready for oshi, that is ok but opening all these pull requests is spamming us with emails with no apparent value.

pandalxb:
> @pandalxb Please don't spam us. You can test on your own fork. If what you are doing is not ready for oshi, that is ok but opening all these pull requests is spamming us with emails with no apparent value.

I am so sorry for not realizing that it would send you a large amount of spam emails. I will build my own fork on appveyor until it successes.

PR #2829, opened 2025-02-27
Modify API description of Sensors

pandalxb:
Add a link to jLibreHardwareMonitor so users know how to include this dependency and remove the duplicate description of getCpuTemperature about OHM.

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/Sensors.java
+++ b/oshi-core/src/main/java/oshi/hardware/Sensors.java
@@ -28,6 +28,8 @@ public interface Sensors {
      * CPU Temperature
      *
      * @return CPU Temperature in degrees Celsius if available, 0 or {@link Double#NaN} otherwise.
+     *         <p>
+     *         See notes on <a href="https://www.oshi.ooo/oshi-core-java11/apidocs/com.github.oshi/oshi/hardware/Sensors.html">Sensors</a>.
      */
     double getCpuTemperature();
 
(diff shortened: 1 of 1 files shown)


oshi-core/src/main/java/oshi/hardware/Sensors.java
```diff
@@ -28,11 +28,6 @@ public interface Sensors {
      * CPU Temperature
      *
      * @return CPU Temperature in degrees Celsius if available, 0 or {@link Double#NaN} otherwise.
-     *         <p>
-     *         On Windows, if not running Open Hardware Monitor, requires elevated permissions and hardware BIOS that
```
dbwiddis, line 32: Rather than deleting, maybe "not using optional jLibreHardwareMonitor dependency or running..." ?

Or maybe "See notes on " and a link to the Sensors class javadoc?

pandalxb: Done.

dbwiddis: I think rather than an HTML link there is a `{@link Sensors}` option?

pandalxb: Sure, I'm sorry for my insufficient experience with Javadoc.

dbwiddis: We're all learning!

dbwiddis:
New javadocs are live: https://www.oshi.ooo/oshi-core-java11/apidocs/com.github.oshi/oshi/hardware/Sensors.html

Issue #2834, opened 2025-03-03
The new non-optional dependencies on io.github.pandalxb.jlibrehardwaremonitor packages in the OSGi metadata are problematic


merks:
The just-released 6.7.0 version specifies additional new package requirements compared to the 6.6.0 version:

![Image](https://github.com/user-attachments/assets/6c853111-8ad5-48cb-a298-68bab7b9ff3b)

I have a feeling this is accidental because the *.pom specifies it as an optional dependency:

![Image](https://github.com/user-attachments/assets/67fc73f1-75ea-469e-a841-65285e0d07e3)

Moreover that dependency:

https://repo1.maven.org/maven2/io/github/pandalxb/jLibreHardwareMonitor/1.0.5/

is not available as an OSGi artifact:

![Image](https://github.com/user-attachments/assets/2a29f094-20ae-4366-b56f-3b69dc0046c1)

so the newly released artifact is not easily usable in an OSGi runtime.

dbwiddis:
Fixed in #2835

merks:
Thank you for the release of the new version:

https://repo1.maven.org/maven2/com/github/oshi/oshi-core/6.7.1/

PR #2835, opened 2025-03-03
Configure bnd-maven-plugin to make librehardwaremonitor optional

merks:
https://github.com/oshi/oshi/issues/2834

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -323,7 +323,7 @@
                     <version>${bnd-maven-plugin.version}</version>
                     <configuration>
                         <bnd><![CDATA[Export-Package: oshi.*;-noimport:=true;-split-package:=merge-first
-                        Import-Package: ${osgi.slf4j.import.packages}, *
+                        Import-Package: io.github.pandalxb.jlibrehardwaremonitor.*;resolution:=optional, ${osgi.slf4j.import.packages}, *
                         Bundle-SymbolicName: ${project.groupId}.${project.artifactId}
                         -snapshot: SNAPSHOT]]></bnd>
                     </configuration>


merks:
FYI, I built it locally with this change and it produces this result:

image

merks:
@ruspl-afed

FYI, I ran into this issue updating Orbit in preparation for the 4.36 / 2025-06 release cycle where currently we/you are using 6.6.0:

https://github.com/eclipse-orbit/orbit-simrel/blob/e35c5e8fba5e9475e5902fd5840b8f17a3fd126c/maven-osgi/tp/Maven.target#L227-L232

mprins (approved):
LGTM, fwiw, Maven dependency management already has this defined as optional: https://github.com/oshi/oshi/blob/15fd9d48133971899d4649663e6bef013e876f44/oshi-core/pom.xml#L49

It seems the bnd-maven-plugin doesn't respect the pom file though

laeubi:
> It seems the bnd-maven-plugin doesn't respect the pom file though

First a package being optional is completely different from a maven jar marked as optional.

**In OSGi**, optional is a runtime claim and says you can work completely without it and you have taken actions to handle the absence of that code, this is really hard work and most often wrong. Usage of this likely indicates that your bundle is not having properly modularization.

**In Maven**, optional means _something completely different that one intuitively would assume_, and **is most likely used wrong here** see the [example from the docs](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html):

> Suppose there is a project named X2 that has similar functionality to Hibernate. It supports many databases such as MySQL, PostgreSQL, and several versions of Oracle. Each supported database requires an additional dependency on a driver jar. All of these dependencies are needed at compile time to build X2. However your project only uses one specific database and doesn't need drivers for the others. X2 can declare these dependencies as optional, so that when your project declares X2 as a direct dependency in its POM, all the drivers supported by the X2 are not automatically included in your project's classpath. **Your project will have to include an explicit dependency on the specific driver** for the one database it does use.

So (maven) optional effectively means:

1. You have some kind of API (examples are slf4j, JPA, JMS, ...)
2. There are multiple providers (eg. logback, hibernate, activemq) and you want to bind in your build to one of them
3. The users of your jar **have to choose one of them** but not necessarily the one you chosen
4. The project choice will be marked `optional`, the consumer must include it as `runtime` (not optional!)

If you want to express that downstream consumers don't need to include the dependency, then mark it as `provided`, or if it is needed to run your product only use `runtime` in all cases **don't use optional**!

mprins:
We use `optional` in the Maven sense of (quoting that page) 
> ... The idea is that some of the [runtime] dependencies are only used for certain features in the project and will not be needed if that feature isn't used. Ideally...

I don't see the difference with your osgi description

merks:
@mprins 

It's hard to say why you don't see a difference.  I did a Maven build locally and can definitely see the impact of the changes on the generated MANIFEST.MF:

image

Where/how are you looking?

mprins:
@merks I was talking about the remarks made by @laeubi

dbwiddis (approved):
LGTM and sorry for missing this.  Is there a deadline to get this merged/released?

dbwiddis:
Responding to @laeubi:

> In OSGi, optional is a runtime claim and says you can work completely without it and you have taken actions to handle the absence of that code,

That is precisely what we are doing here. If that dependency is not present, we query WMI tables, either the built-in Microsoft ones (which tend to not work) or WMI tables populated by a running instance of OHM program (which we otherwise would not know existed). 

> this is really hard work and most often wrong

It's been the way the code has worked for the last several years. The only difference here is that if we find the new LHM "driver" classes on the classpath, then we can use that new "driver" rather than the default setup.

Further, the description for `optional` that you cite is exactly our usage. It's effectively a driver for sensor information.

> Suppose there is a project named X2 that has similar functionality to Hibernate. 

OSHI is the project here.

> It supports many databases such as MySQL, PostgreSQL, and several versions of Oracle. 

It supports LHM via DLL, OHM via DLL, OHM via WMI, and BIOS via WMI.

> Each supported database requires an additional dependency on a driver jar.

The first two (DLL) require an additional dependency on a driver jar.

>  All of these dependencies are needed at compile time to build X2. 

The LHM dependency is needed at compile time to build OSHI.

> However your project only uses one specific database and doesn't need drivers for the others. 

The LHM dependency only uses one of the two DLLs when present; the "drivers" for the other two are COM/WMI which is already present in OSHI for other usages.

Compare this to [the docs for `provided`](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope)

> This is much like compile, but indicates you expect the JDK or a container to provide the dependency at runtime. For example, when building a web application for the Java Enterprise Edition, you would set the dependency on the Servlet API and related Java EE APIs to scope provided because the web container provides those classes. 

We do not expect the JDK or a container to provide the LHM dependency.  That situation would indicate no downstream user action is required, because someone/something else is providing the dependency.   But in our case, the user needs to explicitly define the dependency themselves to include it. It is not provided by a container or any other program.

I still believe `optional` is correct here.  Can you cite other docs than the ones you have which indicate otherwise?

dbwiddis:
Further docs [from here](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html):

> The idea is that some of the dependencies are only used for certain features in the project and will not be needed if that feature isn't used. Ideally, such a feature would be split into a sub-module that depends on the core functionality project.

and 

> If a user wants to use functionality related to an optional dependency, they have to redeclare that optional dependency in their own project. 

and 

> Optional dependencies save space and memory. They prevent problematic jars that violate a license agreement or cause classpath issues from being bundled into a WAR, EAR, fat jar, or the like.

merks:
> LGTM and sorry for missing this. Is there a deadline to get this merged/released?

Nothing to be sorry about! For consumption by eclipse orbit I’ve simply omitted the problematic version. When you release a new version at your convenience the analyzer will incorporate it.  

Thanks to all for your responsiveness and for providing useful technology.

PR #2836, opened 2025-03-04
Linux ARM 64 runner

mprins:
Add a Linux arm64 / Java 17 combo to the build matrix

These are now available in "Public Preview": https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -19,6 +19,9 @@ jobs:
       matrix:
         java: [11, 17, 21]
         os: [ubuntu-latest]
+        include:
+          - java: 17
+            os: 'ubuntu-24.04-arm'
       fail-fast: false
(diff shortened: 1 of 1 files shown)


mprins (commented):
workflow run: https://github.com/oshi/oshi/actions/runs/13654499074 (passed)

mprins:
In theory we could also try out running some unix vm's on arm64, eg. OpenBSD has `aarch64` available: https://github.com/vmactions/openbsd-vm/tree/main?tab=readme-ov-file#6-select-architecture though for now it seems that it does not use native support "3. Support Linux ARM runner." as listed in https://github.com/vmactions/openbsd-vm/tree/main?tab=readme-ov-file#upcoming-features

hazendaz (approved):

Issue #2826, opened 2025-02-26
Optimize calls to netstat by using /proc files


dbwiddis:
Coming from https://github.com/oshi/oshi/discussions/2433:
> I am curious if there has ever been any discussion about optimizing the Linux network metrics by utilizing the various /proc files rather than executing `netstat` and parsing its output? It looks like for the most part on Linux that `netstat` is parsing the /proc files itself and just reformatting it. The execution of `netstat` results in a noticeable latency in sampling these metrics that is not seem direct parsing of /proc files. 
>
> If it is just a matter of lack of contribution I would love to contribute in any way but don't want to tread on something that may be intentionally implemented in a particular way.

_Originally posted by @jake-at-work in https://github.com/oshi/oshi/discussions/2433_

rohan-coder02:
Hi @dbwiddis,

Is it ok if I could present my analysis and code changes done to optimize calls to netstat by using /proc files?

Below is my analysis and code changes.

**OSHI Existing Code Analysis:**

- OSHI originally collected TCPv4, UDPv4, and UDPv6 statistics using netstat commands, as seen in [LinuxInternetProtocolStats.java](https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/software/os/linux/LinuxInternetProtocolStats.java)

```
@Override
public TcpStats getTCPv4Stats() {
    return NetStat.queryTcpStats("netstat -st4");
}

@Override
public UdpStats getUDPv4Stats() {
    return NetStat.queryUdpStats("netstat -su4");
}

@Override
public UdpStats getUDPv6Stats() {
    return NetStat.queryUdpStats("netstat -su6");
}
```

**Linux /proc Analysis for Optimized Data Retrieval:**

- Instead of executing netstat, we directly read from the /proc filesystem, which provides the same information with lower latency
- I have validated the outputs from various flavors of Linux such as Redhat, Amazon Linux 2, etc

_Sample Output of /proc/net/snmp (TCPv4 & UDPv4):_

```
cat /proc/net/snmp
Tcp: RtoAlgorithm RtoMin RtoMax MaxConn ActiveOpens PassiveOpens AttemptFails EstabResets CurrEstab InSegs OutSegs RetransSegs InErrs OutRsts InCsumErrors
Tcp: 1 200 120000 -1 65229 30305 1 17865 1 2004091 2251354 1617 0 8815 0
[...]

PR #2846, opened 2025-03-09
Optimize netstat calls using /proc files (Issue #2826)

rohan-coder02:
### Title:

Optimize calls to netstat by using /proc files (Issue #2826)

### Description:

This PR optimizes the retrieval of network statistics by replacing external netstat calls with direct file reads from /proc/net/snmp (TCPv4 and UDPv4) and /proc/net/snmp6 (UDPv6).

### Changes Made:

- Replaced netstat calls with parsing /proc/net/snmp (TCPv4 and UDPv4) and /proc/net/snmp6 (UDPv6)
- Updated LinuxInternetProtocolStats.java to extract TCP and UDP statistics 

### Testing Done:

- Verified that the extracted TCP/UDP statistics match the output from cat /proc/net/snmp and cat /proc/net/snmp6
- Confirmed correctness by comparing values before and after changes
- Tested on various flavors of Linux such as Redhat, Amazon Linux 2, etc by creating a simple Java App and printing the TCP and UDP stats

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/linux/LinuxInternetProtocolStats.java
+++ b/oshi-core/src/main/java/oshi/software/os/linux/LinuxInternetProtocolStats.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2024 The OSHI Project Contributors
+ * Copyright 2020-2025 The OSHI Project Contributors
  * SPDX-License-Identifier: MIT
  */
 package oshi.software.os.linux;
@@ -18,14 +18,12 @@ import static oshi.software.os.InternetProtocolStats.TcpState.TIME_WAIT;
 import static oshi.software.os.InternetProtocolStats.TcpState.UNKNOWN;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
 
 import oshi.annotation.concurrent.ThreadSafe;
 import oshi.driver.linux.proc.ProcessStat;
-import oshi.driver.unix.NetStat;
 import oshi.software.common.AbstractInternetProtocolStats;
 import oshi.util.FileUtil;
 import oshi.util.ParseUtil;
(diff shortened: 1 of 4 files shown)


oshi-core/src/main/java/oshi/software/os/linux/LinuxInternetProtocolStats.java
```diff
@@ -36,19 +38,103 @@
...
+
     @Override
     public TcpStats getTCPv4Stats() {
-        return NetStat.queryTcpStats("netstat -st4");
+        byte[] fileBytes = FileUtil.readAllBytes(ProcPath.SNMP, true);
+        String content = new String(fileBytes);
+        List<String> lines = Arrays.asList(content.split("\\n"));
+        Map<TcpStat, Long> tcpData = new EnumMap<>(TcpStat.class);
+
+        for (int line = 0; line < lines.size() - 1; line++) {
```
dbwiddis, line 60: Given the format of this file with pairs of headers/values, there should probably be an iteration by 2 here.  You're relying on a whole row of numbers throwing an exception on the enum `valueOf()` here.
```
cat /proc/net/snmp
Tcp: RtoAlgorithm RtoMin RtoMax MaxConn ActiveOpens PassiveOpens AttemptFails EstabResets CurrEstab InSegs OutSegs RetransSegs InErrs OutRsts InCsumErrors
Tcp: 1 200 120000 -1 65229 30305 1 17865 1 2004091 2251354 1617 0 8815 0
Udp: InDatagrams NoPorts InErrors OutDatagrams RcvbufErrors SndbufErrors InCsumErrors IgnoredMulti
Udp: 249873 2 0 249940 0 0 0 0
```

That said, it looks from the Linux source code that the headers are in a fixed order (see https://github.com/torvalds/linux/blob/9712d38c87087ae34f8cede6583092c272507c52/net/ipv4/proc.c#L144-L160 and https://github.com/torvalds/linux/blob/9712d38c87087ae34f8cede6583092c272507c52/net/ipv4/proc.c#L163-L174) and thus predictable. There's no need to actually read the header row, which would let you use the traditional all-caps enum naming and then use the `ParseUtil.stringToEnumMap()` implementation [I suggested here](https://github.com/oshi/oshi/issues/2826#issuecomment-2708441062).

oshi-core/src/main/java/oshi/software/os/linux/LinuxInternetProtocolStats.java
```diff
@@ -36,19 +38,103 @@
...
+
+    private enum UdpStat {
+        OutDatagrams, InDatagrams, NoPorts, InErrors
+    }
+
     @Override
     public TcpStats getTCPv4Stats() {
-        return NetStat.queryTcpStats("netstat -st4");
+        byte[] fileBytes = FileUtil.readAllBytes(ProcPath.SNMP, true);
+        String content = new String(fileBytes);
```
dbwiddis, line 56: Haven't looked at checkstyle but I suspect it's going to yell at you for leaving off a character encoding here.

But again, that's unnecessary, `ParseUtil.parseByteArrayToStrings()` will produce a list for you so you won't need to do the next line.

dbwiddis (commented):
First, thanks for contributing!  This will be a great improvement.

I've made a few suggestions here: try to use existing parsing functions when possible.  The header values are predictable so you don't need to parse them.

oshi-core/src/main/java/oshi/software/os/linux/LinuxInternetProtocolStats.java
rohan-coder02: **Analysis:**

- I found that ParseUtil.parseByteArrayToStrings() assumes null (\0) termination (`bytes[end] == 0`). However, _/proc/net/snmp_ and _/proc/net/snmp6_ actually uses newlines (\n)
- Because of this, when reading _/proc/net/snmp_ or _/proc/net/snmp6_, the method treats it as a single giant string, leading to lines.size() == 1 (Refer below view)
```
byte[] fileBytes = FileUtil.readAllBytes(ProcPath.SNMP, true);
List<String> lines = ParseUtil.parseByteArrayToStrings(fileBytes);
System.out.println("lines size is " + lines.size());
```
- parseByteArrayToStrings-before

**Solution:**

I see two solutions here

Use the below code done in the previous commit of PR.
```
List<String> lines = Arrays.asList(content.split("\\n"));
```
(OR)

In oshi.util.ParseUtil#parseByteArrayToStrings, add the following -> `bytes[end] == '\n'`
```
(end == bytes.length || bytes[end] == 0 || bytes[end] == '\n')
```
Upon doing this change with additional logging it treats it as a separate line unlike before, refer below view
```
byte[] fileBytes = FileUtil.readAllBytes(ProcPath.SNMP, true);
List<String> lines = ParseUtil.parseByteArrayToStrings(fileBytes);
System.out.println("lines size is " + lines.size());
```
parseByteArrayToStrings-after

In the new commit, I've used the second approach 

Let me know if any queries.

oshi-core/src/main/java/oshi/software/os/linux/LinuxInternetProtocolStats.java
```diff
@@ -43,98 +41,101 @@ public class LinuxInternetProtocolStats extends AbstractInternetProtocolStats {
...
-            if (parts.length == 2 && parts[0].startsWith(UDP6)) {
-                try {
-                    // Remove "Udp6" prefix and map to enum
-                    String keyName = parts[0].substring(4);
-                    UdpStat key = UdpStat.valueOf(keyName);
-                    long value = Long.parseLong(parts[1]);
-                    udpData.put(key, value);
-                } catch (IllegalArgumentException e) {
-                    // Ignore fields that are not in UdpStat enum
+        List<String> lines = ParseUtil.parseByteArrayToStrings(fileBytes);
```
rohan-coder02, line 106: Also, in the latest commit, I've modified the **getUDPv6Stats()** function

Previously, **getUDPv6Stats()** scanned the entire _/proc/net/snmp6_ file from top to bottom, checking each line for UDPv6 statistics.

However, _/proc/net/snmp6_ has a structured order:

- Ip6 statistics
- Icmp6 statistics
- Udp6 statistics (which we need)
- udplite6 statistics

Refer https://github.com/torvalds/linux/blob/9712d38c87087ae34f8cede6583092c272507c52/net/ipv6/proc.c#L217

Here, use a counter to track matched values (we only need 4 UDPv6 stats) and then break. 

**Findings:**

- Validated the above scenario in Redhat and Ubuntu
```
Redhat
cat /proc/net/snmp6 | wc -l
89
grep -E -c '^(Ip6|Icmp6)' /proc/net/snmp6
72
grep -E -c '^(UdpLite6)' /proc/net/snmp6
8
```
- Here, bottom to top traversal we reach UDPv6 stats after 8 iterations whereas top to bottom after 72 iterations
```
Ubuntu
cat /proc/net/snmp6 | wc -l
89
grep -E -c '^(Ip6|Icmp6)' /proc/net/snmp6
72
grep -E -c '^(UdpLite6)' /proc/net/snmp6
8
```

oshi-core/src/main/java/oshi/software/os/linux/LinuxInternetProtocolStats.java
```diff
@@ -36,19 +36,106 @@
...
+        Map<UdpStat, Long> udpData = new EnumMap<>(UdpStat.class);
+
+        for (int line = 0; line < lines.size() - 1; line += 2) {
+            if (lines.get(line).startsWith(UDP_COLON) && lines.get(line + 1).startsWith(UDP_COLON)) {
+                Map<UdpStat, String> parsedData = ParseUtil.stringToEnumMap(UdpStat.class, lines.get(line + 1), ' ');
+                for (Map.Entry<UdpStat, String> entry : parsedData.entrySet()) {
+                    if (entry.getKey().equals(UdpStat.UdpColon)) {
+                        continue; // Skip "Udp:"
+                    }
+                    udpData.put(entry.getKey(), Long.parseLong(entry.getValue()));
```
dbwiddis, line 92: This will fail our [forbidden apis check](https://github.com/oshi/oshi/blob/8d9eda2c70b715af26b95630d487c4891c5508b4/config/forbidden-apis.txt#L8) as you don't handle exceptions.  But we have `ParseUtil.parseLongOrDefault()` which does.

oshi-core/src/main/java/oshi/software/os/linux/LinuxInternetProtocolStats.java
```diff
@@ -36,19 +36,106 @@
...
-        return NetStat.queryUdpStats("netstat -su4");
+        byte[] fileBytes = FileUtil.readAllBytes(ProcPath.SNMP, true);
+        List<String> lines = ParseUtil.parseByteArrayToStrings(fileBytes);
+        Map<UdpStat, Long> udpData = new EnumMap<>(UdpStat.class);
+
+        for (int line = 0; line < lines.size() - 1; line += 2) {
+            if (lines.get(line).startsWith(UDP_COLON) && lines.get(line + 1).startsWith(UDP_COLON)) {
+                Map<UdpStat, String> parsedData = ParseUtil.stringToEnumMap(UdpStat.class, lines.get(line + 1), ' ');
+                for (Map.Entry<UdpStat, String> entry : parsedData.entrySet()) {
+                    if (entry.getKey().equals(UdpStat.UdpColon)) {
```
dbwiddis, line 89: Why not just strip off the prefix using `substring(UDP_COLON.length())`?  (Same for TCP. Then you don't need that entry in the enum.)

oshi-core/src/main/java/oshi/software/os/linux/LinuxInternetProtocolStats.java
```diff
@@ -36,19 +36,106 @@
...
+        // Traverse bottom-to-top for efficiency as the /etc/proc/snmp6 file follows sequential format -> ip6, icmp6,
+        // udp6 stats
+        for (int line = lines.size() - 1; line >= 0 && foundUDPv6StatsCount < 4; line--) {
+            if (lines.get(line).startsWith(UDP6)) {
+                String[] parts = lines.get(line).split("\\s+");
+                switch (parts[0]) {
+                case "Udp6InDatagrams":
+                    inDatagrams = Long.parseLong(parts[1]);
+                    foundUDPv6StatsCount++;
+                    break;
```
dbwiddis, line 119: `ParseUtil` is your friend.  In this case, `ParseUtil.parseLastLong()` (after a "startsWith" check)

dbwiddis (commented):
Thanks for your patience continuing to work on this!  Logic looks good here, just some final tweaks needed.

Also please add a change log entry.

oshi-core/src/main/java/oshi/software/os/linux/LinuxInternetProtocolStats.java
rohan-coder02: Thanks for the feedback.

I've incorporated it in the latest commit.

oshi-core/src/main/java/oshi/software/os/linux/LinuxInternetProtocolStats.java
rohan-coder02: Thanks for the feedback.

I've incorporated it in the latest commit.

oshi-core/src/main/java/oshi/software/os/linux/LinuxInternetProtocolStats.java
rohan-coder02: Thanks for the feedback.

I've incorporated it in the latest commit.

dbwiddis (approved):
LGTM!   Thanks for this contribution, looking forward to more! 🎉

PR #2849, opened 2025-03-13
Include 3D controller ([0302]) in Linux GPU detection

lesley29:
This PR extends the Linux GPU detection logic to recognize 3D controller devices (class [0302]) in addition to the existing VGA class check. Many discrete NVIDIA GPUs (such as the RTX A2000) are now reported as "3D controller" instead of "VGA", causing OSHI to miss them when multiple GPUs are present, for instance, in systems with both an integrated GPU and a discrete GPU. In cases where only a single GPU is present, the current code falls back to using `lshw` to retrieve GPU info, which usually works as expected. However, because the lspci-based detection is used first, the discrete dGPU isn’t detected unless we also check for the 3d controller class. For example, consider this `lspci` output for two cards:
```
Class: VGA compatible controller [0300] 
Vendor: Intel Corporation [8086] 
Device: Alder Lake-P GT2 [Iris Xe Graphics] [46a6]
...
Class: 3D controller [0302]
Vendor: NVIDIA Corporation [10de]
Device: GA107GLM [RTX A2000 ...]
```

**Note on tests**
I noticed in the commit history that there was an attempt to add a mocking library for testing `ExecutingCommand.runNative`, but it was eventually removed. Since this is a small fix, I validated it locally instead of introducing a new mocking framework. If there’s a preferred approach to testing this functionality, please let me know and I’ll adjust accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 6.7.2 (in progress)
 
-* Your contribution here!
+##### Bug fixes / Improvements
+* [#2849](https://github.com/oshi/oshi/pull/2846): Include 3D controller ([0302]) in Linux GPU detection - [@lesley29](https://github.com/lesley29).
 
 # 6.7.0 (2025-02-25), 6.7.1 (2025-03-12)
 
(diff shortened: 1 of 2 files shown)


dbwiddis (commented):
LGTM.

Since 6.7.1 has been released, put the new change log entry under 6.7.2.  That's the section I copy and paste into the release notes before appending it to the long list of patch releases.

Regarding mocking, it's somewhat complicated.  We can capture output and put it in a text file and mock that as a return, but if it never changes, then the mock is mostly useless (sure, it might catch a change in how we process that data but that's pretty rare).  I much prefer trying to test on as many actual systems as possible.

I'm not one who wants to have coverage for the sake of coverage.  If there's a useful reason to mock, let's do it.  If it's just to check a box of code coverage, let's not.

lesley29:
> Since 6.7.1 has been released, put the new change log entry under 6.7.2. 

Oops, my bad. I’ve moved the changelog entry to the 6.7.2.

Ok, I see your point regarding mocking, let’s keep it as is for now

Issue #35, opened 2015-06-15
Version in pom.xml for v1.2


alessiofachechi:
@dbwiddis probably you forget to set in `pom.xml` of the tag v1.2 (and of the master) the version at 1.2.

dbwiddis:
I didn't forget to set it, so much as forget to commit it.  :wink: 

But seriously -- I've made many more changes than the version tag in that file, and have a few more I need to make that have to do with publishing to the Central Repository, which unfortunately requires a 2 business day wait the first time to get the project set up.

It's updated properly in the zipped release.   And when I do finally update it, I think it should actually be set for the next version (possibly with the -SNAPSHOT appended.)

I'm still learning Maven. If you know more and would like to teach me, please do so!

dbwiddis:
https://issues.sonatype.org/browse/OSSRH-16053

dbwiddis:
OK!  Merged #36. It's now 1.2.1-SNAPSHOT, and you can see all the other pom.xml changes I made to get us [published on the Central Repository (a.k.a. Maven Central)](https://oss.sonatype.org/content/repositories/releases/com/github/dblock/oshi-core/1.2/).   A few bumps in this release road but the next release should be a lot more smooth.

alessiofachechi:
Well done @dbwiddis! :+1: :)

Anyway I know only the M2 basics, so I couldn't help you so much... :stuck_out_tongue_winking_eye:

PR #2890, opened 2025-05-22
Handle 4 years digit format on macOS system_profiler output

dyorgio:
Some macOS installations have 4 digits years on Last Modified dates of system_profiler SPApplicationsDataType output.
This commit includes a fallback to theses scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 6.8.2 (in progress)
 
-* Your contribution here!
+##### Bug fixes / Improvements
+* [#2889](https://github.com/oshi/oshi/pull/2889): Include fallback on macos getInstalledApps for systems with 4 year digits on date format - [@dyorgio](https://github.com/dyorgio).
 
 # 6.8.0 (2025-03-22), 6.8.1 (2025-04-15)
 
--- a/oshi-core/src/main/java/oshi/software/os/mac/MacInstalledApps.java
+++ b/oshi-core/src/main/java/oshi/software/os/mac/MacInstalledApps.java
@@ -68,6 +68,10 @@ public final class MacInstalledApps {
 
         String lastModified = details.getOrDefault("Last Modified", Constants.UNKNOWN);
         long lastModifiedEpoch = ParseUtil.parseDateToEpoch(lastModified, "dd/MM/yy, HH:mm");
+        if (lastModifiedEpoch == 0) {
+            // Fallback to 4 digits year format
+            lastModifiedEpoch = ParseUtil.parseDateToEpoch(lastModified, "dd/MM/yyyy, HH:mm");
+        }
 
         // Additional info map
         Map<String, String> additionalInfo = new HashMap<>();


dyorgio:
Example of "system_profiler SPApplicationsDataType" output:

```
    Apache NetBeans:

      Version: 26
      Obtained from: Identified Developer
      Last Modified: 21/05/2025, 12:04
      Kind: iOS
      Signed by: Developer ID Application: Codelerity Ltd. (3KH8VFSK7Q), Developer ID Certification Authority, Apple Root CA
      Location: /Applications/Apache NetBeans.app
```

oshi-core/src/main/java/oshi/software/os/mac/MacInstalledApps.java
```diff
@@ -68,6 +68,10 @@ private static ApplicationInfo createAppInfo(String name, Map<String, String> de
 
         String lastModified = details.getOrDefault("Last Modified", Constants.UNKNOWN);
         long lastModifiedEpoch = ParseUtil.parseDateToEpoch(lastModified, "dd/MM/yy, HH:mm");
+        if (lastModifiedEpoch == 0) {
+            // Fallback to 4 digits year format
+            lastModifiedEpoch = ParseUtil.parseDateToEpoch(lastModified, "dd/MM/yyyy, HH:mm");
+        }
```
dbwiddis, line 74: Perhaps we can add this fallback to the `parseDateToEpoch()` method, so any other similar errors can be caught.  

For example:
```java
public static long parseDateToEpoch(String dateString, String datePattern) {
    if (dateString.equals(Constants.UNKNOWN) || dateString.isEmpty()) {
        return 0; // Default value if date is unknown or empty
    }
    try {
        DateTimeFormatter formatter = DateTimeFormatter.ofPattern(datePattern, Locale.ROOT);
        if (datePattern.contains("H") || datePattern.contains("m") || datePattern.contains("s")) {
            LocalDateTime localDateTime = LocalDateTime.parse(dateString, formatter);
            return localDateTime.atZone(ZoneId.systemDefault()).toInstant().toEpochMilli();
        } else {
            LocalDate localDate = LocalDate.parse(dateString, formatter);
            return localDate.atStartOfDay(ZoneId.systemDefault()).toInstant().toEpochMilli();
        }
    } catch (DateTimeParseException e) {
        if (datePattern.contains("yy") && !datePattern.contains("yyyy")) {
            return parseDateToEpoch(dateString, datePattern.replace("yy", "yyyy"));
        }
        LOG.trace("Unable to parse date string: " + dateString);
        return 0;
    }
}
```

dbwiddis (commented):
Thanks for this PR!

I've suggested an alternative fix. Let me know what you think.

oshi-core/src/main/java/oshi/software/os/mac/MacInstalledApps.java
dbwiddis: > like yours, it don't "cache" the correct pattern to future parsing...

Good point.

> I found a this code too ...  that could be used to detect ...

Nice find!   Yeah, we can probably do that.

Declare a local JNA sublcass (and later contribute [upstream to JNA](https://github.com/java-native-access/jna/blob/master/contrib/platform/src/com/sun/jna/platform/mac/CoreFoundation.java) so we can delete this):
```java
public interface CoreFoundation extends com.sun.jna.platform.mac.CoreFoundation {
    CoreFoundation INSTANCE = Native.load("CoreFoundation", CoreFoundation.class);

    // not sure of enum type here, I think int is right
    int kCFDateFormatterNoStyle = 0;
    int kCFDateFormatterShortStyle = 1;
    int kCFDateFormatterMediumStyle = 2;

    CFTypeRef CFLocaleCopyCurrent();
    CFTypeRef CFDateFormatterCreate(CFAllocatorRef allocator, CFTypeRef locale, long dateStyle, long timeStyle);
    CFStringRef CFDateFormatterGetFormat(CFTypeRef formatter);

}
```
Then add these methods to our CFUtil class (with appropriate shortening of CoreFoundation.INSTANCE to a constant CF for readability, and possible exception handling):
```java
public static String getMacDateFormat() {
    CFTypeRef localeRef = CoreFoundation.INSTANCE.CFLocaleCopyCurrent();
    try {
        return getMacFormat(localeRef, kCFDateFormatterShortStyle, kCFDateFormatterNoStyle);
    } finally {
        INSTANCE.CFRelease(localeRef);
    }
}

private static String getMacFormat(CFTypeRef localeRef, long dateStyle, long timeStyle) {
    CFTypeRef formatter = CoreFoundation.INSTANCE.CFDateFormatterCreate(null, localeRef, dateStyle, timeStyle);
    if (formatter == null) {
        throw new IllegalStateException("CFDateFormatterCreate: null");
    }
    try {
        CFStringRef format = CoreFoundation.INSTANCE.CFDateFormatterGetFormat(formatter);
        return format.stringValue();
    } finally {
        cf.CFRelease(formatter);
    }
}
```

dbwiddis: > like yours, it don't "cache" the correct pattern to future parsing...

Good point.

> I found a this code too ...  that could be used to detect ...

Nice find!   Yeah, we can probably do that.  I just threw this together as a test.

Declare a local JNA sublcass (and later contribute [upstream to JNA](https://github.com/java-native-access/jna/blob/master/contrib/platform/src/com/sun/jna/platform/mac/CoreFoundation.java) so we can delete this):
```java
package oshi.jna.platform.mac;

import com.sun.jna.Native;

public interface CoreFoundation extends com.sun.jna.platform.mac.CoreFoundation {

    CoreFoundation INSTANCE = Native.load("CoreFoundation", CoreFoundation.class);

    int kCFDateFormatterNoStyle = 0;
    int kCFDateFormatterShortStyle = 1;
    int kCFDateFormatterMediumStyle = 2;
    int kCFDateFormatterLongStyle = 3;
    int kCFDateFormatterFullStyle = 4;

    CFTypeRef CFLocaleCopyCurrent();

    CFTypeRef CFDateFormatterCreate(CFAllocatorRef allocator, CFTypeRef locale, long dateStyle, long timeStyle);

    CFStringRef CFDateFormatterGetFormat(CFTypeRef formatter);
}
```
With this test class
```java
package oshi.jna.platform.mac;

import com.sun.jna.platform.mac.CoreFoundation.CFStringRef;
import com.sun.jna.platform.mac.CoreFoundation.CFTypeRef;

public class CFTest {
    private static CoreFoundation CF = CoreFoundation.INSTANCE;

        System.out.format("%11s %-18s%n", "Style enum", "Format");
        for (int style = 0; style <5;style++) {
            System.out.format("%11d %-18s%n", style, getMacDateFormat(style));
        }
    }
    
    public static String getMacDateFormat(int style) {
        CFTypeRef localeRef = CF.CFLocaleCopyCurrent();
        try {
            return getMacFormat(localeRef, style, 0);
        } finally {
            CF.CFRelease(localeRef);
        }
    }

    private static String getMacFormat(CFTypeRef localeRef, long dateStyle, long timeStyle) {
        CFTypeRef formatter = CF.CFDateFormatterCreate(null, localeRef, dateStyle, timeStyle);
        if (formatter == null) {
            throw new IllegalStateException("CFDateFormatterCreate: null");
        }
        try {
            CFStringRef format = CF.CFDateFormatterGetFormat(formatter);
            return format.stringValue();
        } finally {
            CF.CFRelease(formatter);
        }
    }
}
```
Output:

```
 Style enum Format            
          0                   
          1 M/d/yy            
          2 MMM d, y          
          3 MMMM d, y         
          4 EEEE, MMMM d, y   
```

dyorgio: HI @dbwiddis, my machine output:

```
0           
1           dd/MM/y, HH:mm
2           d 'de' MMM 'de' y, HH:mm:ss
3           d 'de' MMMM 'de' y 'às' HH:mm:ss z
4           EEEE, d 'de' MMMM 'de' y 'às' HH:mm:ss zzzz
```

I'm not sure about this only one char for year.... never saw it before

dyorgio: But works!
```.java
System.out.println(ParseUtil.parseDateToEpoch("21/05/2025, 12:04", getMacDateTimeFormat(CoreFoundation.kCFDateFormatterShortStyle)));
```

output:
```
1747839840000
```

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # 6.8.2 (in progress)
 
-* Your contribution here!
+##### Bug fixes / Improvements
+* [#2889](https://github.com/oshi/oshi/pull/2889): Handle system date format on macos getInstalledApps output parsing - [@dyorgio](https://github.com/dyorgio).
```
dbwiddis, line 4: Nit:
```suggestion
* [#2889](https://github.com/oshi/oshi/pull/2889): Handle system date format on macOS getInstalledApps output parsing - [@dyorgio](https://github.com/dyorgio).
```

oshi-core/src/main/java/oshi/jna/platform/mac/CoreFoundation.java
```diff
@@ -0,0 +1,28 @@
...
+
+    int kCFDateFormatterNoStyle = 0;
+    int kCFDateFormatterShortStyle = 1;
+    int kCFDateFormatterMediumStyle = 2;
+    int kCFDateFormatterLongStyle = 3;
+    int kCFDateFormatterFullStyle = 4;
+
+    CFTypeRef CFLocaleCopyCurrent();
+
+    CFTypeRef CFDateFormatterCreate(CFAllocatorRef allocator, CFTypeRef locale, long dateStyle, long timeStyle);
```
dbwiddis, line 25: While `long` works in the 64-bit mac world, we should probably use the `CFIndex` type since it's available.
```suggestion
    CFTypeRef CFDateFormatterCreate(CFAllocatorRef allocator, CFTypeRef locale, CFIndex dateStyle, CFIndex timeStyle);
```

oshi-core/src/main/java/oshi/jna/platform/mac/CoreFoundation.java
```diff
@@ -0,0 +1,28 @@
...
+
+    CFStringRef CFDateFormatterGetFormat(CFTypeRef formatter);
+}
```
dbwiddis, line 28: Nit: This file doesn't end in a newline.  I think `mvn spotless:apply` will fix that.

oshi-core/src/main/java/oshi/software/os/mac/MacInstalledApps.java
```diff
@@ -17,6 +21,7 @@
 public final class MacInstalledApps {
 
     private static final String COLON = ":";
+    private static CoreFoundation CF = CoreFoundation.INSTANCE;
```
dbwiddis, line 24:
```suggestion
    private static final CoreFoundation CF = CoreFoundation.INSTANCE;
```

oshi-core/src/main/java/oshi/software/os/mac/MacInstalledApps.java
```diff
@@ -31,6 +36,7 @@ private static List<ApplicationInfo> parseMacAppInfo(List<String> lines) {
         String appName = null;
         Map<String, String> appDetails = null;
         boolean collectingAppDetails = false;
+        String dateFormat = getMacDateTimeFormat(CoreFoundation.kCFDateFormatterShortStyle);
```
dbwiddis, line 39: 1. This shouldn't ever change. Let's make this a static class field.
2. Consider a static import for this.

oshi-core/src/main/java/oshi/software/os/mac/MacInstalledApps.java
```diff
@@ -78,4 +84,22 @@ private static ApplicationInfo createAppInfo(String name, Map<String, String> de
         return new ApplicationInfo(name, ParseUtil.getValueOrUnknown(details, "Version"), vendor, lastModifiedEpoch,
                 additionalInfo);
     }
+
+    private static String getMacDateTimeFormat(int style) {
+        CFTypeRef localeRef = CF.CFLocaleCopyCurrent();
+        try {
+            CFTypeRef formatter = CF.CFDateFormatterCreate(null, localeRef, style, style);
+            if (formatter == null) {
+                throw new IllegalStateException("CFDateFormatterCreate: null");
```
dbwiddis, line 93: While I copied this from the provided code, OSHI's pattern is to not throw exceptions.  Just return an empty string here.

dbwiddis (commented):
Looks great! Just a few nits.

oshi-core/src/main/java/oshi/jna/platform/mac/CoreFoundation.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+public interface CoreFoundation extends com.sun.jna.platform.mac.CoreFoundation {
+
+    CoreFoundation INSTANCE = Native.load("CoreFoundation", CoreFoundation.class);
+
+    int kCFDateFormatterNoStyle = 0;
+    int kCFDateFormatterShortStyle = 1;
+    int kCFDateFormatterMediumStyle = 2;
+    int kCFDateFormatterLongStyle = 3;
+    int kCFDateFormatterFullStyle = 4;
```
dbwiddis, line 21: Let's make this an enum like https://github.com/java-native-access/jna/blob/ec70f9484b3c7c0adf7444b9397df71df35c6957/contrib/platform/src/com/sun/jna/platform/mac/CoreFoundation.java#L260-L274

oshi-core/src/main/java/oshi/software/os/mac/MacInstalledApps.java
dbwiddis: > I'm not sure about this only one char for year.... never saw it before

Apparently "unicode technical standard #35".

See https://unicode.org/reports/tr35/tr35-dates.html#Date_Format_Patterns

oshi-core/src/main/java/oshi/software/os/mac/MacInstalledApps.java
dyorgio: 1. User could alter region on-the-fly in macOS...
2. ok!

dbwiddis: > 1. User could alter region on-the-fly in macOS...

Fair point.  I guess one call before the app loop isn't horrible.

oshi-core/src/main/java/oshi/jna/platform/mac/CoreFoundation.java
dbwiddis: On second thought, looking at [the method signature](https://developer.apple.com/documentation/corefoundation/cfdateformatterstyle) we just need the constants (which should be `long` or `NativeLong`) and the shortened names:
[case fullStyle](https://developer.apple.com/documentation/corefoundation/cfdateformatterstyle/fullstyle)
[case longStyle](https://developer.apple.com/documentation/corefoundation/cfdateformatterstyle/longstyle)
[case mediumStyle](https://developer.apple.com/documentation/corefoundation/cfdateformatterstyle/mediumstyle)
[case noStyle](https://developer.apple.com/documentation/corefoundation/cfdateformatterstyle/nostyle)
[case shortStyle](https://developer.apple.com/documentation/corefoundation/cfdateformatterstyle/shortstyle)

dbwiddis: See implementation here:
 -  https://github.com/java-native-access/jna/pull/1672

PR #2902, opened 2025-06-04
Fix: Read both 32-bit and 64-bit Windows registry keys when fetching installed apps

rohan-coder02:
### Title:

 Read both 32-bit and 64-bit Windows registry keys when fetching installed apps

**Description of the Code Change**:

Updated the getRegistryValueOrUnknown method to improve compatibility when accessing Windows registry keys,  in environments where 32-bit Java is used on 64-bit Windows systems.

- Tries both WinNT.KEY_WOW64_64KEY and WinNT.KEY_WOW64_32KEY access flags.
- Returns the first non-UNKNOWN result it successfully reads.
- Logs if it encounters a Win32Exception, helping with debugging access failures to the specific registry path

**Replication**:

Created a simple Java program that outputs the size of installed apps using latest OSHI lib and after code change.
Testing done on Windows 2022 using 32 and 64 bit Java
```
List<ApplicationInfo> installedApplications = si.getOperatingSystem().getInstalledApplications();
System.out.println("Size is " + installedApplications.size());
```

Using 32 bit Java with  latest OSHI lib:

```
java -jar Test-oshi.jar
Size is 2
```

Using 32 bit Java with custom OSHI build with code change

```
java -jar Test-oshi.jar
Size is 8
```

### DIFF
--- a/oshi-core/src/main/java/oshi/driver/windows/registry/InstalledAppsData.java
+++ b/oshi-core/src/main/java/oshi/driver/windows/registry/InstalledAppsData.java
@@ -13,12 +13,12 @@ import org.slf4j.LoggerFactory;
 import oshi.software.os.ApplicationInfo;
 import oshi.util.ParseUtil;
 
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.LinkedHashSet;
+import java.util.LinkedHashMap;
 import java.util.Arrays;
 
 public final class InstalledAppsData {
@@ -39,7 +39,7 @@ public final class InstalledAppsData {
                 Arrays.asList("SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall"));
     }
 
-    public static List<ApplicationInfo> queryInstalledApps() {
+    public static Set<ApplicationInfo> queryInstalledApps() {
         Set<ApplicationInfo> appInfoSet = new LinkedHashSet<>();
 
         // Iterate through both HKLM and HKCU paths
@@ -68,7 +68,7 @@ public final class InstalledAppsData {
 
                             long installDateEpoch = ParseUtil.parseDateToEpoch(installDate, "yyyyMMdd");
 
-                            Map<String, String> additionalInfo = new HashMap<>();
+                            Map<String, String> additionalInfo = new LinkedHashMap<>();
                             additionalInfo.put("installLocation", installLocation);
                             additionalInfo.put("installSource", installSource);
 
@@ -83,7 +83,7 @@ public final class InstalledAppsData {
             }
         }
 
-        return new ArrayList<>(appInfoSet);
+        return appInfoSet;
     }
 
     private static String getRegistryValueOrUnknown(WinReg.HKEY rootKey, String path, String key, int accessFlag) {
--- a/oshi-core/src/main/java/oshi/driver/windows/registry/InstalledAppsData.java
+++ b/oshi-core/src/main/java/oshi/driver/windows/registry/InstalledAppsData.java
@@ -13,13 +13,7 @@ import org.slf4j.LoggerFactory;
 import oshi.software.os.ApplicationInfo;
 import oshi.util.ParseUtil;
 
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.LinkedHashSet;
-import java.util.LinkedHashMap;
-import java.util.Arrays;
+import java.util.*;
 
 public final class InstalledAppsData {
     private static final Logger LOG = LoggerFactory.getLogger(InstalledAppsData.class);
@@ -39,7 +33,7 @@ public final class InstalledAppsData {
                 Arrays.asList("SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Uninstall"));
     }
 
-    public static Set<ApplicationInfo> queryInstalledApps() {
+    public static List<ApplicationInfo> queryInstalledApps() {
         Set<ApplicationInfo> appInfoSet = new LinkedHashSet<>();
 
         // Iterate through both HKLM and HKCU paths
@@ -83,7 +77,7 @@ public final class InstalledAppsData {
             }
         }
 
-        return appInfoSet;
+        return new ArrayList<>(appInfoSet);
     }
 
     private static String getRegistryValueOrUnknown(WinReg.HKEY rootKey, String path, String key, int accessFlag) {
--- a/oshi-core/src/main/java/oshi/driver/windows/registry/InstalledAppsData.java
+++ b/oshi-core/src/main/java/oshi/driver/windows/registry/InstalledAppsData.java
@@ -6,8 +6,12 @@ package oshi.driver.windows.registry;
 
 import com.sun.jna.platform.win32.Advapi32Util;
 import com.sun.jna.platform.win32.Win32Exception;
+import com.sun.jna.platform.win32.WinNT;
 import com.sun.jna.platform.win32.WinReg;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import oshi.software.os.ApplicationInfo;
+import oshi.util.Constants;
 import oshi.util.ParseUtil;
 
 import java.util.ArrayList;
@@ -17,6 +21,7 @@ import java.util.Map;
 import java.util.Arrays;
 
 public final class InstalledAppsData {
+    private static final Logger LOG = LoggerFactory.getLogger(InstalledAppsData.class);
     private InstalledAppsData() {
     }
 
@@ -46,6 +51,9 @@ public final class InstalledAppsData {
                     String fullPath = registryPath + "\\" + key;
                     try {
                         String name = getRegistryValueOrUnknown(rootKey, fullPath, "DisplayName");
+                        if (name.equals(Constants.UNKNOWN)) {
+                            continue;
+                        }
                         String version = getRegistryValueOrUnknown(rootKey, fullPath, "DisplayVersion");
                         String publisher = getRegistryValueOrUnknown(rootKey, fullPath, "Publisher");
                         String installDate = getRegistryValueOrUnknown(rootKey, fullPath, "InstallDate");
@@ -73,7 +81,16 @@ public final class InstalledAppsData {
     }
 
     private static String getRegistryValueOrUnknown(WinReg.HKEY rootKey, String path, String key) {
-        String value = Advapi32Util.registryGetStringValue(rootKey, path, key);
-        return ParseUtil.getStringValueOrUnknown(value);
+        for (int accessFlag : new int[] { WinNT.KEY_WOW64_64KEY, WinNT.KEY_WOW64_32KEY }) {
+            try {
+                String value = Advapi32Util.registryGetStringValue(rootKey, path, key, accessFlag);
+                if (!ParseUtil.getStringValueOrUnknown(value).equals(Constants.UNKNOWN)) {
+                    return value;
+                }
+            } catch (Win32Exception e) {
+                LOG.trace("Unable to access " + path + " with flag " + accessFlag + ": " + e.getMessage());
+            }
+        }
+        return Constants.UNKNOWN;
     }
 }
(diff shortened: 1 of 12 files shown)


dyorgio:
Very nice @rohan-coder02 ! And how about ```Advapi32Util.registryGetKeys```? This method also has a version with samDesiredExtra parameter... It's safe to keep it without ```KEY_WOW64_64KEY```?

oshi-core/src/main/java/oshi/driver/windows/registry/InstalledAppsData.java
```diff
@@ -73,7 +81,16 @@ public static List<ApplicationInfo> queryInstalledApps() {
...
+        for (int accessFlag : new int[] { WinNT.KEY_WOW64_64KEY, WinNT.KEY_WOW64_32KEY }) {
+            try {
+                String value = Advapi32Util.registryGetStringValue(rootKey, path, key, accessFlag);
+                if (!ParseUtil.getStringValueOrUnknown(value).equals(Constants.UNKNOWN)) {
+                    return value;
+                }
+            } catch (Win32Exception e) {
+                LOG.trace("Unable to access " + path + " with flag " + accessFlag + ": " + e.getMessage());
+            }
+        }
```
dbwiddis: While this `int[]` array was the LLM suggestion I'm not sure this implementation requires us to iterate an array compared to just "try one key, if it's unknown try the other key".   This implementation would also declare and throw away this new array every time.

Given we are probably more likely to get more apps with `WinNT.KEY_WOW64_64KEY` I'd suggest
```suggestion
            try {
                String value = Advapi32Util.registryGetStringValue(rootKey, path, key, WinNT.KEY_WOW64_64KEY);
                if (ParseUtil.getStringValueOrUnknown(value).equals(Constants.UNKNOWN)) {
                    value = Advapi32Util.registryGetStringValue(rootKey, path, key, WinNT.KEY_WOW64_32KEY);
                }
                if (!ParseUtil.getStringValueOrUnknown(value).equals(Constants.UNKNOWN)) {
                    return value;
                }
            } catch (Win32Exception e) {
                LOG.trace("Unable to access " + path + ": " + e.getMessage());
            }
        }
```

dbwiddis: Also going to the trouble of creating an "UNKNOWN" just to get rid of it may be overkill here. Simple null/empty conditionals might be preferred.

rohan-coder02: 1)

@dyorgio thanks for your point, In the newly published commit I've taken care of this.
Below is the output using 32 and 64 bit Java respectively

```
.\zulu11.80.21-ca-jre11.0.27-win_i686\bin\java -jar Test2.jar
Size is 19

.\jre\bin\java -jar Test2.jar
Size is 19
```

2)

@dbwiddis I've added simple null/empty conditionals in the new commit

oshi-core/src/main/java/oshi/driver/windows/registry/InstalledAppsData.java
```diff
@@ -40,30 +45,36 @@ public static List<ApplicationInfo> queryInstalledApps() {
...
-                        additionalInfo.put("installLocation", installLocation);
-                        additionalInfo.put("installSource", installSource);
-
-                        ApplicationInfo app = new ApplicationInfo(name, version, publisher, installDateEpoch,
-                                additionalInfo);
-
-                        appInfoList.add(app);
-                    } catch (Win32Exception e) {
-                        // Skip keys that are inaccessible or have missing values
+                for (int accessFlag : new int[] { WinNT.KEY_WOW64_64KEY, WinNT.KEY_WOW64_32KEY }) {
```
dbwiddis: Let's move this array to a class field.

oshi-core/src/main/java/oshi/driver/windows/registry/InstalledAppsData.java
```diff
@@ -72,8 +83,15 @@ public static List<ApplicationInfo> queryInstalledApps() {
...
-        String value = Advapi32Util.registryGetStringValue(rootKey, path, key);
-        return ParseUtil.getStringValueOrUnknown(value);
+    private static String getRegistryValueOrUnknown(WinReg.HKEY rootKey, String path, String key, int accessFlag) {
+        try {
+            String value = Advapi32Util.registryGetStringValue(rootKey, path, key, accessFlag);
+            if (value != null && !value.trim().isEmpty()) {
+                return value;
+            }
+        } catch (Win32Exception e) {
+            LOG.trace("Unable to access " + path + ": " + e.getMessage());
```
dbwiddis: I had removed the accessFlag with my suggestion but since you have that avialable in this helper method you might want to put it back.

dbwiddis:
> I was thinking about implement `hashcode` on applicationinfo class and sort list before return it...

As you mentioned using `LinkedHash<foo>` rather than sorting is preferred.  In this case a `LinkedHashSet`.

>  I found more than 80 references to new HashMap on this repository, any chance to replace it all to LinkedHashMap?

While I don't think changing any of them to `Linked` would break any API, I would only do it if you think there's a need for sorted output.

oshi-core/src/main/java/oshi/driver/windows/registry/InstalledAppsData.java
rohan-coder02: Made the change in the recent commit

oshi-core/src/main/java/oshi/driver/windows/registry/InstalledAppsData.java
rohan-coder02: Made the change in the recent commit

dyorgio:
@rohan-coder02 without a ```java.util.Set``` the results will have duplicates.

dbwiddis:
> @rohan-coder02 without a ```java.util.Set``` the results will have duplicates.

Yep.  Use a LinkedHashSet.

dyorgio:
@rohan-coder02 Please note that ApplicationInfo does not implements equals and hashcode methods...
When you implement it, pay attention on additional Info, the order is unexpected.

dbwiddis:
> given this PR was raised to read both 32-bit and 64-bit Windows registry keys when fetching installed apps.
> 
> Should I raise a separate PR or continue in the existing PR for adding equals and hashcode methods in the ApplicationInfo class and return a Set instead of a List in OperatingSystem#getInstalledApplications()?

Since de-duplication of the outputs is necessary, and that requires a hashing function, I think that's appropriate for this PR.

Making output consistent with a LinkedHashSet project-wide (in cases where it matters) can/should be in a different PR, but doesn't hurt to make that change here in this PR anyway.

rohan-coder02:
Thanks for the update, I'll proceed with the changes in this PR

dbwiddis:
Thanks for your patience!  It's been a busy week but I should have more time to review this weekend and get this merged.

rohan-coder02:
Hi @dbwiddis 

Deduplication using Set has been added and is handled across Windows, Linux, and macOS in the recent commit.

oshi-core/src/main/java/oshi/software/os/ApplicationInfo.java
```diff
@@ -29,7 +30,7 @@ public class ApplicationInfo {
      * A map containing additional application details such as install location, source, etc. Keys are field names, and
      * values are corresponding details.
      */
-    private final Map<String, String> additionalInfo;
+    private final LinkedHashMap<String, String> additionalInfo;
```
dbwiddis: Nit: I don't think we need to change the type of `Map` here.  The implementation is what matters.

oshi-core/src/main/java/oshi/software/os/ApplicationInfo.java
```diff
@@ -46,7 +47,7 @@ public ApplicationInfo(String name, String version, String vendor, long timestam
         this.version = version;
         this.vendor = vendor;
         this.timestamp = timestamp;
-        this.additionalInfo = additionalInfo != null ? new HashMap<>(additionalInfo) : new HashMap<>();
+        this.additionalInfo = additionalInfo != null ? new LinkedHashMap<>(additionalInfo) : new LinkedHashMap<>();
```
dbwiddis:
```suggestion
        this.additionalInfo = additionalInfo != null ? new LinkedHashMap<>(additionalInfo) : Collections.emptyMap();
```
I'm not really sure we need the defensive copy here either, but why not keep at least one `Linked` in this class, I guess...

oshi-core/src/main/java/oshi/software/os/OperatingSystem.java
```diff
@@ -372,8 +373,8 @@ default List<OSDesktopWindow> getDesktopWindows(boolean visibleOnly) {
      *
      * @return A list of installed applications or an empty list if unsupported.
      */
-    default List<ApplicationInfo> getInstalledApplications() {
-        return Collections.emptyList();
+    default Set<ApplicationInfo> getInstalledApplications() {
```
dbwiddis: We follow SemVer so you can't change the type of an API return here without bumping a major version, which I don't think is really needed anyway.

Creating a list from a (Linked) set will still preserve the order.

oshi-core/src/main/java/oshi/software/os/OperatingSystem.java
```diff
@@ -372,8 +373,8 @@ default List<OSDesktopWindow> getDesktopWindows(boolean visibleOnly) {
...
+        return Collections.EMPTY_SET;
```
dbwiddis: This will go back to `.emptyList()` but for future reference, the constants don't infer types so it's preferable to use `Collections.emptySet()`.  I think what you have, without a cast, generates a compiler warning about an unchecked conversion.

oshi-core/src/main/java/oshi/software/os/linux/LinuxInstalledApps.java
```diff
@@ -72,8 +74,8 @@ private static boolean isPackageManagerAvailable(String packageManager) {
         return !result.isEmpty();
     }
 
-    private static List<ApplicationInfo> parseLinuxAppInfo(List<String> output) {
-        List<ApplicationInfo> appInfoList = new ArrayList<>();
+    private static Set<ApplicationInfo> parseLinuxAppInfo(List<String> output) {
```
dbwiddis: (Repeat comment for all OS's)

Following along with keeping `List` at the interface level, this should also return a `List` (and be memoized as such).

oshi-core/src/main/java/oshi/software/os/linux/LinuxInstalledApps.java
```diff
@@ -94,10 +96,10 @@ private static List<ApplicationInfo> parseLinuxAppInfo(List<String> output) {
                         ParseUtil.parseLongOrDefault(parts[4], 0L), // Date Epoch
                         additionalInfo);
 
-                appInfoList.add(app);
+                appInfoSet.add(app);
             }
         }
 
-        return appInfoList;
+        return appInfoSet;
```
dbwiddis: After changing return types, create a (order preserving) list from the (insertion order preserving) set.
```suggestion
        return new ArrayList<>(appInfoSet);
```

dbwiddis (commented):
LGTM but keep the interface a List, only use the set internally in the query method.

oshi-core/src/main/java/oshi/software/os/ApplicationInfo.java
rohan-coder02: Addressed in the latest commit

oshi-core/src/main/java/oshi/software/os/ApplicationInfo.java
rohan-coder02: Addressed in the latest commit

oshi-core/src/main/java/oshi/software/os/OperatingSystem.java
rohan-coder02: Addressed in the latest commit

oshi-core/src/main/java/oshi/software/os/linux/LinuxInstalledApps.java
rohan-coder02: Addressed in the latest commit

oshi-core/src/main/java/oshi/software/os/linux/LinuxInstalledApps.java
rohan-coder02: Addressed in the latest commit

PR #2911, opened 2025-06-09
feat: add jbang catalog for demo apps

maxandersen:
you can run demo apps using jbang like `jbang -m oshi.demo.OshiGui com.github.oshi:oshi-demo:RELEASE` which is nice, but this commit just makes it a tad easier to run some of the demo apps using jbang.

when this is committed you should be able to do: `jbang alias list oshi/oshi` and get something like:

```
detectvm@oshi/oshi
   com.github.oshi:oshi-demo:RELEASE
         Main Class: oshi.demo.DetectVM
diskstore@oshi/oshi
   com.github.oshi:oshi-demo:RELEASE
         Main Class: oshi.demo.DiskStoreForPath
gui@oshi/oshi
   com.github.oshi:oshi-demo:RELEASE
         Main Class: oshi.demo.OshiGui
id@oshi/oshi
   com.github.oshi:oshi-demo:RELEASE
         Main Class: oshi.demo.ComputerID
json@oshi/oshi
   com.github.oshi:oshi-demo:RELEASE
         Main Class: oshi.demo.Json
```

which means you can now run each of these with just `jbang gui@oshi/oshi` 

Which I think is nice :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,29 @@ git clone https://github.com/oshi/oshi.git && cd oshi
 
 Some settings are configurable in the [`oshi.properties`](https://github.com/oshi/oshi/blob/master/oshi-core/src/main/resources/oshi.properties) file, which may also be manipulated using the [`GlobalConfig`](https://oshi.github.io/oshi/oshi-core/apidocs/oshi/util/GlobalConfig.html) class or using Java System Properties. This should be done at startup, as configuration is not thread-safe and OSHI does not guarantee re-reading the configuration during operation.
 
+Demo
+----
+
 The `oshi-demo` artifact includes [several proof-of-concept examples](https://github.com/oshi/oshi/blob/master/oshi-demo/src/main/java/oshi/demo/) of using OSHI to obtain information, including a basic Swing GUI.
+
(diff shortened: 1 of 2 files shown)


maxandersen:
> I don't see much benefit of a scripting tool to run code that is provided as-is to demonstrate how the API could be used. Any IDE will to that just fine

sorry for delay - sure, any IDE will do that just fine if you actually know those exists and are easy to run..these commands let you see what oshi can do withot even having to touch or setup your IDE ...I think that having different ways to assist developers at different skill levels and stages makes it more will become aware and use it.

dbwiddis:
OK I think I get the value.  Can we try to make the README entry more concise and integrate it with the existing SystemInfoTest example?   

Basically, we have the usage section and then maybe a new header "sample output" giving the maven option, and then including the jbang option below it... but try to get fewer line breaks, perhaps just one code block using `#` for comments on what the commands do?

maxandersen:
@dbwiddis makes sense - updated the PR to do what I think you suggested.

dbwiddis:
> @dbwiddis makes sense - updated the PR to do what I think you suggested.

Thanks! Looks good, I'll take a look later today and get this merged.

PR #2912, opened 2025-06-10
Installed apps feature for AIX

rohan-coder02:
**Context**:

The standard `lslpp -Lc` command lists installed software in AIX.

**Code Changes**:

- Use `lslpp -Lc` to extract installed application data.
- Parse the following fields from each colon-separated line:
   - name → field index 0
   - version → index 2
   - description → index 7
   - installPath → index 16
   - timestamp → index 17 (can be either yyww or full datetime)
- Skip the first line (column headers).
- https://www.ibm.com/docs/en/aix/7.1.0?topic=l-lslpp-command

Column names -> https://community.unix.com/t/list-of-installed-applications/338886
The command lists installed software in AIX, but lacks explicit vendor information
In the PR `AixInstalledApps` class consists of the sample outputs from the command in comments

**Replication:**

Used AIX 7.1 ppc64.

Created a simple Java program that outputs the installed apps using the latest OSHI lib, after the following code change.
```
        List<ApplicationInfo> installedApplications = si.getOperatingSystem().getInstalledApplications();
        System.out.println("Size is " + installedApplications.size());
        for (ApplicationInfo applicationInfo : installedApplications) {
            System.out.println(applicationInfo.toString());
        }
```

Output:

```
Size is 918
AppInfo{name=ICU4C.rte, version=7.1.4.0, vendor=unknown, timestamp=1445212800000, additionalInfo={architecture=ppc64, description=International Components for Unicode, installPath=/}}
AppInfo{name=Java5.sdk, version=5.0.0.620, vendor=unknown, timestamp=0, additionalInfo={architecture=ppc64, description=Java SDK 32-bit, installPath=/}}
...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,9 @@
 # 6.8.3 (in progress)
 
-##### New Features
-
-* [#2912](https://github.com/oshi/oshi/pull/2912): Installed apps feature for AIX - [@rohan-coder02](https://github.com/rohan-coder02).
-
 ##### Bug fixes / Improvements
 * [#2902](https://github.com/oshi/oshi/pull/2902): Deduplicate installed applications using Set across Windows, Linux, MacOS and Read both 32-bit, 64-bit Windows registry keys when fetching installed apps - [@rohan-coder02](https://github.com/rohan-coder02).
 * [#2908](https://github.com/oshi/oshi/pull/2908): Fix Windows cache size - [@lesley29](https://github.com/dbwiddis).
+* [#2912](https://github.com/oshi/oshi/pull/2912): Installed apps feature for AIX - [@rohan-coder02](https://github.com/rohan-coder02).
 
 # 6.8.0 (2025-03-22), 6.8.1 (2025-04-15), 6.8.2 (2025-05-31)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # 6.8.3 (in progress)
 
+##### New Features
+
+* [#2912](https://github.com/oshi/oshi/pull/2912): Installed apps feature for AIX - [@rohan-coder02](https://github.com/rohan-coder02).
+
 ##### Bug fixes / Improvements
 * [#2902](https://github.com/oshi/oshi/pull/2902): Deduplicate installed applications using Set across Windows, Linux, MacOS and Read both 32-bit, 64-bit Windows registry keys when fetching installed apps - [@rohan-coder02](https://github.com/rohan-coder02).
 * [#2908](https://github.com/oshi/oshi/pull/2908): Fix Windows cache size - [@lesley29](https://github.com/dbwiddis).
--- a/oshi-core/src/main/java/oshi/software/os/unix/aix/AixInstalledApps.java
+++ b/oshi-core/src/main/java/oshi/software/os/unix/aix/AixInstalledApps.java
@@ -17,7 +17,7 @@ import java.util.Map;
 import java.util.LinkedHashMap;
 import java.util.regex.Pattern;
 
-public class AixInstalledApps {
+public final class AixInstalledApps {
 
     private static final Pattern COLON_PATTERN = Pattern.compile(":");
 
--- a/oshi-core/src/main/java/oshi/software/os/unix/aix/AixInstalledApps.java
+++ b/oshi-core/src/main/java/oshi/software/os/unix/aix/AixInstalledApps.java
@@ -21,6 +21,9 @@ public class AixInstalledApps {
 
     private static final Pattern COLON_PATTERN = Pattern.compile(":");
 
+    private AixInstalledApps() {
+    }
+
     public static List<ApplicationInfo> queryInstalledApps() {
         // https://www.ibm.com/docs/en/aix/7.1.0?topic=l-lslpp-command
         List<String> output = ExecutingCommand.runNative("lslpp -Lc");
@@ -55,7 +58,9 @@ public class AixInstalledApps {
             long timestamp = 0;
             if (!buildDate.equals(Constants.UNKNOWN)) {
                 if (buildDate.matches("\\d{4}")) {
-                    timestamp = ParseUtil.parseDateToEpoch(buildDate, "YYWW");
+                    // Convert to ISO week date string (e.g., 1125 -> 2011-W25-2 for Monday)
+                    String isoWeekString = "20" + buildDate.substring(0, 2) + "-W" + buildDate.substring(2) + "-2";
+                    timestamp = ParseUtil.parseDateToEpoch(isoWeekString, "YYYY-'W'ww-e");
                 } else {
                     timestamp = ParseUtil.parseDateToEpoch(buildDate, "EEE MMM dd HH:mm:ss yyyy");
                 }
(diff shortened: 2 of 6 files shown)


oshi-core/src/main/java/oshi/software/os/unix/aix/AixInstalledApps.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.Map;
+import java.util.LinkedHashMap;
+import java.util.regex.Pattern;
+
+public class AixInstalledApps {
+
+    private static final Pattern COLON_PATTERN = Pattern.compile(":");
+
```
dbwiddis, line 23: (Checkstyle) Please add a private constructor here to prevent instantiation.

oshi-core/src/main/java/oshi/util/ParseUtil.java
```diff
@@ -10,13 +10,16 @@
...
+import java.time.DayOfWeek;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.time.LocalDate;
 import java.time.LocalTime;
 import java.time.OffsetDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
+import java.time.temporal.IsoFields;
+import java.time.temporal.TemporalAdjusters;
```
dbwiddis: (Checkstyle) These 3 new imports are changing the line number of a checkstyle exclusion.  If we keep these new imports, add 3 to both line numbers here (104-124): https://github.com/oshi/oshi/blob/d1643ff0dc5e498d83edf04422e1a25ce0a9bd11/config/checkstyle-suppressions.xml#L9

But see below for a suggestion.

oshi-core/src/main/java/oshi/util/ParseUtil.java
```diff
@@ -1397,6 +1400,14 @@ public static long parseDateToEpoch(String dateString, String datePattern) {
...
         }
         try {
+            if (datePattern.equals("YYWW")) {
+                // convert this to a date i.e. the start of the week -> Monday
+                int year = 2000 + Integer.parseInt(dateString.substring(0, 2));
+                int week = Integer.parseInt(dateString.substring(2));
+                LocalDate weekStart = LocalDate.ofYearDay(year, 1).with(IsoFields.WEEK_OF_WEEK_BASED_YEAR, week)
+                        .with(TemporalAdjusters.previousOrSame(DayOfWeek.MONDAY));
+                return weekStart.atStartOfDay(ZoneId.systemDefault()).toInstant().toEpochMilli();
+            }
```
dbwiddis: I don't really like adding more arbitrary conditionals here.  The existing code already does a sort of "if we don't have time fields, default to start of day" and that's essentially all we're doing here, defaulting to start of week (and presumably start of day of Monday of that week).

WDYT about replacing the whole inside of this try block with a formatter with defaults?
```java
DateTimeFormatter formatter = new DateTimeFormatterBuilder().appendPattern(datePattern)
        // Defaulting missing time components to start of period
        .parseDefaulting(ChronoField.HOUR_OF_DAY, 0).parseDefaulting(ChronoField.MINUTE_OF_HOUR, 0)
        .parseDefaulting(ChronoField.SECOND_OF_MINUTE, 0).parseDefaulting(ChronoField.MILLI_OF_SECOND, 0)
        // Following ISO-8601 for week definitions (weeks start on Monday)
        .parseDefaulting(ChronoField.DAY_OF_WEEK, WeekFields.ISO.getFirstDayOfWeek().getValue())
        .toFormatter(Locale.ROOT);

LocalDateTime localDateTime = LocalDateTime.parse(dateString, formatter);
return localDateTime.atZone(ZoneId.systemDefault()).toInstant().toEpochMilli();
```
As CodeCov nagged you we need to add a unit test the YYWW pattern.

dbwiddis (changes requested):
Thanks for this contribution!   

A few checkstyle complaints and a refactoring suggestion, plus need a unit test for the new format.

oshi-core/src/main/java/oshi/software/os/unix/aix/AixInstalledApps.java
rohan-coder02: addressed in the latest commit

oshi-core/src/main/java/oshi/util/ParseUtil.java
rohan-coder02: addressed in the latest commit

oshi-core/src/main/java/oshi/util/ParseUtil.java
rohan-coder02: 1)

- In the recent commit, DateTimeFormatter is with default time components.
- Upon adding the line below, `DateTimeParseException` was observed on Mac and Windows resulting in epoch time as 0.
```
.parseDefaulting(ChronoField.DAY_OF_WEEK, WeekFields.ISO.getFirstDayOfWeek().getValue())
```
- In the class AIXInstalledApps, the ISO week is passed as below in the recent commit.
```
// Convert to ISO week date string (e.g., 1125 -> 2011-W25-2 for Monday)
String isoWeekString = "20" + buildDate.substring(0, 2) + "-W" + buildDate.substring(2) + "-2";
timestamp = ParseUtil.parseDateToEpoch(isoWeekString, "YYYY-'W'ww-e");
```
- It converts AIX build date in YYWW format to ISO week date (`YYYY-'W'ww-e`) with Monday as the default day for accurate parsing.
- Retested on AIX, Windows, and Mac using the OSHI custom jar with the above code changes, and it accurately displayed the epoch time.

2)

Also, added the unit tests.

dbwiddis: > Upon adding the line below, DateTimeParseException was observed on Mac and Windows resulting in epoch time as 0.

Ah, right, the W is a special case.  But...

> String isoWeekString = "20" + buildDate.substring(0, 2) + "-W" + buildDate.substring(2) + "-2";
timestamp = ParseUtil.parseDateToEpoch(isoWeekString, "YYYY-'W'ww-e");

This works I guess.  I'm curious if the original format would work if we just used a conditional on the builder (only using it if W is in the string)?

dbwiddis: And, no, it doesn't work.  Your approach is actually pretty good.   I'll do a full re-review later and likely merge this.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # 6.8.3 (in progress)
 
+##### New Features
+
+* [#2912](https://github.com/oshi/oshi/pull/2912): Installed apps feature for AIX - [@rohan-coder02](https://github.com/rohan-coder02).
```
dbwiddis: Minor nit: should be under "Improvements" since the feature is already there, we're just fixing the OS incompatibility.

dbwiddis (approved):

oshi-core/src/main/java/oshi/util/ParseUtil.java
rohan-coder02: Sure.

CHANGELOG.md
rohan-coder02: I'll quickly do another commit with this statement as a part of improvements

PR #2915, opened 2025-06-17
Rewrite macOS getInstalledApps using system_profiler xml output format

dyorgio:
Previous parsing version get some erros depending of output (missing Version values per example).

Altering it to xml make it more predictable.

I also improved vendor decoding signed_by information and included fallback to version info (reading app Info.plist).

### DIFF
--- a/.github/workflows/site.yaml
+++ b/.github/workflows/site.yaml
@@ -21,7 +21,7 @@ jobs:
         with:
           cache: maven
           distribution: temurin
-          java-version: 21
+          java-version: 24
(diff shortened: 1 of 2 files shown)


dyorgio:
Hi @rohan-coder02 , please if you can take a look on this.

dyorgio:
@dbwiddis I agree with you, I only use JRE xml parsing classes because I was afraid to do it "by my hands" will not be accepted. I will remove theses classes and create util methods to do the job.

PR #2925, opened 2025-06-30
OSHI FFM support

rohan-coder02:
### Overview:

This PR introduces a Service Provider Interface (SPI) based pluggable core architecture to OSHI, enabling dynamic loading of JDK-specific implementations at runtime. The primary goal is to leverage JEP 454 (Foreign Function & Memory API) for performance improvements in JDK24+ while maintaining backward compatibility

### Key Changes:

**1. New Factory Interface**

- Added OperatingSystemFactory interface in oshi.software.os package
- Creates platform-specific OperatingSystem implementations
- Includes isSupportedOnThisJdk() method for runtime compatibility checking

**2. SPI-Based Implementation Selection**

- Created OperatingSystemFactoryProvider bootstrapper class
- Uses ServiceLoader to dynamically discover and select appropriate factory implementations
- Automatically chooses the implementation based on current JDK version

**3. Modular Architecture**

- oshi-core: Contains API interfaces, classic implementations (JDK8+), and ClassicOperatingSystemFactory
- oshi-core-java24: New module containing FFM-based implementations and FfmOperatingSystemFactory
- Each module registers its factory implementation via SPI

**4. Maven Profile Configuration**

- Added separate Maven profiles for different JDK versions
- oshi-core-java24 module only builds with JDK24+ to prevent class version conflicts
- Maintains existing build compatibility for JDK8-23

### SPI Registration:

- JDK8-23: ServiceLoader selects ClassicOperatingSystemFactory (existing behavior)
- JDK24+: ServiceLoader selects FfmOperatingSystemFactory (FFM benefits)
- API remains unchanged: Users continue using existing SystemInfo and OperatingSystem interfaces

### Directory Structure:

```
oshi/
 pom.xml
 oshi-core/src/main/java/oshi/software/os/
 OperatingSystemFactory.java // Factory interface
 ClassicOperatingSystemFactory.java // Classic implementation
 OperatingSystemFactoryProvider.java // Bootstrapper
 ... (platform-specific OS classes)
 src/main/resources/META-INF/services/
 oshi.software.os.OperatingSystemFactory

 oshi-core-java24/src/main/java/oshi/software/os/
 FfmOperatingSystemFactory.java // FFM implementation (JDK24+)
 ... (FFM platform-specific OS classes)
 src/main/resources/META-INF/services/
 oshi.software.os.OperatingSystemFactory
```

### Testing:

OshiSpiTest.java is a simple Java program to test.

Below are sample outputs that prints java version and factory selected i.e. classic or FFM based on java version

```
java -cp oshi-core-6.8.3-SNAPSHOT.jar OshiSpiTest.java 
11.0.26
Loaded factory: oshi.software.os.ClassicOperatingSystemFactory

java -cp oshi-core-java24-6.8.3-SNAPSHOT.jar OshiSpiTest.java
24.0.1
Loaded factory: oshi.software.os.FfmOperatingSystemFactory
```

### DIFF
--- a/oshi-core-java24/.gitignore
+++ b/oshi-core-java24/.gitignore
@@ -1,0 +1,38 @@
+target/
+!.mvn/wrapper/maven-wrapper.jar
+!**/src/main/**/target/
+!**/src/test/**/target/
+
+### IntelliJ IDEA ###
+.idea/modules.xml
+.idea/jarRepositories.xml
+.idea/compiler.xml
+.idea/libraries/
+*.iws
+*.iml
+*.ipr
+
+### Eclipse ###
+.apt_generated
+.classpath
+.factorypath
+.project
+.settings
+.springBeans
+.sts4-cache
+
+### NetBeans ###
+/nbproject/private/
+/nbbuild/
+/dist/
+/nbdist/
+/.nb-gradle/
+build/
+!**/src/main/**/build/
+!**/src/test/**/build/
+
+### VS Code ###
+.vscode/
+
+### Mac OS ###
+.DS_Store
--- a/oshi-core-java24/pom.xml
+++ b/oshi-core-java24/pom.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.github.oshi</groupId>
+        <artifactId>oshi-parent</artifactId>
+        <version>6.8.3-SNAPSHOT</version>
+    </parent>
+
+    <groupId>org.example</groupId>
+    <artifactId>oshi-core-java24</artifactId>
+
+    <properties>
+        <maven.compiler.source>24</maven.compiler.source>
+        <maven.compiler.target>24</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>com.github.oshi</groupId>
+            <artifactId>oshi-core</artifactId>
+            <version>6.8.3-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.github.oshi</groupId>
+            <artifactId>oshi-core</artifactId>
+            <version>6.8.3-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
+    </dependencies>
+
+</project>
--- a/oshi-core-java24/src/main/java/oshi/software/os/FfmOperatingSystemFactory.java
+++ b/oshi-core-java24/src/main/java/oshi/software/os/FfmOperatingSystemFactory.java
@@ -1,0 +1,16 @@
+package oshi.software.os;
+
+import oshi.software.os.OperatingSystemFactory;
+
+public class FfmOperatingSystemFactory implements OperatingSystemFactory{
+    @Override
+    public boolean isSupportedOnThisJdk() {
+        String version = System.getProperty("java.specification.version");
+        try {
+            int v = Integer.parseInt(version);
+            return v >= 24;
+        } catch (NumberFormatException e) {
+            return false;
+        }
+    }
+}
--- a/oshi-core/src/main/java/oshi/software/os/OperatingSystemFactory.java
+++ b/oshi-core/src/main/java/oshi/software/os/OperatingSystemFactory.java
@@ -1,0 +1,8 @@
+package oshi.software.os;
+
+public interface OperatingSystemFactory {
+    /**
+     * Returns true if this factory supports the current JDK.
+     */
+    boolean isSupportedOnThisJdk();
+}
--- a/oshi-core/src/main/java/oshi/software/os/OperatingSystemFactoryProvider.java
+++ b/oshi-core/src/main/java/oshi/software/os/OperatingSystemFactoryProvider.java
@@ -1,0 +1,21 @@
+package oshi.software.os;
+
+import java.util.ServiceLoader;
+
+public class OperatingSystemFactoryProvider {
+    private static final OperatingSystemFactory INSTANCE = loadInstance();
+
+    private static OperatingSystemFactory loadInstance() {
+        ServiceLoader<OperatingSystemFactory> loader = ServiceLoader.load(OperatingSystemFactory.class);
+        for (OperatingSystemFactory factory : loader) {
+            if (factory.isSupportedOnThisJdk()) {
+                return factory;
+            }
+        }
+        throw new IllegalStateException("No compatible OperatingSystemFactory found");
+    }
+
+    public static OperatingSystemFactory getInstance() {
+        return INSTANCE;
+    }
+}
--- a/pom.xml
+++ b/pom.xml
@@ -43,6 +43,7 @@
     <modules>
         <module>oshi-core</module>
         <module>oshi-demo</module>
+        <module>oshi-core-java24</module>
     </modules>
 
     <scm>
@@ -983,6 +984,19 @@
             <properties>
                 <maven.compiler.release>${maven.compiler.target}</maven.compiler.release>
                 <maven.compiler.testRelease>${maven.compiler.testTarget}</maven.compiler.testRelease>
+            </properties>
+        </profile>
+        <!-- Java 24+ profile -->
+        <profile>
+            <id>java24</id>
+            <activation>
+                <jdk>[24,)</jdk>
+            </activation>
+            <modules>
+                <module>oshi-core-java24</module>
+            </modules>
+            <properties>
+                <maven.compiler.release>24</maven.compiler.release>
             </properties>
         </profile>
         <profile>
--- a/OshiSpiTest.java
+++ b/OshiSpiTest.java
@@ -1,0 +1,13 @@
+import oshi.software.os.OperatingSystem;
+import oshi.software.os.OperatingSystemFactory;
+import oshi.software.os.OperatingSystemFactoryProvider;
+
+public class OshiSpiTest {
+    public static void main(String[] args) {
+        System.out.println(System.getProperty("java.version"));
+        OperatingSystemFactory factory = OperatingSystemFactoryProvider.getInstance();
+        boolean os = factory.isSupportedOnThisJdk();
+
+        System.out.println("Loaded factory: " + factory.getClass().getName());
+    }
+}
--- a/oshi-core-java24/src/main/resources/META-INF/services/oshi.software.os.OperatingSystemFactory
+++ b/oshi-core-java24/src/main/resources/META-INF/services/oshi.software.os.OperatingSystemFactory
@@ -1,0 +1,1 @@
+oshi.software.os.FfmOperatingSystemFactory
--- a/oshi-core/src/main/java/oshi/software/os/ClassicOperatingSystemFactory.java
+++ b/oshi-core/src/main/java/oshi/software/os/ClassicOperatingSystemFactory.java
@@ -1,0 +1,8 @@
+package oshi.software.os;
+
+public class ClassicOperatingSystemFactory implements OperatingSystemFactory{
+    @Override
+    public boolean isSupportedOnThisJdk() {
+        return true;
+    }
+}
--- a/oshi-core/src/main/resources/META-INF/services/oshi.software.os.OperatingSystemFactory
+++ b/oshi-core/src/main/resources/META-INF/services/oshi.software.os.OperatingSystemFactory
@@ -1,0 +1,1 @@
+oshi.software.os.ClassicOperatingSystemFactory


oshi-core-java24/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+        <artifactId>oshi-parent</artifactId>
+        <version>6.8.3-SNAPSHOT</version>
+    </parent>
+
+    <groupId>org.example</groupId>
+    <artifactId>oshi-core-java24</artifactId>
+
+    <properties>
+        <maven.compiler.source>24</maven.compiler.source>
+        <maven.compiler.target>24</maven.compiler.target>
```
mprins: these should be replaced with [`maven.compiler.release`](https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html), becase we use a newer version of the compiler plugin (should be done in the parent pom as well

oshi-core-java24/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>com.github.oshi</groupId>
+            <artifactId>oshi-core</artifactId>
+            <version>6.8.3-SNAPSHOT</version>
+            <scope>compile</scope>
```
mprins: this is the default
```suggestion
```

oshi-core-java24/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.github.oshi</groupId>
+            <artifactId>oshi-core</artifactId>
+            <version>6.8.3-SNAPSHOT</version>
+            <scope>compile</scope>
```
mprins: this is the default
```suggestion
```

oshi-core-java24/pom.xml
mprins: I see that is added in the profile

oshi-core-java24/.gitignore
```diff
@@ -0,0 +1,38 @@
+target/
```
dbwiddis: Hopefully all these are handled at the top project-level .gitIgnore so we don't need a subproject file.

oshi-core-java24/src/main/java/oshi/software/os/FfmOperatingSystemFactory.java
```diff
@@ -0,0 +1,16 @@
...
+public class FfmOperatingSystemFactory implements OperatingSystemFactory{
+    @Override
+    public boolean isSupportedOnThisJdk() {
+        String version = System.getProperty("java.specification.version");
+        try {
+            int v = Integer.parseInt(version);
+            return v >= 24;
+        } catch (NumberFormatException e) {
+            return false;
+        }
```
dbwiddis: On JDK8 "java.specification.version" there is `1.8`.  This is somewhat handled in the catch block but using exceptions for flow control isn't a good practice.  And we have parsing utils that handle these exceptions.
```suggestion
        if (version.startsWith("1.")) {
            return false;
        }
        return ParseUtil.parseIntOrDefault(version, 0) >= 24;
```
Or you could even use a one-liner to combine that.

oshi-core/src/main/java/oshi/software/os/OperatingSystemFactoryProvider.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class OperatingSystemFactoryProvider {
+    private static final OperatingSystemFactory INSTANCE = loadInstance();
+
+    private static OperatingSystemFactory loadInstance() {
+        ServiceLoader<OperatingSystemFactory> loader = ServiceLoader.load(OperatingSystemFactory.class);
+        for (OperatingSystemFactory factory : loader) {
+            if (factory.isSupportedOnThisJdk()) {
+                return factory;
+            }
```
dbwiddis: As @mprins noted this just returns the first service it finds, but ordering is not guaranteed.  Rather than reporting `isSupportedOnThisJDK()` how about the factory just reports its minimum version?   Then here you could save the factory to a temp variable along its "currentMinVersion" and update it if you find a newer option.

pom.xml
```diff
@@ -43,6 +43,7 @@
     <modules>
         <module>oshi-core</module>
         <module>oshi-demo</module>
+        <module>oshi-core-java24</module>
```
dbwiddis: Don't add this here, only add it in the `java24` profile.

pom.xml
```diff
@@ -985,6 +986,19 @@
...
+        <profile>
+            <id>java24</id>
+            <activation>
+                <jdk>[24,)</jdk>
+            </activation>
+            <modules>
+                <module>oshi-core-java24</module>
+            </modules>
+            <properties>
+                <maven.compiler.release>24</maven.compiler.release>
```
dbwiddis: Let's keep the same property pair as in jdk11 module (and since it's still executiong you can just omit this here).  You've already defined `maven.compiler.target` to be 24 in that POM.   So really we just need to say "only build and release this module in JDK24".

Related, a couple of things.  24 isn't out yet...
1. We can keep it "named' 24 for now, but keep the activation as 22 during development (with clear documentation we intend to raise it to LTS and keep it there).
2. We probably want to move `<oshi-dist>` here and change its configuration to include the jdk24 artifacts.  This can probalby wait until jdk24, so we don't "release" it early.

oshi-core-java24/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+
+    <properties>
+        <maven.compiler.source>24</maven.compiler.source>
+        <maven.compiler.target>24</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>com.github.oshi</groupId>
+            <artifactId>oshi-core</artifactId>
```
dbwiddis: Using `oshi-core` as a dependency here implies oshi-java24 is a standalone product, the only dependency you need to add.  (In which case we should depend on `oshi-core-java11` to inherit the modular descriptors.)

If we go this route we need to duplicate a lot of code (the entire codebase essentially) as we migrate.   I don't like this idea.

I had in mind something closer to an MRJAR, where when on higher JDK's you'd add/replace some classes.  (I'd have probably just gone the MRJAR route earlier but it seemed community consensus was to release different modules which is what we're doing!)

Take a look at the oshi-java11 pom.   We use the same codebase, we just add more files to it.   That's what I'd like to do here.   The base (oshi-core) artifact will have the SPI loader, but the files won't be present on that release so it'll default to existing classes.  On the oshi-core-java24, we'll add extra files that do provide a better version of the services (more performant).

oshi-core/src/main/java/oshi/software/os/OperatingSystemFactory.java
```diff
@@ -0,0 +1,8 @@
+package oshi.software.os;
+
+public interface OperatingSystemFactory {
+    /**
+     * Returns true if this factory supports the current JDK.
+     */
+    boolean isSupportedOnThisJdk();
```
dbwiddis: As mentioned in an earlier comment, better to provide an `int` minJdkVersionSupported here.

oshi-core-java24/.gitignore
rohan-coder02: Addressed in the latest commit

oshi-core/src/main/java/oshi/software/os/OperatingSystemFactoryProvider.java
rohan-coder02: Have implemented the logic of minimum version in the latest commit.
Let me know the feedback.

pom.xml
rohan-coder02: Addressed in the latest commit

pom.xml
rohan-coder02: Addressed in the latest commit

oshi-core-java24/pom.xml
rohan-coder02: In the latest commit, have done the changes in oshi-core-java24's pom.xml.
Let me know the feedback.

dbwiddis: > In the latest commit, have done the changes in oshi-core-java24's pom.xml. Let me know the feedback.

So just took a quick look.  I mostly see the factory providers which is fine.   But I don't see any intention on how to actually implement this in a side-by-side implementation.

Option A (my preference):
 - We keep the existing implementation for the most part, without changing the (JDK8 compatible) code
 - We do an attempt to read the latest SPI implementation prior to defaulting to the JDK8 implementaiton, and if so, we execute it instead.  If no SPI is loaded nothing changes.
 - The SPI implementations are a limited number of classes in the JDK24 project which is loaded *alongside* oshi-core.  THe main intent here is to offer more performant alternatives while still maintaining compatibility.

Option B (where I think you're headed):
 - Create an entirely new subproject for JDK24 that depends on the JDK8 implementation, so only one import is needed
 - Always use the "new" code in this implementation, with "fallback" to the oshi-core implementation when it hasn't been written.
 - Eventually get rid of the oshi-core (and JNA) dependency.

I think Option B is an eventual end state but it will take a long time to get here.  I think. Option A is a near term way to directionally head there while the proportion of non-JNA code is relatively small.

WDYT?

rohan-coder02: I was thinking of going with Option B i.e. separating out FFM-based code into a JDK 24+ specific module.
The idea was that we could just use the new module directly when running on newer JDKs and fall back to oshi-core implementation

That said, Option A helps, especially since FFM specific code for OS'es still needs to be added and we want to keep things working as-is for everyone else.

I'm listing the down the changes for option A below, let me know your views.

- Added SPI-based loading using ServiceLoader to load OperatingSystemFactory implementations.

- **Created**:
-- ClassicOperatingSystemFactory (for Java 8–23).
-- FfmOperatingSystemFactory (for Java 24+).

- **The OperatingSystemFactoryProvider**:
   -- Loads factories via SPI.
   -- Picks the one with the highest supported Java version ≤ current runtime.
   -- Falls back to classic if none found.

- **In the oshi-core-java24 POM**:
    -- Changed dependency to use oshi-core instead of oshi-core-java11.
    -- Added META-INF/services support.
    --Included shared resource/test source directories from oshi-core

dbwiddis: > I was thinking of going with Option B i.e. separating out FFM-based code into a JDK 24+ specific module.

I'm OK if you want to go this way, but doing the SPI stuff is probably not needed.

One thing to think about it cross-platform compatibility, particularly with OS's who don't follow openjdk.

> That said, Option A helps

The reason I prefer this is that I honestly think there's not that much interest in option B, and it's a lot of work.  A lets us offer newer features and better performance in a few cases where it matters (mostly process-reading stuff).

rohan-coder02: I've thought of below design for option A and wanted your feedback before proceeding with the same

1)

- In oshi.SystemInfo#createOperatingSystem, we check the Java version as below
```
private static OperatingSystem createOperatingSystem() {
    int javaVersion = Runtime.version().feature();
    
    // Use FFM-based implementation for Java 24+ if applicable
    if (javaVersion >= 24) {
        // You can later expand this switch with actual FFM-based implementations
        switch (CURRENT_PLATFORM) {
            case LINUX:
                return new FfmLinuxOperatingSystem();
            // case WINDOWS:
            //     return new FfmWindowsOperatingSystem();
            default:
                // fallback to classic for unsupported platforms
                break;
        }
    }
...
```
- There could be features say available in LinuxOperatingSystem but not in Ffm, hence, we need to ensure the fallback in FfmLinuxOperatingSystem
```
public class FfmLinuxOperatingSystem extends AbstractOperatingSystem {
    private final LinuxOperatingSystem fallback;

    public FfmLinuxOperatingSystem() {
        this.fallback = new LinuxOperatingSystem(); // classic fallback
    }

    @Override
    public List<ApplicationInfo> getInstalledApplications() {
        return fallback.getInstalledApplications(); // fallback to classic
    }

    @Override
    public GlobalMemory getMemory() {
        // Fully FFM-based implementation
        return new FfmGlobalMemory(); 
    }

    // ... and so on
}
```

2)

Should We Retain oshi-core-java24 as a Separate Module?

Keep the separate module (oshi-core-java24) for Java 24+ classes

(OR)

Move the FFM-backed implementations into oshi-core, possibly under a distinct oshi.ffm.* package, guarded by version checks.

dbwiddis: > Should We Retain oshi-core-java24 as a Separate Module?
> 
> Keep the separate module (oshi-core-java24) for Java 24+ classes
> 
> (OR)
> 
> Move the FFM-backed implementations into oshi-core, possibly under a distinct oshi.ffm.* package, guarded by version checks.

We definitely should keep a separate module, compiled for and with compatibility required for JDK 24+ (for now we can use 22+ and bump it until LTS).

The question I proposed was "should we have oshi-core with a jdk24 optional dependency to load improvements under SPI" and I think after looking at this, it's probably more complicated to try to maintain the SPI model than just a clean new module as you propose.

So let's head in the Option B direction:
- new JDK24 module as the only dependency people use
- try to remove any reference to JNA in this module (so the SystemInfo class will not rely on `Platform` but instead to much the same thing... see the oshi-ffm project for how that was done there)
- have JDK24+ (22+ for now) code when available. We'll probably want full copies of the OperatingSystem and HardwareAbstractionLayer classes in 24.  
- From there, the platform specific implementations will either have their own classes called in the JDK24 project, or will call the oshi-core equivalent class, which will be a dependency.

WDYT?

rohan-coder02: 1)

- With the previous design, say, LinuxOperatingSystem extends AbstractOperatingSystem that implements OperatingSystem, suppose there is a method here that’s only meaningful in JNA context
 -- Implement a dummy method in oshi-core-java 24 version
 -- The oshi-core JNA specific methods/APIs get into the FFM module
- In the new module approach with its own interfaces, we can avoid adding JNA specific methods and iterate freely

2)

Without SPI approach, easy for user to just load oshi-core-java24.jar and access the APIs

3)

- The below directory structure is present oshi-ffm, I'll proceed with the same.
```
oshi-core-java24/
├── pom.xml
├── src/
│   ├── main/
│   │   ├── java/
│   │   │   └── oshi/
│   │   │       ├── SystemInfo.java               <-- Custom for Java 22+
│   │   │       ├── hardware/
│   │   │       │   └── HardwareAbstractionLayer.java
│   │   │       └── software/
│   │   │           └── os/
│   │   │               └── OperatingSystem.java
```

dbwiddis: > * With the previous design, say, LinuxOperatingSystem extends AbstractOperatingSystem that implements OperatingSystem, suppose there is a method here that’s only meaningful in JNA context

I don't think we have any (API) methods that expose dependencies on anything JNA.  I may be wrong in some edge cases.

> * The below directory structure is present oshi-ffm, I'll proceed with the same.

Probably a good start, other than my attempt to rename the packages with an ooo prefix. Probably best to leave as-is for a seamless transition.

rohan-coder02: I’ve set up the initial structure for the oshi-core-java24 module

- Independent module for JDK 22+/24+ using FFM.
- SystemInfo.java added with modern switch pattern for platform detection.
- Currently just includes top-level interfaces and basic structure.

For next step, shall I proceed by adding classes like LinuxOperatingSystem, LinuxHardwareAbstractionLayer, etc., and fill them out incrementally?

dbwiddis: > For next step, shall I proceed by adding classes like LinuxOperatingSystem, LinuxHardwareAbstractionLayer, etc., and fill them out incrementally?

I'm wondering if we're going to have a problem here with the same-named classes in multiple projects once you import oshi-core.   We may need to do that `ooo.` prefix.

rohan-coder02: Yes.

1)

Could we have the prefix as `ffm` instead of `ooo`?

2)

Before I proceed with the OS-level implementations (e.g., LinuxOperatingSystem), shall I include those changes in this commit or open a separate PR?

I can also prepare a list of APIs for each OS, indicating whether they're JNA or non-JNA based, to implement accordingly.

dbwiddis: > Could we have the prefix as ffm instead of ooo?

It's generally a standard to have package names beginning with reverse DNS, and since I have the oshi.ooo domain I figured that was an option.  Of course, it also obligates me to maintain that domain. :)

If we're going to use ffm it should not be the prefix, but perhaps a second field, such as `oshi.ffm` for the `SystemInfo` class and `oshi.ffm.hw`, `oshi.ffm.os`,  etc.   It should be in a simple, predictable place that consumers of the API can do a search/replace.

All that said:

One big thing to think about is modularity.  Ideally we want to import `oshi-core-java11` here but the platform-specific implementations there are not exposed in the module-info (by design).  How are we going to handle this?  At some point we need to expose some non-API classes to modules... or copy over the files like we currently do with the java11 build, selectively including replacements once we have them (and thus avoiding the dependency).

Which also means if we're copying over files we don't need to rename the packages at all...

> Before I proceed with the OS-level implementations (e.g., LinuxOperatingSystem), shall I include those changes in this commit or open a separate PR?

For this PR, let's set up the new class so that it compiles and runs tests, etc. with the new API/Interface classes but all of them delegating to the oshi-core implementations.  Then each OS implementation can proceed independently (and possibly in parallel).

rohan-coder02: > It's generally a standard to have package names beginning with reverse DNS, and since I have the oshi.ooo domain I figured that was an option. Of course, it also obligates me to maintain that domain. :)

Thanks, got it.

1)

I'll rename the packages to follow oshi.ffm.* convention — e.g., oshi.ffm, oshi.ffm.hw, oshi.ffm.software.os, etc., for predictability and alignment.

2)
On modularity: Agreed

Shall I copy the necessary platform-specific classes from `oshi-core` into the `oshi-core-java24` module as needed, similar to the Java 11 build?
This will help avoid requiring internal exports and keeps modules clean.

dbwiddis: > I'll rename the packages to follow oshi.ffm.* convention

As mentioned, if you can figure out how to do this by copying source files rather than importing oshi-core(-java11), this may not be needed.  And might be preferred to leave the values unchanged.

> Shall I copy the necessary platform-specific classes from oshi-core into the oshi-core-java24 module as needed, similar to the Java 11 build?

I think this is probably the best way forward.  A self-contained build using build-helper-maven-plugin to copy the "default" files from oshi-core.  Then All you'd need to do is exclude the classes you don't want to copy ([although I'm not sure that works](https://stackoverflow.com/questions/64973148/excluding-files-from-jar-with-build-helper-maven-plugin)) or more narrowly target which classes you do copy over.

It's a hack, but I think the only other real option is to fork to a branch, have a separate release, and do a lot of backporting.  Which may be a reasonable approach as well.

rohan-coder02: Summary of the latest commit:

- Introduced the oshi-core-java24 module with package structure oshi.ffm.* to separate FFM-based implementations.
- Set up modular structure using build-helper-maven-plugin to copy source files from oshi-core, avoiding direct dependency and allowing selective overrides.
- Interfaces OperatingSystem and HardwareAbstractionLayer created under oshi.ffm.* and are wired through SystemInfo using memoized delegates.
- Current implementation delegates to existing oshi-core classes for all platforms as a starting point.
- Build and tests compile and run successfully using Java 24.

Let me know your views

dbwiddis: > Let me know your views

I'm a bit confused here.  How does this code even run?
```java
private final Supplier<OperatingSystem> os = memoize(SystemInfo::createOperatingSystem);
private final Supplier<HardwareAbstractionLayer> hardware = memoize(SystemInfo::createHardware);
```

Those methods aren't in the existing `SystemInfo` class and you don't import the one from the dependency (or copied over).  And I think your `PlatformEnum` class will have a package/classname conflict.

> Set up modular structure using build-helper-maven-plugin to copy source files from oshi-core, avoiding direct dependency and allowing selective overrides.

I don't see that code and I still see oshi-core? Did you forget to commit the new pom.xml?

> Interfaces OperatingSystem and HardwareAbstractionLayer created under oshi.ffm.*

If we're not using the oshi-core dependency we can actually keep the same package names without ffm.  The different packages are only needed if we use a dependency which, I think with modules, overly complicates things.

> Build and tests compile and run successfully using Java 24.

Perhaps locally with your pom.xml but they woudln't here (our PR github workflow max version is 21). :)

rohan-coder02: I just rechecked again, it looks commit didn't happen properly.

Have done it again, let me know.

oshi-core-java24/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+                    <execution>
+                        <id>add-sources</id>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>add-source</goal>
+                        </goals>
+                        <configuration>
+                            <sources>
+                                <!-- Copy source files from oshi-core to override selectively -->
+                                <source>${project.basedir}/../oshi-core/src/main/java</source>
```
dbwiddis: So this works right now with the `.ffm` package naming, and means we don't have to be picky about what we copy over, it'll all just be here.   While good for the short term, as we transition more and more implementations these will mostly end up unused classes.  And right now we're even including the non Windows/Mac/Linux OS's which will never be used.

My thought is we have in this project the top level interface classes in exactly the same package naming, and this copying occurs only at the "platform" level for the implementations for the supported OS's.

oshi-core-java24/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+    <parent>
+        <groupId>com.github.oshi</groupId>
+        <artifactId>oshi-parent</artifactId>
+        <version>6.8.3-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>oshi-core-java24</artifactId>
+    <packaging>jar</packaging>
+
+    <name>oshi-core-java24</name>
```
dbwiddis: So, I just now realized, it's JDK 25 that's LTS!  So we should target that, not 24.  Can't believe I haven't looked that up lately.

oshi-core-java24/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+                            </sources>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-javadoc-plugin</artifactId>
+                <configuration>
+                    <source>24</source>
```
dbwiddis: Eventually we'll bump this to 25.

oshi-core-java24/src/main/java/oshi/PlatformEnum.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2025 The OSHI Project Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package oshi;
```
dbwiddis: with the current copying over of `/src/main/java` this package will get overwritten.  One more reason to copy over lower level (platform directory) implementations and keep the interfaces in this project.

oshi-core-java24/src/main/java/oshi/SystemInfo.java
```diff
@@ -0,0 +1,90 @@
...
+package oshi;
+
+import oshi.hardware.HardwareAbstractionLayer;
+import oshi.hardware.linux.LinuxHardwareAbstractionLayer;
+import oshi.hardware.mac.MacHardwareAbstractionLayer;
+import oshi.hardware.windows.WindowsHardwareAbstractionLayer;
+import oshi.software.os.OperatingSystem;
+import oshi.software.os.linux.LinuxOperatingSystem;
+import oshi.software.os.mac.MacOperatingSystem;
+import oshi.software.os.windows.WindowsOperatingSystem;
```
dbwiddis: Here's where I think we keep the `oshi.hardware.HAL` and `oshi.software.os.OS` classes here and copy over the `.linux`, `.mac`, and `.windows` files from the old implementation until we replace them.

oshi-core-java24/src/main/java/oshi/hardware/HardwareAbstractionLayer.java
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright 2025 The OSHI Project Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package oshi.hardware;
+
+public interface HardwareAbstractionLayer {
```
dbwiddis: We *could* copy over the `oshi-core` interface classes...

I don't see a reason for this empty interface though.

oshi-core-java24/src/main/java/oshi/hardware/linux/LinuxHardwareAbstractionLayer.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright 2025 The OSHI Project Contributors
+ * SPDX-License-Identifier: MIT
+ */
+package oshi.hardware.linux;
+
+import oshi.hardware.HardwareAbstractionLayer;
+
+public class LinuxHardwareAbstractionLayer implements HardwareAbstractionLayer {
```
dbwiddis: Whether we copy these over or keep our own copy may be dependent on how different the implementation is.  For now, the empty class is useless if we're copying over it.

oshi-core-java24/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
```diff
@@ -0,0 +1,17 @@
...
+package oshi.software.os.windows;
+
+import oshi.software.os.common.AbstractOperatingSystem;
+
+public class WindowsOperatingSystem extends AbstractOperatingSystem {
+
+    private final oshi.software.os.OperatingSystem delegate;
+
+    public WindowsOperatingSystem() {
+        this.delegate = new oshi.software.os.windows.WindowsOperatingSystem();
```
dbwiddis: I like the idea of a delegate, but not sure how we're going to use it?

dbwiddis (commented):
I like this! We're getting close!

Currently you have a lot of classes we're overwriting with the copy.  I'm not quite sure where the boundary of copying needs to be, and we can fine tune that later.

oshi-core-java24/pom.xml
rohan-coder02: In the latest commit, I plan to include specific files as below.
```
                            <sources>
                                <!-- Copy source files from oshi-core to override selectively -->
                                <source>${project.basedir}/../oshi-core/src/main/java/oshi/software/os/linux</source>
                                <source>${project.basedir}/../oshi-core/src/main/java/oshi/software/os/windows</source>
                                <source>${project.basedir}/../oshi-core/src/main/java/oshi/software/os/mac</source>
                                <source>${project.basedir}/../oshi-core/src/main/java/oshi/hardware/linux</source>
                                <source>${project.basedir}/../oshi-core/src/main/java/oshi/hardware/windows</source>
                                <source>${project.basedir}/../oshi-core/src/main/java/oshi/hardware/mac</source>
                            </sources>
```

oshi-core-java24/pom.xml
rohan-coder02: Few queries before I proceed with the changes in the latest commit

1)

Shall I rename `oshi-core-java24` to `oshi-core -java25`?

2)

In **pom.xml** of `oshi-core-java24`, shall I proceed with the below changes?
```
    <properties>
        <maven.compiler.source>25</maven.compiler.source>
        <maven.compiler.target>25</maven.compiler.target>
        <maven.compiler.testSource>25</maven.compiler.testSource>
        <maven.compiler.testTarget>25</maven.compiler.testTarget>
    </properties>
```

3)

I'm currently assuming that the new module will be called `oshi-core-java25`, targeting JDK 25 since it's the next LTS.

Just to confirm — if we name it -java25 but avoid using JDK 25-specific features (stick to source/target 24), would we still be able to support users running Java 24 as well?

checking on this as there could be queries from users on compatibility.

oshi-core-java24/src/main/java/oshi/PlatformEnum.java
rohan-coder02: Sure, this will be addressed by selective copy by including only platform directories

oshi-core-java24/src/main/java/oshi/SystemInfo.java
rohan-coder02: I'll proceed with below directory structure as the base for now
```
oshi-core-java24/
├── pom.xml
└── src/
    └── main/
        └── java/
            └── oshi/
                ├── SystemInfo.java
                ├── PlatformEnum.java
                ├── hardware/
                │   └── HardwareAbstractionLayer.java
                └── software/
                    └── os/
                        └── OperatingSystem.java
```

oshi-core-java24/src/main/java/oshi/hardware/linux/LinuxHardwareAbstractionLayer.java
rohan-coder02: I'll address in the latest commit where SystemInfo class will include platform specific implementations from oshi-core for now

Hence, the empty class will be removed

oshi-core-java24/src/main/java/oshi/software/os/windows/WindowsOperatingSystem.java
rohan-coder02: won't be using delegate approach as currently not extending or modifying any behavior from the oshi-core platform-specific implementations as SystemInfo class will include platform specific implementations from oshi-core for now

Once I begin with FFM-specific enhancements, will think on this

oshi-core-java24/pom.xml
dbwiddis: > Shall I rename `oshi-core-java24` to `oshi-core -java25`?

Yes please.  We plan to support the LTS version.  In theory we could do 22 based on our FFM support but it's going to take us 6 months to get this to any sort of releaseable shape so this is good timing.

> In **pom.xml** of `oshi-core-java24`, shall I proceed with the below changes?

Only if you plan on using 25-ea for the build for the next 6 months.  We can keep it as 24 for now and bump to 25 when that's released.

> Just to confirm — if we name it -java25 but avoid using JDK 25-specific features (stick to source/target 24), would we still be able to support users running Java 24 as well?

For now, yes.   Once JDK 25 is released, even if things could work, let's just stick with the LTS version (it's the same reason why the modular version is java11 instead of java9 which in theory would work as all we're adding is the module-info and no other non-jdk8 features!)

dbwiddis: TO be more clear on 25 vs. 22 (what we're really doing this for), JDK 22 and 23 are no longer supported. We should not encourage people to be using them.

rohan-coder02: Thanks for the points.

Before I commit, had one doubt, upon executing `mvn clean test` in `oshi-core-java25` module 

+ I'm selectively including only the needed shared sources from oshi-core using build-helper-maven-plugin
```
Example:

<source>${project.basedir}/../oshi-core/src/main/java/oshi/software/os/linux</source>
 <source>${project.basedir}/../oshi-core/src/main/java/oshi/software/os/windows</source>
<source>${project.basedir}/../oshi-core/src/main/java/oshi/software/os/mac</source>
<source>${project.basedir}/../oshi-core/src/main/java/oshi/hardware/linux</source>
<source>${project.basedir}/../oshi-core/src/main/java/oshi/hardware/windows</source>
<source>${project.basedir}/../oshi-core/src/main/java/oshi/hardware/mac</source>
<source>${project.basedir}/../oshi-core/src/main/java/oshi/software/common</source>
<source>${project.basedir}/../oshi-core/src/main/java/oshi/util</source>
...
```
+ When running `mvn clean test` from `oshi-core-java25`, (It includes OS'es other than Mac, Windows, Linux)
```
...
oshi/driver/unix/aix/PsInfo.java:[40,26] cannot find symbol
symbol: class AixLibc
...
```
+ Ideally, we would need isolated testing of oshi-core-java25 i.e. currently `oshi/SystemInfoTest.java`
+ Tried using ` mvn clean test -Pjava25` but observed same issue

dbwiddis: I'm assuming you do *not* have an entry for `<source>${project.basedir}/../oshi-core/src/main/java/oshi/driver/unix</source>`?    Can you get more insight into which test class is failing?  Perhaps those test classes also need selective importing?

rohan-coder02: 1)

Was able to resolve the earlier issue by selectively importing only the required directories.

Now, the only issue remaining is with a few classes/interfaces under `oshi.software.os` and `oshi.hardware` that are referenced elsewhere but not present in `oshi-core-java25`, since we're intentionally not including the full package to avoid overriding OperatingSystem and HardwareAbstractionLayer prematurely.

```
[ERROR] .../NetSessionData.java:[17,24] cannot find symbol
  symbol:   class OSSession
  location: package oshi.software.os

[ERROR] .../InstalledAppsData.java:[13,24] cannot find symbol
  symbol:   class ApplicationInfo
  location: package oshi.software.os
```

Would it be acceptable to omit these interfaces for now and proceed with the following minimal initial commit? 
I plan to include these interfaces and related implementations once OperatingSystem and HardwareAbstractionLayer are built out for `oshi-core-java25`
```
oshi-core-java25/
├── pom.xml
└── src/
    └── main/
        └── java/
            └── oshi/
                ├── SystemInfo.java
                ├── PlatformEnum.java
```

2)

Upon approaching the issue as per 1) with selective directory imports, I'm now hitting errors where some classes in `oshi-core` refer to `PlatformEnum.SOLARIS`, `AIX`, and `FREEBSD`, which aren't currently defined in the minimal PlatformEnum class in `oshi-core-java25`
```
[ERROR] AbstractPowerSource.java:[227,14] cannot find symbol
  symbol:   variable SOLARIS
  location: class oshi.hardware.common.AbstractPowerSource
```
To resolve this, I plan to add the following constants to oshi-core-java25's PlatformEnum:
```
SOLARIS("Solaris"),
AIX("Aix"),
FREEBSD("Freebsd"),
```
Keep SystemInfo's static detection logic limited to LINUX, MACOS, and WINDOWS, mapping all others to UNSUPPORTED

dbwiddis: > Was able to resolve the earlier issue by selectively importing only the required directories.

Nice!

> Now, the only issue remaining is with a few classes/interfaces under `oshi.software.os` and `oshi.hardware` that are referenced elsewhere but not present in `oshi-core-java25`, since we're intentionally not including the full package to avoid overriding OperatingSystem and HardwareAbstractionLayer prematurely.
> 
> ```
> [ERROR] .../NetSessionData.java:[17,24] cannot find symbol
>   symbol:   class OSSession
>   location: package oshi.software.os
> 
> [ERROR] .../InstalledAppsData.java:[13,24] cannot find symbol
>   symbol:   class ApplicationInfo
>   location: package oshi.software.os
> ```
> 
> Would it be acceptable to omit these interfaces for now and proceed with the following minimal initial commit? I plan to include these interfaces and related implementations once OperatingSystem and HardwareAbstractionLayer are built out for `oshi-core-java25`

Well, the *ideal* approach is to copy over the API interfaces and classes as they, in theory, shouldn't be changing (to maintain API compatibility), and then be selective with implementations.  But as we move forward in this I'm realizing we're using build-helper-maven-plugin for far more than it was designed to do.   It seems you can exclude subdirectories for resources but not for sources.

So my thought is that we should just have full copies of the API interfaces and classes in these top-level directories.  Yes, it's added effort to update the API in two places if it changes, but that should be rare.  So in this case just copy over the eisting OSSession and ApplicationInfo classes.

We can fine tune this later, or figure out another way of copying over files we need.  Honestly we're just being lazy at this point and it's possible we need a different maven plugin to do what we're trying to do. (CC @hazendaz, our Maven expert, if you have any insight here.)


> Upon approaching the issue as per 1) with selective directory imports, I'm now hitting errors where some classes in `oshi-core` refer to `PlatformEnum.SOLARIS`, `AIX`, and `FREEBSD`, which aren't currently defined in the minimal PlatformEnum class in `oshi-core-java25`

Yeah, that's poor coding on my part.

> To resolve this, I plan to add the following constants to oshi-core-java25's PlatformEnum:

Seems a good workaround!

rohan-coder02: I'll check this weekend if any maven plugin suits the use-case we're trying to do.

I'll keep you posted else will go ahead with the approach of having full copies of the API interfaces and classes in these top-level directories.

rohan-coder02: I thought through this and below is the analysis.

Our priority for `oshi-core-java25` is to leverage FFM features introduced in Java 25.

**Approach:**

+ Use oshi-core as a dependency initially
+ Exclude classes from `oshi-core` that will be reimplemented in `oshi-core-java25`
+ Gradually replace remaining dependencies with FFM-based/Latest-Java-feature implementations

**Pros:**

+ Immediate focus on FFM development
+ Utilise required code snippets from `oshi-core` 
+ Lower upfront rewrite effort

**Cons & Handling:**

+ Potential test/classpath conflicts → Handled via Maven exclusions and selective test configs
+ Exclusion management overhead → We can have a roadmap for this to add the excluded files as we proceed with implementation
+ Dependency coupling on oshi-core → Long-term goal is full decoupling once the FFM version are ready

Using `build-helper-maven-plugin` there is no inclusion/exclusion option for individual classes and available at package level as observed previously. Any time we override a single class in `oshi-core-java25`, we must exclude its entire package from `oshi-core`.

Using the current approach i.e. `maven-shade-plugin` exclusion happens at individual classes and we will be able to replaces classes progressively and no package removals.

The end goal in this approach is `oshi-core-java25` will be standalone just that initially we prioritise FFM development and later on other dependencies from `oshi-core`.

rohan-coder02: Hi @dbwiddis 

wanted to know your view on the above else I'll proceed with having full copies of the API interfaces and classes in these top-level directories.

dbwiddis: Hey @rohan-coder02 sorry for the delay here.

Let me come at this from top level tenets: 
 - it should be a standalone dependency
 - it should be modular
 - it should have the same API (including package paths)
 - it should permit class-by-class (or even method-by-method) replacement
 - eventually we should eliminate the oshi-core dependency

Originallyl I thought we could/should not include `oshi-core` because that would give us class/package name conflicts.  However, excluding or renaming classes using the shade plugin would allow us to bring it in and use whichever classes we want at, forexample, `legacy.oshi.hardware.platform.windows....".  So a shaded/renamed strategy would work.

Alternately, copying files over with build-helper can accomplish the same thing if we choose to use new package naming for the ffm classes until all classes in the package are copied over.  This means we would have new copies of all the API interfaces and classes (platform-agnostic) in all the top-level classes.  These would have to _both_ be updated in the rare case that we change the API (generally only new features do this to add something).  New FFM classes would be in new packages that wouldn't conflict with/overwrite the old classes.  Here's a concrete example of what I'm thinking:
 - new copies of packages `oshi` and `oshi.hardware` and `oshi.os.software` (API classes).  Everything else under that is copied over.
 - top-level "logical tree" classes in a new package:
   - In `SystemInfo` class the new implementations for `WindowsOperatingSystem`, `MacOperatingSystem`, and `LinuxOperatingSystem` live in new `oshi.ffm.os.software` package.  
   - Initially these are just copies of the original classes which call the appropriate `oshi.os.software` class.
   - As we create new implementations (e.g., `WindowsFileSystem` we move the new implementation into the `oshi.ffm.os.software` package.  The old class is still there, just unreferenced.
   - Eventually once all classes are removed we can exclude that legacy package from copying.  There's no rush to do that, they can happily live unused for a long time (unlike the shading version where we have to constantly update the excluded classes)
   - There are probably a few rare cases where implementations reach across package and this may break things but we can handle those with appropriate util methods.
   - We could even keep the same package structure and just have a new class name for the FFM stuff (`WindowsHardwareAbstractionLayerFFM` and `WindowsOperatingSystemFFM`, for example).  

I think both approaches generally accomplish the same thing.  I think copying and adding new classes with either a different name or different (new) package is much easier than shading.

But you're implementing it and at this point I'm happy to let you pick which way to go!

rohan-coder02: Thanks for the points.

In the recent commit, below approach is used

- Introduced SystemInfo and PlatformEnum classes in oshi.ffm package
- Configured Maven Build Helper plugin to include source files from oshi-core
- Added required dependencies: JNA, JNA-Platform, and SLF4J API
- Verified build with `mvn clean package` 
- Prepared for further implementation of hardware and software.os packages (currently, its empty as its initial commit, will start working on this in coming weeks)

Wrote a simple Java program to test `oshi-core-java25-6.8.3-SNAPSHOT.jar` and it runs fine.
```
import oshi.ffm.SystemInfo;

class ListInstalledApps {
        System.out.println(System.getProperty("java.version"));
        System.out.println(si.getOperatingSystem().getInstalledApplications().size());
    }
}
```

Output:

```
24.0.1

384
```

dbwiddis: Looks great, just one big decision point.

It looks like you're changing the package naming for the new dependency to `oshi.ffm.*`.  That's fine if we want to make this decision; anyone switching to the new dependency just needs to do a search/replace of all their imports.  Then we can just build the new structure parallel to the old one.

I've got no problem with this if we just want to make that decision. It's much simpler than any other alternative.

rohan-coder02: Yes, users switching to `oshi-core-java25` just do a simple search/replace on imports.

I’ve added the API classes under `oshi-core-java25` → hardware and software.os, with implementations still delegating to the legacy oshi code where applicable. This keeps things compiling and functional for now, while allowing us to build out the FFM-specific internals incrementally.

let me know if there’s anything else need to be included in this PR for the initial commit

dbwiddis: > let me know if there’s anything else need to be included in this PR for the initial commit

Flip it out of draft? :)

Full disclosure: we've got a few pending changes that I want to cut a release on before merging this, but after that I'm fine incuding this and bumping the minor version (at least).  I'd want to have at least some sort of indication in the README documenting this new dependency and the caveats on our eventual JDK25 requirement.

rohan-coder02: 1)

Have flipped the PR out of draft. 

2)

Does the snippet below in **README.md** after `Downloads and Dependency Management` looks ok?
Do let me know if any changes for the same.

OSHI Java 25+ Module
----------------------------
A new module, **`oshi-core-java25`**, is now available.

- **Purpose:** This module provides API-compatible implementations using the JDK Foreign Function & Memory (FFM) API, replacing JNA for native access.
- **Compatibility:**  
  - Compiles on **JDK 24+**.  
  - Will require **JDK 25+** at runtime once the full migration is complete.  
- **API Structure:**  
  - The same top-level APIs remain under the `oshi.*` package for source compatibility.  
  - New FFM-based implementations are in `oshi.ffm.*`.  
- **Status:**  
  - Some methods still delegate to legacy JNA-based internals until the migration is complete.

3)

I'll also update the CHANGELOG and update the PR.

dbwiddis: > Does the snippet below in README.md after Downloads and Dependency Management looks ok?

Given the direction we're going, I think we need a bit more clarity:
1. Rather than "This module provides ... using the JDK FFM" would like to qualify with "intends to provide". A comment about migration "over time with community contributions" would be a bonus.
2. API structure seems confusing. Let's be clear about "use this dependency in place of oshi-core, and add .ffm to your import statements".  Not exactly those words but conveying that meaning.

rohan-coder02: I've updated the readme in the recent commit, let me know if any changes needed

dbwiddis: Sorry I've been swamped for the last week/weekend. Will get to it in the next couple of days.

README.md
```diff
@@ -52,6 +52,21 @@ Current Development (SNAPSHOT) downloads
...
+A new module, **`oshi-core-java25`**, is now available.
+
+- **Purpose:** This module intends to provide API-compatible implementations using the JDK Foreign Function & Memory (FFM) API, replacing JNA for native access over time with community contributions.
+- **Compatibility:**  
+  - Compiles on **JDK 24+**.  
+  - Will require **JDK 25+** at runtime once the full migration is complete.  
+- **Usage:**  
+  - Use this dependency **in place of** `oshi-core`.  
+  - Update your imports to use the equivalent `oshi.ffm.*` packages when accessing implementation classes.  
+  - Top-level API classes remain in the same `oshi.*` packages to minimize code changes for most applications.
```
dbwiddis: These seem to conflict with each other.  "Update your imports" implies a global search and replace of `oshi.` with `oshi.ffm.`.  This is what the contribution below looks like, copying over all the API classes (which are actually the few things that likely *don't* need to change.

But then the second line indicates the same `oshi` interfaces/top-level packages which we don't have here.

README.md
```diff
@@ -52,6 +52,21 @@ Current Development (SNAPSHOT) downloads
   * JDK8: [oshi-core-6.8.4-SNAPSHOT](https://central.sonatype.com/service/rest/repository/browse/maven-snapshots/com/github/oshi/oshi-core/6.8.4-SNAPSHOT)
   * JPMS: [oshi-core-java11-6.8.4-SNAPSHOT](https://central.sonatype.com/service/rest/repository/browse/maven-snapshots/com/github/oshi/oshi-core-java11/6.8.4-SNAPSHOT/)
 
+OSHI Java 25+ Module
+----------------------------
+A new module, **`oshi-core-java25`**, is now available.
+
+- **Purpose:** This module intends to provide API-compatible implementations using the JDK Foreign Function & Memory (FFM) API, replacing JNA for native access over time with community contributions.
+- **Compatibility:**
```
dbwiddis: Should mention a reduction in OS support (at least initially) until other operating systems have JDK25 options.

oshi-core-java25/src/main/java/oshi/ffm/SystemInfo.java
```diff
@@ -0,0 +1,90 @@
...
+        default -> throw new UnsupportedOperationException(NOT_SUPPORTED);
+        };
+    }
+
+    private static HardwareAbstractionLayer createHardware() {
+        return switch (CURRENT_PLATFORM) {
+        case LINUX -> new LinuxHardwareAbstractionLayer();
+        case MACOS -> new MacHardwareAbstractionLayer();
+        case WINDOWS -> new WindowsHardwareAbstractionLayer();
+        default -> throw new UnsupportedOperationException(NOT_SUPPORTED);
```
dbwiddis: This indentation looks off (or non-existent).  Are we running spotless on this repo?

dbwiddis (commented):
A few documentation questions but the structure looks good.

I wonder if there's any automation we can create to check on PRs if the interfaces are changed in one location that they are changed in another...

oshi-core-java25/src/main/java/oshi/ffm/SystemInfo.java
rohan-coder02: Yes `mvn spotless:apply` command was run before proceeding with the commit

README.md
rohan-coder02: Is the below in line under `Compatibilty` fine?

```
Initial support may be limited to operating systems with JDK 25 builds; broader OS support will follow as availability expands.
```

dbwiddis: Yeah, saw that later. That's good!

README.md
rohan-coder02: Yes.

Shall I proceed with the change below?
This is easy for users using `oshi-core-java25` to understand and do change.

```
Update imports to reference `oshi.ffm.*` packages instead of `oshi.*`
```

dbwiddis: > Shall I proceed with the change below?

I'm not sure... we seem to keep going in circles.

The way I see it there are essentially two options, and you can pick whichever one you find easier.

Option A. Keep existing API packages and API/interfaces (no change for users) and only change/overwrite/replace the `SystemInfo.java` entry point (or perhaps all the classes in that top level `oshi` package.  

From inside that top-level System-Info we can call out to `.ffm.platform` implementations (or just implementations with FFM in the class name but in the same package structure to avoid overwriting) or the original implementations, transparently to the user, all of which inherit from/implement the existing interfaces (and even abstract/default parent classes).

Advantages:
 - Keep the same old classes by copying over everything under the top level.
 - Guaranteed to keep same API and not have to make changes in two places if/when the API is updated

Disadvantage:
 - Potential confusion with multiple similar classes, but this would be temporary.
 - I think there's at least one API class that comes from JNA that we'd need to replace.

B. Change every package import to be `oshi.ffm` essentially creating an entirely new project.  Either import oshi-core as a dependency (this works with the changed filename!) or copy over the oshi-core files temporarily until all new implementations have been written.

Advantages:
 - Clear separation of the projects and easier to get rid of JNA at the end.
 - Allows users to actually import both, using them for different types of applications

Disadvantages:
 - Old (copied) implementations still "import" the old interfaces, so if users update their code to the new `oshi.ffm` interface (like we tell them to) the copied-over code wouldn't inherit properly and it would break their code.  We'd need to temporarily have copies of the old interfaces inherit from the new interfaces to make sure this all works.  Which prevents us cleanly overwriting/copying them.

It looks like you're headed toward option B but it still seems to me that A would be easier to do with a lot less work.   Just copy over everything as-is, and build new classes and/or packages alongside it that don't get overwritten.  The only "special" thing any user has to do is change which SystemInfo class they start the whole chain with.  We can have the `oshi.ffm.SystemInfo` version (or `oshi.SystemInfoFFM`) but still refer to `oshi.hardware.*` and `oshi.software.os.*` API interfaces and classes.

rohan-coder02: Let me know if the below understanding is correct in the case of Option A

Previous user code using `oshi-core`

```
import oshi.SystemInfo;  
import oshi.software.os.OperatingSystem;

public class Demo {
        System.out.println(os.getFamily() + " " + os.getVersion());
    }
}
```

Once user decides to use `oshi-core-java25`

```
import oshi.ffm.SystemInfo;   // new entrypoint
import oshi.software.os.OperatingSystem;

public class Demo {
        SystemInfo si = new SystemInfo();  // FFM-backed
        System.out.println(os.getFamily() + " " + os.getVersion());
    }
}
```

under the hood SystemInfo delegates to ffm based implementations

So, the below update in README should help the user in option A

```
Import oshi.ffm.SystemInfo instead of oshi.SystemInfo as the entrypoint.
All other imports (oshi.hardware.*, oshi.software.os.*) remain unchanged.
```

dbwiddis: > Let me know if the below understanding is correct in the case of Option A

Yes, that's precisely it. 

Just make that top-level class different in some way and it can refer internally to old API classes (that don't change and we don't have to worry about making copies of API classes) and (temporarily) old implementations until they're replaced.

Up to you if you want to change the package as you've proposed, or just have a different class name.  `SystemInformation` in the same package could be the new version.  Or `SystemInfo25`, or `SystemInfoFFM` etc.  Remember the old SystemInfo class will still be around from the copy. 
 - Unless you only add sources from all the sub-packages of `oshi` (annotation, driver, hardware, jna, software, util) but have entirely new `oshi` content (Just the SystemInfo and PlatformEnum classes) for the FFM version. In this case, you don't even need to rename/repackage it!  The user literally needs to make no changes.

I do like the idea of keeping the API packages the same, though.

But the above comment I think is the simplest way forward.

rohan-coder02: Introduced SystemInfoFFM as a new entry-point that delegates to FFM-based implementations.

Users only need to swap oshi.SystemInfo → oshi.SystemInfoFFM; all other API packages remain unchanged.

Sample code with oshi-core-java25

```
import oshi.SystemInfoFFM;

public class ListInstalledApps {
    public static void getApps() {
        SystemInfoFFM si = new SystemInfoFFM();
        System.out.println(si.getOperatingSystem().getInstalledApplications().size());
    }
}
```

dbwiddis:
> Sure, I'll address in the next PR/commit.

Already handled as I was trying to set up my IDE.  I really like this setup and you can see in my in-progress PR how easy it is for me to add new stuff.  Excited about OSHI's ability to quickly adopt FFM ... thanks so much for your patience in seeing this PR through to completion!

PR #2926, opened 2025-06-30
Exclude CD-ROM from Windows FileSystem localOnly query

cesnekmichal:
Improve reading logical local drives.

from: 
wmic logicaldisk where "DriveType!=4" get DeviceID, VolumeName, FileSystem, DriveType 
to: 
wmic logicaldisk where "DriveType=3" get DeviceID, VolumeName, FileSystem, DriveType

where
| DriveType | Meaning | 
| 0 | Unknown | 
| 1 | No root directory | 
| 2 | Removable disk (e.g., USB) | 
| 3 | Fixed disk (e.g., internal HDD/SSD) | 
| 4 | Network drive (e.g., \server\share) | 
| 5 | CD-ROM | 
| 6 | RAM disk |

### DIFF
--- a/oshi-core/src/main/java/oshi/driver/windows/wmi/Win32LogicalDisk.java
+++ b/oshi-core/src/main/java/oshi/driver/windows/wmi/Win32LogicalDisk.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2022 The OSHI Project Contributors
+ * Copyright 2020-2025 The OSHI Project Contributors
  * SPDX-License-Identifier: MIT
  */
 package oshi.driver.windows.wmi;
@@ -41,7 +41,7 @@ public final class Win32LogicalDisk {
         StringBuilder wmiClassName = new StringBuilder(WIN32_LOGICAL_DISK);
         boolean where = false;
         if (localOnly) {
-            wmiClassName.append(" WHERE DriveType = 3");
+            wmiClassName.append(" WHERE DriveType = 2 OR DriveType = 3 OR DriveType = 6");
             where = true;
         }
         if (nameToMatch != null) {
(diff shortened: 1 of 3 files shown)


oshi-core/src/main/java/oshi/driver/windows/wmi/Win32LogicalDisk.java
```diff
@@ -41,7 +41,7 @@ public static WmiResult<LogicalDiskProperty> queryLogicalDisk(String nameToMatch
         StringBuilder wmiClassName = new StringBuilder(WIN32_LOGICAL_DISK);
         boolean where = false;
         if (localOnly) {
-            wmiClassName.append(" WHERE DriveType != 4");
+            wmiClassName.append(" WHERE DriveType = 3");
```
dbwiddis: I get the intent to exclude CDROM (5) but not sure we want to narrowly restrict to just 3.  Seems 2 (USB/thumb) and 6 (RAM) should be acceptable here?  How about we just add CDROM to the disallowed options?
```suggestion
            wmiClassName.append(" WHERE DriveType != 4 AND DriveType != 5");
```
or add alternatives
```suggestion
            wmiClassName.append(" WHERE DriveType = 2 OR DriveType = 3 OR DriveType = 6");
```

dbwiddis (commented):
Since we're changing user-facing behavior, this definitely needs a change log.

dbwiddis:
Actually, digging into this code, I don't see where OSHI ever calls this code with the `localOnly` parameter set to true. Is this one you're calling directly?

oshi-core/src/main/java/oshi/driver/windows/wmi/Win32LogicalDisk.java
cesnekmichal: Yes, the implementation can simply be:
`wmiClassName.append(" WHERE DriveType = 2 OR DriveType = 3 OR DriveType = 6");`
The main thing is to exclude the CD-ROM, which listing it can take several minutes.

PR #2939, opened 2025-07-28
Add preferred resolution to EDID Util

ayonization:
Added preferred resolution to the EDIDUtil class

### DIFF
--- a/oshi-core/src/main/java/oshi/util/EdidUtil.java
+++ b/oshi-core/src/main/java/oshi/util/EdidUtil.java
@@ -203,6 +203,20 @@ public final class EdidUtil {
     }
 
     /**
+     * Get the preferred resolution for the monitor (Eg: 1920x1080)
+     *
+     * @param edid The edid Byte array
+     * @return Plain text preferred resolution
+     */
+
+    public static String getPreferredResolution(byte[] edid) {
+        int dtd = 54;
+        int horizontalRes = (edid[dtd + 4] & 0xF0) << 4 | edid[dtd + 2] & 0xFF;
+        int verticalRes = (edid[dtd + 7] & 0xF0) << 4 | edid[dtd + 5] & 0xFF;
+        return horizontalRes + "x" + verticalRes;
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


oshi-core/src/main/java/oshi/util/EdidUtil.java
```diff
@@ -201,6 +201,19 @@ public static String getDescriptorRangeLimits(byte[] desc) {
...
+     * Get the preferred resolution for the monitor (Eg: 1920x1080)
+     *
+     * @param edid The edid Byte array
+     * @return Plain text preferred resolution
+     */
+
+    public static String getPreferredResolution(byte[] edid) {
+        int dtd = 54;
+        int horizontalRes = (edid[dtd + 4] & 240) << 4 | edid[dtd + 2] & 255;
+        int verticalRes = (edid[dtd + 7] & 240) << 4 | edid[dtd + 5] & 255;
```
dbwiddis: Can we change the 255 bitmasking to `& 0xFF` (or `0xff`) for consistency with other usages in this file?  And `0xF0` for 240?

dbwiddis (commented):
Thanks for the PR!

One request below, and can you also add a CHANGELOG entry?

oshi-core/src/main/java/oshi/util/EdidUtil.java
dbwiddis: (It might be more clear of the intent to do the `<< 4` prior to masking with `0xF00` as well.)

ayonization: Sure! Also do you need me to add unit tests? Looks like thats been flagged by the codecov plugin.

dbwiddis: Yes, we need tests. Hopefully you can grab a real EDID to get the expected output.

dbwiddis:
Failing checkstyle. Can you run `./mvnw spotless:apply` to correct?  I tried but it won't let me push to your master branch :)

ayonization:
@dbwiddis  Done!

PR #2959, opened 2025-08-28
Windows FFM support

rohan-coder02:
### Title

Add initial Windows FFM support 

### Description

This PR introduces the start for Windows support in the new FFM-backed implementation of OSHI.

Added *WindowsOperatingSystemFFM* entrypoint under **oshi.software.os.windows**.

Introduced low-level FFM bindings for:

- Kernel32FFM
- Advapi32FFM
- WinNTFFM
- WinFFM (common linker/symbol lookup utilities).

Implemented privilege enabling (enableDebugPrivilege) with FFM equivalents of OpenProcessToken, LookupPrivilegeValue, and AdjustTokenPrivileges.

Verified initial APIs against native output (PowerShell/CMD).

#### Implemented in the recent commit:

- getProcessId() via FFM (Kernel32.GetCurrentProcessId).
- getSystemUptime() via FFM (Kernel32.GetTickCount64 / GetTickCount).

#### In progress:

- Add more APIs 
- Extend to cover additional Windows OS-level functions with FFM.


oshi-core-java25/src/main/java/oshi/ffm/windows/Advapi32FFM.java
```diff
@@ -0,0 +1,45 @@
...
+import java.lang.foreign.SymbolLookup;
+import java.lang.invoke.MethodHandle;
+
+import static java.lang.foreign.ValueLayout.ADDRESS;
+import static java.lang.foreign.ValueLayout.JAVA_INT;
+
+public final class Advapi32FFM {
+
+    private static final SymbolLookup ADV = WinFFM.lib("Advapi32");
+
```
dbwiddis, line 18: I notice you've listed the library downcall methods together and then all the implementations together.  

In my implementation I paired them up, with a comment above each pair showing the native signature.   I think I'd prefer my approach because it's easier to look and see that the parameters line up (I had actually switched two parameters in one of my early PRs, that was fun to debug!).    Also easier to add new methods in pairs, or organize methods by similarity/logic, etc.

oshi-core-java25/src/main/java/oshi/ffm/windows/Advapi32FFM.java
```diff
@@ -0,0 +1,45 @@
...
+    private static final MethodHandle LookupPrivilegeValueW = WinFFM.downcall(ADV, "LookupPrivilegeValueW",
+            FunctionDescriptor.of(JAVA_INT, ADDRESS, ADDRESS, ADDRESS));
+
+    private static final MethodHandle AdjustTokenPrivileges = WinFFM.downcall(ADV, "AdjustTokenPrivileges",
+            FunctionDescriptor.of(JAVA_INT, ADDRESS, JAVA_INT, ADDRESS, JAVA_INT, ADDRESS, ADDRESS));
+
+    public static boolean openProcessToken(MemorySegment process, int desiredAccess, MemorySegment hTokenOut)
+            throws Throwable {
+        int ok = (int) OpenProcessToken.invokeExact(process, desiredAccess, hTokenOut);
+        return ok != 0;
```
dbwiddis:
```suggestion
        return isSuccess((int) OpenProcessToken.invokeExact(process, desiredAccess, hTokenOut));
```

And add this helper method to `WinFFM` (and make it a superclass so you can call it without the WinFFM prefix).
```java
    /**
     * Converts Windows BOOL return value to Java boolean.
     * In Windows APIs, 0 = FALSE (failure), non-zero = TRUE (success)
     *
     * @param winBool Windows BOOL value (4 byte JAVA_INT)
     * @return true if successful (non-zero), false if failed (zero)
     */
    public static boolean isSuccess(int winBool) {
        return winBool != 0;
    }
```

oshi-core-java25/src/main/java/oshi/ffm/windows/Advapi32FFM.java
```diff
@@ -0,0 +1,45 @@
...
+    private static final MethodHandle OpenProcessToken = WinFFM.downcall(ADV, "OpenProcessToken",
+            FunctionDescriptor.of(JAVA_INT, ADDRESS, JAVA_INT, ADDRESS));
+
+    private static final MethodHandle LookupPrivilegeValueW = WinFFM.downcall(ADV, "LookupPrivilegeValueW",
+            FunctionDescriptor.of(JAVA_INT, ADDRESS, ADDRESS, ADDRESS));
+
+    private static final MethodHandle AdjustTokenPrivileges = WinFFM.downcall(ADV, "AdjustTokenPrivileges",
+            FunctionDescriptor.of(JAVA_INT, ADDRESS, JAVA_INT, ADDRESS, JAVA_INT, ADDRESS, ADDRESS));
+
+    public static boolean openProcessToken(MemorySegment process, int desiredAccess, MemorySegment hTokenOut)
```
dbwiddis: I know this follows Java conventions, but in this project we've kept a convention of using the same casing as the native method, including leading capitalization, underscores, etc.

oshi-core-java25/src/main/java/oshi/ffm/windows/Advapi32FFM.java
```diff
@@ -0,0 +1,45 @@
...
+            throws Throwable {
+        int ok = (int) OpenProcessToken.invokeExact(process, desiredAccess, hTokenOut);
+        return ok != 0;
+    }
+
+    public static boolean lookupPrivilegeValue(String name, MemorySegment luid, Arena arena) throws Throwable {
+        MemorySegment nameSeg = WinFFM.utf16(arena, name);
+        int ok = (int) LookupPrivilegeValueW.invokeExact(MemorySegment.NULL, nameSeg, luid);
+        return ok != 0;
```
dbwiddis:
```suggestion
        return isSuccess((int) LookupPrivilegeValueW.invokeExact(MemorySegment.NULL, nameSeg, luid));
```

oshi-core-java25/src/main/java/oshi/ffm/windows/Advapi32FFM.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public static boolean adjustTokenPrivileges(MemorySegment hToken, MemorySegment tkp) throws Throwable {
+        int ok = (int) AdjustTokenPrivileges.invokeExact(hToken, 0, tkp, 0, MemorySegment.NULL, MemorySegment.NULL);
+        return ok != 0;
```
dbwiddis:
```suggestion
        return isSuccess((int) AdjustTokenPrivileges.invokeExact(hToken, 0, tkp, 0, MemorySegment.NULL, MemorySegment.NULL));
```

oshi-core-java25/src/main/java/oshi/ffm/windows/Advapi32FFM.java
```diff
@@ -0,0 +1,45 @@
...
+import java.lang.foreign.Arena;
+import java.lang.foreign.FunctionDescriptor;
+import java.lang.foreign.MemorySegment;
+import java.lang.foreign.SymbolLookup;
+import java.lang.invoke.MethodHandle;
+
+import static java.lang.foreign.ValueLayout.ADDRESS;
+import static java.lang.foreign.ValueLayout.JAVA_INT;
+
+public final class Advapi32FFM {
```
dbwiddis: Make `WinFFM` an abstract class so you can use its utils here.
```suggestion
public final class Advapi32FFM extends WinFFM {
```

oshi-core-java25/src/main/java/oshi/ffm/windows/Kernel32FFM.java
```diff
@@ -0,0 +1,88 @@
...
+import java.lang.invoke.MethodHandle;
+
+import static java.lang.foreign.ValueLayout.ADDRESS;
+import static java.lang.foreign.ValueLayout.JAVA_INT;
+import static java.lang.foreign.ValueLayout.JAVA_LONG;
+
+public final class Kernel32FFM {
+
+    private static final SymbolLookup K32 = WinFFM.lib("Kernel32");
+
```
dbwiddis: Similar comments for this class of pairing methods with their invocations and matching native function casing.

oshi-core-java25/src/main/java/oshi/ffm/windows/Kernel32FFM.java
```diff
@@ -0,0 +1,88 @@
...
+            throw new RuntimeException("Kernel32FFM.getCurrentProcess failed", t);
+        }
+    }
+
+    public static int getCurrentProcessId() {
+        try {
+            return (int) GetCurrentProcessId.invokeExact();
+        } catch (Throwable t) {
+            throw new RuntimeException("Kernel32FFM.getCurrentProcessId failed", t);
+        }
```
dbwiddis: For most of the implementations we are just throwing Throwable and forcing the caller to handle the exception.  Otherwise they now have to deal with (unchecked) runtime exceptions that we don't even give them a signal they need to handle.

oshi-core-java25/src/main/java/oshi/ffm/windows/Kernel32FFM.java
```diff
@@ -0,0 +1,88 @@
...
+        try {
+            return (int) GetLastError.invokeExact();
+        } catch (Throwable t) {
+            return -1;
+        }
+    }
+
+    public static long querySystemUptime(boolean isVistaOrGreater) {
+        try {
+            if (isVistaOrGreater) {
```
dbwiddis: I don't like putting this logic in the mapping class.  Just have separate `GetTickCount` and `GetTickCount64` methods and let the user figure out which one to call.

If you want a util function like this, put it in `WinFFM` class and define a constant in it.
```
private static final boolean IS_VISTA_OR_GREATER = VersionHelpers.IsWindowsVistaOrGreater();
```

oshi-core-java25/src/main/java/oshi/ffm/windows/WinFFM.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Utility class for working with the Foreign Function & Memory API (Java 24+). Provides helpers for library lookup,
+ * downcalls, and UTF-16 string conversion.
+ */
+public final class WinFFM {
+
+    private static final Linker LINKER = Linker.nativeLinker();
+
+    private WinFFM() {
+    }
```
dbwiddis: Linker is already defined in `ForeignFunctions` superclass with some general util functions.  You can put this in the middle.  Or, I don't see anything in here that's Windows specific except the utf16 method, which can stay if it's not the only method or be elevated as a sort-of-general-but-windows-only exception.  TLDR: add most (or all?) of these methods to `ForeignFunctions`.

Consider renaming to WindowsForeignFunctions also.
```suggestion
public abstract class WinFFM extends ForeignFunctions {

    protected WinFFM() {
    }
```

oshi-core-java25/src/main/java/oshi/ffm/windows/WinFFM.java
```diff
@@ -0,0 +1,54 @@
...
+     * Allocate a null-terminated UTF-16 string in the given arena.
+     */
+    public static MemorySegment utf16(Arena arena, String s) {
+        char[] chars = (s + "\0").toCharArray();
+        MemorySegment seg = arena.allocate(JAVA_CHAR, chars.length);
+        for (int i = 0; i < chars.length; i++) {
+            seg.setAtIndex(JAVA_CHAR, i, chars[i]);
+        }
+        return seg;
+    }
```
dbwiddis: Use the more efficient built-in method here.  Some additional thoughts:
 - While null termination is probably needed most of the time, might be good to have an overload that allows a boolean `appendNull` parameter (defaulting to true if left off) for the (rare) case we don't need to add a null.  This can wait until such a need arises.
 - I think the name `allocateWideString()` or `allocateLPWSTR()` would be better
 - Most methods put the `arena` parameter as the last one, let's try to stick with that pattern

```suggestion
    public static MemorySegment utf16(Arena arena, String s) {
        return arena.allocateFrom(s + "\0", StandardCharsets.UTF_16LE);
    }
```

oshi-core-java25/src/main/java/oshi/ffm/windows/WinNTFFM.java
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: MIT
+ */
+package oshi.ffm.windows;
+
+import java.lang.foreign.GroupLayout;
+import java.lang.foreign.MemoryLayout;
+
+import static java.lang.foreign.ValueLayout.JAVA_INT;
+
+public final class WinNTFFM {
```
dbwiddis: I've followed the pattern of putting constants and structs in a `WinNT` interface (to get automatic `public static final`.  Since I think `WinNT` is just a header file this seems to bmatch.

Open to changing that pattern.  No need to do so now, we can refactor later. Interested in your thoughts.

oshi-core-java25/src/main/java/oshi/ffm/windows/WinNTFFM.java
```diff
@@ -0,0 +1,41 @@
...
+    // }
+    public static final GroupLayout LUID_AND_ATTRIBUTES = MemoryLayout.structLayout(LUID.withName("Luid"),
+            JAVA_INT.withName("Attributes"));
+
+    // typedef struct _TOKEN_PRIVILEGES {
+    // DWORD PrivilegeCount;
+    // LUID_AND_ATTRIBUTES Privileges[ANYSIZE_ARRAY];
+    // }
+    // Here we define a layout with 1 privilege entry
+    public static final GroupLayout TOKEN_PRIVILEGES_1 = MemoryLayout.structLayout(JAVA_INT.withName("PrivilegeCount"),
```
dbwiddis: Don't think we need the _1 here
```suggestion
    public static final GroupLayout TOKEN_PRIVILEGES = MemoryLayout.structLayout(JAVA_INT.withName("PrivilegeCount"),
```

oshi-core-java25/src/main/java/oshi/software/os/windows/WindowsOperatingSystemFFM.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(WindowsOperatingSystemFFM.class);
+
+    static {
+        enableDebugPrivilege();
+    }
+
+    private static boolean enableDebugPrivilege() {
+        try (Arena arena = Arena.ofConfined()) {
+            MemorySegment hTokenOut = arena.allocate(ADDRESS);
```
dbwiddis: Recommend naming this `hTokenPtr` to make clear this is a pointer to the HTOKEN (another pointer).  In JNA-land it's a PointerByReference. That's not obvious without those constructs so we need to use variable naming to make it more clear.

If this becomes a common pattern might want a `allocateHTokenPtr(arena)` helper method in a superclass.  Same for other handles.  Or maybe we just need a `allocatePointerByReference(arena)` method in general?  I'm sure this will change over time as we figure out what we need.

oshi-core-java25/src/main/java/oshi/software/os/windows/WindowsOperatingSystemFFM.java
```diff
@@ -0,0 +1,95 @@
...
+
+            MemorySegment hProcess = Kernel32FFM.getCurrentProcess();
+            boolean success = Advapi32FFM.openProcessToken(hProcess,
+                    WinNTFFM.TOKEN_QUERY | WinNTFFM.TOKEN_ADJUST_PRIVILEGES, hTokenOut);
+            if (!success) {
```
dbwiddis:
```suggestion
            if (!Advapi32FFM.openProcessToken(hProcess,
                    WinNTFFM.TOKEN_QUERY | WinNTFFM.TOKEN_ADJUST_PRIVILEGES, hTokenOut)) {
```

oshi-core-java25/src/main/java/oshi/software/os/windows/WindowsOperatingSystemFFM.java
```diff
@@ -0,0 +1,95 @@
...
+                int err = Kernel32FFM.getLastError();
+                LOG.error("OpenProcessToken failed, error: " + err);
```
dbwiddis:
```suggestion
                LOG.error("OpenProcessToken failed, error: " + GetLastError());
```
(with a static import)

oshi-core-java25/src/main/java/oshi/software/os/windows/WindowsOperatingSystemFFM.java
```diff
@@ -0,0 +1,95 @@
...
+                return false;
+            }
+            MemorySegment hToken = hTokenOut.get(ADDRESS, 0);
+
+            MemorySegment luid = arena.allocate(WinNTFFM.LUID);
+            success = Advapi32FFM.lookupPrivilegeValue("SeDebugPrivilege", luid, arena);
+            if (!success) {
+                int err = Kernel32FFM.getLastError();
+                LOG.error("LookupPrivilegeValue failed, error: " + err);
+                Kernel32FFM.closeHandle(hToken);
```
dbwiddis: Consider putting these all in a `finally` block so you only have to do it once.  (You may need to pre-declare it as null and do a null check before closing.)

oshi-core-java25/src/main/java/oshi/software/os/windows/WindowsOperatingSystemFFM.java
```diff
@@ -0,0 +1,95 @@
...
+                return false;
+            }
+
+            MemorySegment tkp = arena.allocate(WinNTFFM.TOKEN_PRIVILEGES_1);
+            tkp.set(JAVA_INT, 0, 1);
+            tkp.asSlice(4, WinNTFFM.LUID.byteSize()).copyFrom(luid);
+            tkp.setAtIndex(JAVA_INT, 3, WinNTFFM.SE_PRIVILEGE_ENABLED);
```
dbwiddis: Magic Numbers Everywhere!    

Use `PathElement` for this.  See IOKit class for examples.

`setAtIndex()` is evil unless you're dealing with an array.  No one wants to do structure byte math.

oshi-core-java25/src/main/java/oshi/software/os/windows/WindowsOperatingSystemFFM.java
```diff
@@ -0,0 +1,95 @@
...
+    public int getProcessId() {
+        return Kernel32FFM.getCurrentProcessId();
+    }
+
+    @Override
+    public long getSystemUptime() {
+        return Kernel32FFM.querySystemUptime(isVistaOrGreater());
+    }
+
+    public static boolean isVistaOrGreater() {
```
dbwiddis: This is already a public constant elsewhere, natively determined from Windows API . Don't reinvent the wheel. :)

oshi-core-java25/src/main/java/oshi/ffm/windows/Kernel32FFM.java
rohan-coder02: I see the below approach, let me know your view

log the exception and return Optional like below
```
try {
        return OptionalInt.of(kernel32.getCurrentProcessId());
    } catch (Exception e) {
        logger.debug("Process ID unavailable: {}", e.getMessage());
        return OptionalInt.empty();
    }
```

dbwiddis: Yes, it's fine to catch an exception and return something: in the above case we're returning an Optional indicating whether we got something or not.

The main point is to not throw unchecked exceptions.  Either throw a checked exception (or throwable) so the caller knows to handle it, which might generally be the case when a method might be called in a variety of circumstances, , or catch and handle it, appropriate for this case when there's really only one expected value that should in theory never fail.   We are still sending a signal to the caller of "hey, check that this value exists".

oshi-core-java25/src/main/java/oshi/software/os/windows/WindowsOperatingSystemFFM.java
rohan-coder02: I'll be adding the below function in WinNTFFM class.
```
/**
     * Builds a TOKEN_PRIVILEGES struct with a single privilege enabled.
     *
     */
    public static MemorySegment setupTokenPrivileges(Arena arena, MemorySegment luid) {

        MemorySegment tkp = arena.allocate(TOKEN_PRIVILEGES);

        // Set privilege count
        long privilegeCountOffset = TOKEN_PRIVILEGES.byteOffset(MemoryLayout.PathElement.groupElement("PrivilegeCount"));
        tkp.set(JAVA_INT, privilegeCountOffset, 1);

        // Copy LUID
        long luidOffset = TOKEN_PRIVILEGES.byteOffset(
            MemoryLayout.PathElement.groupElement("Privileges"),
            MemoryLayout.PathElement.groupElement("Luid")
        );

        MemorySegment luidSegment = tkp.asSlice(luidOffset, luid.byteSize());
        luidSegment.copyFrom(luid);

        // Set attributes
        long attributesOffset = TOKEN_PRIVILEGES.byteOffset(
            MemoryLayout.PathElement.groupElement("Privileges"),
            MemoryLayout.PathElement.groupElement("Attributes")
        );
        tkp.set(JAVA_INT, attributesOffset, SE_PRIVILEGE_ENABLED);

        return tkp;
    }
```

calling this in WindowsOperatingSystemFFM class as below.
```
MemorySegment tkp = WinNTFFM.setupTokenPrivileges(arena, luid);
```

oshi-core-java25/src/main/java/oshi/ffm/windows/WinNTFFM.java
rohan-coder02: + WinNT as an interface works if there are only constants but further use-cases may involve having helper function using the constants then a class would be more appropriate
+ Like adding `setupTokenPrivileges` function that uses constants -> https://github.com/oshi/oshi/pull/2959#discussion_r2312510920

oshi-core-java25/src/main/java/oshi/software/os/windows/WindowsOperatingSystemFFM.java
dbwiddis: I'm not convinced that needs to be outside of the one class where it's used; it's mostly an initialization step.

And if you do add it, as a function/method it fits better in the superclass with the other util methods.

You can refer to the (interface) constants with static imports and it looks identical in your code.

Think native C:  the constants and structs are in the header (`.h`) files while functions are in the `.c` files (which import appropriate `.h` files).  I'm somewhat replicating that setup here (I had originally used the "Headers" and "Structs" names and then combined them).

dbwiddis: Also see my struct layouts where I defined the PathElements in the interface (header) files as constants: so your byte offsets can just be a single constant in the static method.

oshi-core-java25/src/main/java/oshi/ffm/ForeignFunctions.java
```diff
@@ -44,4 +46,19 @@ public static byte[] getByteArrayFromNativePointer(MemorySegment pointer, long l
...
+     */
+    public static SymbolLookup lib(String name) {
+        return SymbolLookup.libraryLookup(name, Arena.global());
+    }
+
+    /**
+     * Create a downcall handle for a symbol in a library.
+     */
+    public static MethodHandle downcall(SymbolLookup lib, String symbol, FunctionDescriptor fd) {
+        MemorySegment sym = lib.find(symbol).orElseThrow(() -> new UnsatisfiedLinkError("Missing symbol: " + symbol));
```
dbwiddis: You can simplify by using the existing:
```suggestion
        MemorySegment sym = lib.findOrThrow(symbol);
```
This throws a `NoSuchElementException` which is more appropriate than `UnsatisfiedLinkError`.  While an unchecked exception, we only really call the `invokeExact()` wrapper from our own classes, which throws a `Throwable` (so even the unchecked exception has to be handled downstream).

oshi-core-java25/src/main/java/oshi/ffm/windows/Advapi32FFM.java
dbwiddis: Thanks for pairing up the methods.  I think it's still useful to document the native signature.  In particular the return value for `OpenProcessToken` is `BOOL` but here we see `JAVA_INT` so there's no obvious signal to anyone using this method that they need to treat the resulting integer as success/failure.  For now a simple `//` comment will probably do.  But it really really really helps review so I don't have to manually look up each function. :)

oshi-core-java25/src/main/java/oshi/ffm/ForeignFunctions.java
```diff
@@ -44,4 +46,19 @@ public static byte[] getByteArrayFromNativePointer(MemorySegment pointer, long l
...
+     * Lookup a library by name in the global arena.
+     */
+    public static SymbolLookup lib(String name) {
+        return SymbolLookup.libraryLookup(name, Arena.global());
+    }
+
+    /**
+     * Create a downcall handle for a symbol in a library.
+     */
+    public static MethodHandle downcall(SymbolLookup lib, String symbol, FunctionDescriptor fd) {
```
dbwiddis: I like this helper method. I'll update my mac methods to use it!

As long as we're doing a helper method, I like the idea of replacing the `FunctionDescriptor` argument by separating the return value and argument values.
```suggestion
    public static MethodHandle downcall(SymbolLookup lib, String symbol, MemoryLayout resLayout, List<MemoryLayout> argLayouts) {
```

You can then pass these directly to `FunctionDescriptor.ofVoid()` (if `resLayout` is null) or `FunctionDescriptor.of()` otherwise, and it's more clear in the implementing code.

oshi-core-java25/src/main/java/oshi/ffm/windows/Kernel32FFM.java
```diff
@@ -0,0 +1,100 @@
...
+            return OptionalInt.of((int) GetLastError.invokeExact());
+        } catch (Throwable t) {
+            LOG.debug("Kernel32FFM.GetLastError failed: {}", t.getMessage());
+            return OptionalInt.empty();
+        }
+    }
+
+    private static final MethodHandle GetTickCount = downcall(K32, "GetTickCount", FunctionDescriptor.of(JAVA_INT) // 32-bit
+    // return
+    );
```
dbwiddis: I'm on the fence as to whether we should support any 32-bit options since JDK 25 is 64-bit only.  I've already made a 64-bit only assumption on macOS; this would up our minimum supported version in Windows but I'm kind of OK with not allowing the WindowsXP + JDK25 combination.  :)

oshi-core-java25/src/main/java/oshi/ffm/windows/WinNTFFM.java
```diff
@@ -0,0 +1,32 @@
...
+
+import static java.lang.foreign.ValueLayout.JAVA_INT;
+
+public interface WinNTFFM {
+
+    int SE_PRIVILEGE_ENABLED = 0x00000002;
+    int TOKEN_QUERY = 0x0008;
+    int TOKEN_ADJUST_PRIVILEGES = 0x0020;
+
+    GroupLayout LUID = MemoryLayout.structLayout(JAVA_INT.withName("LowPart"), JAVA_INT.withName("HighPart"));
```
dbwiddis: All of these can probably use the subclass `StructLayout`.  Not that important but why not use the more specific type?

oshi-core-java25/src/main/java/oshi/ffm/windows/WinNTFFM.java
```diff
@@ -0,0 +1,32 @@
...
+    GroupLayout TOKEN_PRIVILEGES = MemoryLayout.structLayout(JAVA_INT.withName("PrivilegeCount"),
+            LUID_AND_ATTRIBUTES.withName("Privileges"));
+
+    long OFFSET_PRIVILEGE_COUNT = TOKEN_PRIVILEGES.byteOffset(MemoryLayout.PathElement.groupElement("PrivilegeCount"));
+
+    long OFFSET_LUID = TOKEN_PRIVILEGES.byteOffset(MemoryLayout.PathElement.groupElement("Privileges"),
+            MemoryLayout.PathElement.groupElement("Luid"));
+
+    long OFFSET_ATTRIBUTES = TOKEN_PRIVILEGES.byteOffset(MemoryLayout.PathElement.groupElement("Privileges"),
+            MemoryLayout.PathElement.groupElement("Attributes"));
```
dbwiddis: A few thoughts.

When I did the mac code, I just put the `PathElements` here as a constant and called `byteOffset` in that code.  This looks generally like an improvement, but I'm worried about the case where we have competing names of nested fields in multiple structures.  I'm not sure we won't see "Attributes" somewhere else.  So if we go this route I'd want a more verbose name (and I'd prefer OFFSET at the end) such as `PRIVELEGES_LUID_OFFSET` and `PRIVILEGES_ATTRIBUTES_OFFSET`).

WDYT?

oshi-core-java25/src/main/java/oshi/ffm/windows/WindowsForeignFunctions.java
```diff
@@ -0,0 +1,79 @@
...
+
+        tkp.set(JAVA_INT, OFFSET_ATTRIBUTES, SE_PRIVILEGE_ENABLED);
+
+        return tkp;
+    }
+
+    /**
+     * Allocate a null-terminated UTF-16 string in the given arena.
+     */
+    public static MemorySegment utf16(Arena arena, String s) {
```
dbwiddis: still prefer a method name using "wide string" or "LPWSTR".

oshi-core-java25/src/main/java/oshi/ffm/windows/WindowsForeignFunctions.java
```diff
@@ -0,0 +1,79 @@
...
+     * (success)
+     */
+    public static boolean isSuccess(int winBool) {
+        return winBool != 0;
+    }
+
+    /**
+     * Checks if the current OS is Windows Vista or greater.
+     */
+    public static boolean isVistaOrGreater() {
```
dbwiddis: I think you can delete this as you're using the constant elsewhere.

rohan-coder02: My approach was

+ Since we're working on the FFM replacement and moving away from JNA dependencies, I implemented a non-JNA version of `isVistaOrGreater()` so we’re not tied to JNA
+ Functionally, this is equivalent to what VersionHelpers does internally (string + major version check), just without JNA
+ `JNA` -> https://github.com/java-native-access/jna/blob/master/contrib/platform/src/com/sun/jna/platform/win32/VersionHelpers.java#L58
+ `oshi-core` -> https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/software/os/windows/WindowsOSProcess.java#L32
```
import com.sun.jna.platform.win32.VersionHelpers;
...
private static final boolean IS_VISTA_OR_GREATER = VersionHelpers.IsWindowsVistaOrGreater();
```

Let me know your views

dbwiddis: It's currently JNA but we'll replace that class with FFM eventually.

dbwiddis (approved):
LGTM!  Merging this so you can iterate on the next one!

PR #2969, opened 2025-09-13
Small Code Improvements

khmarbaise:
- Added JavaDoc's
 - Fixed some typos in comments.
 - Simplified RegEx.
 - Using Files instead of hand crafted reading.

### DIFF
--- a/oshi-core/src/main/java/oshi/driver/linux/proc/ProcessStat.java
+++ b/oshi-core/src/main/java/oshi/driver/linux/proc/ProcessStat.java
@@ -4,13 +4,6 @@
  */
 package oshi.driver.linux.proc;
 
-import static oshi.software.os.OSProcess.State.OTHER;
-import static oshi.software.os.OSProcess.State.RUNNING;
-import static oshi.software.os.OSProcess.State.SLEEPING;
-import static oshi.software.os.OSProcess.State.STOPPED;
-import static oshi.software.os.OSProcess.State.WAITING;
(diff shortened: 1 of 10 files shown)


oshi-core/src/main/java/oshi/driver/linux/proc/ProcessStat.java
```diff
@@ -30,13 +23,20 @@
...
+import static oshi.software.os.OSProcess.State.ZOMBIE;
+
 /**
  * Utility to read process statistics from {@code /proc/[pid]/stat}
  */
 @ThreadSafe
 public final class ProcessStat {
 
-    private static final Pattern SOCKET = Pattern.compile("socket:\\[(\\d+)\\]");
+    private static final Pattern SOCKET = Pattern.compile("socket:[(\\d+)]");
```
dbwiddis: I think this change is incorrect.

dbwiddis:
> Are those failings on Ubuntu in CI JDK 11/21/24 caused by my change?

Fixed the socket regex. The other failure looks to be caused by the FileUtil change, but not sure yet exactly how.

dbwiddis:
The FileUtil failure looks like it's using native random access to the file (`pread0()`) but proc filesystems don't support those low level calls.

oshi-core/src/main/java/oshi/driver/linux/proc/ProcessStat.java
khmarbaise: Dammed. Sorry for that... Thanks for fixing that...

PR #2975, opened 2025-09-19
Query Mac apps in xml formatting using internal xml parsing utils.

dyorgio:
This fix actual code that miss last app listed.
Its also load more info from internal app plist (xml or binary formats).
Performance is the same of current code (local tests shows only ~10% of speed gain, irrelevant), but old Xml pull request is much more slower (JDK xml parsing libs is the guilty).

Please @rohan-coder02, take a look :).

### DIFF
--- a/oshi-core/src/main/java/oshi/software/os/mac/MacInstalledApps.java
+++ b/oshi-core/src/main/java/oshi/software/os/mac/MacInstalledApps.java
@@ -8,9 +8,9 @@ import java.util.ArrayList;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.List;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


rohan-coder02 (commented):
1)

> Its also load more info from internal app plist

It prints installed apps output in my mac fine.

2)

> Performance is the same of current code

I tested your code in my local by creating a simple Java app.
```
long startMillis = System.currentTimeMillis();
SystemInfo systemInfo = new SystemInfo();
System.out.println(systemInfo.getOperatingSystem().getInstalledApplications());
long fetchEndMillis = System.currentTimeMillis();
System.out.println("Time to fetch: " + (fetchEndMillis - startMillis) + " ms");
```

The time taken to run using existing approach -> `Time to fetch: 3167 ms`

The time taken to run using approach in this PR -> `Time to fetch: 3379 ms`

The performance is around the same level.

dyorgio:
> It prints installed apps output in my mac fine.

Are the item count equals? In my test its bring one more.

dbwiddis:
Hey @dyorgio can you resolve the conflicts here?  

I was briefly considering using json for the mac parsing in preference to xml; different parsing headaches of course with quotation marks vs. more clearly defined xml syntax.

dyorgio:
Depends, json output breaks old macOS compatibility.

https://apple.stackexchange.com/questions/388377/in-which-macos-version-was-the-json-flag-switch-added-in-system-profiler

what minimum version of macOS that you would like to keep?

dbwiddis:
> what minimum version of macOS that you would like to keep?

Huh, didn't know it was that recent.  I'm happy with the XML parsing, though... my JSON experiments had too many escaped quotation marks for my liking without actually using a library for it.

dyorgio:
Captura de Tela 2025-10-29 às 16 07 37
Very strange, this option already is turned on

PR #2992, opened 2025-09-28
Allow using MacOperatingSystemFFM without loading JNA

wolfs:
Currently, since `MacOperatingSystem` is a super class of `MacOperatingSystemFFM`, instantiating `MacOperatingSystemFFM` causes the initialization of JNA. That also caused loading the `System` library in the classloader, hiding some problems when using poor JNA.

In this PR, I introduce `AbstractMacOperatingSystem` that doesn't reference any JNA classes. The `MacOperatingSystemFFM` extends this. Moreover, I add code to load the `System` library via ffm methods in `MacSystemFunctions`.


oshi-core-java25/src/main/java/oshi/ffm/mac/MacSystemFunctions.java
```diff
@@ -23,18 +24,20 @@ public final class MacSystemFunctions extends ForeignFunctions {
     private MacSystemFunctions() {
     }
 
+    private static final SymbolLookup SYSTEMB = libraryLookup("System");
```
wolfs, line 27: 💭 I am not sure if there is some more lifecycle management necessary for loading a library with ffm, though the tests seem to work with that.

oshi-core-java25/src/main/java/oshi/ffm/ForeignFunctions.java
```diff
@@ -18,11 +18,16 @@
 public abstract class ForeignFunctions {
 
     protected static final Linker LINKER = Linker.nativeLinker();
+    protected static final Arena LIBRARY_ARENA = Arena.ofShared();
```
wolfs, line 21: 💭 Same here, I hope having a shared arena as a static field will be enough of lifecycle management with ffm.

dbwiddis:
Hey @wolfs thanks for helping out here.  Some quick comments/questions:
 - inheriting from `MacOperatingSystem` was intended as temporary during migration, with the intention to eventually remove it.  
 - I'm not sure what problems loading JNA's System library was causing. Can you be more specific?  It's still likely going to be loaded by other classes in OSHI until all JNA usage is removed.
 - I don't think you need to explicitly load the System Library (or any of the core frameworks) in FFM. They're already included [since macOS Big Sur](https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-release-notes).  The existing MacOperatingSystemFFM code works just fine without specifying libraries.

oshi-core/src/main/java/oshi/software/os/mac/AbstractMacOperatingSystem.java
```diff
@@ -0,0 +1,179 @@
...
+import static oshi.software.os.OSService.State.RUNNING;
+import static oshi.software.os.OSService.State.STOPPED;
+import static oshi.util.Memoizer.installedAppsExpiration;
+
+/**
+ * macOS, previously Mac OS X and later OS X) is a series of proprietary graphical operating systems developed and
+ * marketed by Apple Inc. since 2001. It is the primary operating system for Apple's Mac computers.
+ */
+@ThreadSafe
+public abstract class AbstractMacOperatingSystem extends AbstractOperatingSystem {
```
dbwiddis: Wouldn't it be cleaner to keep the existing `MacOperatingSystem` inheritance and just extract any JNA-related things to a new `MacOperatingSystemJNA` subclass?

wolfs:
> I'm not sure what problems loading JNA's System library was causing. Can you be more specific? It's still likely going to be loaded by other classes in OSHI until all JNA usage is removed.

My goal in using the FFM variant is that I can avoid loading the JNA native library since that may be conflicting with other usages of JNA since my "application" is a Gradle plugin. So as soon as enough code paths are supported in the FFM code, I would be able to use that instead the JNA one on Java 25 and avoid a few of the problems I am seeing with JNA.

> I don't think you need to explicitly load the System Library (or any of the core frameworks) in FFM. They're already included [since macOS Big Sur](https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-release-notes). The existing MacOperatingSystemFFM code works just fine without specifying libraries.

The System libraries are included in macOS but not loaded in the JVM. That is why I need to load them explicitly. If you revert my second commit you'll see (or at least I saw) a failure with the symbols not being found. The JNA initialization also initializes JNA's `SYSTEMB` constant, that causes loading the `System` library on the classloader the "old" way. And that is why the default symbol lookup will then find the symbols.

oshi-core/src/main/java/oshi/software/os/mac/AbstractMacOperatingSystem.java
wolfs: Sounds good to me. The only change I see in my code would be that I'd rename `AbstractMacOperatingSystem -> MacOperatingSystem` and `MacOperatingSystem` -> `MacOperatingSystemJNA`, right?

dbwiddis: I haven't looked in detail at your changes yet but based on a quick glance that seems right.

My comment was based on the current (before this PR) status with `MacOperatingSystem` remaining a superclass, and just moving any *.jna.* classes into a subclass (and instantiating that subclass from SystemInfo).

It may be the same thing you have done.

dbwiddis:
> My goal in using the FFM variant is that I can avoid loading the JNA native library since that may be conflicting with other usages of JNA since my "application" is a Gradle plugin. So as soon as enough code paths are supported in the FFM code, I would be able to use that instead the JNA one on Java 25 and avoid a few of the problems I am seeing with JNA.

That's the eventual goal; however in the longer term we'd actually exclude all the JNA dependencies from the `build.gradle` (or `pom.xml`) file completely.  

Currently during the transition, we're somewhat hacking the build using the build-heper plugin to just add the old `oshi-core` classes to the buildpath/classpath.  It's somewhat of a "sledge hammer" approach (all or nothing) but it's simple.  And since not many people have (yet) been contributing to the migration, I was lazy.

Later (when we don't need to support JDK 8) we can either completely separate the modules (ideal) or use other build hacks to copy files rather than add a file path.

I think migrating JNA code to JNA subclasses may help in this eventual end goal, but please somewhat keep that in mind.  

> The System libraries are included in macOS but not loaded in the JVM. That is why I need to load them explicitly. If you revert my second commit you'll see (or at least I saw) a failure with the symbols not being found.

OK, digging into the details, it does seem that the existing `SymbolLookup.loaderLookup()` loads libraries that were previously loaded using `System.load()` which is why removing the JNA-based load caused it to fail.  

However, I think we should put [`Linker.defaultLookup()`](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/lang/foreign/Linker.html#defaultLookup()) as the definition of `SYMBOL_LOOKUP` as I think it will work for most cases.

oshi-core-java25/src/main/java/oshi/ffm/ForeignFunctions.java
dbwiddis: Looking into this, it seems the best fit and similar to JNA (load once and forget).

oshi-core-java25/src/main/java/oshi/ffm/mac/MacSystemFunctions.java
dbwiddis: This is probably fine for static initialization like JNA has done.  Note there are some use cases where people want to unload libraries after use, so we might consider some sort of name-to-library map where we can load (if absent) and permit unloading, but no need to do that now.

Small request.  The reason it's named SystemB is because that's how I named it when first contributing it to JNA.  `System` is an existing class so I couldn't use it, and at the time in the frameworks library there was some sort of B subdirectory... but since then I've seen it in A subdirectories as well.   Can we just name this SYSTEM_LIB or something more appropriate?

oshi-core-java25/src/main/java/oshi/ffm/ForeignFunctions.java
wolfs: I wonder now if `ofAuto()` would be the right thing to do. The `ofShared()` needs to be closed manually, while the `ofAuto()` works in tandem with Garbage collection. See [here](https://docs.oracle.com/en/java/javase/25/core/memory-segments-and-arenas.html).

dbwiddis: Yeah I looked into that, but we don't want this to be closed.  It would lose the library mappings.

wolfs: Well, if that class is being unloaded from the classloader, the library bindings will also be gone. So I think `ofAuto()` would be better.

dbwiddis: > I think `ofAuto()` would be better.

I don't think `ofAuto()` provides any added benefits over `ofShared()`.  
 - It would lose the loadings if the Arena was ever closed, although since you made it a static variable that would never happen
 - It's not thread safe, while `ofShared()` is. 
 - `ofShared()` lets us explicitly close if we want (even as a static variable) while `ofAuto()` relies on GC to (possibly) run which we'd have to make the arena unreachable (set the static variable null) to release the memory (at some unpredictable future time).

If we want to keep the load-once-and-forget behavior I think either `ofShared()` or `global()` works.

If we want to enable unloading I think we should have a whole new class (similar to JNA's `NativeLibrary` class) that uses a different `ofShared()` for each library and allows closing them.  OR.... use the same `System.load()` pattern as JNA and the original `loaderLookup()`.

wolfs: I don't understand.
> It would lose the loadings if the Arena was ever closed, although since you made it a static variable that would never happen
You cannot close the `ofAuto()` Arena manually, so as long as somebody references it, like the loaded library, it will not be closed.

> It's not thread safe, while ofShared() is.
It has the same thread safety conditions as `ofShared()`. The [Javadocs](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/lang/foreign/Arena.html) have a nice table:
image

> ofShared() lets us explicitly close if we want (even as a static variable) while ofAuto() relies on GC to (possibly) run which we'd have to make the arena unreachable (set the static variable null) to release the memory (at some unpredictable future time).

`ofShared()` would require us to call `close()` at some point, if not we'd have some memory leaks when the oshi is being used in a long running process (as I am doing it right now in the Gradle daemon when adding oshi to a plugin). I would rely on Garbage collection to clean up things as I do for everything else. I don't see an advantage of manually cleaning this up, if not we'd need to have something like JNA cleaner threads, and those are again a different kind of problem, see e.g. https://github.com/java-native-access/jna/issues/1521. Since `ofAuto()` takes care of that for us, I would not go into trying to implement some manual closing.

dbwiddis: > `ofShared()` would require us to call `close()` at some point, if not we'd have some memory leaks when the oshi is being used in a long running process

We're using the loaded libraries as static final variables in a class. 
```java
private static final SymbolLookup SYSTEMB = libraryLookup("System");
```
We currently have no mechanism for ever unloading those `SymbolLookup` instances, and allowing their memory to go away (by closing the `Arena` we used to look them up) will permanently break them.  Given this use case, we never want to close the `Arena` as long as the JVM is running. Given your table above, the only difference is explicitly closing it vs. allowing GC to close it, and in our current usage the explicit vs. automatic closing is irrelevant, as they will never (as static class variables) become unreachable.  In our current usage, `gloabal()` may just as well be a better choice.

> `ofShared()` would require us to call `close()` at some point

Precisely.  Which, if we do want to change the implementation to allow unloading the native libraries (like JNA currently allows) would permit us to immediately `close()` the Arena to release the memory allocations, rather than just waiting for GC to (maybe, eventually) release them.   

While not directly comparable, we previously ran into memory issues with our registry reading code (for Windows threads and processes) which auto-grew a buffer to the needed size. (see https://github.com/oshi/oshi/discussions/2073). Originally we just reassigned the buffer variable and allowed GC to (eventually) release the buffer, and these native allocations hung around a long time.  

When JNA made `Memory` closeable in 5.12.0, we even [added explicit `CloseableFooByRef` implementations](https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/jna/ByRef.java) for most of the JNA native allocations (paralleling `ofConfined()`) precisely to ensure prompt releasing of native memory. (And because JNA also needed a two-step GC + finalize/cleaner process.)

Anyway if we are going to implement any code to recycle the native library `SymbolLookup` instance, adding a line to `close()` the arena at the same time ensures that memory is immediately available and not subject to the whim of the garbage collector.

TLDR: there is no benefit in the "library loading" scenario, where we need to maintain reachability of the `Arena` until we're done with it, of using `ofAuto()` other than not having to call `close()`; but in the case we actually do want to unload the library to free the memory, we shouldn't wait for GC to (maybe, eventually) free up that memory.

wolfs: The main problem with JNI (and that is somewhat what JNA uses + code generation) is that Garbage collecting the Java objects does not free the native counterparts. Like you need to manually unload native libraries and free memory regions. FFM fixes that (at least that is my understanding)! And that is why `ofAuto()` should be the right thing to do and we don't need to do all the different things that JNA is doing.

> And because JNA also needed a two-step GC + finalize/cleaner process.
And my understanding is that this isn't necessary with FFM.

> Given your table above, the only difference is explicitly closing it vs. allowing GC to close it, and in our current usage the explicit vs. automatic closing is irrelevant, as they will never (as static class variables) become unreachable.

The difference is that (to my understanding) is, that `ofAuto()`will release the native memory when Garbage collected, while `ofShared()` will _not_.

In our current setup automatic closing is relevant, since the static class variables can become Garbage collected, when the class object is Garbage collected, aka when the classloader goes away. And classloaders in long running applications with plugin systems (like Gradle), need to make sure that the classes are actually unloaded. I agree that for simple applications that run for some time and then the JVM exits we don't need to do anything, and it doesn't really matter whether we use `ofShared()`, `global()`, or `ofAuto()`. Though if we want to allow some cleanup, and don't won't to go into the difficult area of implementing our own cleaner, we should rely on the Garbage collector IMO and use `ofAuto()`.

Would you be opposed to use `ofAuto()` for now until we see that causing any problems? All of this is so new that I don't know what the best practices are long term and I think we need to find out.

dbwiddis: > The main problem with JNI (and that is somewhat what JNA uses + code generation) is that Garbage collecting the Java objects does not free the native counterparts. Like you need to manually unload native libraries and free memory regions.

Correct, except that in "you" we're referring to the JNA library itself which handled this with finalizer (previously) and cleaner (now) implementations to do that, but those are still two-step processes as I referenced.

Compare to using `closeable()` on `Memory` allocations which allows us to explicitly release the native memory when we want to rather than waiting for (maybe) GC followed by (queued) Cleaner ops.

My point is: _why would you wait for GC to handle releasing memory when you can do it yourself_, in the case that you actually want to unload this library mapping? 

> In our current setup automatic closing is relevant, since the static class variables can become Garbage collected

No they can't.  They are `protected static final` class instances.  Per [JLS §12.7](https://docs.oracle.com/javase/specs/jls/se25/html/jls-12.html#jls-12.7):
> A class or interface may be unloaded if and only if its defining class loader may be reclaimed by the garbage collector as discussed in [§12.6](https://docs.oracle.com/javase/specs/jls/se25/html/jls-12.html#jls-12.6).
> 
> Classes and interfaces loaded by the bootstrap loader may not be unloaded.

And clarified below that:
> If the defining loader can be reclaimed, then there can never be any live references to it... This, in turn, can only be true if there are can never be any live references to any of the classes defined by that loader

This means that:
- A class's static fields remain reachable as long as the class itself is loaded
- The class remains loaded as long as its ClassLoader is reachable
- For standard application classes loaded by the system ClassLoader (Such as `SymbolLookup`), this means the entire application lifetime

I'll repeat: 
1. for our use case of never unloading `SymbolLookup` instances, they will _never_ be closed during the lifetime of the JVM.  So it doesn't matter which we use.
2. for the use case where we want to intentionally unload a `SymbolLookup` instance, it is a better practice to proactively free the memory rather than waiting for (never guaranteed to run) GC.

dbwiddis: > Would you be opposed to use ofAuto() for now until we see that causing any problems? All of this is so new that I don't know what the best practices are long term and I think we need to find out.

Actually after rereading, I'm fine with that, given your discussion of plugin classloaders.

PR #3000, opened 2025-10-05
Use ofAuto() instead of ofShared()

wolfs:
To allow unloading the native libraries when the classloaders are being Garbage collected, e.g. when using the library in a plugin when some plugin system is in place.

See the discussion [here](https://github.com/oshi/oshi/pull/2992#discussion_r2395780953).

### DIFF
--- a/oshi-core-java25/src/main/java/oshi/ffm/mac/MacSystemFunctions.java
+++ b/oshi-core-java25/src/main/java/oshi/ffm/mac/MacSystemFunctions.java
@@ -24,12 +24,12 @@ public final class MacSystemFunctions extends ForeignFunctions {
     private MacSystemFunctions() {
     }
 
-    private static final SymbolLookup SYSTEMB = libraryLookup("System");
+    private static final SymbolLookup SYSTEM_LIBRARY = libraryLookup("System");
 
     public static final ValueLayout.OfLong SIZE_T = ValueLayout.JAVA_LONG;
 
     // int proc_listpids(uint32_t type, uint32_t typeinfo, void *buffer, int buffersize)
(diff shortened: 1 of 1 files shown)


oshi-core-java25/src/main/java/oshi/ffm/ForeignFunctions.java
```diff
@@ -15,10 +13,12 @@
 import java.lang.foreign.SymbolLookup;
 import java.lang.invoke.MethodHandle;
 
+import static java.lang.foreign.ValueLayout.JAVA_BYTE;
+
 public abstract class ForeignFunctions {
 
     protected static final Linker LINKER = Linker.nativeLinker();
-    protected static final Arena LIBRARY_ARENA = Arena.ofShared();
+    protected static final Arena LIBRARY_ARENA = Arena.ofAuto();
```
dbwiddis, line 21: Approving this for now as in most use cases it won't matter but in the edge case of a plugin it may.

Curious if we can change this back to shared in the future in a way that allows us to explicitly close it before unloading.  JNA, for example, has a `disposeAll()` method for its libraries that a plugin could call.

wolfs: If we want to do that, then I suggest that we'd have the base types to be closable and not using static fields for this. Another option would be to have some system property that decides what kind of Arena to use.

dbwiddis: > If we want to do that, then I suggest that we'd have the base types to be closable and not using static fields for this.

Of course.  It warrants a full discussion of how we're doing this end to end.  I still don't think relying on GC to release native memory allocations is an ideal long term plan.

PR #3007, opened 2025-10-11
Add ffm implementation for hal.memory

wolfs:
This adds an implementation for hal.memory (and virtualmemory). I also added the errno handling that should work for POSIX systems instead of relying on the return code of the method.

### DIFF
--- a/oshi-core-java25/src/test/java/oshi/hardware/GlobalMemoryFFMTest.java
+++ b/oshi-core-java25/src/test/java/oshi/hardware/GlobalMemoryFFMTest.java
@@ -9,7 +9,7 @@ import org.junit.jupiter.api.condition.EnabledForJreRange;
 import org.junit.jupiter.api.condition.JRE;
 import oshi.SystemInfoFFM;
 
-@EnabledForJreRange(min = JRE.JAVA_24)
+@EnabledForJreRange(min = JRE.JAVA_25)
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 public class GlobalMemoryFFMTest extends GlobalMemoryTest {
 
--- a/oshi-core-java25/src/test/java/oshi/SystemInfoFFMTest.java
+++ b/oshi-core-java25/src/test/java/oshi/SystemInfoFFMTest.java
@@ -19,7 +19,7 @@ import static org.hamcrest.Matchers.not;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
-@EnabledForJreRange(min = JRE.JAVA_24)
+@EnabledForJreRange(min = JRE.JAVA_25)
 public class SystemInfoFFMTest extends SystemInfoTest {
 
     private static final Logger logger = LoggerFactory.getLogger(SystemInfoFFMTest.class);
--- a/oshi-core-java25/src/test/java/oshi/hardware/VirtualMemoryFFMTest.java
+++ b/oshi-core-java25/src/test/java/oshi/hardware/VirtualMemoryFFMTest.java
@@ -9,7 +9,7 @@ import org.junit.jupiter.api.condition.EnabledForJreRange;
 import org.junit.jupiter.api.condition.JRE;
 import oshi.SystemInfoFFM;
 
-@EnabledForJreRange(min = JRE.JAVA_24)
+@EnabledForJreRange(min = JRE.JAVA_25)
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 public class VirtualMemoryFFMTest extends VirtualMemoryTest {
 
--- a/oshi-core-java25/src/test/java/oshi/software/os/OperatingSystemFFMTest.java
+++ b/oshi-core-java25/src/test/java/oshi/software/os/OperatingSystemFFMTest.java
@@ -20,7 +20,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 /**
  * Test OS
  */
-@EnabledForJreRange(min = JRE.JAVA_24)
+@EnabledForJreRange(min = JRE.JAVA_25)
 @TestInstance(Lifecycle.PER_CLASS)
 public class OperatingSystemFFMTest extends OperatingSystemTest {
     @Override
--- a/oshi-core/src/test/java/oshi/SystemInfoTest.java
+++ b/oshi-core/src/test/java/oshi/SystemInfoTest.java
@@ -60,7 +60,7 @@ import oshi.util.Util;
 /**
  * A demonstration of access to many of OSHI's capabilities
  */
-@EnabledForJreRange(max = JRE.JAVA_23)
+@EnabledForJreRange(max = JRE.JAVA_25)
 public class SystemInfoTest { // NOSONAR squid:S5786
 
     private static final Logger logger = LoggerFactory.getLogger(SystemInfoTest.class);
--- a/oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
+++ b/oshi-core/src/test/java/oshi/software/os/OperatingSystemTest.java
@@ -47,7 +47,7 @@ import oshi.software.os.OperatingSystem.ProcessSorting;
 /**
  * Test OS
  */
-@EnabledForJreRange(max = JRE.JAVA_23)
+@EnabledForJreRange(max = JRE.JAVA_25)
 @TestInstance(Lifecycle.PER_CLASS)
 class OperatingSystemTest {
 


oshi-core-java25/src/main/java/oshi/ffm/ForeignFunctions.java
```diff
@@ -70,4 +72,12 @@ public static MethodHandle downcall(SymbolLookup lib, String symbol, MemoryLayou
...
         return LINKER.downcallHandle(sym, fd);
     }
+
+    protected static final Linker.Option CAPTURE_CALL_STATE = Linker.Option.captureCallState("errno");
+    public static final StructLayout CAPTURED_STATE_LAYOUT = Linker.Option.captureStateLayout();
+    protected static final VarHandle ERRNO_HANDLE = CAPTURED_STATE_LAYOUT.varHandle(PathElement.groupElement("errno"));
+
+    public static int getErrno(MemorySegment callState) {
+        return (int) ERRNO_HANDLE.get(callState, 0);
+    }
```
dbwiddis, line 82: Thanks for this implementation! I learned something new!

oshi-core-java25/src/main/java/oshi/ffm/mac/MacSystemFunctions.java
```diff
@@ -157,4 +163,42 @@ public static int getfsstat64(MemorySegment buffer, int bufsize, int flags) {
...
+    private static final MethodHandle mach_host_self_handle = LINKER.downcallHandle(SYSTEM_LIBRARY.findOrThrow("mach_host_self"),
+        FunctionDescriptor.of(JAVA_INT));
+
+
+    public static int mach_host_self() {
+        try {
+            return (int) mach_host_self_handle.invokeExact();
+        } catch (Throwable e) {
+            return -1;
+        }
```
dbwiddis: Shouldn't this return MACH_PORT_NULL (0) on failure?

If we're catching a Throwable and returning anything not explicitly defined, we should use an Optional; otherwise we should usually just let the caller handle the throwable and decide on the default.

oshi-core-java25/src/main/java/oshi/ffm/mac/MacSystemFunctions.java
```diff
@@ -157,4 +163,42 @@ public static int getfsstat64(MemorySegment buffer, int bufsize, int flags) {
...
+    }
+
+    private static final MethodHandle host_page_size = LINKER.downcallHandle(SYSTEM_LIBRARY.findOrThrow("host_page_size"),
+        FunctionDescriptor.of(JAVA_INT, JAVA_INT, ADDRESS));
+
+    public static int host_page_size(int hostPort, MemorySegment pPageSize) {
+        try {
+            return (int) host_page_size.invokeExact(hostPort, pPageSize);
+        } catch (Throwable e) {
+            return -1;
```
dbwiddis: We already have this default return in the calling method, we should just throw here and use a catch block to assign the -1.

oshi-core-java25/src/main/java/oshi/ffm/mac/MacSystemFunctions.java
```diff
@@ -157,4 +163,42 @@ public static int getfsstat64(MemorySegment buffer, int bufsize, int flags) {
...
+        FunctionDescriptor.of(JAVA_INT, JAVA_INT, JAVA_INT, ADDRESS, ADDRESS),
+        CAPTURE_CALL_STATE
+    );
+
+
+    public static int host_statistics(MemorySegment callState, int hostPort, int hostStat, MemorySegment stats, MemorySegment count) {
+        try {
+            return (int) host_statistics.invokeExact(callState, hostPort, hostStat, stats, count);
+        } catch (Throwable e) {
+            return -1;
```
dbwiddis: We took the time to capture the call state/errno here. How does the caller know -1 = failure?  We should just throw the throwable from this method.

oshi-core-java25/src/main/java/oshi/hardware/platform/mac/MacMemoryUtil.java
```diff
@@ -0,0 +1,43 @@
...
+import java.lang.foreign.MemorySegment;
+
+import static java.lang.foreign.ValueLayout.JAVA_INT;
+import static oshi.ffm.ForeignFunctions.CAPTURED_STATE_LAYOUT;
+import static oshi.ffm.ForeignFunctions.getErrno;
+import static oshi.ffm.mac.MacSystem.VM_STATISTICS;
+import static oshi.ffm.mac.MacSystemFunctions.host_statistics;
+import static oshi.ffm.mac.MacSystemFunctions.mach_host_self;
+
+final class MacMemoryUtil {
```
dbwiddis: We've been using the FFM name in the Util classes to help distinguish them.... at some point we may want to completely separate all the code and making either file names or packages clearly named will help that.

oshi-core-java25/src/main/java/oshi/util/platform/mac/SysctlUtilFFM.java
```diff
@@ -83,11 +80,8 @@ public static long sysctl(String name, long def) {
         try (Arena arena = Arena.ofConfined()) {
             MemorySegment nameSeg = arena.allocateFrom(name);
             MemorySegment valueSeg = arena.allocate(JAVA_LONG);
-            MemorySegment sizeSeg = arena.allocateFrom(SIZE_T, Integer.BYTES);
-            int result = MacSystemFunctions.sysctlbyname(nameSeg, valueSeg, sizeSeg, MemorySegment.NULL, 0L);
-
-            if (result != 0) {
-                LOG.warn(SYSCTL_FAIL, name, result);
+            MemorySegment sizeSeg = arena.allocateFrom(SIZE_T, JAVA_LONG.byteSize());
```
dbwiddis, line 83: Thanks for fixing this bug!

oshi-core-java25/src/main/java/oshi/SystemInfoFFM.java
```diff
@@ -24,10 +24,10 @@ public class SystemInfoFFM {
...
         String osName = System.getProperty("os.name");
         CURRENT_PLATFORM = switch (osName) {
-        case String name when name.startsWith("Linux") -> PlatformEnumFFM.LINUX;
-        case String name when name.startsWith("Mac") || name.startsWith("Darwin") -> PlatformEnumFFM.MACOS;
-        case String name when name.startsWith("Windows") -> PlatformEnumFFM.WINDOWS;
-        default -> PlatformEnumFFM.UNSUPPORTED;
+            case String name when name.startsWith("Linux") -> PlatformEnumFFM.LINUX;
+            case String name when name.startsWith("Mac") || name.startsWith("Darwin") -> PlatformEnumFFM.MACOS;
+            case String name when name.startsWith("Windows") -> PlatformEnumFFM.WINDOWS;
+            default -> PlatformEnumFFM.UNSUPPORTED;
```
dbwiddis, line 30: Thanks for this indentation.  I'm not sure our current formatter does this... could just be an older config from JDK8 days that didn't even include this new format.   Can you either update the xml file or open an issue with a suggestion for me to do so?

oshi-core-java25/src/main/java/oshi/ffm/mac/MacSystemFunctions.java
```diff
@@ -100,22 +100,28 @@ public static int getpid() throws Throwable {
...
-            MemorySegment newp, long newlen) throws Throwable {
-        return (int) sysctl.invokeExact(name, namelen, oldp, oldlenp, newp, newlen);
+    private static final MethodHandle sysctl = LINKER.downcallHandle(
+        SYSTEM_LIBRARY.findOrThrow("sysctl"),
+        FunctionDescriptor.of(JAVA_INT, ADDRESS, JAVA_INT, ADDRESS, ADDRESS, ADDRESS, SIZE_T),
+        CAPTURE_CALL_STATE
+    );
+
+    public static int sysctl(MemorySegment callState, MemorySegment name, int namelen, MemorySegment oldp, MemorySegment oldlenp,
+                             MemorySegment newp, long newlen) throws Throwable {
```
dbwiddis: Similar indentation question.  I want to make sure our spotless format matches whatever you've got in your IDE to minimize the diffs.  I'm not sure it will be consistent at this point.

oshi-core-java25/src/test/java/oshi/hardware/GlobalMemoryFFMTest.java
```diff
@@ -0,0 +1,16 @@
+package oshi.hardware;
+
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.condition.EnabledForJreRange;
+import org.junit.jupiter.api.condition.JRE;
+import oshi.SystemInfoFFM;
+
+@EnabledForJreRange(min = JRE.JAVA_24)
```
dbwiddis: We can make this 25 now.  We only had 24 until JDK25 GA, when we were able to bump the GHA matrix to test on 25.

oshi-core-java25/src/main/java/oshi/SystemInfoFFM.java
wolfs: I updated the xml file and formatted the code base.

oshi-core-java25/src/main/java/oshi/hardware/platform/mac/MacMemoryUtil.java
wolfs: Added FFM as a postfix.

oshi-core-java25/src/main/java/oshi/ffm/mac/MacSystemFunctions.java
wolfs: Should be consistent now.

oshi-core-java25/src/test/java/oshi/hardware/GlobalMemoryFFMTest.java
wolfs: It is now JDK 25 everywhere.

oshi-core-java25/src/main/java/oshi/ffm/mac/MacSystemFunctions.java
wolfs: All of the native function calls now throw the throwable.

Issue #3023, opened 2025-11-12
Unexpected registry type on Windows 7


mts2200:
Hello! I am an application developer and am using OSHI to retrieve hardware data about a user's computer. One of my users encountered the problem `java.lang.RuntimeException: Unexpected registry type 3, expected REG_SZ or REG_EXPAND_SZ`. After a detailed study of the problem, it turned out that he had Windows 7 32-bit, some variables in `regedit` are in binary format (REG_BINARY).

Stacktrace:
```
java.lang.RuntimeException: Unexpected registry type 3, expected REG_SZ or REG_EXPAND_SZ
	at com.sun.jna.platform.win32.Advapi32Util.registryGetStringValue(Advapi32Util.java:774) ~[jna-platform-5.17.0.jar:5.17.0 (b0)]
	at com.sun.jna.platform.win32.Advapi32Util.registryGetStringValue(Advapi32Util.java:744) ~[jna-platform-5.17.0.jar:5.17.0 (b0)]
	at com.sun.jna.platform.win32.Advapi32Util.registryGetStringValue(Advapi32Util.java:718) ~[jna-platform-5.17.0.jar:5.17.0 (b0)]
	at oshi.hardware.platform.windows.WindowsGraphicsCard.getGraphicsCards(WindowsGraphicsCard.java:83) ~[oshi-core-6.8.2.jar:6.8.2]
	at oshi.hardware.platform.windows.WindowsHardwareAbstractionLayer.getGraphicsCards(WindowsHardwareAbstractionLayer.java:87) ~[oshi-core-6.8.2.jar:6.8.2]
```

User `regedit` screenshot:

<img width="1121" height="117" alt="Image" src="https://github.com/user-attachments/assets/6cee4582-65b5-48eb-9a5d-98dcbca1cf65" />

Can someone help to fix it?

rohan-coder02:
Hi @mts2200 ,

**Analysis:**

+ https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsGraphicsCard.java
+ In WindowsGraphicsCard class, OSHI uses `Advapi32Util#registryGetStringValue` from JNA for the below variables
```
String name = Advapi32Util.registryGetStringValue(WinReg.HKEY_LOCAL_MACHINE, fullKey, DRIVER_DESC);

[...]

PR #3024, opened 2025-11-14
Add RegistryUtil class in oshi-core to handle unexpected registry types

rohan-coder02:
### Description:

+ RegistryUtil class can be used for reading Windows Registry values both StringValue and LongValue
+ In the case of StringValue it handles both String (REG_SZ and REG_EXPAND_SZ) and Binary (REG_BINARY)
+ This should help fix issue -> https://github.com/oshi/oshi/issues/3023
+ In the case of LongValue it handles both String (REG_SZ and REG_EXPAND_SZ) and Integer (REG_DWORD)
+ This ensures the fix made in https://github.com/oshi/oshi/pull/2973 is intact 

### Testing:

Compared the outputs using oshi 6.9.1 and with the current code changes (oshi-core-6.9.2-SNAPSHOT) in ideal scenario, the outputs match.
```

System.out.println(si.getOperatingSystem().getInstalledApplications());

System.out.println(si.getHardware().getGraphicsCards());
```

To fix https://github.com/oshi/oshi/issues/3023

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ##### Bug fixes / Improvements
 * [#2974](https://github.com/oshi/oshi/pull/2974): Use simple xml parsing on MacInstalledApps, support more info/version from app plist, inclusing binary format - [@dyorgio](https://github.com/dyorgio).
 * [#2975](https://github.com/oshi/oshi/pull/2975): Query Mac apps in xml formatting using internal xml parsing utils - [@dyorgio](https://github.com/dyorgio).
+* [#3024](https://github.com/oshi/oshi/pull/3024): Add RegistryUtil class in oshi-core to handle unexpected registry types - [@rohan-coder02](https://github.com/rohan-coder02).
 
 # 6.9.0 (2025-09-18), 6.9.1 (2025-10-18)
 
--- a/oshi-core/src/main/java/oshi/driver/windows/registry/InstalledAppsData.java
+++ b/oshi-core/src/main/java/oshi/driver/windows/registry/InstalledAppsData.java
@@ -18,10 +18,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import com.sun.jna.platform.win32.Advapi32;
 import com.sun.jna.platform.win32.Advapi32Util;
 import com.sun.jna.platform.win32.Win32Exception;
 import com.sun.jna.platform.win32.WinReg.HKEY;


oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
```diff
@@ -0,0 +1,181 @@
...
+
+    public static Object getRegistryValueOrNull(HKEY root, String path, String key, int accessFlag) {
+        HKEY hKey = null;
+        try {
+            hKey = getRegistryHKey(root, path, accessFlag);
+            Object value = Advapi32Util.registryGetValue(root, path, key);
+            if ((value instanceof Integer) || (value instanceof String && !((String) value).trim().isEmpty())
+                    || value instanceof byte[]) {
+                return value;
+            }
```
dbwiddis:
```suggestion
            return Objects.isNull(value) ? null : value;
```
~In addition to removing the unneeded `getRegistryHkey()` and the whole finally block.~

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+        return null;
+    }
+
+    private static HKEY getRegistryHKey(HKEY rootKey, String path, int accessFlag) {
+        HKEYByReference phkKey = new HKEYByReference();
+        int rc = ADV.RegOpenKeyEx(rootKey, path, 0, KEY_READ | accessFlag, phkKey);
+        if (rc != ERROR_SUCCESS) {
+            throw new Win32Exception(rc);
+        }
```
dbwiddis:
```suggestion
        HKEYByReference phkKey = Advapi32Util.registryGetKey(rootKey, path, KEY_READ | accessFlag);
```

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
```diff
@@ -0,0 +1,181 @@
...
+        return phkKey.getValue();
+    }
+
+    private static long registryValueToLong(Object val) {
```
dbwiddis, line 78: Let's add a javadoc (optional for private) or at least comments here indicating the purpose of this method

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
```diff
@@ -0,0 +1,181 @@
...
+        Object val = getRegistryValueOrNull(root, path, key, accessFlag);
+        return registryValueToString(val);
+    }
+
+    /**
+     * Decodes registry value to String using multiple fallback encodings.
+     */
+    private static String registryValueToString(Object val) {
+        if (val == null)
+            return null;
```
dbwiddis: Style nit (and elsewhere):
```suggestion
        if (val == null) {
            return null;
        }
```

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    /**
+     * Attempts to decode REG_BINARY into a string, trying UTF-16LE → UTF-8 → Windows-1252 → Hex fallback.
+     */
+    private static String decodeBinaryToString(byte[] bytes) {
+        if (bytes == null || bytes.length == 0)
+            return null;
+
+        String[] encodings = { "UTF-16LE", "UTF-8", "Windows-1252" };
```
dbwiddis: Let's use `StandardCharseets.UTF_16LE` and create a constant using `Charset.forName("Windows-1252")` rather than parsing the charset every time

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
```diff
@@ -0,0 +1,181 @@
...
+
+        for (String enc : encodings) {
+            try {
+                String decoded = new String(bytes, enc).trim();
+                if (!decoded.isEmpty()) {
+                    return decoded;
```
dbwiddis: I'm not sure iterating these charsets and just testing for an empty string will work.  With the existing code, UTF-16LE will not produce an empty string or throw an exception if the string is a standard "C String" and I don't think a C string would fail UTF-8 parsing.

Since we're on Windows we can only check for the two charsets, and can probably use a heuristic to pick which one:
 - UTF-16LE will end with two null bytes
 - Windows-1252 (C String) will end with one null byte
 - Else return hex

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
```diff
@@ -0,0 +1,182 @@
...
+                }
+            } catch (UnsupportedEncodingException e) {
```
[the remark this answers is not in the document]
dbwiddis: This exception won't be thrown if we use StandardCharsets or precompute Windows-1252 as a constant (it would throw an exception on initialization but all JVMs have this.)

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
```diff
@@ -0,0 +1,181 @@
...
+            }
+        }
+
+        // fall back to Hex
+        StringBuilder sb = new StringBuilder();
+        for (byte b : bytes) {
+            sb.append(String.format(Locale.ROOT, "%02X", b));
+        }
+
+        return sb.length() == 0 ? null : sb.toString();
```
dbwiddis:
```suggestion
        return Arrays.stream(bytes)
            .mapToObj(b -> String.format(Locale.ROOT, "%02X", b))
            .collect(Collectors.joining(" "));
```
More readable with a space.

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
```diff
@@ -0,0 +1,181 @@
...
+import com.sun.jna.platform.win32.WinReg.HKEY;
+import com.sun.jna.platform.win32.WinReg.HKEYByReference;
+
+import java.io.UnsupportedEncodingException;
+import java.util.Locale;
+
+import static com.sun.jna.platform.win32.WinError.ERROR_SUCCESS;
+import static com.sun.jna.platform.win32.WinNT.KEY_READ;
+
+public final class RegistryUtil {
```
dbwiddis, line 23: Let's use the JNA name since we probably will need/want a similar FFM version later.

rohan-coder02: Currently, we've followed the pattern like WindowsInstalledApps class in `oshi-core` and WindowsInstalledAppsFFM class in FFM.

So, we can name the class in FFM as RegistryUtilFFM.

Let me know your view.

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
rohan-coder02: Arrays.stream supports int[], long[], double[]
https://www.studytonight.com/java-util/java-arrays-stream-method

Arrays-stream-error-with-bytes


Arrays-stream-datatype-support

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
rohan-coder02: I'll modify as below, let me know your view.
```
    private static String decodeBinaryToString(byte[] bytes) {
        if (bytes == null || bytes.length == 0) {
            return null;
        }

        int len = bytes.length;

        // Check for UTF-16LE null terminator (00 00)
        if (len >= 2 && bytes[len - 1] == 0x00 && bytes[len - 2] == 0x00) {
            return new String(bytes, StandardCharsets.UTF_16LE).trim();
        }

        // Check for Windows-1252 (single null terminator)
        if (bytes[len - 1] == 0x00) {
            return new String(bytes, Charset.forName("Windows-1252")).trim();
        }

        // fall back to Hex
        StringBuilder sb = new StringBuilder();
        for (byte b : bytes) {
            sb.append(String.format(Locale.ROOT, "%02X", b));
        }

        return sb.length() == 0 ? null : sb.toString();
    }
```

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
rohan-coder02: I'll modify the code as below.
```
public static Object getRegistryValueOrNull(HKEY root, String path, String key, int accessFlag) {
        HKEY hKey = null;
        try {
            hKey = Advapi32Util.registryGetKey(root, path, KEY_READ | accessFlag).getValue();
            Object value = Advapi32Util.registryGetValue(root, path, key);
            return Objects.isNull(value) ? null : value;
        } catch (Win32Exception e) {
            LOG.trace("Unable to access " + path + " with flag " + accessFlag + ": " + e.getMessage());
        }
       return null;
}
```

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
dbwiddis: Huh.  That's silly.

How about
```java
        return IntStream.range(0, bytes.length)
            .mapToObj(i -> String.format(Locale.ROOT, "%02X", bytes[i]))
            .collect(Collectors.joining(" "));
```

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
dbwiddis: > let me know your view.

LGTM except:
 - make Charset.forName("Windows-1252") a class constant so you only do it once
 - you need a length check for the windows-1252 version
 - we already return null for empty array in the first line so I think returning empty string at the end is ok
 - may want to move this to `ParseUtil` and add a few tests for it

oshi-core/src/main/java/oshi/util/platform/windows/RegistryUtil.java
dbwiddis: Ignore my comment about the finally block: I was trusting the copilot auto-comment (who turned that on, anyway? 😁).   It's needed.  Otherwise looks good.

oshi-core/src/main/java/oshi/util/ParseUtil.java
```diff
@@ -48,6 +51,8 @@ public final class ParseUtil {
 
     private static final Logger LOG = LoggerFactory.getLogger(ParseUtil.class);
 
+    private static final Charset CP1252 = Charset.forName("Windows-1252");
```
dbwiddis, line 54: Love the shortened name here :)

PR #3057, opened 2026-01-12
File system fixes and optimizations

dyorgio:
Delay name filtering on MacFileSystem.getFileStoreMatching to fix non-matching that could be found with DiskArbitration.
Include local boolean on all OSFileStore objects. It is used to optimize updateAttributes() fileStore lookup.

## Summary by CodeRabbit

* **New Features**
  * Added OSFileStore.isLocal() to identify local (non-network) file stores.

* **Bug Fixes / Improvements**
  * Improved file store detection and matching, with consistent local-vs-network filtering across macOS, Linux, Windows, FreeBSD, OpenBSD, Solaris, and AIX.
  * More reliable attribute refreshes and optimizations for filesystem enumeration and reporting.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 # 6.9.3 (in progress)
 
-
 ##### Bug fixes / Improvements
 * [#3057](https://github.com/oshi/oshi/pull/3057): FileSystem fixes and optimization - [@dyorgio](https://github.com/dyorgio).
 * [#3064](https://github.com/oshi/oshi/pull/3064): Handle binary registry values in WindowsGraphicsCard - [@dbwiddis](https://github.com/dbwiddis).
--- a/oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisFileSystem.java
+++ b/oshi-core/src/main/java/oshi/software/os/unix/solaris/SolarisFileSystem.java
@@ -142,8 +142,8 @@ public class SolarisFileSystem extends AbstractFileSystem {
             }
 
             fsList.add(new SolarisOSFileStore(name, volume, name, path, options, "", isLocal, "", description, type,
-                    freeSpace, usableSpace, totalSpace, inodeFreeMap.containsKey(path) ? inodeFreeMap.get(path) : 0L,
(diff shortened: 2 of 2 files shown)


oshi-core/src/main/java/oshi/software/os/linux/LinuxFileSystem.java
```diff
@@ -132,7 +128,8 @@ private static List<OSFileStore> getFileStoreMatching(String nameToMatch, Map<St
             String type = split[2];
 
             // Skip non-local drives if requested, and exclude pseudo file systems
-            if ((localOnly && NETWORK_FS_TYPES.contains(type))
+            boolean nonLocal = NETWORK_FS_TYPES.contains(type);
+            if ((localOnly && nonLocal)
```
dbwiddis: (Style nit) (Here and elsewhere) If we only used `nonLocal` in this context this would be fine, but the later need to negate this in the constructors (`!nonLocal`) is slightly less readable.  I think it might be better to define the boolean as `isLocal` particularly since you added that name as a new API.

oshi-core/src/main/java/oshi/software/os/OSFileStore.java
```diff
@@ -82,6 +82,13 @@ public interface OSFileStore {
      */
     String getUUID();
 
+    /**
+     * Is this Filesystem local?
+     * 
+     * @return <code>true</code> if the drive is local.
```
dbwiddis: Let's add some caveats to this indicating it's OS-dependent, and give examples of what types it includes and excludes.  I think the goal here is to avoid latency associated with remote drives, but I think we do include "ram disks"?

oshi-core/src/main/java/oshi/software/os/mac/MacFileSystem.java
```diff
@@ -232,9 +225,13 @@ private static List<OSFileStore> getFileStoreMatching(String nameToMatch, boolea
                         }
                     }
 
+                    if (nameToMatch != null && !nameToMatch.equals(name)) {
+                        continue;
+                    }
```
dbwiddis, line 230: I think you need to move this name matching in the java25 implementation as well.

PR #3082, opened 2026-03-01
Separate RSS from the private/footprint memory shown by graphical system monitors

dyorgio:
Add getMemoryFootprint method on OSProcess class.

The objective is get a memory size value like in end user task manager.

In linux and Windows RSS value appears to be ok, but on macOS values showed in "Activity Monitor" are very different of it.

## Summary by CodeRabbit

* **New Features**
  * Per-process memory split into Resident and Private (footprint) values across platforms; Windows exposes both private and total working set. Added sorting by private memory and a Private memory column in process views. Old RSS-based metric retained for compatibility but deprecated.

* **Tests**
  * Updated tests to assert both resident and private memory values.

* **Documentation**
  * Changelog and download snapshots updated to 6.11.0-SNAPSHOT.


dyorgio:
@dbwiddis thank you so much, you are a code machine :).

PR #3096, opened 2026-03-13
Correct PowerSupply units on Linux

benplotke:
I noticed the units for batteries on Linux were off. I found the incoming units from https://docs.kernel.org/power/power_supply_class.html and added the appropriate conversions to the units specified in the PowerSupply interface.

I checked no tests failed, but I did not add any new tests. Let me know if you want me to add a test or change anything.

Thanks for reviewing my PR!

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of Linux power-source readings via corrected unit conversions, scaling, sign semantics, and fallback computations.
  * Adjusted Mac power-source charging/discharging logic to produce more accurate status reporting.
* **Improvements**
  * Better handling of alternative sensor data and remaining-time selection between charge/discharge.
  * Safer parsing behavior for numeric, duration, and UUID inputs when values are missing.
* **Tests**
  * Added Linux-focused tests validating parsing, units, sign conventions, fallbacks, and defaults.
* **Chores**
  * Updated changelog and test JVM configuration for reflective access.

### DIFF
--- a/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxPowerSource.java
+++ b/oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxPowerSource.java
@@ -179,7 +179,7 @@ public final class LinuxPowerSource extends AbstractPowerSource {
         int psDesignCapacity = rawDesign >= 0 ? rawDesign / 1000 : -1;
 
         // Debian/Ubuntu provides Voltage and Power. Fedora/RHEL provides Voltage and Current.
-        double psVoltage = ParseUtil.parseDoubleOrDefault(props.get(Prop.POWER_SUPPLY_VOLTAGE_NOW), -1) / 1_000_000d;
+        double psVoltage = ParseUtil.parseDoubleOrDefault(props.get(Prop.POWER_SUPPLY_VOLTAGE_NOW), -1d) / 1_000_000d;
         double psPowerUsageRate = 0d;
         double psAmperage = 0d;
         // From Physics we know P = IV so I = P/V
@@ -188,8 +188,8 @@ public final class LinuxPowerSource extends AbstractPowerSource {
         // (positive = charging, negative = discharging).
         double sign = psDischarging ? -1d : 1d;
         if (psVoltage > 0) {
-            double rawPower = ParseUtil.parseDoubleOrDefault(props.get(Prop.POWER_SUPPLY_POWER_NOW), -1);
-            double rawCurrent = ParseUtil.parseDoubleOrDefault(props.get(Prop.POWER_SUPPLY_CURRENT_NOW), -1);
+            double rawPower = ParseUtil.parseDoubleOrDefault(props.get(Prop.POWER_SUPPLY_POWER_NOW), -1d);
+            double rawCurrent = ParseUtil.parseDoubleOrDefault(props.get(Prop.POWER_SUPPLY_CURRENT_NOW), -1d);
             if (rawPower >= 0) {
                 psPowerUsageRate = sign * rawPower / 1000d;
             }
--- a/oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
+++ b/oshi-core/src/main/java/oshi/hardware/common/AbstractPowerSource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 The OSHI Project Contributors
+ * Copyright 2016-2026 The OSHI Project Contributors
  * SPDX-License-Identifier: MIT
  */
 package oshi.hardware.common;
@@ -218,20 +218,20 @@ public abstract class AbstractPowerSource implements PowerSource {
 
     private static List<PowerSource> getPowerSources() {
         switch (SystemInfo.getCurrentPlatform()) {
-        case WINDOWS:
-            return WindowsPowerSource.getPowerSources();
-        case MACOS:
-            return MacPowerSource.getPowerSources();
-        case LINUX:
-            return LinuxPowerSource.getPowerSources();
-        case SOLARIS:
-            return SolarisPowerSource.getPowerSources();
-        case FREEBSD:
-            return FreeBsdPowerSource.getPowerSources();
-        case AIX:
-            return AixPowerSource.getPowerSources();
-        default:
-            throw new UnsupportedOperationException("Operating system not supported: " + Platform.getOSType());
+            case WINDOWS:
+                return WindowsPowerSource.getPowerSources();
+            case MACOS:
+                return MacPowerSource.getPowerSources();
+            case LINUX:
+                return LinuxPowerSource.getPowerSources();
+            case SOLARIS:
+                return SolarisPowerSource.getPowerSources();
+            case FREEBSD:
+                return FreeBsdPowerSource.getPowerSources();
+            case AIX:
+                return AixPowerSource.getPowerSources();
+            default:
+                throw new UnsupportedOperationException("Operating system not supported: " + Platform.getOSType());
         }
     }
 
@@ -266,7 +266,7 @@ public abstract class AbstractPowerSource implements PowerSource {
         sb.append("SerialNumber: ").append(getSerialNumber()).append(", ");
         sb.append("Temperature: ");
         if (getTemperature() > 0) {
-            sb.append(getTemperature()).append("°C");
+            sb.append(getTemperature()).append("C");
         } else {
             sb.append(Constants.UNKNOWN);
         }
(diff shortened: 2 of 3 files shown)


oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxPowerSource.java
```diff
@@ -102,41 +102,45 @@ public static List<PowerSource> getPowerSources() {
 
                                         // Debian/Ubuntu provides Energy. Fedora/RHEL provides Charge.
                                         psCurrentCapacity = ParseUtil.parseIntOrDefault(
-                                                device.getPropertyValue("POWER_SUPPLY_ENERGY_NOW"), -1);
+                                                device.getPropertyValue("POWER_SUPPLY_ENERGY_NOW"), -1) / 1000;
```
dbwiddis: As CodeRabbit pointed out, this integer division breaks the sentinel value.  

If this key is not present, -1 (int) divided by 1000 (int) will result in 0 (int), so the next line will fail to attempt the fallback.  

You could make the default -1000 but that seems like a completely wasted division in the failure case, vs. leaving it as -1 and then doing the division-by-1000 where you later assign the CapacityUnits enum.

The same pattern repeats below.

All this repetition (and the same logic in the file-based fallback repeating all the strings) is screaming for a helper method to do this division, in addition to making those strings constants.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxPowerSource.java
```diff
@@ -102,41 +102,45 @@ public static List<PowerSource> getPowerSources() {
...
-                                                device.getPropertyValue("POWER_SUPPLY_VOLTAGE_NOW"), -1);
+                                        psVoltage = ParseUtil.parseDoubleOrDefault(
+                                                device.getPropertyValue("POWER_SUPPLY_VOLTAGE_NOW"), -1) / 1_000_000d;
                                         // From Physics we know P = IV so I = P/V
                                         if (psVoltage > 0) {
                                             String power = device.getPropertyValue("POWER_SUPPLY_POWER_NOW");
                                             String current = device.getPropertyValue("POWER_SUPPLY_CURRENT_NOW");
                                             if (power == null) {
-                                                psAmperage = ParseUtil.parseIntOrDefault(current, 0);
+                                                psAmperage = ParseUtil.parseDoubleOrDefault(current, 0) / 1000d;
```
dbwiddis: may as well make the default `0d` to save a cast

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxPowerSource.java
```diff
@@ -200,32 +204,43 @@ public static List<PowerSource> getPowerSources() {
...
+                                / 1000;
+                        psCapacityUnits = CapacityUnits.MAH;
                     }
                     if (psMap.containsKey("POWER_SUPPLY_ENERGY_FULL")) {
-                        psCurrentCapacity = ParseUtil.parseIntOrDefault(psMap.get("POWER_SUPPLY_ENERGY_FULL"), 1);
+                        psCurrentCapacity = ParseUtil.parseIntOrDefault(psMap.get("POWER_SUPPLY_ENERGY_FULL"), 1)
+                                / 1000;
                     } else if (psMap.containsKey("POWER_SUPPLY_CHARGE_FULL")) {
-                        psCurrentCapacity = ParseUtil.parseIntOrDefault(psMap.get("POWER_SUPPLY_CHARGE_FULL"), 1);
+                        psCurrentCapacity = ParseUtil.parseIntOrDefault(psMap.get("POWER_SUPPLY_CHARGE_FULL"), 1)
```
dbwiddis: (Pre exissting bug). As CodeRabbit points out, these two should be max.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxPowerSource.java
```diff
@@ -200,32 +204,43 @@ public static List<PowerSource> getPowerSources() {
...
+                                / 1000;
                     }
                     if (psMap.containsKey("POWER_SUPPLY_ENERGY_FULL_DESIGN")) {
-                        psMaxCapacity = ParseUtil.parseIntOrDefault(psMap.get("POWER_SUPPLY_ENERGY_FULL_DESIGN"), 1);
+                        psMaxCapacity = ParseUtil.parseIntOrDefault(psMap.get("POWER_SUPPLY_ENERGY_FULL_DESIGN"), 1)
+                                / 1000;
                     } else if (psMap.containsKey("POWER_SUPPLY_CHARGE_FULL_DESIGN")) {
-                        psMaxCapacity = ParseUtil.parseIntOrDefault(psMap.get("POWER_SUPPLY_CHARGE_FULL_DESIGN"), 1);
+                        psMaxCapacity = ParseUtil.parseIntOrDefault(psMap.get("POWER_SUPPLY_CHARGE_FULL_DESIGN"), 1)
```
dbwiddis: (Pre exissting bug). As CodeRabbit points out, these two should be design.

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxPowerSource.java
```diff
@@ -200,32 +204,43 @@ public static List<PowerSource> getPowerSources() {
...
+                                / 1000;
                     }
                     // Debian/Ubuntu provides Voltage and Power.
                     // Fedora/RHEL provides Voltage and Current.
                     if (psMap.containsKey("POWER_SUPPLY_VOLTAGE_NOW")) {
-                        psVoltage = ParseUtil.parseIntOrDefault(psMap.get("POWER_SUPPLY_VOLTAGE_NOW"), -1);
+                        psVoltage = ParseUtil.parseDoubleOrDefault(psMap.get("POWER_SUPPLY_VOLTAGE_NOW"), -1)
+                                / 1_000_000d;
```
dbwiddis: This is documented to return -1 if not available, so the conversion breaks that.  (Similar issue as the integer division, but with different results.)

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxPowerSource.java
```diff
@@ -200,32 +204,43 @@ public static List<PowerSource> getPowerSources() {
...
                     }
                     // From Physics we know P = IV so I = P/V
                     if (psVoltage > 0) {
                         if (psMap.containsKey("POWER_SUPPLY_POWER_NOW")) {
-                            psPowerUsageRate = ParseUtil.parseIntOrDefault(psMap.get("POWER_SUPPLY_POWER_NOW"), -1);
+                            psPowerUsageRate = ParseUtil.parseDoubleOrDefault(psMap.get("POWER_SUPPLY_POWER_NOW"), -1)
```
dbwiddis: (pre-existing bug, or at least very confusing) 

Both current and power can be negative for a discharging battery.  Using the -1 default here is probably wrong.  Since we are already checking for the key we would only hit this default in the case of a parse failure.   Still, I think maybe we should make the default/sentinel value (and original assigned value) `Double.NaN` or perhaps -0.0 (note the need to use `equals()` to differentiate either of these for code logic, as `==` has different behavior)

oshi-core/src/main/java/oshi/hardware/platform/linux/LinuxPowerSource.java
```diff
@@ -102,41 +102,45 @@ public static List<PowerSource> getPowerSources() {
...
                                         }
                                         psDesignCapacity = ParseUtil.parseIntOrDefault(
-                                                device.getPropertyValue("POWER_SUPPLY_ENERGY_FULL_DESIGN"), 1);
+                                                device.getPropertyValue("POWER_SUPPLY_ENERGY_FULL_DESIGN"), 1) / 1000;
                                         if (psDesignCapacity < 0) {
                                             psDesignCapacity = ParseUtil.parseIntOrDefault(
-                                                    device.getPropertyValue("POWER_SUPPLY_CHARGE_FULL_DESIGN"), 1);
+                                                    device.getPropertyValue("POWER_SUPPLY_CHARGE_FULL_DESIGN"), 1)
+                                                    / 1000;
                                         }
```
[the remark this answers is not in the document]
benplotke: We expect the number from POWER_SUPPLY_ENERGY_NOW to either be non-negative or not set, in which case we would get negative one, which will not go to zero when divided by 1000. So the change will never cause us to miss the fallback case.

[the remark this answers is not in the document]
benplotke: Oh, my mistake. I was thinking this was a double. I'll fix

Issue #3100, opened 2026-03-24
Configurable Privileged Command Execution (Least Privilege Support)


dbwiddis:
## Summary

Enable OSHI users to safely access features requiring elevated permissions without running their entire application as root, by providing a configurable mechanism to execute specific commands with privilege escalation.
 - Coming from #3099

## What problem are we solving?

OSHI includes several features that require elevated permissions (e.g., `sudo` or Administrator equivalent).

On Linux systems in particular, OSHI relies on system utilities such as:
- `dmidecode` (e.g., serial number, BIOS, memory hardware details)
- `lshw`, `lsblk`, and similar utilities
- Direct reads from `/proc` and `/sys` virtual filesystems (via Java file I/O, e.g., `BufferedReader` over files)

Currently, users must choose between:
1. Running their entire application as root  
2. Losing access to OSHI features that require elevated permissions  

Running an entire application with elevated privileges violates the **Principle of Least Privilege**, and introduces unnecessary risk, especially since OSHI exposes utilities like `ExecutingCommand`, which intentionally allow arbitrary command execution (although the same functionality exists in core Java, it's just less convenient).

We need a way to grant **fine-grained, minimal privilege access** to only the specific commands required for OSHI's scope of OS/hardware information.

### 1. Global configuration for privilege escalation prefix

Introduce `GlobalConfig` properties to prepend a command prefix for specific native command execution.

Example (`oshi.properties`):

```properties
oshi.sudocommand.prefix=sudo -n
oshi.sudocommand.allowlist=/usr/sbin/dmidecode, /usr/bin/lsblk, /usr/bin/lshw
```
Or programmatically:
```
GlobalConfig.set("oshi.sudocommand.prefix", "sudo -n");
GlobalConfig.set("oshi.sudocommand.allowlist", "/usr/sbin/dmidecode, /usr/bin/lsblk, /usr/bin/lshw");
```
This would transform:

```
dmidecode -t system
```
into:

```
sudo -n dmidecode -t system
```
This approach:
 - Keeps OSHI unprivileged by default
 - Uses existing OS-level privilege delegation

### 2. Document passwordless sudo for specific commands

Users can configure /etc/sudoers to allow specific commands without requiring a password.

Example:

```
sudo visudo
```
Add:

```
oshiuser ALL=(ALL) NOPASSWD: /usr/sbin/dmidecode, /usr/bin/lsblk, /usr/bin/lshw
```
Then OSHI can safely execute:
```
[...]

PR #3107, opened 2026-03-30
Add configurable privilege escalation on Linux

rohan-coder02:
- Add GlobalConfig constants: OSHI_SUDOCOMMAND_PREFIX, OSHI_SUDOCOMMAND_ALLOWLIST, OSHI_SUDOCOMMAND_FILE_ALLOWLIST
- Create PrivilegedUtil class for privileged command/file operations
- Add ExecutingCommand.runPrivilegedNative() for privilege-aware execution
- Update Linux drivers (Dmidecode, Lshw, Sysfs) to use privileged methods
- Update LinuxCentralProcessor, LinuxOSProcess, AbstractGlobalMemory
- Added the tests for the functionality

Resolves: oshi/oshi#3100

## Summary by CodeRabbit

* **New Features**
  * Optional privileged execution/fallbacks for commands and file reads on Linux.

* **Configuration**
  * Added three properties to configure privileged prefix and allowlists for commands and files.

* **Improvements**
  * More reliable hardware, CPU, memory, graphics, and process data on restricted Linux systems.

* **Refactor**
  * Centralized privileged helpers and unified file/key-value parsing utilities.

* **Tests**
  * New unit tests covering privileged gating, allowlist parsing, file access, and config behavior.

### DIFF
--- a/oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
+++ b/oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
@@ -4,6 +4,7 @@
  */
 package oshi.util;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static oshi.util.GlobalConfig.OSHI_OS_LINUX_PRIVILEGED_ALLOWLIST;
 import static oshi.util.GlobalConfig.OSHI_OS_LINUX_PRIVILEGED_FILE_ALLOWLIST;
 import static oshi.util.GlobalConfig.OSHI_OS_LINUX_PRIVILEGED_PREFIX;
@@ -268,14 +269,27 @@ public final class PrivilegedUtil {
         try {
             Process p = Runtime.getRuntime().exec(cmdArray);
             try {
-                return FileUtil.readAllBytes(p.getInputStream());
+                byte[] stdout = FileUtil.readAllBytes(p.getInputStream());
+                byte[] stderr = FileUtil.readAllBytes(p.getErrorStream());
+                int exitCode = p.waitFor();
+                if (exitCode == 0) {
+                    return stdout;
+                }
+                if (reportError) {
+                    LOG.error("Privileged cat exited with code {}: {}", exitCode, new String(stderr, UTF_8).trim());
+                } else {
+                    LOG.debug("Privileged cat exited with code {}: {}", exitCode, new String(stderr, UTF_8).trim());
+                }
             } finally {
                 ExecutingCommand.destroyProcess(p);
             }
-        } catch (SecurityException | IOException e) {
+        } catch (SecurityException | IOException | InterruptedException e) {
+            if (e instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
             LOG.debug("Failed to execute privileged cat command: {}", e.getMessage());
-            return new byte[0];
         }
+        return new byte[0];
     }
 
     private static String[] buildCatCommand(String filePath) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   [#3089](https://github.com/oshi/oshi/pull/3089),
   [#3092](https://github.com/oshi/oshi/pull/3092),
   [#3095](https://github.com/oshi/oshi/pull/3095): Add GPU metrics to GraphicsCard - [@dbwiddis](https://github.com/dbwiddis).
+* [#3107](https://github.com/oshi/oshi/pull/3107): Add configurable privilege escalation on Linux - [@rohan-coder02](https://github.com/rohan-coder02).
 
 ##### Bug fixes / Improvements
 * [#3086](https://github.com/oshi/oshi/pull/3086): DXGI-based VRAM detection for Windows GPU adapters - [@dbwiddis](https://github.com/dbwiddis).
--- a/oshi-core/src/main/java/oshi/util/ParseUtil.java
+++ b/oshi-core/src/main/java/oshi/util/ParseUtil.java
@@ -1461,9 +1461,9 @@ public final class ParseUtil {
     public static Map<String, String> parseStringListToMap(List<String> lines, String separator) {
         Map<String, String> map = new HashMap<>();
         for (String line : lines) {
-            String[] parts = line.split(separator, 2);
-            if (parts.length == 2) {
-                map.put(parts[0], parts[1].trim());
+            int idx = line.indexOf(separator);
+            if (idx >= 0) {
+                map.put(line.substring(0, idx), line.substring(idx + separator.length()).trim());
             }
         }
         return map;


oshi-core/src/main/java/oshi/util/ExecutingCommand.java
```diff
@@ -159,4 +159,53 @@ public static String getAnswerAt(String cmd2launch, int answerIdx) {
...
+        if (UserGroupInfo.isElevated()) {
+            return runNative(cmdToRun);
+        }
+
+        // Get configured prefix
+        String prefix = PrivilegedUtil.getPrefix();
+        if (prefix.isEmpty()) {
+            LOG.debug("No sudo prefix configured, skipping privileged command: {}", cmdToRun);
+            return Collections.emptyList();
+        }
```
dbwiddis: I don't think this early return is needed.  There's a possibility someone has given permissions to run a command enabling it for everyone without sudo.  You should probably combine the "blank prefix" with the above conditional and just do `runNative()`. The command may still fail, but the code is already set up to handle these situations.

Similar for the below conditional.

oshi-core/src/main/java/oshi/util/ExecutingCommand.java
```diff
@@ -159,4 +159,53 @@ public static String getAnswerAt(String cmd2launch, int answerIdx) {
...
+     *
+     * @param cmd2launch String command to be launched with privilege escalation if needed
+     * @return String or empty string if command failed or privilege requirements were not met
+     */
+    public static String getFirstPrivilegedAnswer(String cmd2launch) {
+        List<String> sa = runPrivilegedNative(cmd2launch);
+        if (!sa.isEmpty()) {
+            return sa.get(0);
+        }
+        return "";
```
dbwiddis: (Nit) (Style) More readable without the negation. Probably compiles to the same bytecode.
```suggestion
        if (sa.isEmpty()) {
            return "";
        }
        return sa.get(0);
```

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,317 @@
+/*
+ * Copyright 2016-2025 The OSHI Project Contributors
```
dbwiddis:
```suggestion
 * Copyright 2026 The OSHI Project Contributors
```

oshi-core/src/test/java/oshi/util/PrivilegedUtilTest.java
```diff
@@ -0,0 +1,227 @@
+/*
+ * Copyright 2016-2025 The OSHI Project Contributors
```
dbwiddis:
```suggestion
 * Copyright 2026 The OSHI Project Contributors
```

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,317 @@
...
+ */
+@ThreadSafe
+public final class PrivilegedUtil {
+
+    private static final Logger LOG = LoggerFactory.getLogger(PrivilegedUtil.class);
+
+    private static volatile Set<String> cachedCommandAllowlist;
+    private static volatile Set<String> cachedFileAllowlist;
+    private static volatile String lastCommandAllowlistConfig = "";
+    private static volatile String lastFileAllowlistConfig = "";
```
dbwiddis: Eslewhere in OSHI we use memoized suppliers to handle this type of caching.  So here you'd define the suppliers and reference the `queryFoo()` methods below.

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,317 @@
...
+            return Collections.emptySet();
+        }
+        Set<String> allowlist = new HashSet<>();
+        for (String entry : allowlistConfig.split(",")) {
+            String trimmed = entry.trim();
+            if (!trimmed.isEmpty()) {
+                allowlist.add(trimmed);
+            }
+        }
+        return allowlist;
```
dbwiddis:
```suggestion
        return Arrays.stream(allowlistConfig.split(","))
            .map(String::trim)
            .filter(s -> !s.isEmpty())
            .collect(Collectors.toSet());
```

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,317 @@
...
+        Path path = Paths.get(filePath);
+        if (!Files.exists(path)) {
+            return Collections.emptyList();
+        }
+
+        // File exists but not readable - try privileged read
+        String prefix = getPrefix();
+        if (prefix.isEmpty()) {
+            return Collections.emptyList();
+        }
```
dbwiddis: Similar comment here about attempting a priveleged read even without a prefix.

And I'm wondering if we need to also validate that `cat` is in the allowlist.

oshi-core/src/main/resources/oshi.properties
```diff
@@ -288,3 +288,42 @@ oshi.os.linux.sensors.hwmon.names=coretemp,k10temp,zenpower,k8temp,via-cputemp
...
+# - Users must configure passwordless sudo for the specified commands
+# - For production use, restrict sudo access to specific commands only
+#
+# Example sudoers configuration (run: sudo visudo):
+#   oshiuser ALL=(ALL) NOPASSWD: /usr/sbin/dmidecode, /usr/bin/lshw, /usr/bin/lsblk
+#
+# The command prefix to prepend for privileged command execution.
+# Common values: "sudo -n" (non-interactive sudo)
+# Default: empty string (no privilege escalation)
+oshi.sudocommand.prefix=
```
dbwiddis: I think we may want these at the operating system level, such as `oshi.os.linux.privileged.*`:
 - FreeBSD/OpenBSD may need `doas`
 - Windows may use third party tools like `gsudo` since the built-in one may not work well with inline
 - Different OS's may need to use different command line switches
 - Different OS's may have different paths to the command for the allowlist

You're already running into this issue with BSD systems using `dmidecode` for `AbstractGlobalMemory`.

oshi-core/src/main/resources/oshi.properties
```diff
@@ -288,3 +288,42 @@ oshi.os.linux.sensors.hwmon.names=coretemp,k10temp,zenpower,k8temp,via-cputemp
...
+# Supports both bare command names (e.g., dmidecode) and full paths
+# (e.g., /usr/sbin/dmidecode). A bare name will match commands invoked
+# by that name regardless of path.
+# Default: empty string (no commands allowed)
+oshi.sudocommand.allowlist=
+#
+# Comma-separated list of file paths eligible for privileged read via
+# "sudo cat" fallback when normal file read fails due to permissions.
+# Supports patterns with %d as a PID placeholder (e.g., /proc/%d/io
+# matches /proc/1234/io).
```
dbwiddis: Suggest we use [Java-standard glob patterns here](https://docs.oracle.com/javase/7/docs/api/java/nio/file/FileSystem.html#getPathMatcher(java.lang.String)).  They are designed for exactly this use case, well documented , and save you the headache of parsing it all yourself (and documenting it)

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,317 @@
...
+            return false;
+        }
+        String cmdPath = tokens[0];
+
+        // Extract bare command name from path
+        String cmdName = cmdPath;
+        int lastSlash = cmdPath.lastIndexOf('/');
+        if (lastSlash >= 0 && lastSlash < cmdPath.length() - 1) {
+            cmdName = cmdPath.substring(lastSlash + 1);
+        }
```
dbwiddis, line 91: Let's add a method to FileUtil that does this using the correct OS-dependent file separator
```java
Path p = Paths.get(<filename string, with / on *nix and \ on windows>);
String file = p.getFileName().toString();
```
And apply it [here](https://github.com/search?q=repo%3Aoshi%2Foshi%20getfilename&type=code) as well.

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,317 @@
...
+            // Handle %d placeholder for PID matching
+            if (allowed.contains("%d")) {
+                // Convert pattern like /proc/%d/io to regex /proc/\d+/io
+                String regex = Pattern.quote(allowed).replace("%d", "\\E\\d+\\Q");
+                // Clean up empty quote sections
+                regex = regex.replace("\\Q\\E", "");
+                if (filePath.matches(regex)) {
+                    return true;
+                }
+            }
```
dbwiddis: As mentioned on `oshi.properties` let's use built-in glob matching patterns.

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,317 @@
...
+     *
+     * @param filePath The file to read
+     * @return A list of Strings representing each line of the file, or empty list if unreadable
+     */
+    public static List<String> readFilePrivileged(String filePath) {
+        // First attempt normal read (silent failure)
+        List<String> result = FileUtil.readFile(filePath, false);
+        if (!result.isEmpty()) {
+            return result;
+        }
```
dbwiddis: In the `ExecutingCommand.runPrivilegedNative()` implementation, you check whether the user is elevated or not.  You don't do that here, wasting an attempt at a sudo cat every time; and the only way you actually get this empty string is due to exception handling (which is generally "expensive" computationally).

You should do the same `isElevated()` check before this line and just delegate to the existing read (this code)... basically try to mirror the `ExecutingCommand` implementation.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -138,8 +138,8 @@ public Map<String, String> getEnvironmentVariables() {
     }
 
     private Map<String, String> queryEnvironmentVariables() {
-        return Collections.unmodifiableMap(ParseUtil.parseByteArrayToStringMap(FileUtil
-                .readAllBytes(String.format(Locale.ROOT, ProcPath.PID_ENVIRON, getProcessID()), LOG_PROCFS_WARNING)));
+        return Collections.unmodifiableMap(ParseUtil.parseByteArrayToStringMap(PrivilegedUtil
+                .readAllBytesPrivileged(String.format(Locale.ROOT, ProcPath.PID_ENVIRON, getProcessID()))));
```
dbwiddis: You dropped `LOG_PROCFS_WARNING`

oshi-core/src/main/java/oshi/driver/linux/Lshw.java
```diff
@@ -26,19 +25,17 @@ private Lshw() {
...
 
-            for (String checkLine : ExecutingCommand.runNative("lshw -C system")) {
-                if (checkLine.contains(modelMarker)) {
-                    model = checkLine.split(modelMarker)[1].trim();
-                } else if (checkLine.contains(serialMarker)) {
-                    serial = checkLine.split(serialMarker)[1].trim();
-                } else if (checkLine.contains(uuidMarker)) {
-                    uuid = checkLine.split(uuidMarker)[1].trim();
-                }
+        for (String checkLine : ExecutingCommand.runPrivilegedNative("lshw -C system")) {
```
dbwiddis, line 32: This is good but you also do `lshw` calls for:
 - LinuxGraphicsCard.java:123 ExecutingCommand.runNative("lshw -C display")` (possibly move that implementation to this file)
 - Lshw.java:80 `ExecutingCommand.runNative("lshw -class processor")` (in this file)

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,317 @@
...
+    public static java.util.Map<String, String> getKeyValueMapFromFilePrivileged(String filePath, String separator) {
+        java.util.Map<String, String> map = new java.util.HashMap<>();
+        List<String> lines = readFilePrivileged(filePath);
+        for (String line : lines) {
+            String[] parts = line.split(separator);
+            if (parts.length == 2) {
+                map.put(parts[0], parts[1].trim());
+            }
+        }
+        return map;
```
dbwiddis: This essentially duplicates code in `FileUtil.getKeyValueMapFromFile()`.  Should probably combine the common code into a ParseUtil helper method taking the string list argument and returning a map.  We already have `parseByteArrayToStringMap()`, `parseCharArrayToStringMap()` and `stringToEnumMap()` so why not another similar method?

We already do this in `ProcUtil.parseStatistics()` which could also benefit from refactoring to that helper method.

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,317 @@
...
+        }
+
+        // Execute sudo cat using array to handle paths with spaces
+        List<String> cmdList = new ArrayList<>();
+        cmdList.addAll(Arrays.asList(prefix.split("\\s+")));
+        cmdList.add("cat");
+        cmdList.add(filePath);
+        String[] cmdArray = cmdList.toArray(new String[0]);
+        LOG.debug("Attempting privileged file read: {}", Arrays.toString(cmdArray));
+        List<String> lines = ExecutingCommand.runNative(cmdArray);
```
dbwiddis: I don't think we can use `runNative()` for binary file reading.  It loses newlines when creating the `List<String>` and doesn't document whether it was `\n\r` or `\n` which you manually add back.   I think you need to use your own parsing of the return of `cat` to collect all the bytes.

You may need to spawn your own process here to execute the command and directly read the byte stream.

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,317 @@
...
+
+    /**
+     * Reads a file and returns a map of string keys to string values, with privileged fallback.
+     *
+     * @param filePath  The file to read
+     * @param separator Character(s) in each line of the file that separate the key and the value
+     * @return The map contained in the file, delimited by the separator, with the value whitespace trimmed
+     */
+    public static java.util.Map<String, String> getKeyValueMapFromFilePrivileged(String filePath, String separator) {
+        java.util.Map<String, String> map = new java.util.HashMap<>();
```
dbwiddis: Make this an import. :)

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,317 @@
...
+public final class PrivilegedUtil {
+
+    private static final Logger LOG = LoggerFactory.getLogger(PrivilegedUtil.class);
+
+    private static volatile Set<String> cachedCommandAllowlist;
+    private static volatile Set<String> cachedFileAllowlist;
+    private static volatile String lastCommandAllowlistConfig = "";
+    private static volatile String lastFileAllowlistConfig = "";
+
+    private PrivilegedUtil() {
```
dbwiddis, line 50: This should probably be `PrivilegedFileUtil`?

oshi-core/src/test/java/oshi/util/ExecutingCommandTest.java
```diff
@@ -37,4 +46,56 @@ void testRunNative() {
         assertThat("bad command", ExecutingCommand.runNative(BAD_COMMAND), is(empty()));
         assertThat("bad command first answer", ExecutingCommand.getFirstAnswer(BAD_COMMAND), is(emptyString()));
     }
+
+    @Test
+    void testRunPrivilegedNativeNoConfig() {
+        // Without config, runPrivilegedNative should return empty list when not elevated
+        // (unless already running as root, which we can't control in tests)
+        if (!UserGroupInfo.isElevated()) {
```
dbwiddis: Prefer Junit 5 `@DisabledIf` rather than this conditional (all the new tests)

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
rohan-coder02: For `cat` in allowlist, I think the file allowlist is to check for file reads. 

Requiring cat in command allowlist would be redundant, hence, explicitly adding it in function `readAllBytesPrivileged` when doing the file read
```
cmdList.add("cat");
```

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
rohan-coder02: PrivilegedUtil looks fine I think since it handles both privileged commands (runPrivilegedNative) and privileged file operations

oshi-core/src/main/java/oshi/hardware/common/AbstractGlobalMemory.java
```diff
@@ -26,7 +26,7 @@ public List<PhysicalMemory> getPhysicalMemory() {
         // dmidecode requires sudo permission but is the only option on Linux
         // and Unix
         List<PhysicalMemory> pmList = new ArrayList<>();
-        List<String> dmi = ExecutingCommand.runNative("dmidecode --type 17");
+        List<String> dmi = ExecutingCommand.runPrivilegedNative("dmidecode --type 17");
```
dbwiddis, line 29: ~Since we're only implementing on Linux with this PR, this abstract superclass could break on the BSD implementations.~

~Possible solution:~
 - ~refactor the parsing to a `protected` helper method~
 - ~in the abstract class call with `runNative()` as it currently is~
 - ~in the Linux subclass, call with `runPrivilegedNative()` and call the new superclass helper method.~

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    /**
+     * Gets the configured sudo command prefix.
+     *
+     * @return The prefix string, or empty string if not configured
+     */
+    public static String getPrefix() {
+        return GlobalConfig.get(OSHI_OS_LINUX_PRIVILEGED_PREFIX, "");
+    }
```
dbwiddis, line 191: This is an alternate solution for fixing the BSD `AbstractGlobalMemory` issue and probably a better one.  We should do some sort of OS version check here and just not have any prefix for anything other than Linux.

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,313 @@
...
+        Set<String> fileAllowlist = getFileAllowlist();
+        if (!isFileAllowed(filePath, fileAllowlist)) {
+            return FileUtil.readFile(filePath, false);
+        }
+
+        // First attempt normal read (silent failure)
+        List<String> result = FileUtil.readFile(filePath, false);
+        if (!result.isEmpty()) {
+            return result;
+        }
```
dbwiddis: By the time we get here, we've checked:
 - not elevated, otherwise we'd just use readFile as root
 - file is in allowlist, otherwise we'd just use readFile as root (could be combined with previous conditional)

So at this point we should not do a normal read: we know we're not root, and the file is in the allowlist.   

An empty string here is not a valid test of getting real information, for example you'd get a one-string list here:
```
$ cat /proc/1/io
cat: /proc/1/io: Permission denied
```

So these lines should probably be removed entirely.

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
```diff
@@ -138,8 +138,8 @@ public Map<String, String> getEnvironmentVariables() {
     }
 
     private Map<String, String> queryEnvironmentVariables() {
-        return Collections.unmodifiableMap(ParseUtil.parseByteArrayToStringMap(FileUtil
-                .readAllBytes(String.format(Locale.ROOT, ProcPath.PID_ENVIRON, getProcessID()), LOG_PROCFS_WARNING)));
+        return Collections.unmodifiableMap(ParseUtil.parseByteArrayToStringMap(PrivilegedUtil
+                .readAllBytesPrivileged(String.format(Locale.ROOT, ProcPath.PID_ENVIRON, getProcessID()), LOG_PROCFS_WARNING)));
     }
```
dbwiddis, line 144: Looking at this again, I'm wondering if we should exempt `/proc/<pid>/environ` from the file reading by cat.   Unlike most of the other paths, this path in particular exposes potential security-sensitive information such as access keys or session tokens.   

In order to use the privileged `read*()` methods, the server admin must add `cat` to the allowed commands.   I'm not sure we have a use case where a non-root user should have this level of access, even if it's OSHI.  In particular, if this privileged user is providing API responses to anyone else, we probably shouldn't show them this.

oshi-core/src/main/resources/oshi.properties
```diff
@@ -288,3 +288,45 @@ oshi.os.linux.sensors.hwmon.names=coretemp,k10temp,zenpower,k8temp,via-cputemp
...
+#
+# Comma-separated list of file paths eligible for privileged read via
+# "sudo cat" fallback when normal file read fails due to permissions.
+# Supports Java glob patterns:
+#   * matches any sequence of characters within a single path segment
+#   ? matches any single character
+#   [abc] matches any character in the brackets
+# Example: /proc/*/io matches /proc/1234/io, /proc/1/io, etc.
+# Example paths:
+#   /proc/*/io, /proc/*/environ, /sys/devices/virtual/dmi/id/product_serial
```
dbwiddis: Similar to the comments on `getEnvironmentVariables()` let's remove this path as an example.

oshi-core/src/main/resources/oshi.properties
```diff
@@ -288,3 +288,45 @@ oshi.os.linux.sensors.hwmon.names=coretemp,k10temp,zenpower,k8temp,via-cputemp
...
+# commands without running the entire application as root.
+#
+# SECURITY CONSIDERATIONS:
+# - The prefix is NOT applied when already running as root (uid=0)
+# - Only commands explicitly listed in the allowlist will use the prefix
+# - Users must configure passwordless sudo for the specified commands
+# - For production use, restrict sudo access to specific commands only
+#
+# Example sudoers configuration (run: sudo visudo):
+#   oshiuser ALL=(ALL) NOPASSWD: /usr/sbin/dmidecode, /usr/bin/lshw, /usr/bin/lsblk
```
dbwiddis: we actually don't need sudo for `lsblk`...

oshi-core/src/test/java/oshi/util/ExecutingCommandTest.java
```diff
@@ -9,24 +9,38 @@
...
     private static final String BAD_COMMAND = "noOSshouldHaveACommandNamedThis";
 
+    static boolean isElevated() {
+        return UserGroupInfo.isElevated();
+    }
+
+    @BeforeEach
+    void setUp() {
+        GlobalConfig.clear();
+    }
```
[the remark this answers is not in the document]
dbwiddis: Suggested fix: add a package-private method in `PrivilegedUtil` to re-initialize the memorizers:
```java
static void clearCaches() {
    commandAllowlist = memoize(PrivilegedUtil::loadCommandAllowlist, defaultExpiration());
    fileAllowlist = memoize(PrivilegedUtil::loadFileAllowlist, defaultExpiration());
}
```
Then in the tests, just reset in sequence after clearing the caches:
```java
@BeforeEach
void setUp() {
    GlobalConfig.clear();
    PrivilegedUtil.clearCaches();
}
```

oshi-core/src/main/java/oshi/util/PrivilegedUtil.java
```diff
@@ -0,0 +1,313 @@
...
+        try {
+            p = Runtime.getRuntime().exec(cmdArray);
+            try (java.io.InputStream is = p.getInputStream();
+                 java.io.ByteArrayOutputStream baos = new java.io.ByteArrayOutputStream()) {
+                byte[] buffer = new byte[8192];
+                int bytesRead;
+                while ((bytesRead = is.read(buffer)) != -1) {
+                    baos.write(buffer, 0, bytesRead);
+                }
+                return baos.toByteArray();
```
dbwiddis: I think you need to `flush()` before returning.

https://github.com/oshi/oshi/blob/cbfba7a2e05a8fa312aa4f40e2e670aa12fa6ec0/oshi-core/src/main/java/oshi/util/FileUtil.java#L472

Another [example here](https://github.com/openjdk/jol/blob/db2cc744485031c8149991a004ab02f2ea9b24dd/jol-core/src/main/java/org/openjdk/jol/util/IOUtils.java#L78)

oshi-core/src/main/java/oshi/software/os/linux/LinuxOSProcess.java
dbwiddis: After thinking about this, we do need this proc path explicitly configured for this to actually work.  

So really I think it boils down to calling out this risk in the config file but leaving the ultimate decision up to users.

PR #3158, opened 2026-04-14
Fix getOpenFileDescriptors() to use _Total PDH counter instance

rohan-coder02:
Fix PdhUtilFFM.getOpenFileDescriptors() to query the pre-aggregated _Total PDH counter instance instead of wildcard.

Aligns FFM implementation with the working JNA version for better efficiency.

Removes loop iteration and directly returns the single handle count value.

## Summary by CodeRabbit

* **Refactor**
  * Improved Windows handle-count retrieval by switching to a direct aggregate counter query, simplifying logic and reducing overhead.
  * No public API signatures changed.
  * Existing error behavior (returning an error indicator on failure) is preserved.

* **Fix**
  * When merging filesystem entries on Windows, prefer the name from system/WMI data so store names are more accurate.

* **Chore**
  * Internal native binding adjustments to support the new retrieval approach.

### DIFF
--- a/oshi-core-java25/src/main/java/oshi/util/platform/windows/PdhUtilFFM.java
+++ b/oshi-core-java25/src/main/java/oshi/util/platform/windows/PdhUtilFFM.java
@@ -6,15 +6,13 @@ package oshi.util.platform.windows;
 
 import static java.lang.foreign.MemoryLayout.PathElement.groupElement;
 import static java.lang.foreign.ValueLayout.ADDRESS;
-import static java.lang.foreign.ValueLayout.JAVA_INT;
 import static java.lang.foreign.ValueLayout.JAVA_LONG;
-import static oshi.ffm.windows.PdhFFM.PDH_FMT_COUNTERVALUE_ITEM_LAYOUT;
+import static oshi.ffm.windows.PdhFFM.PDH_FMT_COUNTERVALUE_LAYOUT;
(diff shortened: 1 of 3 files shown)


dbwiddis:
This is still failing the JNA == FFM test because the actual counter is `ProcessInformation` not `Process`.  That was originally linked in the JNA code by the class name; extracting all those enums into a separate class lost that context.

oshi-core-java25/src/main/java/oshi/util/platform/windows/PdhUtilFFM.java
```diff
@@ -45,7 +43,7 @@ public static long getOpenFileDescriptors() {
 
             try {
                 MemorySegment counterPtr = arena.allocate(ADDRESS);
-                checkSuccess(PdhAddEnglishCounter(query, toWideString(arena, "\\Process(*)\\Handle Count"),
+                checkSuccess(PdhAddEnglishCounter(query, toWideString(arena, "\\Process(_Total)\\Handle Count"),
```
dbwiddis, line 42:
```suggestion
                checkSuccess(PdhAddEnglishCounter(query, toWideString(arena, "\\ProcessInformation(_Total)\\Handle Count"),
```

dbwiddis: Nevermind, I think I'm confused here...

Issue #3224, opened 2026-05-03
Add cgroup resource limits API for container-aware applications


dbwiddis:
## Summary

Add a `CgroupInfo` API to oshi-common that exposes container resource limits and usage from the Linux cgroup (control groups) filesystem. This gives users explicit access to container-scoped CPU, memory, and PID constraints alongside OSHI's existing host-level APIs.

### What are cgroups?

[Control groups (cgroups)](https://man7.org/linux/man-pages/man7/cgroups.7.html) are a Linux kernel feature for organizing processes into hierarchical groups and applying resource limits. They are the foundational mechanism behind virtually all Linux container runtimes: Docker, Kubernetes (containerd, CRI-O), Podman, LXC, and systemd-nspawn all use cgroups to enforce CPU, memory, and I/O constraints.

cgroups expose resource limits and usage counters as plain text files under `/sys/fs/cgroup/`. No native calls are required to read them, making this a natural fit for oshi-common with zero native dependencies.

### cgroup v2 (unified hierarchy)

cgroup v2 ([kernel docs](https://docs.kernel.org/admin-guide/cgroup-v2.html)) is the current standard, using a single unified hierarchy mounted at `/sys/fs/cgroup/`. It is the default in:
- Ubuntu 21.10+, Fedora 31+, Debian 11+, RHEL 9+, Arch Linux
- Docker Engine 20.10+ (with systemd cgroup driver)
- Kubernetes 1.25+ (graduated to stable)
- containerd 1.4+, Podman 3.0+

Key files (relative to the cgroup path):
- `cpu.max` (quota and period on one line, e.g., `200000 100000` or `max 100000`)
- `cpu.stat` (usage_usec, user_usec, system_usec, etc.)
- `memory.max` (bytes or `max` for unlimited)
- `memory.current` (current usage in bytes)
- `memory.stat` (detailed breakdown)
- `pids.max` (limit or `max`)
- `pids.current` (current count)
- `io.stat` (per-device I/O counters)

### cgroup v1 (legacy)

cgroup v1 uses separate hierarchies per resource controller, mounted at paths like `/sys/fs/cgroup/cpu/`, `/sys/fs/cgroup/memory/`, etc. It is still present on older distributions and some container platforms. The file names and formats differ from v2 (e.g., `cpu.cfs_quota_us` and `cpu.cfs_period_us` instead of `cpu.max`).

[...]

PR #3225, opened 2026-05-03
Add CgroupInfo API for container resource monitoring

rohan-coder02:
## Summary

Implements a new `CgroupInfo` interface to expose cgroup resource limits and usage metrics, enabling OSHI users to detect containerized environments and retrieve resource constraints.

Relates to #3224

## Usage Examples

```java
CgroupInfo cgroup = os.getCgroupInfo();

// Check if running in a container
if (cgroup.isContainerized()) {
    System.out.println("Running in cgroup v" + cgroup.getVersion());
    
    // CPU limits
    double effectiveCpus = cgroup.getEffectiveCpus();
    if (effectiveCpus > 0) {
        System.out.println("CPU limit: " + effectiveCpus + " cores");
    }
    
    // Memory limits
    long memLimit = cgroup.getMemoryLimit();
    if (memLimit < CgroupInfo.UNLIMITED_MEMORY) {
        System.out.println("Memory limit: " + memLimit / (1024 * 1024) + " MB");
    }
    System.out.println("Memory usage: " + cgroup.getMemoryUsage() / (1024 * 1024) + " MB");
    
    // PID limits
    long pidLimit = cgroup.getPidLimit();
    if (pidLimit != CgroupInfo.UNLIMITED) {
        System.out.println("PID limit: " + pidLimit);
    }
    System.out.println("Current PIDs: " + cgroup.getPidCurrent());
}
```

### `CgroupInfo` interface

- All methods have `default` implementations returning sentinel values, so non-Linux platforms get correct no-op behavior without a separate class
- Named constants for sentinel values: `UNLIMITED` (-1), `DEFAULT_CPU_PERIOD` (100000), `UNLIMITED_MEMORY` (Long.MAX_VALUE), `UNLIMITED_CPUS` (-1.0d)
- `getEffectiveCpus()` is a default method computed from `getCpuQuota() / getCpuPeriod()`
- `OperatingSystem.getCgroupInfo()` returns an anonymous instance using the defaults

### `LinuxCgroupInfo` implementation

- **Version detection**: Reads `/proc/filesystems` for cgroup2 support
- **v2 support** (primary): Parses `cpu.max`, `cpu.stat`, `memory.max`, `memory.current`, `pids.max`, `pids.current`
- **v1 fallback**: Parses controller-specific files (`cpu.cfs_quota_us`, `cpuacct.usage`, etc.)
- **Memoization**: Version and cgroup path cached permanently; limit values (quota, period, memory, PIDs) use `defaultExpiration()` TTL since they can be reconfigured at runtime
- **Testability**: V2 read methods have package-private overloads accepting a base path, enabling unit tests with `@TempDir`

### Other changes

- Version bumped to 7.1.0-SNAPSHOT (new public API)
- `ProcPath.java` — added `SELF_CGROUP` and `FILESYSTEMS` constants
- `SysPath.java` — added `CGROUP` constant
- `LinuxOperatingSystem.java` — overrides `getCgroupInfo()` with memoized instance

## Summary by CodeRabbit

* **New Features**
  * Added CgroupInfo API and Linux cgroup support (v1/v2): reports container status, cgroup version, CPU quota/period/usage, effective CPUs, memory limit/usage, and PID limits/current. OS integration surfaces cgroup info in system output.

* **Tests**
  * Added unit tests and gated container integration tests covering v1/v2, edge cases, and defaults.

* **Chores**
  * Bumped project version to 7.1.0, updated changelog, and added CI job for cgroup integration tests.


dbwiddis:
The lint failure is because we added to the API which requires a minor version bump to 7.1.0 :)

dbwiddis:
Implementation looks good. We need some tests!

oshi-common/src/main/java/oshi/software/common/os/linux/LinuxCgroupInfo.java
```diff
@@ -0,0 +1,366 @@
...
+
+    private static final String[] CONTAINER_MARKERS = { "/docker/", "/kubepods/", "/lxc/", "/containerd/", "/crio-",
+            "/buildkit/" };
+
+    private final Supplier<Integer> versionSupplier = Memoizer.memoize(this::detectVersion);
+    private final Supplier<String> cgroupPathSupplier = Memoizer.memoize(this::parseCgroupPath);
+    private final Supplier<Long> cpuQuotaSupplier = Memoizer.memoize(this::readCpuQuota);
+    private final Supplier<Long> cpuPeriodSupplier = Memoizer.memoize(this::readCpuPeriod);
+    private final Supplier<Long> memoryLimitSupplier = Memoizer.memoize(this::readMemoryLimit);
+    private final Supplier<Long> pidLimitSupplier = Memoizer.memoize(this::readPidLimit);
```
dbwiddis: While the version and path won't change, I think it's possible to change some of the other parameters during the container's operation.  So these need some sort of expiration.

oshi-common/src/main/java/oshi/software/common/os/linux/LinuxCgroupInfo.java
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    private long readCpuPeriod() {
+        int version = getVersion();
+        if (version == 2) {
+            return readCpuPeriodV2();
+        } else if (version == 1) {
+            return readCpuPeriodV1();
+        }
+        return 100000L;
```
dbwiddis: We're returning this magic number a lot. We should make it a constant (`DEFAULT_CPU_PERIOD` perhaps?) and we could reference that constant in the interface javadoc.

Similar comment applies to the other sentinel values (-1L, -1d) that are not self explanatory (0, empty string, max value).

oshi-common/src/main/java/oshi/software/common/os/linux/LinuxCgroupInfo.java
```diff
@@ -0,0 +1,366 @@
...
+        }
+        String[] parts = cpuMax.split("\\s+");
+        if (parts.length >= 1) {
+            if ("max".equalsIgnoreCase(parts[0])) {
+                return -1L;
+            }
+            return ParseUtil.parseLongOrDefault(parts[0], -1L);
+        }
+        return -1L;
+    }
```
dbwiddis, line 257: Let's make all the methods like these testable by having the no-arg  version delegate to a package-private version taking a string path value that we can then test.

So here we would have 
```java
    private long readCpuQuotaV2() {
        return readCpuQuotaV2(getV2CGroupBase());
    }
```
and then
```java
    // package private for testing
    long readCpuQuotaV2(String basePath) {
        // the existing method
    }
```

Then you could test them by passing strings from test files.

See the pattern here:
https://github.com/oshi/oshi/blob/459e2242eb672cbf6a492547022105249ccc504d/oshi-common/src/main/java/oshi/hardware/common/platform/linux/LinuxCentralProcessor.java#L505-L509

Tested like this:
https://github.com/oshi/oshi/blob/459e2242eb672cbf6a492547022105249ccc504d/oshi-common/src/test/java/oshi/hardware/common/platform/linux/LinuxCentralProcessorTest.java#L145-L158

oshi-common/src/main/java/oshi/software/common/NoOpCgroupInfo.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public long getCpuQuota() {
+        return -1L;
+    }
+
+    @Override
+    public long getCpuPeriod() {
+        return 100000L;
```
dbwiddis: This class might be a great place to define these constants.

Alternately instead of `NoOpCgroupInfo` could call it `DefaultCgroupInfo` or just `AbstractCgroupInfo` and inherit from it and then your default return values could be `super.getCpuPeriod()`.

Alternately define the sentinel values on the interface as `default` and skip this class entirely.

PR #3246, opened 2026-05-11
Update UPGRADING.md with additional dependency

Friedrich2:
The described issue came up during regular after-library-upgrade-tests within my project: The Pair class was not more accessible, therefore I had to additionally import the oshi-commons JAR file.

## Summary by CodeRabbit

* **Documentation**
  * Usage and upgrade docs updated with guidance for manual JAR setups: a companion runtime library is required alongside the core implementation for recent versions, and a minimum native-access library version is needed to avoid runtime linkage issues. README now covers recommended dependency manager usage plus instructions for downloading distribution ZIPs (see upgrade notes for details).

[Review Change Stack](https://app.coderabbit.ai/change-stack/oshi/oshi/pull/3246)

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,3 +1,23 @@
+# Project Dependencies
+
+Dependency management tools such as Maven or Gradle automatically handle transitive dependencies. If you manage JAR files manually, the following notes may help.
+
+## JNA version coupling
+
+OSHI's `oshi-core` module is tightly coupled with [JNA](https://github.com/java-native-access/jna) and requires at least the version specified in the POM. Using an older JNA version may cause `NoClassDefFoundError` or `NoSuchMethodError` at runtime due to missing classes or methods. See the [FAQ](FAQ.md#how-do-i-resolve-jna-noclassdeffounderror-or-nosuchmethoderror-issues) for troubleshooting.
+
+## Module split (6.12.0+)
+
+Starting with version 6.12.0, OSHI was split from a single `oshi-core` artifact into multiple modules:
+
+- **`oshi-common`** — shared API interfaces and utilities that do not require JNA or FFM.
+- **`oshi-core`** — JNA-based implementation (depends on `oshi-common`).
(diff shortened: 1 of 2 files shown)


UPGRADING.md
```diff
@@ -76,6 +76,10 @@ The JPMS module name for the FFM module has changed from `com.github.oshi` to `c
 requires com.github.oshi.ffm;
 ```
 
+## Additional need for oshi-commons
+
+If you do not use Maven or similar tools but the raw JAR files, this subject is for you. Since Version 7, oshi-commons is a new dependency of oshi-core. Therefore, in addition to the oshi-core JAR file, you will need the JAR file for oshi-commons in the classpath / modulepath. Otherwise, some classes may not be accessible, since e.g. the subpackage oshi.tuples (used e.g. for the Pair class) will not be found during import.
```
mprins:
```suggestion
If you do not use Maven or similar tools but the raw JAR files, this subject is for you. Since Version 7, `oshi-common` is a new dependency of `oshi-core`. Therefore, in addition to the `oshi-core` JAR file, you will need the JAR file for `oshi-common` in the classpath / modulepath. Otherwise, some classes may not be accessible, since e.g. the subpackage oshi.tuples (used e.g. for the Pair class) will not be found during import.
```

UPGRADING.md
```diff
@@ -76,6 +76,10 @@ The JPMS module name for the FFM module has changed from `com.github.oshi` to `c
 requires com.github.oshi.ffm;
 ```
 
+## Additional need for oshi-commons
```
mprins:
```suggestion
## Additional need for oshi-common
```

mprins (changes requested):

dbwiddis:
Hey @Friedrich2 thanks for calling out this point of confusion!

The module split actually happened in 6.12.  I've expanded your submission to also mention the tight version coupling with JNA and make it more clear in both UPGRADING and README files. 

Let me know if you think this addresses your concern.

mprins (approved):

Friedrich2:
Thank you for your fast and comprehensive solution! This addresses my concern completely.

PR #3283, opened 2026-05-23
Fix macOS JNA CPU tick overflow

dbwiddis:
## Summary
- Treat macOS JNA host CPU load ticks as unsigned 32-bit counters
- Prevent system CPU tick values from becoming negative after long uptimes
- Add changelog entry for the fix

## Validation
- ./mvnw spotless:apply
- ./mvnw -pl oshi-metrics -Dtest=oshi.metrics.OshiMetricsTest#cpuTimeRegistered test -B

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue on macOS where system CPU statistics could overflow to negative values after extended system uptime.

[Review Change Stack](https://app.coderabbit.ai/change-stack/oshi/oshi/pull/3283?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


dbwiddis:
Fun fact: this bug introduced in #27 as one of my first substantive PRs to the project, although in that case it was doing diff math so wouldn't have broken except at the rollover point.  Truly became visible in #51 ... 

Only really found this by running tests on my personal machine with 63 days of up time.

PR #3409, opened 2026-06-25
Fix AIX inode collection: use correct df format and include NFS mounts

jank:
## Problem

\`AixFileSystem.getFileStoreMatching()\` has two bugs in its inode-data collection step that affect AIX systems in production:

**1. Wrong \`df\` command.** The code ran \`df -i\`, which is a Linux/POSIX flag that does not exist on AIX. The AIX-native equivalent is \`df -F %n %l\` (format specifiers: \`%n\` = free file nodes / Ifree, \`%l\` = used file nodes / Iused). On real AIX hosts, \`df -i\` either fails or produces no parseable inode output, causing all file stores to report 0 inodes.

Note: the format string must **not** be quoted. \`df -F "%n %l"\` (with shell quotes) causes AIX to treat the entire string as a single specifier, silently dropping \`%l\` and outputting only the Ifree column.

**2. NFS mounts silently dropped.** The inode-data loop filtered with \`line.startsWith("/")\`. On AIX, NFS entries in \`df\` output appear as \`hostname:/path\` or \`192.168.x.x:/path\`, which do not start with \`/\`. All NFS file stores therefore reported 0 inodes regardless.

Both issues were observed on AIX 7.x production systems.

## Fix

- Switch to \`df -F %n %l\` (AIX-native inode format, unquoted); use \`-T local\` instead of \`-l\` for local-only filtering.
- Replace \`line.startsWith("/")\` with a regex \`^(?:[\w.]+:)?/\` that matches both local device paths (\`/dev/hd4\`) and NFS \`host:/path\` entries.
- Fix column parsing: \`df -F %n %l\` always emits four columns (Filesystem, 512-blocks, Ifree, Iused); parse Ifree as \`split[length-2]\` and Iused as \`split[length-1]\`, and tighten the guard to \`>= 4\`.
- Correct a free/used swap: the previous relative-index code assigned Ifree to \`used\` and Iused to \`free\`, so \`inodeFreeMap\` was populated with the used count.
- Update the sample output comment to reflect actual \`df -F %n %l\` output from an AIX 7.2 host.

## Testing

AIX is not in CI. Manual verification on AIX 7.2 confirms that:
- Local file stores now show correct non-zero inode counts.
- \`/proc\` (which outputs \`-\` for all columns) is handled gracefully via \`ParseUtil.parseLongOrDefault\` returning 0L.
- NFS-mounted file stores are no longer silently skipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filesystem inode reporting on AIX by enhancing how filesystem/mount output is detected and matched, leading to more accurate inode statistics.
  * Fixed inode free/used/total calculation for matched filesystems, improving the correctness of disk usage details shown in system reports.

* **Documentation**
  * Updated the 7.3.2 changelog with a dedicated entry describing the AIX inode reporting fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # 7.3.2 (in progress)
 
-* Your contribution here!
+##### Bug Fixes and Improvements
+
+* [#3409](https://github.com/oshi/oshi/pull/3409): Fix AIX `getFileStores()` inode collection: switch to AIX-native `df -F %n %l`, fix NFS mount filtering, correct column parsing, and fix free/used inode swap - [@jank](https://github.com/jank).
 
 # 7.3.0 (2026-06-06), 7.3.1 (2026-06-11)
 
--- a/oshi-common/src/main/java/oshi/software/common/os/unix/aix/AixFileSystem.java
+++ b/oshi-common/src/main/java/oshi/software/common/os/unix/aix/AixFileSystem.java
@@ -10,6 +10,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 import oshi.annotation.concurrent.ThreadSafe;
 import oshi.software.common.AbstractFileSystem;
@@ -40,6 +41,8 @@ public class AixFileSystem extends AbstractFileSystem {
     private static final List<PathMatcher> FS_VOLUME_INCLUDES = FileSystemUtil
             .loadAndParseFileSystemConfig(OSHI_AIX_FS_VOLUME_INCLUDES);
 
+    private static final Pattern FS_PATTERN = Pattern.compile("^(?:[\\w.]+:)?/");
+
     @Override
     public List<OSFileStore> getFileStores(boolean localOnly) {
         return getFileStoreMatching(null, localOnly);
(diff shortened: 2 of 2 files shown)


jank:
I will do another round of review on AIX regarding parsing of the df results and post my findings here.
AIX behavior confirmed on test system. `df -F "%n %l"` silently swallows the 2nd formatting option. Because the 512-blocks column is always shown this got misinterpreted.
Correction is to us `df -F %n %l`.

dbwiddis:
Thanks for the fix!

> AIX is not in CI. Manual verification on AIX 7.2 ...

We do test AIX in the cfarm GitHub workflow, although that only runs on push to master. 

Still, it represents a missing test or incorrect assumption.

dbwiddis:
> **1. Wrong `df` command.** The code ran `df -i`, which is a Linux/POSIX flag that does not exist on AIX. The AIX-native equivalent is `df -F %n %l` (format specifiers: `%n` = free file nodes / Ifree, `%l` = used file nodes / Iused). On real AIX hosts, `df -i` either fails or produces no parseable inode output, causing all file stores to report 0 inodes.

This may be true on AIX 7.1 and 7.2.  The only AIX access I have is the GCC Compile Farm machine which runs AIX 7.3:

```
$ ssh aix
Last login: Wed Jun 17 04:41:16 CUT 2026 on ssh from 20.172.37.236
*******************************************************************************
*                                                                             *
*                                                                             *
*  Welcome to AIX Version 7.3!                                                *
*                                                                             *
*                                                                             *
*  Please see the README file in /usr/lpp/bos for information pertinent to    *
*  this release of the AIX Operating System.                                  *
*                                                                             *
*                                                                             *
*  Add /opt/freeware/bin to $PATH for additional FOSS software.               *
*                                                                             *
*                                                                             *
*******************************************************************************

$ df -i
Filesystem    512-blocks      Free %Used    Iused %Iused Mounted on
/dev/hd4         8388608   7200008   15%     3279     1% /
/dev/hd2         6160384    523360   92%    45819    36% /usr
/dev/hd9var      5767168   5319792    8%     1301     1% /var
/dev/hd3         5767168   5605376    3%     6544     2% /tmp
/dev/hd1          131072    130304    1%        9     1% /homehd1
/dev/hd11admin     262144    261384    1%        5     1% /admin
/proc                  -         -    -        -      - /proc
/dev/hd10opt    11796480   1861416   85%    35223    12% /opt
/dev/livedump     524288    523552    1%        4     1% /var/adm/ras/livedump
/dev/hd01      880803840 118251768   87%  7816806    30% /home
/ahafs                 -         -    -       35     1% /aha
/dev/ramdisk0  134217728  56321920   59%   739489    10% /scratch
```
The proposed replacement seems to produce the appropriate output:
```
$ df -F %n %l
Filesystem    512-blocks     Ifree    Iused
/dev/hd4         8388608   839601     3279 
/dev/hd2         6160384    81996    45819 
/dev/hd9var      5767168   594103     1301 
/dev/hd3         5767168   634579     6544 
/dev/hd1          131072    14565        9 
/dev/hd11admin     262144    29133        5 
/proc                  -        -        - 
/dev/hd10opt    11796480   260337    35223 
/dev/livedump     524288    58200        4 
/dev/hd01      880803840 18662559  7816806 
/ahafs                 -    32733       35 
/dev/ramdisk0  134217728  7470790   739489
```

Given this block of code is only getting the inode values, it's a reasonable compatibility adjustment: please add a note in a comment about the choice to use this command line for AIX 7.1-7.2 compatibility.

dbwiddis (approved):

PR #3423, opened 2026-07-01
Add NVD cache warmer and make the site build reuse a warm cache

dbwiddis:
## Problem

The Site (release) build is the project's main OWASP dependency-check / NVD consumer, but it runs at most every few weeks. That cadence is longer than **both**:

- GitHub's **7-day cache-eviction** window (caches unused for 7 days are dropped), and
- dependency-check's **7-day incremental-update** window (`nvdValidForHours`).

So nearly every release hits a **cold, full NVD download** (~362k records) that is fully exposed to NVD API throttling. The latest release's Site run failed 6–7 times in a row; one attempt stalled mid-download and was killed at GitHub's hard **6-hour job cap** (`The operation was canceled`). The API key is present and correctly wired via `.mvn/settings.xml` — the failure is the cold download, not auth.

## Changes

1. **New `NVD Cache Warmer` workflow** (`nvd-cache.yaml`), scheduled every 3 days:
   - Keeps the NVD database warm. Since **default-branch caches are readable from every branch**, the cache it builds on `master` is restorable by the `site`-branch Site build.
   - Doubles as a **vulnerability alert**: it runs `dependency-check:aggregate` with `failBuildOnCVSS=7`, so a newly disclosed CVE in a current dependency surfaces as a failed scheduled run. (Threshold is easy to tune; `11` disables it.)
   - Rolling cache key + `restore-keys`, a bounded/non-fatal update step, and `save` with `if: always()` so partial progress is never lost.

2. **`site.yaml`**: replaced the static `nvd-db` cache key (write-once, never refreshed, evicted between releases) with a **rolling key + `restore-keys`** so it restores the warmer's cache; added a **bounded, non-fatal update step** that saves partial progress and a **job `timeout-minutes`** so a bad NVD day fails in minutes instead of at the 6-hour cap.

3. **`pom.xml`**: added `nvdValidForHours=168` (+ `nvdApiDelay`/`nvdMaxRetryCount`) to dependency-check in **both** `pluginManagement` (for the warmer's CLI goal) and `reporting` (reporting plugins don't inherit build/pluginManagement config), so a warm cache is treated as fresh and the download is skipped entirely.

## Rollout

Once merged, the warmer runs from the default branch (schedules only fire there) and warms the cache. A subsequent `site` push then restores it and publishes fast. Nothing here touches the `site` branch directly.

## Notes for review

@hazendaz — would value your Maven eyes on:
- The `dependency-check` config split across `pluginManagement` vs `reporting` (needed because reporting plugins don't inherit pluginManagement config — is there a cleaner idiom you'd prefer?).
- The warmer invocation `./mvnw install dependency-check:aggregate ...` — `install` is there so reactor sibling artifacts resolve for the aggregate scan; open to a lighter approach.
- `failBuildOnCVSS=7` as the alert threshold and the 3-day cadence.

## Summary by CodeRabbit

* **New Features**
  * Added an automated **NVD Cache Warmer** GitHub Actions workflow with scheduled refresh and manual trigger to improve security scan speed.

* **Improvements**
  * Enhanced Dependency-Check configuration with new tunable NVD throttling/retry settings and documented cache freshness (~1 week).
  * Updated CI caching for NVD data to restore more safely (restore-key based) and added an explicit site build timeout.
  * Extended Maven/guide reporting setup to use a **Guide API token** via environment configuration.

### DIFF
--- a/.github/workflows/nvd-cache.yaml
+++ b/.github/workflows/nvd-cache.yaml
@@ -37,9 +37,19 @@ jobs:
       - name: Setup Java
         uses: actions/setup-java@1bcf9fb12cf4aa7d266a90ae39939e61372fe520 # v5
         with:
-          cache: maven
           distribution: temurin
           java-version: 25
+      # Cache Maven artifacts but exclude the NVD data dir -- it has its own
+      # rolling cache below, and setup-java's built-in `cache: maven` would
+      # otherwise store the ~1.5 GB database a second time.
+      - name: Cache Maven dependencies
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: |
+            ~/.m2/repository
+            !~/.m2/repository/org/owasp/dependency-check-data
+          key: maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: maven-
       # Rolling cache: restore the newest previous snapshot via the prefix, and
       # always save under a unique key (actions/cache never overwrites a key).
       # Old snapshots LRU-evict; this keeps the DB fresh and incremental.
@@ -63,10 +73,11 @@ jobs:
         with:
           path: ~/.m2/repository/org/owasp/dependency-check-data
           key: nvd-db-${{ github.run_id }}
-      # Scan against the now-warm DB (nvdValidForHours in the pom keeps this from
-      # re-hitting the network). A CVE at or above the threshold fails the run,
-      # which is the alert. Tune the threshold to taste (11 disables it).
+      # Scan against the cached DB only (-DautoUpdate=false): the update step
+      # above owns the network fetch, so the scan can't launch a second full
+      # download under the remaining job budget. A CVE at or above the threshold
+      # fails the run, which is the alert. Tune the threshold to taste (11 off).
       - name: Scan dependencies for vulnerabilities
         env:
           NVD_API_KEY: ${{ secrets.NVD_API_KEY }}
-        run: ./mvnw install dependency-check:aggregate -DskipTests -Djacoco.skip=true -Dmaven.gitcommitid.skip=true -DfailBuildOnCVSS=7 -B --settings ./.mvn/settings.xml
+        run: ./mvnw install dependency-check:aggregate -DskipTests -Djacoco.skip=true -Dmaven.gitcommitid.skip=true -DautoUpdate=false -DfailBuildOnCVSS=7 -B --settings ./.mvn/settings.xml
--- a/.github/workflows/nvd-cache.yaml
+++ b/.github/workflows/nvd-cache.yaml
@@ -1,25 +1,14 @@
 name: NVD Cache Warmer
 
-# Keeps the OWASP dependency-check NVD database warm and alerts on newly
-# disclosed vulnerabilities in current dependencies.
-#
-# The Site (release) build is the only other NVD consumer and it runs at most
-# every few weeks -- longer than BOTH GitHub's 7-day cache-eviction window and
-# dependency-check's 7-day incremental-update window. Without this warmer every
-# release faces a cold, multi-hour full NVD download that is exposed to NVD API
-# throttling (which once ran into the 6-hour job cap and was killed).
-#
-# Running on a schedule keeps the cached database inside both windows, so the
-# release build restores a warm cache and only does a fast incremental update
-# (or none at all). GitHub makes caches created on the default branch readable
-# from every branch, so the cache this builds on master is available to the
-# `site` branch build. A run that fails on a CVE is the vulnerability alert.
+# Keeps the OWASP dependency-check NVD database warm so the Site (release) build
+# never has to do a cold full download -- that repeatedly failed on NVD 503s and
+# timed out. Running every ~3 days stays inside both GitHub's 7-day cache-eviction
+# window and dependency-check's 7-day incremental window, so each run is a fast
+# incremental. A failing run (a CVE at or above the threshold) is the vuln alert.
 
 on:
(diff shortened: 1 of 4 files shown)


.github/workflows/nvd-cache.yaml
```diff
@@ -0,0 +1,83 @@
...
+        with:
+          persist-credentials: false
+      - name: Setup Java
+        uses: actions/setup-java@1bcf9fb12cf4aa7d266a90ae39939e61372fe520 # v5
+        with:
+          distribution: temurin
+          java-version: 25
+      # Cache Maven artifacts but exclude the NVD data dir -- it has its own
+      # rolling cache below, and setup-java's built-in `cache: maven` would
+      # otherwise store the ~1.5 GB database a second time.
```
mprins: it's not actually that big; after full update yesterday the odc.mv.db is almost 250MB which compresses to about half that size...

.github/workflows/site.yaml
```diff
@@ -17,21 +17,57 @@ jobs:
...
+          restore-keys: maven-
+      # Rolling cache: restore the newest snapshot the NVD Cache Warmer saved on
+      # master (default-branch caches are readable from every branch). The old
+      # static `nvd-db` key never updated and evicted between releases, forcing a
+      # cold full download every time.
+      - name: Restore NVD database cache
+        uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: ~/.m2/repository/org/owasp/dependency-check-data
+          key: nvd-db-${{ github.run_id }}
```
mprins, line 39: as far as I understand `github.run_id` is unique for each workflow; so this would get you the last "site" cached db (if it exists) and not the "nvd-cache" cached db

.github/workflows/nvd-cache.yaml
```diff
@@ -0,0 +1,83 @@
...
+        timeout-minutes: 30
+        continue-on-error: true
+        env:
+          NVD_API_KEY: ${{ secrets.NVD_API_KEY }}
+        run: ./mvnw dependency-check:update-only -B --settings ./.mvn/settings.xml -Dmaven.gitcommitid.skip=true
+      - name: Save NVD database cache
+        if: always()
+        uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: ~/.m2/repository/org/owasp/dependency-check-data
```
mprins, line 61: so if the db changes, eg from (currently) 11 to 12 you will need to manually purge the cache to prevent caching the old file.

you could use something like the sha1 hash of the directory, eg. `ls ~/.m2/repository/org/owasp/dependency-check-data | sha1sum  | awk '{print $1}'`, though it does add complexity

.github/workflows/nvd-cache.yaml
dbwiddis: Good catch. I had actually looked at the entire cache size. Will fix the comment.

.github/workflows/site.yaml
dbwiddis: See the next line:
```yaml
restore-keys: nvd-db-
```
You're correct that the given key never hits on restore, it's intentional to avoid overwriting an existing cache (see the next step intended to save a partially downloaded NVD database)

On restore, it'll fetch from the newest `nvd-db-*` key.

.github/workflows/nvd-cache.yaml
dbwiddis: GHA auto-purges after 7 days or LRU based on size, so the old (stale) file will eventually get deleted, and we always deal with `nvm-db-<latest>` and `nvm-db-<old>` get cleaned up later.

But I'm realizing my initial intent to pick up from a partially downloaded DB is wrong.  It's all or nothing. So we're already over-complicating things by using a rolling key....

dbwiddis: Or not.  We can't update an existing cache so the rolling key is needed. Could use any unique key for that, run id works as well as anything else (timestamp, sha)

dbwiddis:
Just pushed new commits:
1. Simplifying the over-engineering
2. Adding a SONATYPE_GUIDE_TOKEN for the OSS Index 

TLDR of the intent here:  
- run an NVD cache warmer every 3 days (allows 1 failed run before 7-day timeout and "cold start")
- needs to run on master branch so cache can be used on site branch on releases
- update site branch to use same (latest) key (read-only).

dbwiddis:
Warm cache worked as designed for latest release: 🎉 
```
Run actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9
  with:
    path: ~/.m2/repository/org/owasp/dependency-check-data
    key: nvd-db-28981869021
    restore-keys: nvd-db-
    enableCrossOsArchive: false
    fail-on-cache-miss: false
    lookup-only: false
  env:
    JAVA_HOME: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/25.0.3-9/x64
    JAVA_HOME_25_X64: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/25.0.3-9/x64
    MAVEN_ARGS: -ntp
Cache hit for restore-key: nvd-db-28851326647
Received 116861151 of 116861151 (100.0%), 178.0 MBs/sec
Cache Size: ~111 MB (116861151 B)
/usr/bin/tar -xf /home/runner/work/_temp/b7ef96e8-49c8-44a5-b29b-669c02bbcb21/cache.tzst -P -C /home/runner/work/oshi/oshi --use-compress-program unzstd
Cache restored successfully
Cache restored from key: nvd-db-28851326647.classpathScope="test"
```

Issue #3430, opened 2026-07-06
7.3.2 - Mandatory OSGi imports on optional pandalxb


MrEasy:
Hi,

While extracting parts of functionality, e.g. with this commit: https://github.com/oshi/oshi/commit/0d9d4b578993bad8e3783d31a60a1ea4951c3a9b
there are classes loaded dynamically, like e.g. classes from package ``io.github.pandalxb.jlibrehardwaremonitor`` [here](https://github.com/oshi/oshi/commit/0d9d4b578993bad8e3783d31a60a1ea4951c3a9b#diff-5c5f27b23525a66992642c30c4e8991f793bcbd57149ee48b8c59c907e37a927R22-R133).

However, the bind plugin detects this by default and adds corresponding mandatory import statements, see this manifest extract:
```
Import-Package: org.slf4j;version="[1.7,3.0)",io.github.pandalxb.jlibrehardwaremonitor.config,io.github.pandalxb.jlibrehardwaremonitor.manager,io.github.pandalxb.jlibrehardwaremonitor.model,java.awt
```

From my point of view, those imports should get declared optional.

7.3.2 otherwise fails in OSGi environments not providing those dependencies:
```
Caused by: org.apache.felix.resolver.reason.ReasonException: Unable to resolve com.github.oshi.oshi-common/7.3.2: missing requirement [com.github.oshi.oshi-common/7.3.2] osgi.wiring.package; filter:="(osgi.wiring.package=io.github.pandalxb.jlibrehardwaremonitor.config)"
        at org.apache.felix.resolver.Candidates$MissingRequirementError.toException(Candidates.java:1341)
        ... 17 more
```

dbwiddis:
Can you please submit a pull request to fix this?

dbwiddis:
Fixed in #3431

PR #3431, opened 2026-07-06
#3430 Declared optional Import-Package instruction in oshi-common

MrEasy:
Moved from oshi-core where now unused

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OSGi compatibility so the shared module can resolve even when an external hardware-monitoring package is not available.
  * Adjusted package import handling to avoid runtime resolution issues in environments that don’t provide that optional dependency.

* **Chores**
  * Updated the changelog with this improvement.


mprins (approved):
LGTM, you might want to add a line to the [CHANGELOG](https://github.com/oshi/oshi/blob/master/CHANGELOG.md)

dbwiddis (approved):
LGTM as well.  Will merge with a changelog added (bug fixes & improvements).

I've been traveling for a few weeks and plan to cut a 7.4.0 release this weekend. Let me know if you need a patch to 7.3.3 earlier than that.

MrEasy:
> (...) Let me know if you need a patch to 7.3.3 earlier than that.

Thanks - all good, not urgent for me.

PR #3483, opened 2026-07-18
Fix Windows backend crash, native-memory leaks, and a registry-type gap

dbwiddis:
Pre-7.4.1 audit fixes for the Windows backends (primarily FFM). These paths are Windows-only, so they were verified by mirroring the existing guard/handle patterns and rely on the Windows CI matrix + `NativeComparisonTest` parity.

### Crash
- **`NetSessionDataFFM` — hard SIGSEGV.** A `SESSION_INFO_10` entry with a null `cname`/`username` (e.g. an anonymous IPC$ session) was passed straight to `readWideString`, faulting on native address 0 — a crash the surrounding `catch(Throwable)` cannot catch. Each pointer is now null-checked before dereference, matching `ProcessWtsDataFFM`.

### Memory / native heap
- **`AdlUtilFFM` — invalid cross-heap free.** The ADL `malloc` callback returned an `Arena.global()` segment, so ADL received JVM-managed memory that it later releases with C `free()` (undefined behavior), and the global arena never reclaimed it. It now allocates from native `malloc` and returns that pointer, matching the JNA backend, so ADL's `free()` is valid.
- **`Advapi32UtilFFM` / `WindowsOperatingSystemJNA` — leaked event-log handles.** `querySystemBootTime` closed the handle only on its normal-exit paths (leaking it if a `Throwable` escaped the record loop) — now wrapped in try/finally. `querySystemLog` (in **both** the FFM and JNA backends) opened a handle only to validate the configured log name and returned without closing it — now closed after validation.

### Divergence
- **`Advapi32UtilFFM.registryGetValue` dropped `REG_QWORD`/`REG_BINARY`.** It returned null for anything past `REG_SZ`/`REG_EXPAND_SZ`/`REG_DWORD`, so the GPU dedicated-VRAM fallback (stored as a `REG_QWORD`) read as 0 under FFM while JNA returned the true value. Added `REG_QWORD` → `Long` and `REG_BINARY` → `byte[]`, matching the JNA `Advapi32Util` contract.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows session/process name decoding by safely reading null-terminated UTF-16 strings, including anonymous or missing names.
  * Fixed native Event Log handle leaks in both FFM and JNA paths when querying boot time and system logs.
  * Improved AMD GPU memory reporting reliability by allocating ADL callback buffers from native C heap.
* **New Features**
  * Added Windows registry support for `QWORD` and `BINARY` values, improving consistency across registry lookups.


dbwiddis:
@coderabbit full review

PR #3525, opened 2026-07-23
Replace org.slf4j.event.Level with an OSHI-owned LogLevel enum

### LINKED ISSUE #3524: Oshi 7.4.1 breaks network capturing with slf4j-api < 1.7.15
Between 7.4.0 and 7.4.1, `AbstractNetworkIF.getAllNetworkInterfaces()` and `isLocalInterface()` (`oshi-common/src/main/java/oshi/hardware/common/AbstractNetworkIF.java`) moved from a plain `try/catch` with `LOG.error(...)` to `ExceptionUtil.getOrDefault(...)`:

```java
// 7.4.0 and earlier
private static List<NetworkInterface> getAllNetworkInterfaces() {
    try {
        ...
    } catch (SocketException ex) {
        LOG.error("...", ex.getMessage());
    }
    ...
}

// 7.4.1
private static List<NetworkInterface> getAllNetworkInterfaces() {
    return ExceptionUtil.getOrDefault(() -> {
        ...
    }, Collections.emptyList(), LOG, "Socket exception when retrieving interfaces: {}");
}
```

The 4-arg `ExceptionUtil.getOrDefault(supplier, defaultValue, log, msg)` overload immediately delegates to the 5-arg one with an explicit `Level.DEBUG`:

```java
public static <T> T getOrDefault(ThrowingSupplier<T> supplier, T defaultValue, Logger log, String msg) {
    return getOrDefault(supplier, defaultValue, log, Level.DEBUG, msg);
}
```

That's a direct, unconditional reference to `org.slf4j.event.Level` — it resolves on every call to `getAllNetworkInterfaces()`/`isLocalInterface()`, regardless of whether a `SocketException` is ever thrown. `org.slf4j.event.Level` was added in **slf4j 1.7.15** (Dec 2014); `Logger`/`LoggerFactory`, which is all earlier oshi versions ever touched, has existed since slf4j 1.0.

This breaks embedders that bundle oshi without also bundling their own `slf4j-api`, relying instead on the host application to provide one — and where that host may provide an `slf4j-api` older than 1.7.15. Concretely: we ship oshi inside a Java agent that Apache Maven loads as a core extension (`-Dmaven.ext.class.path`). Maven core extensions resolve `slf4j-api` from Maven's own bundled copy rather than from the extension's own classpath. Maven's bundled `slf4j-api` was 1.7.5 through Maven 3.3.9 (Nov 2015), and wasn't bumped past 1.7.15 until Maven 3.5.0 (Apr 2017, bundling 1.7.22). Against those older Maven installations, calling `getAllNetworkInterfaces()` now throws:

```
java.lang.NoClassDefFoundError: org/slf4j/event/Level
    at oshi.util.ExceptionUtil.getOrDefault(ExceptionUtil.java:172)
    at oshi.hardware.common.AbstractNetworkIF.getAllNetworkInterfaces(AbstractNetworkIF.java:189)
    at oshi.hardware.common.AbstractNetworkIF.getNetworks(AbstractNetworkIF.java:...)
    at oshi.hardware.platform.mac.MacNetworkIfJNA.getNetworks(MacNetworkIfJNA.java:...)
    at oshi.hardware.platform.mac.MacHardwareAbstractionLayerJNA.getNetworkIFs(MacHardwareAbstractionLayerJNA.java:...)
```

— on every single call, since it's on the non-exceptional path, not just when a `SocketException` actually occurs.

We've worked around this locally by holding at 7.4.0, but since oshi generally seems to care about supporting a broad range of host environments, wanted to flag it in case there's a straightforward fix — e.g. avoiding the hard classload of `org.slf4j.event.Level` on `ExceptionUtil`'s hot/non-exceptional path, or resolving it lazily only when a message is actually about to be logged.

## Environment
- oshi 7.4.1 (also present in 7.4.0's later commits / current `master`, per the file above)
- Reproduced with Apache Maven 3.3.1 / 3.3.9 (bundled `slf4j-api` 1.7.5) as the host; also affects any embedder relying on a host-provided `slf4j-api` older than 1.7.15.

wolfs:
ExceptionUtil and its callers across oshi-common, oshi-core, and oshi-core-ffm carried the log level as org.slf4j.event.Level, a type that only exists from slf4j-api 1.7.15 and was referenced on non-exceptional code paths (e.g. the getOrDefault convenience overloads delegate with an explicit Level.DEBUG, and call sites pass Level.WARN as a plain argument). Embedders relying on a host-provided slf4j older than 1.7.15 -- such as a Java agent loaded as an Apache Maven 3.3.x core extension, which resolves slf4j from Maven's bundled 1.7.5 -- therefore hit NoClassDefFoundError: org/slf4j/event/Level on ordinary calls, not just when an exception was actually logged (#3524).

logAtLevel dispatches only to the classic Logger#debug/info/warn/error/trace methods (present since slf4j 1.0), so carrying the level with an OSHI enum removes the 1.7.15 floor without changing any logging behavior. The now-unused org.slf4j.event OSGi import is dropped as well.

Split the SLF4J 1.7 compatibility check into two profiles, since no single 1.7.x version exercises both constraints: verify-slf4j-17 pins 1.7.5 (the API floor) and skips module-info, because Automatic-Module-Name: org.slf4j only arrived in 1.7.27; verify-slf4j-17-jpms pins 1.7.27 and compiles module-info to keep verifying that `requires org.slf4j` resolves on the module path. Both run in CI.

## Summary by CodeRabbit

* **New Features**
  * Added OSHI-native logging levels for consistent exception and diagnostic handling.

* **Bug Fixes**
  * Improved compatibility with environments providing older SLF4J versions.
  * Prevented unnecessary loading of newer SLF4J logging classes during normal operation.
  * Updated platform-specific diagnostics to preserve existing warning, error, debug, and trace behavior.


dbwiddis (commented):
Overall LGTM.

I don't think we need the `pom.xml` changes (and corresponding added PR test).  

The additional profile and dancing around JPMS complexity adds very little value, given the forbiddenapis check.

The chances of us using any other logging features that this guards against are near-zero.

oshi-common/src/main/java/oshi/util/LogLevel.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * A logging severity level, used by OSHI's exception-handling and native-call helpers (such as {@link ExceptionUtil})
+ * to select which {@code Logger#debug/info/warn/error/trace} method a caught or handled condition is logged with.
+ * <p>
+ * OSHI carries the level with this enum rather than {@code org.slf4j.event.Level} so that it never class-loads a type
+ * introduced in slf4j-api 1.7.15; dispatch uses only classic {@code Logger} methods present since slf4j 1.0. This keeps
+ * OSHI loggable against host-provided slf4j-api versions older than 1.7.15 (e.g. the 1.7.5 bundled by Apache Maven
+ * 3.3.x, which loads OSHI-based extensions against its own slf4j).
+ */
+@ThreadSafe
```
dbwiddis: I don't think we annotate any of our enums in the codebase.  If we did, Immutable would probably be a better choice here, but I think this annotation is unnecessary.

wolfs: Done.

wolfs:
> I don't think we need the pom.xml changes (and corresponding added PR test).

Removed. Thank you for the quick review!

PR #3531, opened 2026-07-27
Reject implausible Apple Silicon SMC temperatures

### LINKED ISSUE #3529: FFM CPU temperature on Apple Silicon is nondeterministic and diverges from JNA
Found incidentally while smoke-testing #3528 on an M2 Max (macOS 15, JDK 25). Not caused by that PR — reproduced on unmodified `master`.

### Symptom

`hal.getSensors().getCpuTemperature()` on the FFM provider returns one of two values across runs, from the same binary with no other change:

```
oshi.ffm.SystemInfo   6.699999809265137
oshi.ffm.SystemInfo  57.995201110839844   /  60.15781784057617  /  61.729740142822266
```

The JNA provider is stable and plausible on the same machine at the same moment:

```
oshi.SystemInfo      59.12498092651367
```

So the FFM path intermittently reports ~6.7 °C where JNA reports ~59 °C. It is not thermal variation — 6.7 is bit-for-bit identical every time it appears, and it alternates run to run rather than drifting.

### Reproduction

Verified the flake is pre-existing by stashing the #3528 changes, rebuilding `oshi-core-ffm` from `master`, and re-running: `master` produced `6.699999809265137` on three consecutive runs and `57.995201110839844` on another. The #3528 branch produced both values too. Both branches show both values, so the variable is not the code under change.

### Suspected cause

`SmcUtilFFM.smcGetFirstFloat(conn, SMC_KEYS_CPU_TEMP_AS)` returns the first key whose value is `> 0`, scanning `Tp09, Tp0T, Tp01, Tp05, Tp0D`. A key that intermittently yields a small bogus reading would be accepted and short-circuit the scan, which matches the exactly-repeating 6.7. Worth checking whether the SMC key-info cache (`KEY_INFO_CACHE`) is being populated with a stale or wrong `dataSize`/`dataType` on some open, since that would change how the same bytes are decoded.

A useful first diagnostic would be logging each key's raw bytes, `dataSize` and `dataType` per read and diffing a 6.7 run against a 59 run.

### Scope

Apple Silicon path only. Not investigated on Intel Macs, and the non-`_AS` key (`TC0P`) was not exercised here.

dbwiddis:
Fixes #3529.

Apple Silicon power-gates a CPU core cluster when it is idle. The SMC then reports a fixed parked value for each die sensor in that cluster instead of a reading — on an M2 Max, `6.7 °C` for `Tp01`/`Tp09` and `4.633 °C` for `Tp05`/`Tp0D`. Those are below room ambient and cannot be die temperatures, but they are *positive*, so the `> 0` test in `smcGetFirstFloat` accepted them. On a genuinely idle Mac, `getCpuTemperature()` reported **6.7 °C**.

### The change

`smcGetFirstFloat` becomes `smcGetFirstTemperature` and requires a reading of at least `MIN_PLAUSIBLE_TEMPERATURE` (20 °C), returning 0 for unavailable per the documented `Sensors` contract. The Intel `TC0P` fallback is held to the same floor. Applied to the JNA and FFM twins, and to the GPU path which shares the helper.

Verified on an M2 Max, identical on both implementations:

```
IDLE    getCpuTemperature() = 0.000    <- reported UNAVAILABLE
LOADED  getCpuTemperature() = 49.339
```

### Why this shape, and not the alternatives

Everything below was measured on hardware rather than reasoned about; details and raw output are in #3529.

- **Not an FFM/JNA divergence.** That was my original (wrong) diagnosis in the issue. Both implementations read byte-identical SMC payloads and agreed on all 19 interleaved samples — both were equally affected.
- **Not nondeterminism.** The value tracks P-cluster power state exactly and reversibly. Sustained load brings the sensors live within ~1.6 s; they park again ~1.5 s after it stops. That is the entire explanation for the apparent run-to-run flakiness.
- **Retry rejected.** At deep idle the parked state is 100% persistent across 40 samples, and a second read recovered **0/15** at delays of 0, 10, 50, 200 and 1000 ms. Even forcing 150 ms of CPU work first recovered 0/10 — one busy thread lands on an E-core, so the P-cluster stays gated.
- **Max-instead-of-first rejected.** All five keys are one power domain and park together, so `max(6.70, 0, 6.70, 4.63, 4.63)` is still `6.70` — no better than the current behaviour.
- **Outlier/median detection rejected.** Across 224 samples spanning a full idle→load→idle transition there were **0 mixed sets**: the keys are always all parked or all live, so there is never a good reading to compare an outlier against.
- **Extending the key list rejected.** `Tp1*` park too at genuine idle; they only looked live in an earlier sample taken on a still-warm machine.

### On the floor, and the absence of a ceiling

The floor is a plausibility guard, not a hardware specification, and is documented as such. A sweep of all **288** temperature sensors found parked values from 0 to 16.6 °C, while idle Apple Silicon runs about 28–35 °C — so 20 °C clears every observed parked value with margin below any genuine reading.

There is deliberately **no upper bound**. Every bad value observed was at the low end (full range across all sensors: −68 to +80.28 °C), and there is no published Apple `Tjmax` to anchor one — the [Linux kernel's Apple SMC hwmon driver](https://lkml.iu.edu/2511.1/11772.html) applies no temperature validation at all. A ceiling near the ~100 °C throttling point risks discarding real readings exactly when they matter most.

### Behaviour change

Users on an idle Apple Silicon Mac will now see **0 (unavailable)** where they previously saw a confident 6.7 °C. That is the documented contract for unavailable data, and it replaces a wrong number with an honest one — but it is user-visible, so it has a CHANGELOG entry. Rejected values are logged at DEBUG with the key name, so anyone asking "why is my temperature 0" can find out.

Intel Macs are unaffected; they use `TC0P`, which is validated but does not park.

### Tests

A plausibility invariant on both twins, enabled on macOS: the reading is either unavailable or at least the floor, never in between. This holds on CI runners too, where SMC access typically fails and returns 0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS CPU and GPU temperature reporting on Apple Silicon by ignoring idle-gated, implausibly low SMC readings using a plausibility threshold.
  * `Sensors#getCpuTemperature` now reports `0` (unavailable/parked) instead of returning fixed sentinel temperatures for idle core clusters.
* **Tests**
  * Expanded macOS test coverage to ensure plausibility filtering accepts only `NaN`, `0`, or temperatures at/above the threshold, while rejecting known parked/sentinel values (including verifying there’s no upper bound).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * [#3526](https://github.com/oshi/oshi/pull/3526): Fix a potential `NullPointerException` in the macOS FFM process backend when an `IOPlatformDevice` entry has no readable name, and guard the native `passwd`/`group` struct dereferences against a null return - [@dbwiddis](https://github.com/dbwiddis).
 * [#3527](https://github.com/oshi/oshi/pull/3527): Consolidate the SLF4J log level dispatch duplicated in `ExceptionUtil` and `PerfDataUtil` into a new `oshi.util.LogUtil`, whose `isEnabled` method offers the slf4j 1.x-compatible equivalent of `Logger#isEnabledForLevel` - [@dbwiddis](https://github.com/dbwiddis).
+* [#3531](https://github.com/oshi/oshi/pull/3531): Stop reporting an implausible CPU or GPU temperature on Apple Silicon. An idle core cluster is power-gated and its die sensors then report a fixed parked value below room ambient, which was previously returned as a reading; `Sensors#getCpuTemperature` now reports 0 (unavailable) instead - [@dbwiddis](https://github.com/dbwiddis).
 
 # 7.4.0 (2026-07-08), 7.4.1 (2026-07-18), 7.4.2 (2027-07-24)
 
--- a/oshi-core-ffm/src/main/java/oshi/ffm/util/platform/mac/SmcUtilFFM.java
+++ b/oshi-core-ffm/src/main/java/oshi/ffm/util/platform/mac/SmcUtilFFM.java
@@ -166,17 +166,30 @@ public final class SmcUtilFFM {
     }
 
     /**
-     * The lowest reading accepted as a genuine die temperature, in degrees Celsius.
+     * The lowest reading accepted as a genuine temperature, in degrees Celsius.
      * <p>
-     * Apple Silicon power-gates a CPU core cluster when it is idle, and the SMC then reports a fixed parked value for
-     * each die sensor in that cluster instead of a reading. On an M2 Max those are 6.7 (Tp01, Tp09) and 4.633 (Tp05,
-     * Tp0D); a sweep of all 288 temperature sensors found parked values from 0 to 16.6. They are below room ambient and
-     * so cannot be real die temperatures, but they are positive, so a simple {@code > 0} test accepts them.
+     * Apple Silicon power-gates a CPU core cluster when it is idle, and the SMC then reports a fixed sentinel for each
+     * die sensor in that cluster instead of a reading. Observed sentinels are 6.7 and 4.633 on an M2 Max (Tp01/Tp09 and
+     * Tp05/Tp0D), up to 8.425 across all of that machine's sensors, and -4.0 through 5.2 in the
+     * <a href="https://github.com/dkorunic/iSMC">iSMC</a> sample reports, which independently describe them as
+     * "firmware sentinels from inactive sensor slots". They are below room ambient and so cannot be real die
+     * temperatures, but most are positive, so a simple {@code > 0} test accepts them.
      * <p>
-     * Idle Apple Silicon runs around 28-35 C, so this floor leaves margin above every parked value observed while
-     * staying well below any genuine reading. It is a plausibility guard, not a hardware specification.
+     * Across those reports, covering M1 through M5, A18 and Intel T2 machines, no genuine sensor read between 8.425 and
+     * 21, so this floor sits in an empty band with margin on both sides. It is a plausibility guard, not a hardware
+     * specification.
      */
-    public static final double MIN_PLAUSIBLE_TEMPERATURE = 20d;
+    public static final double MIN_PLAUSIBLE_TEMPERATURE = 15d;
+
+    /**
+     * Tests whether a reading is plausible as a temperature, rejecting the sentinel an idle-gated sensor reports.
+     *
+     * @param celsius the reading to test, in degrees Celsius
+     * @return true if the reading is at least {@link #MIN_PLAUSIBLE_TEMPERATURE}
+     */
+    public static boolean isPlausibleTemperature(double celsius) {
+        return celsius >= MIN_PLAUSIBLE_TEMPERATURE;
+    }
 
     /**
      * Get the first plausible temperature from a list of SMC keys, skipping sensors that are reporting a parked value
@@ -189,7 +202,7 @@ public final class SmcUtilFFM {
     public static double smcGetFirstTemperature(int conn, List<String> keys) {
         for (String key : keys) {
             double val = smcGetFloat(conn, key);
-            if (val >= MIN_PLAUSIBLE_TEMPERATURE) {
+            if (isPlausibleTemperature(val)) {
                 return val;
             }
             if (val != 0d) {
--- a/oshi-core-ffm/src/main/java/oshi/ffm/util/platform/mac/SmcUtilFFM.java
+++ b/oshi-core-ffm/src/main/java/oshi/ffm/util/platform/mac/SmcUtilFFM.java
@@ -166,17 +166,35 @@ public final class SmcUtilFFM {
     }
 
     /**
-     * Get the first positive value from a list of SMC keys.
+     * The lowest reading accepted as a genuine die temperature, in degrees Celsius.
+     * <p>
+     * Apple Silicon power-gates a CPU core cluster when it is idle, and the SMC then reports a fixed parked value for
+     * each die sensor in that cluster instead of a reading. On an M2 Max those are 6.7 (Tp01, Tp09) and 4.633 (Tp05,
+     * Tp0D); a sweep of all 288 temperature sensors found parked values from 0 to 16.6. They are below room ambient and
+     * so cannot be real die temperatures, but they are positive, so a simple {@code > 0} test accepts them.
+     * <p>
+     * Idle Apple Silicon runs around 28-35 C, so this floor leaves margin above every parked value observed while
+     * staying well below any genuine reading. It is a plausibility guard, not a hardware specification.
+     */
+    public static final double MIN_PLAUSIBLE_TEMPERATURE = 20d;
+
+    /**
+     * Get the first plausible temperature from a list of SMC keys, skipping sensors that are reporting a parked value
+     * because their core cluster is idle. See {@link #MIN_PLAUSIBLE_TEMPERATURE}.
      *
      * @param conn The connection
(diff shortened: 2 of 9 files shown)


dbwiddis:
On the GPU plausibility test suggestion — taking the second half of it (document the exemption), skipping the first (add the assertion), because the proposed assertion would be incorrect on two counts:

1. **Wrong sentinel.** `GpuStats#getTemperature` documents *"temperature in degrees Celsius, or -1 if unavailable"*, not 0 or NaN. An assertion of "unavailable or >= `MIN_PLAUSIBLE_TEMPERATURE`" written against 0 would read the real `-1` return as a failure.
2. **Wrong source.** The fallback reads `Temperature(C)` from the IOAccelerator performance-statistics dictionary — IOKit, not the SMC. It is not power-gated with the GPU cluster, so the SMC parking behaviour this PR fixes does not apply to it, and the floor deliberately does not either. Asserting the floor across the whole method would fail the first time that path returned a legitimate low reading.

The SMC half of the GPU path *is* covered: `MacGpuStats{FFM,JNA}#getTemperature` now call `smcGetFirstTemperature`, so it inherits the floor and the DEBUG logging.

Pushed `dedffcd` documenting the exemption in both twins, which is what the finding's own fallback branch asked for.

dbwiddis:
Cross-validated the 20 C floor against [iSMC's sample reports](https://github.com/dkorunic/iSMC/tree/master/internal/reports) — full SMC dumps from 13 machines spanning M1, M1 Pro/Max/Ultra, M2, M3 Pro/Max, M4, M4 Pro/Max, M5 Pro/Max, A18 and an Intel T2. That is far better coverage than the single M2 Max this PR was developed on.

### The floor never changes a good reading

In every report where OSHI's CPU keys are present, the old `> 0` selection and the new `>= 20` selection pick the **same** value:

| report | OSHI keys found | old `first > 0` | new `first >= 20` |
|---|---|---|---|
| a18 | Tp09=91.43 Tp01=92.76 Tp05=94.28 | 91.43 | 91.43 |
| m2 | Tp09=37.67 Tp01=37.92 Tp05=37.05 | 37.67 | 37.67 |
| m3-max | Tp0T=57 Tp05=48 Tp0D=48 | 57 | 57 |
| m4-max | Tp09=Tp0T=Tp01=Tp05=Tp0D=40 | 40 | 40 |
| m4 | Tp09=37 Tp01=36 Tp05=36 Tp0D=36 | 37 | 37 |

No regression on any of them.

### There is an empty band exactly where the floor sits

Across all 489 `Tp*`/`Tg*` readings in those reports:

```
below 20 C :  53 readings, highest = 5.20
>= 20 C    : 436 readings, lowest  = 21.00   (intel-t2:Tp5a)
             ^^^ nothing at all observed between 5.20 and 21.00
```

The sub-20 values are `-4.0` (x15), `0.0` (x23), `2.5` (x6), `4.0` (x6), `5.2` (x3) — parked sentinels, and note they differ per model, which is why the PR rejects on a value floor rather than by matching sentinels. The `-4.0` on an M4 Max is a useful find: parked values are not even guaranteed positive.

### 20 turns out to be the right number, not just a safe one

Combining with the M2 Max sweep in #3529, which saw a parked value as high as **16.6 C**:

```
highest parked value observed anywhere : 16.6   (M2 Max sweep)
lowest genuine reading observed anywhere: 21.0   (intel-t2:Tp5a)
```

20 is inside the only gap that separates those two populations. A lower floor such as 15 would have admitted the 16.6 park; a higher one would start crowding real readings. Worth recording for anyone tempted to widen the key list later — the margin above the floor is about 1 C in the general `Tp*` space, even though it is much larger for the five keys OSHI actually reads.

dbwiddis:
**Correction to my previous comment, and the floor is now 15 C** (`c0a11ec`).

I claimed 16.6 C was a parked sentinel on the M2 Max and used it to argue that 20 was the only value in the gap. That was wrong. The sensor is `TVMO`, and ten consecutive reads give ten distinct values:

```
TVMO   16.732  15.010  15.765  15.175  16.227  15.651  15.411  15.548  16.090  16.261
Tp09    6.700   6.700   6.700   6.700   6.700   6.700   6.700   6.700   6.700   6.700
```

A sentinel is bit-identical on every read, as `Tp09` is. `TVMO` drifts continuously — it is a live sensor somewhere cool. My sweep had misclassified it because the heuristic was "load-responsive and below 20 C", which a genuinely cool sensor also satisfies.

### Revised evidence

| | value | source |
|---|---|---|
| highest **sentinel** observed | 8.425 | M2 Max sweep (`Tp02`/`Tp0A` family); iSMC reports max 5.2 |
| lowest **genuine** reading observed | 21.0 | `intel-t2:Tp5a`, across 13 machines |

The empty band is 8.425 → 21.0, so 20 sat about 1 C below the genuine population rather than in the middle. **15** is roughly central: 6.6 C above the highest sentinel, 6 C below the lowest genuine reading.

### Prior art

[iSMC](https://github.com/dkorunic/iSMC) reaches the same diagnosis independently — its `isValidReading` rejects *"firmware sentinels from inactive sensor slots (observed: −4..5.2 °C)"* with a `minTempCelsius = 10.0` scoped to temperature units only, the same unit-scoping this PR uses. 15 rather than their 10 because this sweep observed sentinels up to 8.425, higher than they report, which leaves only 1.6 C of margin at 10.

### Also in this push

`isPlausibleTemperature` is extracted so the guard can be tested directly instead of only through a live SMC connection. The new tests table-drive it against values actually captured from hardware — sentinels `-4.0, 0.0, 2.5, 4.0, 4.633, 5.2, 6.033, 6.7, 7.425, 8.425` and genuine readings `21.0 … 110.0` — plus an assertion that the floor stays inside the empty band, so moving it onto either population fails loudly.

Re-verified end to end on the M2 Max: unavailable at deep idle, ~59 C under load.

### Follow-up, not in this PR

iSMC [commit 7e229fa](https://github.com/dkorunic/iSMC/commit/7e229fa6aea6f65d20ee345778a2b5e236c56aba) reports that M4 performance-core keys are `Tp1i`/`Tp1m`/…, not the `Tp01`/`Tp05` family. If so, `SMC_KEYS_CPU_TEMP_AS` reads the wrong sensors on M4 — a key-selection problem independent of this guard, worth its own issue.

dbwiddis:
Both addressed in `53aa2ec`, though the second differently than suggested.

**Boundary assertion — valid, fixed.** `>= 8.425` would have accepted a floor of exactly 8.425, and since the predicate is `celsius >= MIN_PLAUSIBLE_TEMPERATURE`, that floor would make `isPlausibleTemperature(8.425)` return `true` — contradicting `testObservedSentinelsAreRejected`. The guard could not have caught that particular regression. Now strictly greater, with a comment explaining why.

**The 110 C case — real inconsistency, but in my test data rather than in `testSensors`.** The two assertions do not actually conflict at runtime: `testSensors` asserts a bound on what *this machine reports*, while the new test asserts the *predicate's* contract. No hardware reading of 110 is implied by one passing the other.

The genuine problem is that I filed 110 under `GENUINE_READINGS`, whose javadoc says these are readings observed on hardware. 110 was never observed anywhere — it is a probe for the deliberate absence of an upper bound. So it has moved to its own `testNoUpperBoundIsApplied`, leaving the observed-data arrays containing only observed data.

I did not change the 0-100 bound in `testSensors`. It predates this PR, it asserts what real hardware reports rather than what the predicate accepts, and the absence of a ceiling is intentional here: every bad value observed was at the low end, and a ceiling near the ~100 C throttling point would discard real readings exactly when they matter most.

PR #3535, opened 2026-07-27
Read Apple Silicon CPU temp from SMC die-aggregate keys

### LINKED ISSUE #3534: macOS: hardcoded Apple Silicon CPU temperature keys are wrong or absent on M1, M3 and M5
`SmcUtil.SMC_KEYS_CPU_TEMP_AS` (and its FFM twin) hardcodes one list for all Apple Silicon:

```java
{ "Tp09", "Tp0T", "Tp01", "Tp05", "Tp0D" }
```

Those keys were [derived from live SMC enumeration on one machine](https://github.com/oshi/oshi/pull/3091). Cross-checking against [iSMC's key map](https://github.com/dkorunic/iSMC/blob/master/src/temp.txt) — 1077 entries tagged by chip family — shows the same FourCC means different things on different chips, so a single list cannot be right everywhere.

### How iSMC classifies each of our five keys

| key | M1 | M2 | M3 | M4 | A18 | M5 |
|---|---|---|---|---|---|---|
| `Tp09` | **Efficiency Core 1** | Perf Core 3 | – | Perf Core 3 | Perf Cluster Aggregate 1 | – |
| `Tp0T` | **Efficiency Core 2** | – | Perf **Cluster** 1 | Perf Core 8 | – | – |
| `Tp01` | Perf Core 1 | Perf Core 1 | – | Perf Core 1 | Perf Core 1 | – |
| `Tp05` | Perf Core 2 | Perf Core 2 | Perf Core 1 | Perf Core 2 | Perf Core 2 | – |
| `Tp0D` | Perf Core 3 | Perf Core 4 | Perf Core 2 | Perf Core 4 | Perf Cluster Aggregate 2 | – |

Presence by family: M1 5/5, M2 4/5, M3 3/5, M4 5/5, A18 4/5, **M5 0/5**.

### Two concrete consequences

**1. No CPU temperature at all on M5.** Its performance cores are an entirely different key set:

```
Tp0O Tp0R Tp0U Tp0X Tp0a Tp0d Tp0g Tp0j Tp0m Tp0p Tp0u Tp0y
```

None of our five appear. The `report-m5-max.txt` and `report-m5-pro.txt` [sample dumps](https://github.com/dkorunic/iSMC/tree/master/internal/reports) confirm it — none of our keys are present, so every read returns 0 and `getCpuTemperature()` reports unavailable.

**2. An efficiency-core temperature reported as the CPU temperature on M1.** `Tp09` is tried first and is an *Efficiency* Core on M1, so `smcGetFirstTemperature` returns it whenever it is plausible. E-cores run cooler than P-cores, so M1 readings are biased low. Same shape on M3, where `Tp0T` is a cluster aggregate rather than a core.

Also worth noting: iSMC [commit 7e229fa](https://github.com/dkorunic/iSMC/commit/7e229fa6aea6f65d20ee345778a2b5e236c56aba) reworked the M4 performance-core mapping, adding the `Tp1i`/`Tp1m`/… family and cluster sensors. Their M4 list now runs to 60 perf-core entries.

### Approaches, none obviously best

- **Per-chip key tables**, selected from `getProcessorIdentifier()`. Most accurate, but needs a new table for every chip Apple ships, and iSMC's file shows how large that gets.
- **Union of all known keys, take the max plausible reading.** No chip detection, degrades gracefully because an absent key returns 0 and is rejected by the plausibility guard from #3531. "Hottest core" is also a defensible definition of CPU temperature. Costs more SMC round-trips (~0.25 ms per key measured on an M2 Max), which matters little since the value is memoized.
- **Dynamic enumeration** via `SMC_CMD_READ_INDEX` (already implemented in `oshi-demo/SmcDump.java`), filtering to `Tp*`/`Te*`, cached once per JVM. Self-adapting to unreleased chips, but 2409 keys is a slow first call.
- **Config override**, e.g. `oshi.os.mac.sensors.cputemperature.keys`. Precedent exists — `Sensors#getCpuTemperature` already documents `oshi.os.linux.sensors.hwmon.names` and `...cputemperature.types` for Linux. Useful as an escape hatch alongside any of the above.

The union-plus-max option looks like the best effort/robustness trade to me, but I have only an M2 Max to verify on.

Discovered while investigating #3529; the plausibility guard in #3531 is a prerequisite for the union approach, since it is what makes reading a key that is absent or idle-gated safe.

dbwiddis:
## Summary

Fixes #3534. OSHI read the macOS CPU temperature from five hardcoded per-core SMC FourCC keys (`Tp09, Tp0T, Tp01, Tp05, Tp0D`). These are chip-specific, and **all five are absent on the M3 Pro** (which exposes a different set: `Tp06/Tp0E…`, `Te0…`, `Tp3P/Tp3X/Tp3T`), so `Sensors#getCpuTemperature()` returned 0.

This reads the **chip-independent, firmware-computed CPU-die aggregate keys** that the SMC already exposes:

- **`TCMb`** — CPU die **average**, preferred as primary
- **`TCMz`** — CPU die **max**, secondary

Both are `flt` type and decode through the existing `smcGetFirstTemperature` path, which already applies the 15 °C plausibility floor from #3531 — **no new native binding and no private API**. This supersedes the IOHID-fallback idea explored in the issue: across every chip in the iSMC hardware dataset (M1→M5, A18, Intel T2) a usable SMC CPU key is always present — even the M3 Pro, which lacks the per-core keys, still exposes `TCMb`/`TCMz` — so IOHID would only have defended a case that does not occur, at the cost of undocumented private-API bindings reading a different (die-diode) sensor ~20-30 °C lower.

`TCMb` is preferred so the reported value stays close to the historical single-core reading rather than jumping ~20 °C to the die max.

New query order in `queryCpuTemperature` (both backends):
1. SMC CPU-die aggregate keys — `TCMb` then `TCMz` (plausibility-guarded)
2. SMC per-core keys (existing `SMC_KEYS_CPU_TEMP_AS`) — unchanged
3. Intel `TC0P` — unchanged, plausibility-guarded

**Scope: CPU temperature only.** GPU is deferred to a follow-up — it has no analogous `TGM*` die aggregate, and `MacGpuStats.getTemperature()` already has a key-independent IOAccelerator fallback.

## Changes

Changed in lockstep across the JNA (`oshi-core`) and FFM (`oshi-core-ffm`) backends so `NativeComparisonTest.sensors()` stays green:

- `SmcUtil` / `SmcUtilFFM` — add `SMC_KEYS_CPU_TEMP_AGGREGATE_AS = [TCMb, TCMz]`
- `MacSensorsJNA` / `MacSensorsFFM` — try the aggregate keys before the per-core keys
- New JNA `MacSensorsPlausibilityTest` (twin of the existing FFM test) + an aggregate-keys-average-first contract test in both

## Verification

- `mvn -pl oshi-core -am clean test` — 1127 passed
- `mvn -pl oshi-core-ffm -am clean test` — 1127 passed
- `mvn test -pl oshi-benchmark -Pnative-comparison` — `NativeComparisonTest.sensors()` passes (FFM within 25% of JNA)
- **Real M3 Pro hardware** (Mac15,7, macOS 26.5.2, JDK 26): `getCpuTemperature()` went from 0 → the `TCMb` value on both backends (paired same-JVM read: JNA `TCMb`=72.117 → `getCpuTemperature()`=72.117; FFM `TCMb`=73.521 → `getCpuTemperature()`=73.521), confirming it tracks the die average and not `TCMz` or 0.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed CPU temperature reporting on Apple Silicon so `Sensors#getCpuTemperature` no longer incorrectly returns `0` for chips with differing SMC CPU-die keys.
  - Improved temperature detection by prioritizing chip-independent aggregate CPU-die readings, with fallback to existing per-core readings.
  - Ensured implausible temperature sentinel values are not reported as valid readings.
- **Tests**
  - Added/expanded tests to verify aggregate-key preference and plausibility handling for macOS CPU temperature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * [#3526](https://github.com/oshi/oshi/pull/3526): Fix a potential `NullPointerException` in the macOS FFM process backend when an `IOPlatformDevice` entry has no readable name, and guard the native `passwd`/`group` struct dereferences against a null return - [@dbwiddis](https://github.com/dbwiddis).
 * [#3527](https://github.com/oshi/oshi/pull/3527): Consolidate the SLF4J log level dispatch duplicated in `ExceptionUtil` and `PerfDataUtil` into a new `oshi.util.LogUtil`, whose `isEnabled` method offers the slf4j 1.x-compatible equivalent of `Logger#isEnabledForLevel` - [@dbwiddis](https://github.com/dbwiddis).
 * [#3531](https://github.com/oshi/oshi/pull/3531): Stop reporting an implausible CPU or GPU temperature on Apple Silicon. An idle core cluster is power-gated and its die sensors then report a fixed parked value below room ambient, which was previously returned as a reading; `Sensors#getCpuTemperature` now reports 0 (unavailable) instead - [@dbwiddis](https://github.com/dbwiddis).
+* [#3535](https://github.com/oshi/oshi/pull/3535): Fix `Sensors#getCpuTemperature` returning 0 on Apple Silicon chips whose per-core SMC keys differ from the hardcoded set (e.g. the M3 Pro, where all five are absent). Prefer the chip-independent, firmware-computed CPU-die aggregate keys `TCMb` (die average) and `TCMz` (die max) before the per-core keys, reporting the die average - [@dbwiddis](https://github.com/dbwiddis).
 
 # 7.4.0 (2026-07-08), 7.4.1 (2026-07-18), 7.4.2 (2027-07-24)
 
--- a/oshi-core-ffm/src/main/java/oshi/ffm/util/platform/mac/SmcUtilFFM.java
+++ b/oshi-core-ffm/src/main/java/oshi/ffm/util/platform/mac/SmcUtilFFM.java
@@ -76,6 +76,14 @@ public final class SmcUtilFFM {
     /** SMC key for CPU voltage. */
     public static final String SMC_KEY_CPU_VOLTAGE = "VC0C";
 
+    /**
+     * Apple Silicon CPU-die aggregate temperature keys, tried in order until one returns a plausible value. These are
+     * computed by the SMC firmware and are chip-independent, unlike the per-core {@link #SMC_KEYS_CPU_TEMP_AS} keys
+     * which vary between chips (and are entirely absent on some, e.g. the M3 Pro). {@code TCMb} is the CPU-die average
+     * and {@code TCMz} is the CPU-die maximum; the average is preferred because it stays close to the single-sensor
+     * value OSHI historically reported.
+     */
+    public static final List<String> SMC_KEYS_CPU_TEMP_AGGREGATE_AS = List.of("TCMb", "TCMz");
     /** SMC keys for Apple Silicon CPU temperature sensors. */
     public static final List<String> SMC_KEYS_CPU_TEMP_AS = List.of("Tp09", "Tp0T", "Tp01", "Tp05", "Tp0D");
     /** SMC keys for Apple Silicon GPU temperature sensors. */
--- a/oshi-core-ffm/src/main/java/oshi/hardware/platform/mac/MacSensorsFFM.java
+++ b/oshi-core-ffm/src/main/java/oshi/hardware/platform/mac/MacSensorsFFM.java
@@ -4,6 +4,7 @@
  */
 package oshi.hardware.platform.mac;
 
+import static oshi.ffm.util.platform.mac.SmcUtilFFM.SMC_KEYS_CPU_TEMP_AGGREGATE_AS;
 import static oshi.ffm.util.platform.mac.SmcUtilFFM.SMC_KEYS_CPU_TEMP_AS;
 import static oshi.ffm.util.platform.mac.SmcUtilFFM.SMC_KEY_CPU_TEMP;
 import static oshi.ffm.util.platform.mac.SmcUtilFFM.SMC_KEY_CPU_VOLTAGE;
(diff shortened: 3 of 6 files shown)


dbwiddis:
Verified on **M2 Max** (Mac14,6, macOS 15, JDK 25). Works, and it does more here than fix the M3 case.

### The aggregate keys stay live exactly where the per-core keys park

This is the machine where #3529/#3531 was diagnosed, so it has the idle-gating behaviour. At genuine deep idle the per-core keys are parked at their sentinels, but `TCMb`/`TCMz` read normally:

```
state   | TCMb(avg) TCMz(max) | Tp09      Tp01      Tp05      | getCpuTemperature()
--------------------------------------------------------------------------------------
IDLE    | 41.122    42.369    | 6.700     6.700     4.633     | 41.122
IDLE    | 42.905    53.659    | 6.700     6.700     4.633     | 42.905
IDLE    | 41.013    42.369    | 6.700     6.700     4.633     | 41.013
LOADED  | 48.148    53.159    | 46.859    47.473    45.308    | 48.148
LOADED  | 48.672    53.441    | 47.457    47.871    45.642    | 48.672
IDLE2   | 42.030    43.431    | 6.700     6.700     4.633     | 42.030
IDLE2   | 42.348    46.753    | 6.700     6.700     4.633     | 42.348
```

`getCpuTemperature()` returns the `TCMb` value in every row, confirming the aggregate-first ordering and that it is the average rather than `TCMz`.

**This also recovers the idle behaviour #3531 had to give up.** Before this PR an idle M2 Max reported *unavailable*, correctly, because every per-core key was parked. Now it reports a real ~41 °C. So the aggregate keys are not merely chip-independent — on this hardware they are immune to the cluster gating that caused the original bug.

### Both backends agree exactly

```
IDLE    JNA=40.051  FFM=40.051  delta=0.000
IDLE    JNA=40.930  FFM=40.930  delta=0.000
LOADED  JNA=47.309  FFM=47.309  delta=0.000
LOADED  JNA=47.841  FFM=47.841  delta=0.000
```

- `./mvnw clean install` — full reactor green (1127 / 92 / 77 / 30)
- `./mvnw test -pl oshi-benchmark -Pnative-comparison` — `sensors() [OK]`, 39 passed

### One small correction to the description

The rationale says IOHID reads *"~20-30 °C lower"*. On M2 Max, measured at the same moment, it is a smaller gap:

```
SMC  TCMb            40.5 .. 46.3
HID  PMU tdie* avg   34.3 .. 34.5
```

So ~6-12 °C here rather than 20-30. The direction is right and the conclusion is unaffected — the SMC path needs no private-API bindings and works on every chip in the dataset — but the magnitude may be M3-specific and might be worth softening to "reads lower" rather than a specific range.

LGTM otherwise.

PR #3536, opened 2026-07-27
Discover Apple Silicon GPU temperature SMC keys at runtime

dbwiddis:
Follow-up to #3535, which fixed the CPU side. GPU temperature reads a fixed list of four SMC keys, `SMC_KEYS_GPU_TEMP_AS = [Tg05, Tg0D, Tg0f, Tg0j]`, and returns the **first** plausible one.

**Nothing is broken today** — those four cover every Apple Silicon machine in the [iSMC hardware dataset](https://github.com/dkorunic/iSMC/tree/master/internal/reports). But the list survives by luck, not design: on an M2 Max only `Tg0f` and `Tg0j` exist, and six of that machine's sixteen GPU sensors appear in no published key table at all. It also reports one arbitrary cluster out of the 8–84 a modern chip exposes.

There is no GPU analogue of #3535's `TCMb`/`TCMz` firmware die aggregate, so that solution does not transfer.

### What this does

Binary-searches the SMC's sorted key index for the `Tg` block, keeps the keys matching the GPU naming convention, and reports the hottest plausible reading. ~14 ms once, cached for the JVM lifetime.

Measured across the dataset, decoding raw `flt` bytes for the older report format:

| report | today (first of 4) | this PR (max) | keys found |
|---|---|---|---|
| a18 | 66.4 | **68.1** | 8 |
| m1-pro | 62.9 | **63.6** | 4 |
| m3-max | 41.0 | **43.0** | 32 |
| m4-pro | 56.0 | **57.2** | 22 |
| m5-pro | 38.0 | **41.0** | 42 |
| m1-max / m3-pro | 45.7 / 38.2 | unchanged | 8 / 18 |

### The mask is derived from data

Character 3 is always a digit, but character 4 is uppercase in only 42% of published keys, lowercase in 43%, a digit in 15%:

| mask | this M2 Max | iSMC's 100 known GPU keys |
|---|---|---|
| `^Tg[0-9a-z][A-Z]$` | 6/16 | 42/100 |
| **`^Tg[0-9][0-9A-Za-z]$`** | **16/16** | **100/100** |

The narrower mask misses `Tg0f` — the only one of the original four present on this hardware. Zero false positives: every `Tg*` key in the dataset is a GPU sensor.

### Design notes

- **The index logic lives in `oshi-common`**, not `SmcUtil`, whose static `IOKit.INSTANCE` field makes any of its members unloadable off a Mac. Passing an `IntFunction` keeps the two connection handle types out of the shared code, so the binary search, mask, config parsing and aggregation are unit-tested on **any OS** — that is where the edge cases are (18 tests: block at index 0/last/middle/absent, read failures transient and permanent, unsorted input, duplicates, `#KEY` garbage).
- **Caches which keys *exist*, not which currently read plausibly.** Existence is a hardware property; plausibility is not, since an idle sensor reports a sentinel below ambient. Filtering at discovery time would let one unlucky first call cache an empty set and disable GPU temperature for the JVM lifetime.
- **Not `Memoizer`**, which caches whatever the supplier returned including a failure — a transient `smcOpen` failure would be permanent. A failed discovery falls back without caching so a later call retries.
- **No data-type filter.** `smcGetFloat` already decodes every temperature encoding the SMC uses and returns 0 otherwise, which the plausibility floor rejects. The check cost a second read per key and would have wrongly excluded a sensor reported as `sp78`.
- **Max, not average.** Each cluster exposes a pair of sensors at different die locations. Averaging would combine two different measurements, and separating them needs the per-chip table this avoids. Verified the pair is *not* instantaneous/peak-hold — both members rose 44.8/51.4 → 56.1/62.6 under load and fell together afterwards, so the max tracks the hotter location rather than a latched peak.

`SMC_KEYS_GPU_TEMP_AS` is kept as the fallback when the index cannot be read, widened from 4 to the 14 keys most common across published M1–M5 and A18 dumps. `oshi.os.mac.sensors.gputemperature.keys` overrides discovery entirely, parsed lazily so `GlobalConfig.set()` still takes effect.

The Intel IOAccelerator `Temperature(C)` fallback is absent on Apple Silicon, so it is now tried once per card then latched off rather than repeating an IOKit registry walk every call.

### Why not IOReport or IOHID

Both were investigated on hardware and eliminated, detail in #3534:

- **IOReport `GPU Stats → Temperature`** is a driver stub. All 64 channels read 0 as root, under GPU load, and in the raw `RawElements` blob, byte-identical across samples.
- **IOHID** measures the wrong thing: under Metal load the SMC GPU sensors rose +21 °C while the best HID sensor moved +5 °C. It is a damped board/PMU sensor, and the gap widens with load, so it is not a correctable offset.

### Verification

- Full reactor: 1146 + 95 + 80 + 30 tests, 0 failures. Spotless, checkstyle, forbidden-apis, javadoc clean.
- `./mvnw test -pl oshi-benchmark -Pnative-comparison` — `sensors()` and the new `gpuTemperatureKeys()` both pass. The latter asserts the two backends discover **identical** key lists; since each runs its own binary search, list equality is the sharpest available parity check.
- M2 Max, idle → Metal load → idle: `getTemperature()` = 47.8 → 66.5 → 55.1 °C. Discovery finds all 16 keys in both backends.

## Summary by CodeRabbit

* **New Features**
  * macOS now discovers Apple Silicon GPU temperature sensor keys dynamically for broader hardware support.
  * GPU temperature selection now prefers the hottest available GPU cluster.
  * Added an optional configuration (`oshi.os.mac.sensors.gputemperature.keys`) to override which SMC temperature keys are used.

* **Bug Fixes**
  * Improved reliability when GPU temperature data is unavailable by reducing repeated lookups and tightening fallback behavior when sensors/perf stats are missing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * [#3527](https://github.com/oshi/oshi/pull/3527): Consolidate the SLF4J log level dispatch duplicated in `ExceptionUtil` and `PerfDataUtil` into a new `oshi.util.LogUtil`, whose `isEnabled` method offers the slf4j 1.x-compatible equivalent of `Logger#isEnabledForLevel` - [@dbwiddis](https://github.com/dbwiddis).
 * [#3531](https://github.com/oshi/oshi/pull/3531): Stop reporting an implausible CPU or GPU temperature on Apple Silicon. An idle core cluster is power-gated and its die sensors then report a fixed parked value below room ambient, which was previously returned as a reading; `Sensors#getCpuTemperature` now reports 0 (unavailable) instead - [@dbwiddis](https://github.com/dbwiddis).
 * [#3535](https://github.com/oshi/oshi/pull/3535): Fix `Sensors#getCpuTemperature` returning 0 on Apple Silicon chips whose per-core SMC keys differ from the hardcoded set (e.g. the M3 Pro, where all five are absent). Prefer the chip-independent, firmware-computed CPU-die aggregate keys `TCMb` (die average) and `TCMz` (die max) before the per-core keys, reporting the die average - [@dbwiddis](https://github.com/dbwiddis).
+* [#3536](https://github.com/oshi/oshi/pull/3536): Discover Apple Silicon GPU temperature SMC keys at runtime instead of reading a fixed list of four chip-specific keys, and report the hottest GPU cluster rather than the first one found. Adds `oshi.os.mac.sensors.gputemperature.keys` to override the discovered keys - [@dbwiddis](https://github.com/dbwiddis).
 
 # 7.4.0 (2026-07-08), 7.4.1 (2026-07-18), 7.4.2 (2027-07-24)
 
--- a/oshi-benchmark/src/test/java/oshi/comparison/NativeComparisonTest.java
+++ b/oshi-benchmark/src/test/java/oshi/comparison/NativeComparisonTest.java
@@ -20,6 +20,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledOnOs;
 import org.junit.jupiter.api.condition.OS;
 
+import oshi.ffm.util.platform.mac.SmcUtilFFM;
 import oshi.hardware.CentralProcessor;
 import oshi.hardware.ComputerSystem;
 import oshi.hardware.Display;
@@ -48,6 +49,7 @@ import oshi.software.os.OperatingSystem;
 import oshi.util.GlobalConfig;
 import oshi.util.PlatformEnum;
 import oshi.util.Util;
+import oshi.util.platform.mac.SmcUtil;
 
 /**
  * Compares JNA and FFM implementations of the OSHI API to detect regressions, missing implementations, or incorrect
@@ -337,6 +339,16 @@ class NativeComparisonTest {
     }
 
     // ---- Hardware: Graphics Cards ----
+
+    /**
+     * GPU temperature keys are discovered from the SMC at runtime by each backend independently, so comparing the
+     * discovered lists is the sharpest available check that the two binary searches agree.
+     */
+    @Test
+    @EnabledOnOs(OS.MAC)
+    void gpuTemperatureKeys() {
+        assertThat(SmcUtilFFM.getGpuTemperatureKeys()).isEqualTo(SmcUtil.getGpuTemperatureKeys());
+    }
 
     @Test
     void graphicsCards() {
--- a/oshi-common/src/main/java/module-info.java
+++ b/oshi-common/src/main/java/module-info.java
@@ -58,6 +58,7 @@ module com.github.oshi.common {
     exports oshi.software.os;
     exports oshi.spi;
     exports oshi.util;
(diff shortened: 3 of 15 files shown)


dbwiddis:
Both valid, both fixed in `2e5e09d`.

### `Tg0f` missing from the fallback — a regression I introduced

This one was a real bug, not a nit. Widening the fallback list used "keys appearing on at least 4 of the 13 machines in the corpus", and `Tg0f` falls below that threshold — so **widening the list dropped one of the four keys OSHI shipped before this PR**, and specifically the only one of the four present on the M2 Max this work was verified on. Whenever discovery could not complete, the new fallback would have reported *less* than the old hardcoded list on that hardware.

The list now leads with all four historical keys and then the frequency-derived additions:

```java
"Tg05", "Tg0D", "Tg0f", "Tg0j",   // the four read before discovery existed
"Tg0C", "Tg04", "Tg0K", "Tg0L", "Tg0d", "Tg0e", "Tg1k", "Tg0X", "Tg0S", "Tg0y", "Tg0z"
```

Added `testFallbackKeysIncludeTheHistoricalFour` to both twins so the fallback can never again be narrowed below the previous behaviour. Ordering is cosmetic now that the aggregation is a max rather than first-match, but leading with the historical four makes the superset property obvious at a glance.

### Latching on a null `queryPerfStats()`

Agreed, and the distinction is the one the plan intended but the code overshot. A null return means no matching `IOAccelerator` entry was found, which can be transient — a driver reset, an unplugged eGPU — whereas a dictionary that is present but has no `Temperature(C)` key is structural. Latching the former would disable the sensor for the rest of the JVM on a recoverable condition.

Now latched only on the structural case, which is also the one that actually occurs on Apple Silicon: `PerformanceStatistics` is present there and simply has no `Temperature(C)` key, so the common path still latches after one attempt.

Full reactor green (1146 / 96 / 81 / 30), and `-Pnative-comparison` still passes `sensors()` and `gpuTemperatureKeys()`.

dbwiddis:
Fixed in `a9f2e3c`. Valid, though the more interesting problem is not the flake window.

The assertion claimed to prove the result is cached, but it passes **vacuously on exactly the path worth checking**: when discovery cannot complete, `getGpuTemperatureKeys()` deliberately returns the shared `SMC_KEYS_GPU_TEMP_AS` constant *without* caching, so both calls return the same instance and `sameInstance` holds even though nothing was cached. The test would have reported success on a machine where caching never happened.

The flake window you identified is real too, just narrow — a first call that fails followed by one that succeeds returns different instances.

Now:

```java
List<String> second = SmcUtil.getGpuTemperatureKeys();
assertThat("Repeated calls must agree", second, is(equalTo(keys)));
// A completed discovery is cached; a failed one deliberately is not, and returns the shared fallback
// constant instead. Asserting identity unconditionally would pass vacuously in that case.
if (!keys.equals(SmcUtil.SMC_KEYS_GPU_TEMP_AS)) {
    assertThat("A completed discovery must be cached, not repeated", second, is(sameInstance(keys)));
}
```

Content stability is checked unconditionally; identity only where caching was supposed to occur. Mirrored in both twins. Full reactor green (1146 / 96 / 81 / 30).

PR #3541, opened 2026-07-28
Derive Apple Silicon chip details at runtime

### LINKED ISSUE #3537: macOS: derive Apple Silicon chip details at runtime instead of hardcoding per-chip tables
Part of the recurring "macOS changed things again" maintenance. Split out of #3534, where it surfaced during the SMC sensor work but was left out of scope — the sensor work is chip-agnostic by design and does not depend on any of this.

**The framing that matters: none of this needs new per-chip table entries, and none of it needs hardware I don't have.** Everything required is already read by `MacCentralProcessor` and then discarded. Two of the three problems below are cases of the code fetching the right data and ignoring it in favour of a hardcoded list that stops at M4.

## 1. P-core/E-core classification is hardcoded, and stops at M4

`MacCentralProcessor.java:59-61`:

```java
private static final Set<String> ARM_P_CORES = Stream
        .of("apple,firestorm arm,v8", "apple,avalanche arm,v8", "apple,everest arm,v8", "apple,donan arm,v8")
        .collect(Collectors.toSet());
```

Firestorm/Avalanche/Everest/Donan are M1/M2/M3/M4. An M5 or A-series P-core codename is not in the set, so `efficiency` is 0 for **every** core (line 331). That is not only a wrong `PhysicalProcessor#getEfficiency()` — `queryCurrentFreq()` branches on `getEfficiency() > 0`, so per-core frequency reporting degrades with it.

But macOS already reports this directly, chip-independently, and the code already reads it. From an M2 Max:

```
hw.nperflevels: 2
hw.perflevel0.physicalcpu: 8      hw.perflevel0.name: Performance
hw.perflevel1.physicalcpu: 4      hw.perflevel1.name: Efficiency
```

`queryProcessorTopology()` reads `hw.nperflevels` on line 328 — and then uses it only to size the cache list, while line 331 falls back to the hardcoded name set. Confirmed on an M2 Max, cores are ordered efficiency-first:

```
core 0-3    efficiency 0    apple,blizzard arm,v8
core 4-11   efficiency 1    apple,avalanche arm,v8
```

which lines up exactly with `perflevel1.physicalcpu = 4` efficiency cores followed by `perflevel0.physicalcpu = 8` performance cores.

So `ARM_P_CORES` should be **deleted**, not extended. A derivation has options, and the implementer should pick: group cores by their distinct `compatible` string (already in `compatMap`) and label the groups by matching group size against the `perflevel*.physicalcpu` counts; or rely on the efficiency-first ordering. Note the count-matching approach is ambiguous on a chip with an even split such as the base M1 (4+4), where ordering or another tiebreak is needed — worth handling explicitly rather than by accident.

## 2. The microarchitecture table is redundant with data already read

`oshi.architecture.properties` stops at M4:

```
apple.0x1b588bb3=ARM64 SoC: Firestorm + Icestorm
apple.0xda33d83d=ARM64 SoC: Avalanche + Blizzard
apple.0x8765edea=ARM64 SoC: Everest + Sawtooth
apple.0x5f4dea93=ARM64 SoC: M3 Pro (Lobos)
apple.0x6f5129ac=ARM64 SoC: M4 (Donan)
```

so `getMicroarchitecture()` returns `Constants.UNKNOWN` on M5 and on any A-series part.

The entries are, however, mechanically derivable from data the code already has. On this M2 Max the reported value is `ARM64 SoC: Avalanche + Blizzard`, and the two `compatible` strings are `apple,avalanche arm,v8` and `apple,blizzard arm,v8` — the string is exactly the performance codename plus the efficiency codename, capitalised. `compatMap` is built on line 327 and currently used only as a display string on the `PhysicalProcessor`.

Deriving it would cover every future chip with no table maintenance, while keeping the properties file as an override for parts whose marketing name differs from the codename pair (the `M3 Pro (Lobos)` entry is that case). Worth deciding whether the derived value or the table wins when both exist.

## 3. Latent, not currently reachable: the brand-string fallback misclassifies

When `hw.cpufamily` returns 0, `queryProcessorId()` parses the brand string:

```java
int mSeries = ParseUtil.getFirstIntValue(cpuName);
switch (mSeries) {
    case 2:  family = M2_CPUFAMILY; break;
    case 3:  family = M3_CPUFAMILY; break;
    case 4:  family = M4_CPUFAMILY; break;
    default: family = M1_CPUFAMILY;
}
```

`getFirstIntValue` takes the first run of digits, so `"Apple M5"` → 5 → `default` → reported as **M1**, and `"Apple A18 Pro"` → 18 → `default` → also **M1**.

Being straight about severity: `hw.cpufamily` is non-zero on real macOS, so this does not currently execute on hardware. It is a latent trap rather than a live bug. If items 1 and 2 are done by derivation, this fallback matters even less — but silently claiming M1 for anything unrecognised is still the wrong default.

## 4. Housekeeping

`M3_PRO_CPUFAMILY` (`MacCentralProcessor.java:74`) is declared but never referenced in code; it exists only as a key in the properties file. The `M1`/`M2`/`M3`/`M4` constants are likewise used only by the dead fallback in item 3.

## Suggested shape

1. Derive P/E from `hw.perflevel*` plus the existing `compatMap`; delete `ARM_P_CORES`.
2. Derive the microarchitecture string from the `compatible` codename pair, keeping the properties file as an override.
3. Stop defaulting unrecognised chips to M1 in the brand-string fallback.
4. Remove the constants that become unused.

Items 1 and 2 are both verifiable on any Apple Silicon Mac — a correct derivation must keep reporting `efficiency` 0/1 and `ARM64 SoC: Avalanche + Blizzard` on an M2 Max, and equivalents on M1/M3/M4, so existing hardware is enough to prove it. That is a better position than adding table rows for chips nobody in the project owns.

dbwiddis:
Closes #3537.

macOS chip identification came from hardcoded tables that stop at the M4, so an M5 or any A-series part is misreported. Everything needed to avoid those tables is already read from the IORegistry and then discarded.

## Core classification

`ARM_P_CORES` was a set of four core codenames — firestorm/avalanche/everest/donan, i.e. M1–M4. An unlisted codename left `efficiency` **0 for every core**, which is both a wrong `PhysicalProcessor#getEfficiency()` and a degraded `queryCurrentFreq()`, since that branches on `getEfficiency() > 0`.

The `cpuN` nodes already walked for `compatible` also carry **`cluster-type`**, a single letter naming the cluster:

```
cpu0-3    cluster-type = <"E">    compatible = <"apple,blizzard","ARM,v8">
cpu4-11   cluster-type = <"P">    compatible = <"apple,avalanche","ARM,v8">
```

That answers the question directly, for any chip, with no table. It's read in the same pass as `compatible`, and `RegistryEntry.getByteArrayProperty` is already generic — so **neither the JNA nor the FFM backend changes**.

Where `cluster-type` is unavailable the codenames still group the cores, ordered by core number (macOS numbers efficiency cores first) and cross-checked against `hw.perflevel0.physicalcpu` where that count is unambiguous. That reproduces exactly what the deleted table produced, so a machine without `cluster-type` is no worse off than today.

`hw.nperflevels` is deliberately **not** used to detect heterogeneity — it did not exist before macOS 12, so requiring it would misclassify an M1 running macOS 11. It keeps its existing job of sizing the cache loop.

## Microarchitecture

Assembled from the same codenames, so an unlisted chip is described rather than reported `unknown`. The architecture table is consulted **first and still wins**, because an entry may carry a curated name a derivation cannot reproduce — `M3 Pro (Lobos)` and `M4 (Donan)` are exactly that case. Every chip currently in the table is byte-for-byte unchanged.

This needs a new `ProcessorIdentifier` constructor. japicmp reports it as `+++ NEW CONSTRUCTOR` with `Semantic versioning suggestion: 0.0.1`, so **no version bump is required** — verified by running `verify`, not assumed.

## Also

Removes the brand-string fallback that reported any unrecognised chip as an M1 (`"Apple M5"` → 5 → `default` → M1), and the family constants only it used. A missing `hw.cpufamily` now leaves the family zero, which misses the table and falls through to the derived description.

⚠️ This deletes `protected static final` constants (`M1`/`M2`/`M3`/`M3_PRO`/`M4_CPUFAMILY`) from a public but non-`@PublicApi` class. Nothing in the project referenced them; a downstream subclass theoretically could.

## Verification

**Live on an M2 Max — identical to before:**

```
microarch      = ARM64 SoC: Avalanche + Blizzard
core 0-3       eff=0  apple,blizzard arm,v8
core 4-11      eff=1  apple,avalanche arm,v8
currentFreq    E 2.42 GHz / P 3.70 GHz
```

Those `eff` values now come from `cluster-type`, since `ARM_P_CORES` is gone — so the live run exercises the new path end to end.

**15 hardware-free fixture tests** covering M2 Max 8+4, M1's ambiguous 4+4 even split, `cluster-type` absent, partial `cluster-type` propagated by codename, `cluster-type` with no codename, perf-level count overriding core ordering, unrecognised cluster letters, homogeneous cores, no properties at all, and a fabricated future chip (`apple,hydra`/`apple,coyote`) that classifies correctly and derives `ARM64 SoC: Hydra + Coyote` — the case that fails without this change.

The strongest check reads the shipped `oshi.architecture.properties` and asserts the derived string **equals the entry already published** for M1, M2 and M3. It asserts those entries are non-null first, so it cannot pass vacuously if the table fails to load.

Full gate clean on `oshi-common,oshi-core,oshi-core-ffm`, clean full-reactor `install`, 0 tests failed. `NativeComparisonTest` passes including `processorIdentifier()` and `processorTopology()`, so JNA/FFM parity holds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Apple Silicon efficiency core classification on newer chips by using runtime per-core hardware data.
* **Improvements**
  * More accurate Apple processor microarchitecture detection, using derived values when the curated architecture table has no matching entry.
* **Documentation**
  * Clarified that Apple entries in architecture configuration are optional and override derived detection when present.
* **Tests**
  * Added expanded unit coverage for Apple core classification and microarchitecture precedence/fallback behavior.


dbwiddis:
@coderabbitai thanks — addressed in 486633d.

**CHANGELOG ordering** — fixed, `#3541` now follows `#3536`.

**Simplify the perf-level tie-break** — fixed, adopted essentially as proposed. `firstMatches != (...) && firstMatches` is now a plain conjunction.

**`getIdString()` null check** — skipping. `deriveMicroarchitecture` is package-private and its only call site passes `getPhysicalProcessors()`, whose elements this class constructs with a non-null `idString` (`""` at minimum, via `props.getA() == null ? "" : ...`). The guard would be unreachable.

---

Reviewing this on a second machine also turned up two real bugs in `deriveEfficiencyClasses` that the nitpick pass missed, both now fixed:

1. **Partial `cluster-type` readings were discarded.** Strategy 3 opened with `efficiencyMap.clear()`, and it also aborts when any core lacks a codename — so where some `cpuN` nodes expose no readable properties, cores that had reported `cluster-type = P` fell through to the class 0 default:

   ```
   cores 0-3 = apple,everest + cluster-type P; cores 4-7 = no properties
   before:  0 0 0 0 0 0 0 0     <- every P core misclassified
   after:   1 1 1 1 0 0 0 0
   ```

   That was worse than the codename table this replaces, and it re-introduced the degraded `queryCurrentFreq()` the PR sets out to fix. `cluster-type` is exact, so those values are now kept and only the remainder defaults; codename grouping applies only when `cluster-type` classified nothing at all.

2. **Propagation could overwrite a core's own reading.** The strategy-2 loop revisited every core, not just unclassified ones, so a core reporting its own `cluster-type = E` could be overwritten to 1 by a sibling sharing its codename — contradicting the javadoc's "applied to the **other** cores". It now skips already-classified cores.

Two regression tests added for these.

Verified on an M3 Pro (6+6), which is a different chip from the M2 Max in the PR description and exercises the even-split path where `hw.perflevel0.physicalcpu` is ambiguous and core ordering decides: E cores 0-5 @ 2.75 GHz, P cores 6-11 @ 4.06 GHz, `ARM64 SoC: M3 Pro (Lobos)` from the table as intended. `oshi-common` 1168 tests pass, Spotless/Checkstyle clean, and `NativeComparisonTest` `processorIdentifier()`/`processorTopology()` pass so JNA/FFM parity holds.

Issue #3593, opened 2026-08-10
PROPOSAL: Adopt jSpecify nullability annotations


dbwiddis:
## Summary

I'd like to discuss adding [jSpecify](https://jspecify.dev/) (`org.jspecify:jspecify`) as an optional, compile-time-only dependency, and annotating nullability across OSHI — starting with the public API and extending inward over time.

**This is a proposal for discussion, not a PR.** I'd like to leave it open at least a week. I'm specifically re-opening a decision we made in 2020, so I want to lay out what has and hasn't changed.

## We said no to this in 2020

In #1160 I added `@ThreadSafe` / `@Immutable` / `@NotThreadSafe` from JSR-305, then backed them out in #1161 and defined our own instead. Two reasons:

1. **JPMS.** `javax.annotation` was split across multiple artifacts, which Java 9+ modules reject.
2. **Licensing.** The `javax.*` packaging ran afoul of the Oracle Binary Code License.

@hazendaz — you gave the decisive advice there, from the SpotBugs side:

> I would not use JSR305 right now or any of the numerous duplicate solutions. After 'jakarta' is officially
> released in June and sometime after when all containers fully support it, the JSR305 is likely to be
> consumed into jakarta annotations.

I agreed and wrote:

> I can probably just define those within OSHI (to preserve their presence in the javadocs) and hold off on
> any more expansive use of the null-related, or check-return, or similar until JSR305 is in Jakarta. Or,
> better, since I'm looking at an API switch, perhaps just start making use of `Optional`.

Six years on, neither of those happened. Jakarta's common-annotations never took on the nullability annotations. And we never adopted `Optional` in the API — there are currently zero uses of it in `oshi.hardware` or `oshi.software.os`. Our own annotations still carry the javadoc line *"a temporary workaround until it is available in `jakarta.annotations`"*, which is now stale.

So the "wait" was correct, but the thing we were waiting for arrived under a different name.

## What changed: jSpecify

jSpecify is the effort that actually shipped, and it clears both 2020 objections:

| 2020 objection | jSpecify 1.0.1 |
|---|---|
| `javax.annotation` split package breaks JPMS | `org.jspecify.annotations`, single artifact, real `module-info` (module `org.jspecify`) |
| Oracle BCL covers `javax.*` | Apache 2.0 |

It is deliberately tiny — **four annotations, nothing else**:

| Annotation | Target | Purpose |
|---|---|---|
| `@Nullable` | `TYPE_USE` | this type usage may be null |
[...]

PR #3632, opened 2026-08-15
Make the macOS hardware and operating-system layers' nullability contracts honest

dbwiddis:
Seventh step of draining `oshi-core` and `oshi-core-ffm`, and the last macOS group. Covers the
hardware components and the operating-system layer in both implementations, including the
`MacGpuStats` twins deferred from #3626. **Reactor findings 147 → 94.**

### Mostly declarations catching up with documentation

The IO registry views report an absent parent, an exhausted iterator, or a card with no accelerator
entry by returning null, and each already said so. `MacGpuStats.queryPerfStats` even carries a
paragraph explaining why null and an empty map are *different answers* there — `getTemperature`
latches a card as having no sensor only when the dictionary was present and the key was absent,
never when the dictionary itself was missing, which can be transient. The display synthesis, the
medium-type traversal and the localized display name report failure the same way, as does
`getProcess` for a process that is not running, which `OperatingSystem` has always declared
`@Nullable`.

### Two contracts the analyzer read differently from a reader

- **`IOKitProvider.withMatchingService`** returns null when no service matches, *whatever `T` is*,
  and its extractor may itself report an absent property with null. `<T extends @Nullable Object> T`
  said neither of those things; it is now `<T> @Nullable T` taking a
  `Function<RegistryEntry, @Nullable T>`.
- **`RegistryEntry.lookupControllerVidPid`** returns a two-element array whose *elements* may each be
  absent — `@Nullable String[]`, not `String @Nullable []`. The javadoc says "either element may be
  `null`"; the first annotation I wrote said the array itself could be null, and NullAway caught it.

`MacCentralProcessor.platformExpert` tested its result with `Util.isBlank`, which the analyzer cannot
see through; the explicit null-or-empty check is the same test that method performs.

### Verified at runtime on both backends

Run end to end on an M2 Max after the change:

| section | JNA vs FFM |
|---|---|
| Displays | identical |
| USB Devices | identical |
| Graphics Cards | identical but the object hashcode |
| Disks | identical but live read/write counters |

`Apple M2 Max, vendor=Apple (0x106b), vRam=68719476736` on both. The USB tree exercises
`lookupControllerVidPid` and the registry iterator, which are precisely what this PR re-declares, so
those paths are executed rather than merely compiled.

### No behavior change

No CHANGELOG entry. Verified by diffing against master and cancelling every line whose only
difference is an annotation: 19 of the 23 files have no residue, and the rest is the `isBlank`
rewrite above, two array elements read into locals because NullAway does not refine array elements,
and one signature rewrapped across lines.

Full gate on macOS: `spotless:apply sortpom:sort spotless:check checkstyle:check install
forbiddenapis:check javadoc:javadoc` plus `-Perror-prone clean install`. All modules green, 1561
tests, 0 failures.

Part of #3593.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved macOS processor manufacturer reporting: whitespace-only values are now preserved, while missing values still use the standard fallback.
  - Improved handling of unavailable macOS hardware, display, storage, USB, process, network, and power-source information.

- **Refactor**
  - Clarified nullable results throughout macOS hardware and operating-system integrations, improving API accuracy and reliability without changing expected behavior.

### DIFF
--- a/oshi-common/src/main/java/oshi/hardware/common/platform/mac/IOKitProvider.java
+++ b/oshi-common/src/main/java/oshi/hardware/common/platform/mac/IOKitProvider.java
@@ -23,7 +23,7 @@ public interface IOKitProvider {
      * @param extractor   function applied to the matched entry; receives a non-null {@link RegistryEntry}
      * @return the extracted value, or {@code null} if the service was not found
      */
-    <T extends @Nullable Object> T withMatchingService(String serviceName, Function<RegistryEntry, T> extractor);
+    <T> @Nullable T withMatchingService(String serviceName, Function<RegistryEntry, @Nullable T> extractor);
 
     /**
      * Iterates all IORegistry services matching a class name, invoking a consumer for each entry. Entries and the
--- a/oshi-common/src/main/java/oshi/hardware/common/platform/mac/MacCentralProcessor.java
+++ b/oshi-common/src/main/java/oshi/hardware/common/platform/mac/MacCentralProcessor.java
@@ -157,7 +157,7 @@ public abstract class MacCentralProcessor extends AbstractCentralProcessor {
             byte[] data = entry.getByteArrayProperty("manufacturer");
             return data != null ? ParseUtil.decodeNulTerminated(data, StandardCharsets.UTF_8) : null;
         });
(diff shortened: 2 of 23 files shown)


dbwiddis:
@coderabbit review

PR #3652, opened 2026-08-18
Report the OpenTelemetry reserved filesystem state in oshi-metrics

dbwiddis:
Adds the third OpenTelemetry filesystem state, `reserved`, to `oshi-metrics`.

The [filesystem semantic conventions](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#filesystem-metrics) enumerate exactly three values for `system.filesystem.state` — `used`, `free`, `reserved` — and footnote [1] says summing `system.filesystem.usage` across states SHOULD equal `system.filesystem.limit`. We registered two of the three, and computed `used` as `total - usable`, which absorbed the reserve into `used` and left nothing to report as `reserved`.

#3651 gave `OSFileStore` the ordering guarantee `0 <= usable <= free <= total`, which makes the reserve directly expressible: it is exactly `free - usable`, the unused space that is not available to the calling process.

### What changed

The three states now partition the filesystem:

| state | before | after |
|---|---|---|
| `used` | `total - usable` | `total - free` |
| `free` | `usable` | `usable` (unchanged) |
| `reserved` | — | `free - usable` |

They sum to `total` exactly, so `system.filesystem.usage` sums to `system.filesystem.limit` and `system.filesystem.utilization` sums to 1.0.

`free` was already correct: the convention's `free` is `statvfs`'s `f_bavail`, which is `getUsableSpace()`, not `f_bfree`. Only `used` needed redefining, and it had to be redefined — adding `reserved` while leaving `used` alone would over-count by the reserve and break the sum rule.

### User-visible behavior change

`state=used` drops by the reserve on filesystems that hold one back — ext4's default 5% (`tune2fs -m`), and a Windows volume with a quota on the caller. That space is now reported as `reserved` rather than silently counted as used, so an alert on `used` will read slightly lower and one on `used + reserved` reproduces the old number.

It is unchanged on filesystems that reserve nothing at this layer. ZFS and APFS report `f_bfree == f_bavail`, so `reserved` is 0 there. Verified on this machine (macOS/APFS), where every mount reports `free == usable`:

```
/                            total=  494384795648 free=  109363933184 usable=  109363933184 reserved=  0
/System/Volumes/Data         total=  494384795648 free=  109363933184 usable=  109363933184 reserved=  0
/System/Volumes/xarts        total=     524288000 free=     504127488 usable=     504127488 reserved=  0
```

CHANGELOG carries both halves: the new state under *New Features*, the `used` redefinition under *Behavior Changes*.

### Implementation note

Seven near-identical builder chains would have been a lot of repeated tag plumbing around three lines of arithmetic, so the gauge registration collapses into `registerUsage`/`registerUtilization` helpers taking a `Tag` and a `ToDoubleFunction`, with the per-sample `updateAttributes()` refresh factored into one place. The three formulas now sit side by side, where the sum rule is checkable at a glance.

### Testing

New `FileSystemMetricsTest` binds a fixed-value `OSFileStore` and asserts the arithmetic directly — the sum rules, the reserve split, `reserved == 0` when `free == usable`, a zero-capacity filesystem not dividing by zero, and inconsistent OS values not producing a negative state. The live tests in `OshiMetricsTest` cannot assert a sum: a real filesystem's space moves between the three samples. They are extended to cover registration and non-negativity of the new state.

Full gate green locally on macOS: `spotless:apply sortpom:sort checkstyle:check install forbiddenapis:check javadoc:javadoc`, 0 tests failed, `FileSystemMetricsTest [OK]` and `OshiMetricsTest [OK]`.

## Summary by CodeRabbit

* **New Features**
  * Added filesystem metrics for reserved space alongside used and free space.
  * Improved filesystem usage and utilization calculations across Unix and Windows systems.
  * Ensured consistent filesystem sampling within the metrics refresh window.

* **Bug Fixes**
  * Prevented negative metric values and handled zero-capacity filesystems safely.
  * Improved cross-platform ordering of filesystem space values.

* **Documentation**
  * Updated filesystem metrics documentation and the 7.5.1 changelog.
  * Documented Maven Central distribution updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,19 @@
 # 7.5.1 (in progress)
 
-* The `oshi-dist` zip is no longer published to Maven Central; download it from the [GitHub release](https://github.com/oshi/oshi/releases) instead - [@dbwiddis](https://github.com/dbwiddis).
+The `oshi-dist` zip is no longer published to Maven Central; download it from the
+[GitHub release](https://github.com/oshi/oshi/releases) instead.
 
-* `OSFileStore` now guarantees `0 <= getUsableSpace() <= getFreeSpace() <= getTotalSpace()` on every platform. The three values are read by separate queries, so on a ZFS dataset or a swap-backed `tmpfs` they could previously contradict each other; they are now clamped downward to restore the ordering - [@dbwiddis](https://github.com/dbwiddis).
+##### New Features
 
-* Your contribution here!
+* [#3652](https://github.com/oshi/oshi/pull/3652): `oshi-metrics` reports the OpenTelemetry `reserved` state for `system.filesystem.usage` and `system.filesystem.utilization`, alongside the existing `used` and `free`. The three states partition the filesystem, so the `usage` gauges sum to `system.filesystem.limit` and the `utilization` gauges sum to 1.0. `reserved` is unused space unavailable to the calling process: the superuser reserve many UNIX filesystems hold back, or the caller's disk quota on Windows - [@dbwiddis](https://github.com/dbwiddis).
+
+##### Behavior Changes
+
+* [#3652](https://github.com/oshi/oshi/pull/3652): `oshi-metrics` computes the `used` state of `system.filesystem.usage` and `system.filesystem.utilization` as `total - free` rather than `total - usable`, so it no longer counts a filesystem's reserve as used; that space is now reported as `reserved`. Values drop by the reserve on filesystems that have one, such as ext4's default 5%, and are unchanged on ZFS and APFS, which reserve nothing at this layer - [@dbwiddis](https://github.com/dbwiddis).
+
+##### Bug Fixes and Improvements
+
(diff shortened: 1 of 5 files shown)


dbwiddis:
Addressed the review:

**Shared reading across gauges** (`ebd5846`) — good catch, and it applied before this PR too: every gauge called `updateAttributes()` when sampled, so a scrape queried each filesystem five times (seven after adding `reserved`) and each state was measured against a different reading. That is fatal to the partition rule this PR documents, so the refresh is now memoized per file store at `Memoizer.defaultExpiration()`. All of a filesystem's gauges within a scrape see one snapshot, and six of the seven queries go away. `allGaugesOfOneFilesystemShareASingleReading` binds a file store whose space changes on every refresh, and only sums to the limit if the reading is shared.

`DiskMetrics` and `NetworkMetrics` have the same per-gauge refresh shape and would benefit from the same treatment; leaving those for a follow-up since neither documents a cross-gauge invariant.

**CHANGELOG heading levels** — skipping. `#####` under `#` is the established convention for every release section in this file (see 7.5.0 and every section below it); changing three headings to `##` would make the in-progress section the only one that differs. markdownlint is not part of the build.

PR #2951, opened 2025-08-25
Add connection port to the Display class

ayonization:
Added connection port information to the display class.

Currently implemented only for unix, please suggest how this can be implemented on windows and mac. Will update the PR accrodingly.

Brief description of the unix implementation - 

Xrandr verbose returns a list of strings with all the relavant display information.

The connection port information looks like this in the output : 

`DP-1 connected primary 1920x1080+0+0 (0x48) normal (normal left inverted right x axis y axis) 527mm x 296mm`

If you come across the word "connected" the word preceeding it is the connection port.

Also, since xrandr verbose maintains the order of displays being listed, the connection ports list will come in the same order as that of the edid arrays.

Please review the PR and Im open to any suggestions you might have !

PR #3660, opened 2026-08-18
Add getConnectionPort to the Display API

ayonization:
## Summary by CodeRabbit

- **New Features**
  - Added display device-port reporting through the public display API.
  - Added optional output-name reporting for connected displays.
  - Improved Linux and Unix display detection with connector IDs, port names, and EDID-based matching.
  - Display information now resolves more reliably through system graphics tools when available.
  - Unsupported or unavailable metadata returns clear default values without interrupting display detection.

- **Documentation**
  - Added release notes describing the new display metadata fields and platform behavior.

### DIFF
--- a/oshi-common/src/main/java/oshi/hardware/Display.java
+++ b/oshi-common/src/main/java/oshi/hardware/Display.java
@@ -4,6 +4,8 @@
  */
 package oshi.hardware;
 
+import java.util.Optional;
+
 import oshi.annotation.PublicApi;
 import oshi.annotation.concurrent.Immutable;
 
@@ -53,4 +55,22 @@ public interface Display {
      * @return A {@link DisplayInfo} holding the display's decoded attributes.
      */
     DisplayInfo getDisplayInfo();
+
+    /**
+     * The system-level device identification for this display. On Linux this is the DRM connector name extracted from
+     * the sysfs directory (e.g. {@code HDMI-A-1}, {@code eDP-1}, {@code DP-2}).
+     *
+     * @return The device port identifier, or {@code "unknown"} if not available.
+     */
+    String getDevicePort();
+
+    /**
+     * The X11 output name for this display as reported by {@code xrandr} (e.g. {@code HDMI-1}, {@code DP2}). This is
+     * the name to pass to {@code xrandr --output}. Only meaningful on systems running an X server.
+     *
+     * @return An {@link Optional} containing the xrandr output name, or empty if not available.
+     */
+    default Optional<String> getOutputName() {
+        return Optional.empty();
+    }
 }
--- a/oshi-common/src/main/java/oshi/hardware/Display.java
+++ b/oshi-common/src/main/java/oshi/hardware/Display.java
@@ -8,6 +8,7 @@ import java.util.Optional;
 
 import oshi.annotation.PublicApi;
 import oshi.annotation.concurrent.Immutable;
+import oshi.util.Constants;
 
 /**
  * Display refers to the information regarding a video source and monitor identified by the EDID standard.
@@ -58,11 +59,14 @@ public interface Display {
 
     /**
      * The system-level device identification for this display. On Linux this is the DRM connector name extracted from
-     * the sysfs directory (e.g. {@code HDMI-A-1}, {@code eDP-1}, {@code DP-2}).
+     * the sysfs directory (e.g. {@code HDMI-A-1}, {@code eDP-1}, {@code DP-2}), or the {@code xrandr} output name when
+     * DRM sysfs is not available.
      *
-     * @return The device port identifier, or {@code "unknown"} if not available.
+     * @return The device port identifier, or {@link Constants#UNKNOWN} if not available.
      */
-    String getDevicePort();
+    default String getDevicePort() {
+        return Constants.UNKNOWN;
+    }
 
     /**
      * The X11 output name for this display as reported by {@code xrandr} (e.g. {@code HDMI-1}, {@code DP2}). This is
--- a/oshi-common/src/main/java/oshi/hardware/common/AbstractDisplay.java
+++ b/oshi-common/src/main/java/oshi/hardware/common/AbstractDisplay.java
@@ -8,6 +8,7 @@ import oshi.annotation.concurrent.Immutable;
 import oshi.hardware.Display;
 import oshi.hardware.DisplayInfo;
 import oshi.hardware.DisplayInfoImpl;
+import oshi.util.Constants;
 
 /**
  * A Display
@@ -16,14 +17,35 @@ import oshi.hardware.DisplayInfoImpl;
 public abstract class AbstractDisplay implements Display {
 
     private final DisplayInfo displayInfo;
+    private final String devicePort;
 
     /**
-     * Constructor for AbstractDisplay from a raw EDID byte array.
+     * Constructor for AbstractDisplay from a raw EDID byte array, with no device port information.
      *
      * @param edid a byte array representing a display EDID
      */
     protected AbstractDisplay(byte[] edid) {
+        this(edid, Constants.UNKNOWN);
+    }
+
+    /**
+     * Constructor for AbstractDisplay from a raw EDID byte array.
+     *
+     * @param edid       a byte array representing a display EDID
+     * @param devicePort the system-level device port identifier (e.g. DRM connector name {@code HDMI-A-1})
+     */
+    protected AbstractDisplay(byte[] edid, String devicePort) {
         this.displayInfo = new DisplayInfoImpl(edid);
+        this.devicePort = devicePort;
+    }
+
+    /**
+     * Constructor for AbstractDisplay from decoded display information, with no device port information.
+     *
+     * @param displayInfo the decoded display information
+     */
+    protected AbstractDisplay(DisplayInfo displayInfo) {
+        this(displayInfo, Constants.UNKNOWN);
     }
 
     /**
@@ -31,9 +53,11 @@ public abstract class AbstractDisplay implements Display {
      * without providing an EDID. Pass a synthetic {@link DisplayInfo} to expose a synthesized EDID.
      *
      * @param displayInfo the decoded display information
+     * @param devicePort  the system-level device port identifier (e.g. DRM connector name {@code HDMI-A-1})
      */
-    protected AbstractDisplay(DisplayInfo displayInfo) {
+    protected AbstractDisplay(DisplayInfo displayInfo, String devicePort) {
         this.displayInfo = displayInfo;
+        this.devicePort = devicePort;
     }
 
     @Deprecated
@@ -45,6 +69,11 @@ public abstract class AbstractDisplay implements Display {
     @Override
     public DisplayInfo getDisplayInfo() {
         return this.displayInfo;
+    }
+
+    @Override
+    public String getDevicePort() {
+        return this.devicePort;
     }
 
     @Override
--- a/oshi-common/src/main/java/oshi/hardware/common/AbstractDisplay.java
+++ b/oshi-common/src/main/java/oshi/hardware/common/AbstractDisplay.java
@@ -8,7 +8,6 @@ import oshi.annotation.concurrent.Immutable;
 import oshi.hardware.Display;
 import oshi.hardware.DisplayInfo;
 import oshi.hardware.DisplayInfoImpl;
-import oshi.util.Constants;
 
 /**
  * A Display
@@ -17,35 +16,14 @@ import oshi.util.Constants;
 public abstract class AbstractDisplay implements Display {
 
     private final DisplayInfo displayInfo;
-    private final String devicePort;
-
-    /**
-     * Constructor for AbstractDisplay from a raw EDID byte array, with no device port information.
-     *
-     * @param edid a byte array representing a display EDID
-     */
-    protected AbstractDisplay(byte[] edid) {
-        this(edid, Constants.UNKNOWN);
-    }
 
     /**
      * Constructor for AbstractDisplay from a raw EDID byte array.
      *
-     * @param edid       a byte array representing a display EDID
-     * @param devicePort the system-level device port identifier (e.g. DRM connector name {@code HDMI-A-1})
+     * @param edid a byte array representing a display EDID
      */
-    protected AbstractDisplay(byte[] edid, String devicePort) {
+    protected AbstractDisplay(byte[] edid) {
         this.displayInfo = new DisplayInfoImpl(edid);
-        this.devicePort = devicePort;
-    }
-
-    /**
-     * Constructor for AbstractDisplay from decoded display information, with no device port information.
-     *
-     * @param displayInfo the decoded display information
-     */
-    protected AbstractDisplay(DisplayInfo displayInfo) {
-        this(displayInfo, Constants.UNKNOWN);
     }
 
     /**
@@ -53,11 +31,9 @@ public abstract class AbstractDisplay implements Display {
      * without providing an EDID. Pass a synthetic {@link DisplayInfo} to expose a synthesized EDID.
      *
      * @param displayInfo the decoded display information
-     * @param devicePort  the system-level device port identifier (e.g. DRM connector name {@code HDMI-A-1})
      */
-    protected AbstractDisplay(DisplayInfo displayInfo, String devicePort) {
+    protected AbstractDisplay(DisplayInfo displayInfo) {
         this.displayInfo = displayInfo;
-        this.devicePort = devicePort;
     }
 
     @Deprecated
@@ -69,11 +45,6 @@ public abstract class AbstractDisplay implements Display {
     @Override
     public DisplayInfo getDisplayInfo() {
         return this.displayInfo;
-    }
-
-    @Override
-    public String getDevicePort() {
-        return this.devicePort;
     }
 
     @Override
--- a/oshi-common/src/main/java/oshi/hardware/common/platform/linux/LinuxHardwareAbstractionLayer.java
+++ b/oshi-common/src/main/java/oshi/hardware/common/platform/linux/LinuxHardwareAbstractionLayer.java
@@ -4,8 +4,8 @@
  */
 package oshi.hardware.common.platform.linux;
 
+import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collectors;
 
 import oshi.annotation.concurrent.ThreadSafe;
 import oshi.hardware.BluetoothDevice;
@@ -18,6 +18,7 @@ import oshi.hardware.SoundCard;
 import oshi.hardware.common.AbstractHardwareAbstractionLayer;
 import oshi.hardware.common.platform.unix.UnixDisplay;
 import oshi.util.driver.linux.DrmEdid;
+import oshi.util.tuples.Triplet;
 
 /**
  * LinuxHardwareAbstractionLayer class.
@@ -49,9 +50,13 @@ public abstract class LinuxHardwareAbstractionLayer extends AbstractHardwareAbst
 
     @Override
     protected List<Display> createDisplays() {
-        List<byte[]> edids = DrmEdid.getEdidArrays();
-        if (!edids.isEmpty()) {
-            return edids.stream().map(UnixDisplay::new).collect(Collectors.toList());
+        List<Triplet<String, Integer, byte[]>> drmData = DrmEdid.getDisplayData();
+        if (!drmData.isEmpty()) {
+            List<Display> displays = new ArrayList<>(drmData.size());
+            for (Triplet<String, Integer, byte[]> drm : drmData) {
+                displays.add(new UnixDisplay(drm.getC(), drm.getA(), drm.getB()));
+            }
+            return displays;
         }
         return UnixDisplay.getDisplays();
     }
--- a/oshi-common/src/main/java/oshi/util/driver/linux/DrmEdid.java
+++ b/oshi-common/src/main/java/oshi/util/driver/linux/DrmEdid.java
@@ -11,7 +11,9 @@ import java.util.List;
 
 import oshi.annotation.concurrent.ThreadSafe;
 import oshi.util.FileUtil;
+import oshi.util.ParseUtil;
 import oshi.util.linux.SysPath;
+import oshi.util.tuples.Triplet;
 
 /**
  * Utility to read EDID data from the Linux DRM (Direct Rendering Manager) subsystem. The kernel exposes raw EDID bytes
@@ -40,6 +42,32 @@ public final class DrmEdid {
      * @return a list of EDID byte arrays (at least 128 bytes each), or empty if none found
      */
     static List<byte[]> getEdidArrays(File drmDir) {
+        List<Triplet<String, Integer, byte[]>> data = getDisplayData(drmDir);
+        List<byte[]> edids = new ArrayList<>(data.size());
+        for (Triplet<String, Integer, byte[]> t : data) {
+            edids.add(t.getC());
+        }
+        return Collections.unmodifiableList(edids);
+    }
+
+    /**
+     * Read display data from /sys/class/drm for all connected displays.
+     *
+     * @return a list of {@link Triplet} of kernel connector name (e.g. {@code HDMI-A-1}), DRM connector ID ({@code -1}
+     *         if not available), and EDID byte array (at least 128 bytes each), or empty if none found
+     */
+    public static List<Triplet<String, Integer, byte[]>> getDisplayData() {
+        return getDisplayData(new File(SysPath.DRM));
+    }
+
+    /**
+     * Read display data from the given DRM directory.
+     *
+     * @param drmDir the directory containing card*-* subdirectories
+     * @return a list of {@link Triplet} of kernel connector name, DRM connector ID ({@code -1} if not available), and
+     *         EDID byte array (at least 128 bytes each), or empty if none found
+     */
+    static List<Triplet<String, Integer, byte[]>> getDisplayData(File drmDir) {
         if (!drmDir.isDirectory()) {
             return Collections.emptyList();
         }
@@ -47,7 +75,7 @@ public final class DrmEdid {
         if (connectors == null || connectors.length == 0) {
             return Collections.emptyList();
         }
-        List<byte[]> displays = new ArrayList<>();
+        List<Triplet<String, Integer, byte[]>> results = new ArrayList<>();
         for (File connector : connectors) {
             File statusFile = new File(connector, "status");
             if (statusFile.exists()) {
@@ -57,13 +85,22 @@ public final class DrmEdid {
                 }
             }
             File edidFile = new File(connector, "edid");
-            if (edidFile.exists() && edidFile.length() >= 128) {
+            if (edidFile.exists()) {
                 byte[] edid = FileUtil.readAllBytes(edidFile.getPath(), false);
                 if (edid.length >= 128) {
-                    displays.add(edid);
+                    // Extract connector name from directory name: "card0-HDMI-A-1" -> "HDMI-A-1"
+                    String dirName = connector.getName();
+                    String connectorName = dirName.substring(dirName.indexOf('-') + 1);
+                    int connectorId = -1;
+                    File connectorIdFile = new File(connector, "connector_id");
+                    if (connectorIdFile.exists()) {
+                        connectorId = ParseUtil.parseIntOrDefault(
+                                FileUtil.getStringFromFile(connectorIdFile.getPath()).trim(), -1);
+                    }
+                    results.add(new Triplet<>(connectorName, connectorId, edid));
                 }
             }
         }
-        return Collections.unmodifiableList(displays);
+        return Collections.unmodifiableList(results);
     }
 }
--- a/oshi-common/src/main/java/oshi/util/driver/linux/DrmEdid.java
+++ b/oshi-common/src/main/java/oshi/util/driver/linux/DrmEdid.java
@@ -94,8 +94,8 @@ public final class DrmEdid {
                     int connectorId = -1;
                     File connectorIdFile = new File(connector, "connector_id");
                     if (connectorIdFile.exists()) {
-                        connectorId = ParseUtil.parseIntOrDefault(
-                                FileUtil.getStringFromFile(connectorIdFile.getPath()).trim(), -1);
+                        connectorId = ParseUtil
+                                .parseIntOrDefault(FileUtil.getStringFromFile(connectorIdFile.getPath()).trim(), -1);
                     }
                     results.add(new Triplet<>(connectorName, connectorId, edid));
                 }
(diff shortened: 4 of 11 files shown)


oshi-common/src/main/java/oshi/util/driver/unix/Xrandr.java
```diff
@@ -40,29 +43,67 @@ public static List<byte[]> getEdidArrays() {
...
                 }
-                String edidStr = sb.toString();
-                byte[] edid = ParseUtil.hexStringToByteArray(edidStr);
+                byte[] edid = ParseUtil.hexStringToByteArray(sb.toString());
                 if (edid.length >= 128) {
-                    displays.add(edid);
+                    results.put(currentPort, new Pair<>(currentConnectorId, edid));
                 }
                 sb = null;
             }
```
dbwiddis, line 117: I think this is why you concluded the hardware didn't report a connector ID -- I suspect it does!

This entry was written when we were only looking for EDID, to the map the instant the EDID block completes (line 102), so `currentConnectorId` is only ever non-`-1` if `CONNECTOR_ID:` was printed *before* `EDID:`. 

In real output it comes after. X properties are stored by prepending to a list (`prop->next = output->properties` in xserver's `randr/rrproperty.c`) and listed head-first, so `xrandr` prints them in reverse creation order -- and modesetting creates `CONNECTOR_ID` in `drmmode_output_create_resources`, while the EDID property is published later, during mode probing.

Real dumps agree, on both amdgpu and intel:

```
DisplayPort-2 connected primary 1920x1080+0+0 ...
        EDID:                     <- line 98
                00ffffffffffff00...
        ...
        CONNECTOR_ID: 70          <- line 143
```

(from `linuxwacom/wacom-hid-descriptors`, `Ugee EX07/sysinfo.2uu4YqpezM/xrandr.txt`; two other dumps in that repo show the same ordering)

Easy check on your own machine:

```sh
xrandr --verbose | grep -n -E ' connected|CONNECTOR_ID|EDID:'
grep . /sys/class/drm/*/connector_id
```

You might find the IDs are there.

The fix is to make the parser order-independent: buffer the current output's name/id/EDID and emit when you hit the *next* output header (plus once more after the loop). Two traps when you do:

1. **Reset on any output header, not just `connected` ones.** Today a `DP-1 disconnected` line leaves `currentPort` pointing at the previous connected output. Disconnected outputs still print `CONNECTOR_ID` -- the `Huion H640P` dump in that repo has six of them following `eDP`'s EDID block -- so a deferred flush would happily attach a disconnected output's ID to the wrong display.
2. **Use "starts at column 0" as the header test.** Properties are always indented and output headers never are, which makes it unambiguous. `trimmed.contains(" connected")` on line 85 is both looser than that and redundant with the `words[1]` check on line 87.

Roughly:

```java
String currentPort = "";
int currentConnectorId = -1;
byte[] currentEdid = null;
StringBuilder sb = null;
for (String s : xrandr) {
    if (!s.isEmpty() && !Character.isWhitespace(s.charAt(0))) {
        // Output header; flush the previous output before starting a new one
        if (currentEdid != null) {
            results.put(currentPort, new Pair<>(currentConnectorId, currentEdid));
        }
        currentPort = s.split("\\s+")[0];
        currentConnectorId = -1;
        currentEdid = null;
        sb = null;
        continue;
    }
    String trimmed = s.trim();
    ...
}
if (currentEdid != null) {
    results.put(currentPort, new Pair<>(currentConnectorId, currentEdid));
}
```

oshi-common/src/main/java/oshi/util/driver/unix/Xrandr.java
```diff
@@ -40,29 +43,67 @@ public static List<byte[]> getEdidArrays() {
...
+            String trimmed = s.trim();
+            // Lines like "HDMI-1 connected primary 1920x1080+0+0"
+            if (trimmed.contains(" connected")) {
+                String[] words = trimmed.split("\\s+");
+                if (words.length > 1 && "connected".equals(words[1])) {
+                    currentPort = words[0];
+                    currentConnectorId = -1;
+                }
+            } else if (trimmed.startsWith("CONNECTOR_ID:")) {
+                currentConnectorId = ParseUtil.parseIntOrDefault(trimmed.split(":\\s*", 2)[1].trim(), -1);
```
dbwiddis: `ParseUtil` already has this one -- `parseLastInt` does the whole line in one call and can't throw on a value-less `CONNECTOR_ID:`. AGENTS.md asks for a `ParseUtil` check before hand-rolling parsing; it's a big file and easy to miss something in it.

```suggestion
                currentConnectorId = ParseUtil.parseLastInt(trimmed, -1);
```

oshi-common/src/test/java/oshi/util/driver/unix/XrandrTest.java
```diff
@@ -67,6 +70,112 @@ void testGetEdidArraysNoEdidBlock() {
...
+        lines.add("\t\t0d5054254b80714f81809500a9c0b300");
+        lines.add("\t\td1c001010101565e00a0a0a029503020");
+        lines.add("\t\t35000f282100001a000000fd00384b1e");
+        lines.add("\t\t5a1900000a202020202020000000fc00");
+        lines.add("\t\t4c4720554c545241474541520000ff00");
+        lines.add("\t\t3630344e54505a4832313337370a0100");
+        lines.add("  1920x1080 (0x48) 148.500MHz +HSync +VSync *current +preferred");
+        return lines;
+    }
+
```
dbwiddis, line 93: This fixture is what let the bug above through: it puts `CONNECTOR_ID` before `EDID:`, which no real X server does.

Could you rebuild these fixtures from an actual `xrandr --verbose` dump, property order and all? Once they match reality, `testGetDisplayDataSingleWithConnectorId` fails -- and then passes again after the parser change, which is exactly the regression test I'd want here.

Two more cases worth adding while you're in the file:

- an output whose `CONNECTOR_ID` follows the EDID block (the real order);
- a `disconnected` output carrying a `CONNECTOR_ID` after a connected output's EDID -- that's the state-leak trap.

oshi-common/src/main/java/oshi/hardware/Display.java
```diff
@@ -53,4 +53,11 @@ public interface Display {
      * @return A {@link DisplayInfo} holding the display's decoded attributes.
      */
     DisplayInfo getDisplayInfo();
+
+    /**
+     * The connection port type used by this display (e.g. HDMI, DisplayPort, VGA, DVI).
+     *
+     * @return The connection port type, or an empty string if not available.
+     */
+    String getConnectionPort();
```
dbwiddis: Three things here:

FIrst: I think we may need two strings.  You want a very specific xrandr string you can use to issue `xrandr --output` commands, so we need to provide that separately from whatever internal ID (the DRM string).

Second: The doc says port *type* -- "HDMI, DisplayPort, VGA, DVI" -- but every value this actually produces is an *identifier*: `HDMI1`, `DP2`, `DisplayPort-0` on Linux, and the literal `"HDMI"` on the other platforms. You've written this to only return the string and not the X-based identifier.

Third: might want UNKNOWN rather than an empty string.

```suggestion
    /**
     * The name of the connector this display is attached to, such as {@code HDMI-A-1} or {@code eDP-1} on Linux. The
     * exact form is platform-specific, and is not guaranteed to be usable as an argument to any particular tool.
     *
     * @return The connector name, or {@link Constants#UNKNOWN} if not available.
     */
    String getConnectionPort();
```

oshi-common/src/main/java/oshi/hardware/common/platform/linux/LinuxHardwareAbstractionLayer.java
```diff
@@ -49,13 +53,47 @@ public Sensors createSensors() {
...
     protected List<Display> createDisplays() {
-        List<byte[]> edids = DrmEdid.getEdidArrays();
-        if (!edids.isEmpty()) {
-            return edids.stream().map(UnixDisplay::new).collect(Collectors.toList());
+        List<Pair<Integer, byte[]>> drmPairs = DrmEdid.getEdidAndConnectorIdPairs();
+        if (!drmPairs.isEmpty()) {
+            return correlateWithXrandr(drmPairs);
         }
         return UnixDisplay.getDisplays();
     }
```
dbwiddis: This is the one place I'd push back on the overall approach, and it's my fault for not being crisper in the discussion.

The port we report is only ever the xrandr name, and `""` when there's no X server to ask -- yet `DrmEdid` is already iterating directories literally named `card0-HDMI-A-1` and throwing that name away. The kernel connector name is the value that *always* exists: on Wayland, on a headless box, in a container, on a pre-6.5 kernel. What I'd like is:

- `getConnectionPort()` -> the kernel connector name from the sysfs directory (`HDMI-A-1`, `eDP-1`), always populated on the DRM path;
- the xrandr output name -> a separate, X-only value, joined via `CONNECTOR_ID` where available and EDID bytes otherwise, and legitimately absent when there's no X.

Practically that means `DrmEdid` needs to return the connector name alongside the ID and EDID, so its carrier can't stay a `Pair<Integer, byte[]>`. It also gives you a third fallback for the join (normalized name matching) for the drivers that don't publish `CONNECTOR_ID` at all -- intel DDX and nouveau DDX don't.

One cost note while we're here: this path now shells out to `xrandr` unconditionally, where before it spawned no process at all. It's memoized at `slowExpiration()` so it's bounded, but a Wayland-only or headless host now pays a failing `exec` every refresh. Gating on `System.getenv("DISPLAY") != null` would keep the common case free.

oshi-common/src/main/java/oshi/hardware/common/platform/linux/LinuxHardwareAbstractionLayer.java
```diff
@@ -49,13 +53,47 @@ public Sensors createSensors() {
...
+        int len = Math.min(drmEdid.length, 128);
+        for (Map.Entry<String, Pair<Integer, byte[]>> entry : xrandrData.entrySet()) {
+            byte[] xrandrEdid = entry.getValue().getB();
+            int xLen = Math.min(xrandrEdid.length, 128);
+            if (len == xLen && Arrays.equals(Arrays.copyOf(drmEdid, len), Arrays.copyOf(xrandrEdid, xLen))) {
+                return entry.getKey();
+            }
+        }
+        return "";
+    }
```
dbwiddis: This is the only genuinely new logic in the PR, and it's private-static inside a HAL, so nothing can test it. 

The correlation belongs in a driver in `oshi-common`, a pure function taking the DRM list and the xrandr map and returning the join. Then it's fixture-testable with no hardware, and the `CONNECTOR_ID`-hit, `CONNECTOR_ID`-miss, EDID-fallback and nothing-matched paths can each be verified independently.

Two behavioral notes for when you move it:

- **Matched entries aren't consumed**, so two connectors with byte-identical base blocks both resolve to the same X name. Building the lookups once -- `id -> name` and `edid-prefix -> name`, dropping keys that collide -- makes it O(n) *and* makes the ambiguity explicit instead of silently picking the first hit.
- `entry.getValue().getA() == drmConnectorId` on line 79 unboxes an `Integer` out of a `Pair<A extends @Nullable Object, ...>`. It's fine at runtime given how we populate it, but NullAway under `-Perror-prone` may well flag the dereference.

oshi-common/src/main/java/oshi/hardware/common/platform/linux/LinuxHardwareAbstractionLayer.java
```diff
@@ -49,13 +53,47 @@ public Sensors createSensors() {
...
+        // Fallback: match by first 128 bytes of EDID
+        byte[] drmEdid = drm.getB();
+        int len = Math.min(drmEdid.length, 128);
+        for (Map.Entry<String, Pair<Integer, byte[]>> entry : xrandrData.entrySet()) {
+            byte[] xrandrEdid = entry.getValue().getB();
+            int xLen = Math.min(xrandrEdid.length, 128);
+            if (len == xLen && Arrays.equals(Arrays.copyOf(drmEdid, len), Arrays.copyOf(xrandrEdid, xLen))) {
+                return entry.getKey();
+            }
+        }
```
dbwiddis: Small one: both arrays are guaranteed to be at least 128 bytes by their producers, so `len` and `xLen` are both always exactly 128 and the `len == xLen` guard is dead. (Java 8 has no ranged `Arrays.equals`, so the `copyOf` pair is unavoidable -- no complaint there.)

```suggestion
        // Fallback: match by the first 128 bytes of EDID (sysfs includes extension blocks, xrandr does not)
        byte[] drmBase = Arrays.copyOf(drm.getB(), 128);
        for (Map.Entry<String, Pair<Integer, byte[]>> entry : xrandrData.entrySet()) {
            if (Arrays.equals(drmBase, Arrays.copyOf(entry.getValue().getB(), 128))) {
                return entry.getKey();
            }
        }
```

oshi-common/src/main/java/oshi/hardware/common/AbstractDisplay.java
```diff
@@ -16,24 +16,29 @@
...
 
     /**
      * Constructor for AbstractDisplay from a raw EDID byte array.
      *
-     * @param edid a byte array representing a display EDID
+     * @param edid           a byte array representing a display EDID
+     * @param connectionPort the connection port type (e.g. HDMI, DisplayPort)
      */
-    protected AbstractDisplay(byte[] edid) {
+    protected AbstractDisplay(byte[] edid, String connectionPort) {
```
dbwiddis: A suggestion that shrinks this PR by four files: keep the old one-argument constructors as overloads that delegate with the sentinel.

```java
/**
 * Constructor for AbstractDisplay from a raw EDID byte array, with no connection port information.
 *
 * @param edid a byte array representing a display EDID
 */
protected AbstractDisplay(byte[] edid) {
    this(edid, Constants.UNKNOWN);
}
```

(and the same for the `DisplayInfo` constructor below)

Then `MacDisplayJNA`, `MacDisplayFFM`, `WindowsDisplayJNA` and `WindowsDisplayFFM` don't need to change at all, the `public UnixDisplay(byte[])` constructor keeps working for anyone downstream calling it, nothing has to invent a value it doesn't know, and `japicmp` stops reporting the constructor change.

oshi-core/src/main/java/oshi/hardware/platform/mac/MacDisplayJNA.java
```diff
@@ -118,7 +118,7 @@ private static List<Display> getDisplaysFromService(String serviceName, String e
                                 int length = edid.getLength();
                                 Pointer p = edid.getBytePtr();
                                 if (length > 0) {
-                                    displays.add(new MacDisplayJNA(p.getByteArray(0, length)));
+                                    displays.add(new MacDisplayJNA(p.getByteArray(0, length), "HDMI"));
```
dbwiddis: This one does need to change before merge: `"HDMI"` here is a guess, and OSHI's rule is to return a sentinel rather than a plausible-looking wrong answer. Line 210 is the built-in Retina panel, so a MacBook's internal display would report itself as HDMI. Same literal in `MacDisplayFFM`, `WindowsDisplayJNA` and `WindowsDisplayFFM`.

Linux-only is completely fine for this PR -- I'll happily take macOS and Windows as follow-ups, or do them myself -- so these want `Constants.UNKNOWN`. With the delegating overloads suggested on `AbstractDisplay`, all four of these files revert to untouched.

For reference when someone does get to them: macOS can read `ParentBuiltInPortTypeDescription` / `TransportDescription` off the same IORegistry node the EDID already comes from, and Windows has `WmiMonitorConnectionParams.VideoOutputTechnology`. Neither is your problem today.

oshi-common/src/main/java/oshi/util/driver/linux/DrmEdid.java
```diff
@@ -40,14 +43,49 @@ public static List<byte[]> getEdidArrays() {
...
+     *         128 bytes each), or empty if none found
+     */
+    static List<Pair<Integer, byte[]>> getEdidAndConnectorIdPairs(File drmDir) {
+        List<Map.Entry<Integer, byte[]>> entries = getEdidAndConnectorIds(drmDir);
+        List<Pair<Integer, byte[]>> pairs = new ArrayList<>(entries.size());
+        for (Map.Entry<Integer, byte[]> entry : entries) {
+            pairs.add(new Pair<>(entry.getKey(), entry.getValue()));
+        }
+        return pairs;
+    }
```
dbwiddis: I'm not sure of the purpose of this method. The only time you use it is to check if it's empty; a list of map entries seems incorrect.  Should this be a map of connectorID to EDID?

oshi-common/src/main/java/oshi/util/driver/linux/DrmEdid.java
```diff
@@ -40,14 +43,49 @@ public static List<byte[]> getEdidArrays() {
...
+
+    private static List<Map.Entry<Integer, byte[]>> getEdidAndConnectorIds(File drmDir) {
         if (!drmDir.isDirectory()) {
             return Collections.emptyList();
         }
         File[] connectors = drmDir.listFiles(f -> f.isDirectory() && f.getName().matches("card\\d+-.+"));
         if (connectors == null || connectors.length == 0) {
```
dbwiddis: While you're in here: the `listFiles()` call just above returns entries in filesystem order, so the display list ordering is arbitrary from run to run. Sorting `connectors` by name makes `getDisplays()` deterministic. Small fix, but it's the underlying reason the original "pair by index" approach in #2951 could never have worked, so it belongs in this PR.

oshi-common/src/main/java/oshi/util/driver/unix/Xrandr.java
```diff
@@ -40,29 +43,67 @@ public static List<byte[]> getEdidArrays() {
...
+        Map<String, Pair<Integer, byte[]>> results = new LinkedHashMap<>();
+        String currentPort = "";
+        int currentConnectorId = -1;
         StringBuilder sb = null;
         for (String s : xrandr) {
-            if (s.contains("EDID")) {
+            String trimmed = s.trim();
+            // Lines like "HDMI-1 connected primary 1920x1080+0+0"
+            if (trimmed.contains(" connected")) {
+                String[] words = trimmed.split("\\s+");
```
dbwiddis: 1. Error Prone likes using -1 rather than the default 0 for this.  
2. Prefer a precompiled pattern for more than a single character regex.  In this case, you want:
https://github.com/oshi/oshi/blob/4a4a243ff3864b2d32b127b9464ebbe1257a1713/oshi-common/src/main/java/oshi/util/ParseUtil.java#L141

Example: 
https://github.com/oshi/oshi/blob/4a4a243ff3864b2d32b127b9464ebbe1257a1713/oshi-common/src/main/java/oshi/software/common/os/mac/MacNetworkParams.java#L45

oshi-common/src/test/java/oshi/util/driver/unix/XrandrTest.java
```diff
@@ -67,6 +70,112 @@ void testGetEdidArraysNoEdidBlock() {
...
+    // Fixture: two connected displays, one with CONNECTOR_ID, one without
+    private static List<String> createXrandrTwoDisplays() {
+        List<String> lines = new ArrayList<>();
+        lines.add("Screen 0: minimum 8 x 8, current 3840 x 1080, maximum 32767 x 32767");
+        lines.add("DP2 connected primary 1920x1080+0+0 (normal left inverted right x axis y axis) 530mm x 300mm");
+        lines.add("\tCONNECTOR_ID: 96");
+        lines.add("\t\tsupported: 96");
```
dbwiddis: try this test with connector id after EDID.  You might find it fails.

dbwiddis (commented):
Thank you for picking this back up — and for coming back to it after the codebase moved out from under your original PR. That's a real annoyance and I appreciate the patience. 

This is a good first cut: the shape is right, comparing only the first 128 bytes of EDID is exactly right, the LinkedHashMap preserves order as intended, and you updated all four JNA/FFM twins consistently. Nice work.

There's one bug I'd like to fix before anything else, because I think it also explains what you saw on your machine: the parser only picks up CONNECTOR_ID if it appears before the EDID: block, and (some) real xrandr --verbose prints it after. So I suspect your hardware does report a connector ID and we're just dropping it. 

Beyond that, the main design comment is that I'd like getConnectionPort() to be the kernel connector name (HDMI-A-1), with the xrandr name as the optional X-only extra. As implemented, the always-available value is discarded and the X-dependent one is all we report, so Wayland and headless hosts get an empty string even though sysfs already told us the answer.

dbwiddis:
This thread has multiple xrandr outputs you should be robust to: https://forums.linuxmint.com/viewtopic.php?t=421905
  - disconnected (but has connector id)
  - connector id after edid

The parsing logic should reset on both connected and disconnected, or EOF

ayonization:
Hi @dbwiddis ,

I can confirm I do not see the connector ID in my machine at all with xrandr. You can refer to the attached file to check the entire xrandr response from my machine. 


[xrandr.txt](https://github.com/user-attachments/files/31230183/xrandr.txt)


I am thinking that I have over engineered the problem here, and I have a new design in mind. Let me run that through you and if it makes sense to you, Ill start the implementation.

Ive prepared a few rough flow diagrams to help explain what Im trying to do:

https://excalidraw.com/#json=sU7RaVyLz_9k8J1X0Qalk,1Qpsm2r3-KOaILzBzjy74A

What i suggest is : 

- Keep the current logic for display object creation, add the port which DRM reports (can be easily retreived from the DRM directory)
- For the UnixDisplay.getDisplays() fallback, add the connection port which xrandr reports (because here we are parsing xrandr anyways)
- Clients who explicitly need the x server connection port (like myself) can use the display object received.
Using the edid from the display object received, they can call a new method in Xrandr.java, which will take that edid and run xrandr verbose, parse the edid the client sent with the edid from the xrandr verbose output, and then return the corresponding conneciton port. 

A few benifits i see to doing this :

1) Wayland users do not interact with the x server at all, if they do not need to.
2) Connection ports are always the one reported by drm, clients choose when they need the x server port specifically
3) Edid matching ensures results are consistent accross different machines and display drivers.

I think this would be a greatly simpler implementaion with minimal changes to the current design, and at the same time would solve our use case.

Would love to get your feedback, and if you have any concerns, please do let me know

dbwiddis:
> I can confirm I do not see the connector ID in my machine at all with xrandr. 

Got it... I'm seeing conflicting reports at how recently it was added to Linux. In any case the EDID matching works, it's just potentially fallible if you have two identical monitors.  Which is why matching on connector_id should be the primary match if it's available.  

> What i suggest is :
> 
> * Keep the current logic for display object creation, add the port which DRM reports (can be easily retreived from the DRM directory)
> * For the UnixDisplay.getDisplays() fallback, add the connection port which xrandr reports (because here we are parsing xrandr anyways)
> * Clients who explicitly need the x server connection port (like myself) can use the display object received.
>   Using the edid from the display object received, they can call a new method in Xrandr.java, which will take that edid and run xrandr verbose, parse the edid the client sent with the edid from the xrandr verbose output, and then return the corresponding conneciton port.

I appreciate the simplicity here... but don't forget we have a cross-platform API so we not only need to consider DRM (Linux) but also macOS and Windows. 

it's basically "keep what we have but add a new getter for the port" and adding a new method to explicitly call xrandr util on-demand to get the output name.  (Filtering on matching connector id if available, if not, on matching edid.)

> A few benifits i see to doing this :
> 
> 1. Wayland users do not interact with the x server at all, if they do not need to.

Nor do mac users, but the question is, is there a similar "name" for an output that they may want to use for manipulating the display.  And there's no built-in command (at least on recent versions).  Apparently `displayplacer` is a commonly recommended alternative but the IDs for it are UUIDs and that's ... ugly.

And Windows users have displayswitch, but it's got fixed switches. Still we could try to define "internal" and "external".

Or... just ignore that field on non-X operating systems.

> 2. Connection ports are always the one reported by drm, clients choose when they need the x server port specifically
> 3. Edid matching ensures results are consistent accross different machines and display drivers.

Edid matching isn't guaranteed unique, though.

> Would love to get your feedback, and if you have any concerns, please do let me know

Sure.  Let's attack this top down, API level first.

1. Create new methods on the `Display` interface.  You may see that `japicmp` is yelling at you about versions because you added a new (abstract because interface) method, this is ok and if you rebase you'll see I did a minor version bump that will silence it.  Don't use `DisplayInfo`, that's exclusively a human readable representation of the EDID.  You could also add a default method to silence the API change complaint.   But in any case, I'd recommend two methods:
    - `String getDevicePort()` - (I'm not sold on the name, feel free to propose an alternative.) This is the "system level" identification: the DRM path for Linux.  The "TransportDescription" for macOS, the Device path on Windows.  Whatever Wayland produces.
    - `default Optional<String> getOutputName()` - Specific to xrandr (at least for now), the name to be used with `--output`.  Defaults to an empty `Optional` anywhere else.  (We may improve this in the future.)

2. Existing parsing on all operating systems except we need to pull out more information for the new `getDevicePort()` field.
   - DRM path on Linux
   - The parent device name or Description on macOS (`Port-HDMI@1`, `Port-USB-C@1`, for external, `disp0` for internal)
   - Need to do more research on what to put for windows.  PNP Device ID is probably a decent placeholder from existing registry implementation.

3. For the `getOutputName()` field, on-demand when queried, call the Xrandr utility with connectorID and edid, filtering on both (if available) and returning the optional if present.
   - This is the one case we can potentially save two calls if we have initially populated the `Display` via `xrandr` and already have the data stored, thus my original suggestion for a more verbose return to pre-"save" that output name without the need to re-query it.  This is still a possibility, but probably more complexity than we need.

dbwiddis:
Researched WIndows. It's a small change from what we have.

`WindowsDisplayJNA.java:63-65` already allocates a `CloseableSpDeviceInterfaceData` and sets `cbSize`, then never uses it.. I'm guessing this is leftover from an earlier implementation before we switched to the simpler registry. That struct is exactly what's missing:
  - `SP_DEVINFO_DATA` (`info`, which the loop does use) names a device. It gets you the registry key, hence the EDID. It cannot be turned into a path.
  - `SP_DEVICE_INTERFACE_DATA` (`deviceInterfaceData`, the dead code) names one interface exposed by that device, and it's the only thing `SetupDiGetDeviceInterfaceDetail` accepts. It gives the path, `\\?\DISPLAY#DELA1CD#5&2a1c8d3&0&UID4353#{e6f07b5f-...}`.

That path is the join key, because `DISPLAYCONFIG_TARGET_DEVICE_NAME.monitorDevicePath` from `QueryDisplayConfig` is the identical string. The loop already holds info at the moment it reads the EDID, so getting the path costs two calls and no restructuring.

Pseudocode

```python
  // once, before the loop: ask the CCD API for every active connector
  map = {}                                       // normalized device path -> "HDMI-1"
  GetDisplayConfigBufferSizes(ONLY_ACTIVE_PATHS | VIRTUAL_MODE_AWARE, &nPaths, &nModes)
  paths = alloc(72 * nPaths);  modes = alloc(64 * nModes)   // modes required but never read
  QueryDisplayConfig(flags, &nPaths, paths, &nModes, modes, NULL)   // retry on ERROR_INSUFFICIENT_BUFFER
  for each path in paths:
      if not (path.flags & DISPLAYCONFIG_PATH_ACTIVE): continue
      tdn.header = { type: GET_TARGET_NAME, size: 420,
                     adapterId: path.targetInfo.adapterId, id: path.targetInfo.id }
      if DisplayConfigGetDeviceInfo(tdn) == ERROR_SUCCESS:
          map[normalize(tdn.monitorDevicePath)] =
              connectorName(tdn.outputTechnology, tdn.connectorInstance)   // 5, 0 -> "HDMI-1"

  // the existing SetupAPI walk, unchanged except for three added lines
  for memberIndex = 0; SetupDiEnumDeviceInfo(hDevInfo, memberIndex, info); memberIndex++:
      edid = RegQueryValueEx(SetupDiOpenDevRegKey(hDevInfo, info, ...), "EDID")
      if edid == null: continue

      // NEW: the dead deviceInterfaceData finally does its job.
      port = UNKNOWN
      if SetupDiEnumDeviceInterfaces(hDevInfo, info, GUID_DEVINTERFACE_MONITOR, 0, deviceInterfaceData):
          path = SetupDiGetDeviceInterfaceDetail(hDevInfo, deviceInterfaceData)  // size call, then fetch
          port = map.getOrDefault(normalize(path), UNKNOWN)

      displays.add(new WindowsDisplayJNA(edid, port))   // was: new WindowsDisplayJNA(edid)
```

ayonization:
Hi @dbwiddis , 

Thanks for your feedback!

Ive tried to make the changes you suggested.

1) A new getDevicePort() method on the Display class
    Returns "UNKNOWN" on macos and windows for now.
    getDisplays() now also parses the DRM connection port and that is what is returned by the getDevicePort() method.
    Also, getDisplays() now stores the connector ID value while parsing , which is used later for correlationg with the port       
    as reported by x server
    
2) An optional getOutputName() method, for returning the port reported by xrandr on demand.
     Uses the xrandr output and tries find the x connection port by using the connector id reported by drm first, otherwise
     uses edid matching as a fallback.
     
 Note : I noticed something while testing this on my machine, in the existing parsing logic for EDID in the DRM edid class, we first check the size of the EDID file. 
 
 However, i noticed that sysfs reported the file size to be 0, even though the file did actually contain the edid. Please refer to the screenshot attached. I therefore removed the size check before parsing the EDID.

ayonization:
Looks like screenshots arent working for some reason, Heres what I meant when i said the EDID file size appears to be zero :


[edid.txt](https://github.com/user-attachments/files/31283914/edid.txt)

dbwiddis:
Thanks!  Appreciate you coming back to this after the first attempt. This is a genuinely good feature.

I've rebased your commit onto current master and pushed three small tweaks on top, none of which change what you built:
1. Moved devicePort out of AbstractDisplay into UnixDisplay. Display.getDevicePort() is now a default returning Constants.UNKNOWN, matching the getOutputName() default you already wrote, so implementations that can't supply a port don't pass a sentinel up through a constructor. 
2. assertNotNull in XrandrTest where NullAway couldn't see the Map.get results as non-null: that was the Error Prone failure. Your testDevicePortFromTwoArgConstructor moved to a new UnixDisplayTest, since the constructor it covered now lives there.
3. spotless:apply for import ordering, plus a CHANGELOG entry crediting you.

ayonization:
Hi @dbwiddis 

Thanks again for all the guidance on this PR and your exteremely prompt responses !

One quick follow-up question: would you be open to adding something like isPrimary() to the Display class as well?

It looks like this information is available on Windows, macOS, and through XRandR on Linux. The tricky part seems to be the DRM path — DRM itself doesn’t appear to have a concept of a “primary” display.

For Linux, we could potentially populate it when using XRandR (will need to do some more research on this for Wayland), but it would be unavailable when using DRM directly .

Do you think isPrimary() would still make sense as part of the Display abstraction given that limitation?

dbwiddis:
Yes, `isPrimary()` seems like a great feature.  OK if it's "true only if we are sure, false otherwise" given our best-effort approach to most of this!

PR #3681, opened 2026-08-21
Read GPU utilization from NVML on Linux

Krillsson:
LinuxGpuStats#getGpuUtilization() always returns -1 for NVIDIA cards on Linux, even though NVML is available and every other metric on the same class reads from it.

I hit this via a Monitee user report (https://github.com/Krillsson/monitee-agent/issues/341): an RTX A5000 under load reported 97% via nvidia-smi, and Monitee correctly showed temperature, VRAM, fan, clocks and power for the same card. The confusion in the issue about the user still seeing 0% usage is likely due to a bug I found in the app code. I have confirmed that both the jna and ffm paths works as expected using my servers nvidia card. It reports utilization numbers during encoding/decoding work.

NvmlUtilJNA.getGpuUtilization() / NvmlUtilFFM.getGpuUtilization() already wrap nvmlDeviceGetUtilizationRates correctly. But nothing in LinuxGpuStats is calling them.

This adds nvmlGetGpuUtilization(deviceId) as a sixth abstract NVML method (alongside the existing five) and wires it into getGpuUtilization() first, exactly like the other getters, with the sysfs paths kept as the AMD/Intel fallback. Implemented in LinuxGpuStatsJNA, LinuxGpuStatsFFM, and the native-free LinuxGpuStatsNF (returning -1).


oshi-common/src/main/java/oshi/hardware/common/platform/linux/LinuxGpuStats.java
```diff
@@ -102,6 +102,14 @@ protected String getCardName() {
      */
     protected abstract @Nullable String nvmlFindDeviceByName(String name);
 
+    /**
+     * Returns GPU core utilization via NVML, or -1.
+     *
+     * @param deviceId NVML device identifier
+     * @return utilization percentage (0-100) or -1
+     */
+    protected abstract double nvmlGetUtilization(String deviceId);
```
dbwiddis: Nit: the other getters match the NvmlUtil method name.

```suggestion
    protected abstract double nvmlGetGpuUtilization(String deviceId);
```

dbwiddis (commented):
Thanks for this fix!  One tiny naming nit, and CodeRabbit suggests testing the fallback path.

Krillsson:
@dbwiddis could you cut a release with this fix? I was hoping to create a release of monitee-agent to fix this for my users.

dbwiddis:
@Krillsson already in the works, was just finishing a bug-bashing arc first.

Krillsson:
thanks a bunch! 🙏🏻

PR #3698, opened 2026-08-22
Audit FFM struct layouts against the headers, and fix what it found

dbwiddis:
The signature audit compares every `FunctionDescriptor` against the platform's real prototype, but never looked at **struct layouts** — 85 of them across the FFM bindings, with one hand-written test covering three. This extends the same harness to them, and fixes the eleven mismatches it reported: seven on macOS and four on Windows.

## Two were live bugs on macOS

**`OSProcess.getCurrentWorkingDirectory()` returned an empty string** in the FFM implementation. `VNODE_INFO_PATH` padded the leading `vnode_info` with `paddingLayout(152 * 8)` — bit-style sizing left over from the FFM *preview* API, where padding was measured in bits. `vip_path` therefore sat at offset **1216 instead of 152**, and the read found nothing.

```
before:  actual cwd = [/Users/danielwiddis/git/oshi]   FFM getCwd = []
after:   actual cwd = [/Users/danielwiddis/git/oshi]   FFM getCwd = [/Users/danielwiddis/git/oshi]
```

**`NetworkParams.getDomainName()` read the wrong field on macOS and the BSDs.** BSD orders `ai_canonname` before `ai_addr` in `struct addrinfo`; glibc is the reverse. The macOS FFM layout had copied the glibc order, and the shared JNA `CLibrary.Addrinfo` is annotated with it for *every* platform — so both bindings were reading a `struct sockaddr` pointer where a string was expected. Probing a real `getaddrinfo` result on macOS:

```
offset 24 -> "localhost"     <- ai_canonname
offset 32 -> first bytes 1c 1e 00 ...   <- sa_len=28, sa_family=30: a sockaddr
```

Notably the **FreeBSD** FFM layout already had the BSD order, so the two bindings disagreed there and nothing noticed. `Addrinfo` now picks its field order at runtime.

The rest were latent: `timeval` sized 44 rather than 16 by the same bit-style padding; `utmpx.ut_pad` declared 16 bytes where the header says `__uint32_t[16]`; `vm_statistics` missing its three trailing counters; `statfs` missing `f_flags_ext`; `passwd` carrying a `pw_fields` member that exists on FreeBSD but not macOS. JNA's `MacUtmpx` had the same `ut_pad` error.

## How the harness checks layouts

FFM `structLayout` is packed exactly as written — padding is explicit, never inferred — so each named field's offset is the running sum of the sizes before it. That makes the whole layout checkable arithmetically against the header. The pass emits one assert per struct size and per field offset, and reports **both** numbers on a mismatch:

```
VNODE_INFO_PATH: vip_path offset -- header says 152, layout says 1216
```

`structs/<platform>.txt` maps a layout constant to the C type it mirrors. A layout with no entry is reported as **unmapped** rather than silently passing, so adding a line is the act of claiming which struct it models — the review step this file exists to force. Only `structs/mac.txt` is seeded here; the other platforms report their layouts as unmapped until someone makes that claim.

## And a JNA/FFM parity test

`StructLayoutComparisonTest` asserts the two bindings agree on size and field offsets for the same struct. It needs only a JVM, so it runs in the normal suite rather than needing a C toolchain.

**Neither check subsumes the other**, and this PR contains one example of each: both bindings agreed that `ut_pad` was 16 bytes, which only the headers could settle; and the `addrinfo` divergence between the FreeBSD and shared bindings is invisible to a header audit that reads one binding at a time.

## Four more on Windows, all latent

Seeding `structs/windows.txt` turned up four layouts that disagree with the SDK. Each has a *correct* JNA twin — those structs come from `com.sun.jna.platform.win32`, which derives offsets from the field types — so all four are FFM-only divergences.

| Layout | Error | Effect |
|---|---|---|
| `EVENTLOGRECORD` | four bytes of padding after `ReservedFlags` that nothing was due | every field from `ClosingRecordNumber` on sat four bytes late; the record measured 60, not 56 |
| `PDH_RAW_COUNTER` | padding placed before `TimeStamp` instead of after it | `TimeStamp` read from 8 rather than 4 |
| `IP_ADDR_STRING` | missing tail padding | 44 rather than 48 |
| `FIXED_INFO` | missing tail padding, plus the above | 592 rather than 600, and `NodeType`, `ScopeId` and the three `Enable` flags four bytes early |

None changes a value OSHI returns today, which is exactly why they survived. `PDH_RAW_COUNTER` is the instructive one: the misplaced padding compensated exactly, so `FirstValue`, `SecondValue` and the total size were all right and only `TimeStamp` — the one field nothing reads — was wrong. A size-only check cannot see that.

`EVENTLOGRECORD` is the closest to live: the oversized slice could run past the read buffer on the last record.

`PEB` and `RTL_USER_PROCESS_PARAMETERS` are deliberately left unmapped. `winternl.h` publishes both in a reduced form whose size differs from the real, undocumented struct OSHI mirrors, so the SDK is not an authority on either.

## Getting the harness to run off clang

The struct pass was written against clang and failed on both other platforms as soon as their layouts were mapped — loudly, as a harness error rather than a clean run, which is what the check added earlier is for.

- gcc names the two identifiers in a missing-member message in the opposite order and quotes them with `‘ ’`. The same curly quotes meant the *signature* pass had never matched gcc's `was not declared` either, so undeclared symbols were not being skipped on Linux at all.
- MSVC rejected every size check because the `cmp` template took `unsigned long`, 32 bits under LLP64 while `sizeof()` yields 64. It also names the struct **tag** where the mapping names the typedef, so a member the SDK does not declare was reported as an offset mismatch against zero instead of being skipped.
- A mismatch is now identified by the line it is on rather than by clang's wording, because MSVC does not echo the offending source line.

Each job now uploads its generated sources and compiler logs, so the next platform-specific wording can be diagnosed without a debug push.

## Verification

All three jobs green: **macOS** 20 layouts / 163 field offsets, **Linux** 4 / 40, **Windows** 29 / 175 — from 7, 0 and 43 mismatches respectively.

macOS was verified on this machine, including that a deliberately wrong mapping still reports both numbers; Linux was verified in an x86_64 container before pushing. Both implementations now return the working directory and the canonical domain name, agreeing with each other and with the OS.

Full gate from clean with `-Paggregate-coverage`, plus `-Perror-prone`, both green.

The Windows layout changes cannot be exercised from here; AppVeyor covers them on this PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected macOS native structure layouts for process, filesystem, virtual-memory, socket, login, and related system data.
  - Improved working-directory and canonical-domain-name retrieval on macOS and BSD systems.
  - Corrected Windows structure alignment and platform-specific address information ordering.

- **Tests**
  - Added validation comparing native structure sizes and field offsets across bindings.
  - Improved detection and reporting of structure-layout mismatches.

- **Documentation**
  - Updated the changelog and expanded native structure audit mappings for macOS, Linux, and Windows.
  - Added more complete audit diagnostics and preserved platform-specific audit artifacts.


dbwiddis:
Thanks — three of the five were valid against the current tree and are fixed in 5386d9a:

- **Track every layout for each C type.** Confirmed: `windows.txt` maps both `PERFORMANCE_INFORMATION` and `PERFORMANCE_INFORMATION_LAYOUT` to the same C type, and the stem lookup kept only one owner. `ctype_owners` now holds every owner of a stem.
- **Keep file identity in unmapped-layout diagnostics.** Fixed — an unmapped or unresolvable layout prints as `File.NAME` when its simple name is not unique, which is also the key the mapping file needs.
- **Use qualified identities when applying struct skips.** Already fixed in 053ea62 before this review landed; `emit_structs` keys the skip set and the finding label off the mapping key. The `emit_dump` half of the point stood, so `emit_dump`/`DUMP_PRE` are removed — nothing in the repository called them.

Skipping the `E741` rename: there is no Ruff configuration or Ruff CI job in this repository, so the stated consequence does not apply, and `l` is used consistently throughout both scripts. The accompanying ast-grep ReDoS note is a false positive — `tc['undeclared']` is a literal from the `TOOLCHAINS` table, not input.

Re-verified after the change: macOS clean locally, Linux clean in an x86_64 container, and all three audit jobs green.

PR #3709, opened 2026-08-25
Add isPrimary to the Display API

ayonization:
This PR adds a new default method to the Display interface to determine whether a display is the primary display, as defined by the underlying platform.

**Linux**

On systems running an X server, the primary display is determined using the output marked as primary by xrandr.

If the display corresponds to the xrandr primary output, isPrimary() returns true; otherwise, it returns false.

On systems where an X server is not available, isPrimary() defaults to false, since there is no consistent way to determine the primary display from DRM alone.

**macOS**

On macOS, CoreGraphics is used as the authoritative source for determining the main display.

External displays are enumerated through IOKit, so the IOKit display must first be correlated with its corresponding CoreGraphics display. This is done by comparing the display’s numeric manufacturer/vendor ID, product ID, and serial number obtained from its EDID with the values returned by:

* CGDisplayVendorNumber
* CGDisplayModelNumber
* CGDisplaySerialNumber

If these values match the display identified by CoreGraphics as the main display, isPrimary() returns true.

Correlating IOKit and CoreGraphics displays using these identifiers is an established approach because CGDisplayIOServicePort, which previously provided a more direct mapping, is deprecated:

https://stackoverflow.com/questions/20025868/cgdisplayioserviceport-is-deprecated-in-os-x-10-9-how-to-replace

For the Apple Silicon built-in display, where the CGDirectDisplayID can already be identified directly, CGDisplayIsMain is used instead of EDID correlation.

Ambiguity handling: If multiple active displays have the same vendor, product, and serial-number combination, the mapping cannot be determined reliably. In this case, the displays are treated as ambiguous and isPrimary() returns false rather than potentially marking the wrong display as primary.

**Windows**

On Windows, the primary display is determined using the CCD (QueryDisplayConfig) topology.

Windows defines the primary desktop source as the source whose desktop position is (0, 0), as demonstrated in Microsoft’s CCD example:

https://learn.microsoft.com/en-us/windows-hardware/drivers/display/ccd-example-code

The correlation flow is:

QueryDisplayConfig
→ find the active source whose desktop position is (0, 0)
→ obtain its monitorDevicePath
→ normalize the path
→ match it against the monitor device interface path obtained through SetupAPI
→ mark the corresponding WindowsDisplay as primary

This allows the primary status to be associated with the same physical monitor already being enumerated through SetupAPI, without requiring EDID-based correlation.

## Summary by CodeRabbit

* **New Features**
  * Added primary-display detection across Windows, macOS, and Unix systems.
  * Added `isPrimary()` to identify whether a display is the system’s primary display.
  * Improved display identification using vendor, product, serial, connector, and position data.
  * Added EDID accessors for vendor, product, and serial numbers.

* **Bug Fixes**
  * Displays now reliably report primary status, defaulting to false when unavailable.
  * Improved handling of ambiguous or incomplete display information.

### DIFF
--- a/oshi-common/src/main/java/oshi/driver/common/windows/DisplayConnector.java
+++ b/oshi-common/src/main/java/oshi/driver/common/windows/DisplayConnector.java
@@ -64,6 +64,19 @@ public final class DisplayConnector {
     /** Offset of {@code monitorDevicePath} (a {@code WCHAR[128]}) within a {@code DISPLAYCONFIG_TARGET_DEVICE_NAME}. */
     public static final int TDN_MONITOR_DEVICE_PATH_OFFSET = 164;
 
+    // Offsets for determining the primary display from the DISPLAYCONFIG_PATH_INFO and DISPLAYCONFIG_MODE_INFO arrays.
+    // The Windows primary display is the active path whose source mode desktop position is (0, 0).
+    /** Offset of {@code sourceInfo.modeInfoIdx} (a {@code UINT32}) within a {@code DISPLAYCONFIG_PATH_INFO}. */
+    public static final int PATH_SOURCE_MODE_IDX_OFFSET = 12;
+    /** {@code DISPLAYCONFIG_MODE_INFO_TYPE_SOURCE} value for the {@code infoType} field. */
+    public static final int MODE_INFO_TYPE_SOURCE = 1;
+    /** Offset of {@code sourceMode.position.x} (a {@code LONG}) within a {@code DISPLAYCONFIG_MODE_INFO}. */
+    public static final int SOURCE_MODE_POSITION_X_OFFSET = 28;
+    /** Offset of {@code sourceMode.position.y} (a {@code LONG}) within a {@code DISPLAYCONFIG_MODE_INFO}. */
+    public static final int SOURCE_MODE_POSITION_Y_OFFSET = 32;
+    /** Offset of {@code infoType} (a {@code UINT32}) within a {@code DISPLAYCONFIG_MODE_INFO}. */
+    public static final int MODE_INFO_TYPE_OFFSET = 0;
+
     // DISPLAYCONFIG_VIDEO_OUTPUT_TECHNOLOGY values (shared with the D3DKMDT_VIDEO_OUTPUT_TECHNOLOGY enum).
     private static final int VOT_HD15 = 0;
     private static final int VOT_SVIDEO = 1;
--- a/oshi-common/src/main/java/oshi/hardware/Display.java
+++ b/oshi-common/src/main/java/oshi/hardware/Display.java
@@ -88,4 +88,15 @@ public interface Display {
     default Optional<String> getOutputName() {
         return Optional.empty();
     }
+
+    /**
+     * Whether this display is the primary display. Returns {@code true} only when the platform positively identifies
+     * this display as primary; returns {@code false} when the primary display cannot be determined (e.g. under Wayland,
+     * on headless systems, or on platforms that do not expose primary-display information).
+     *
+     * @return {@code true} if this display is the primary display, {@code false} otherwise
+     */
+    default boolean isPrimary() {
+        return false;
+    }
 }
--- a/oshi-common/src/main/java/oshi/hardware/common/platform/unix/UnixDisplay.java
+++ b/oshi-common/src/main/java/oshi/hardware/common/platform/unix/UnixDisplay.java
@@ -17,7 +17,6 @@ import oshi.hardware.Display;
 import oshi.hardware.common.AbstractDisplay;
 import oshi.util.Constants;
 import oshi.util.driver.unix.Xrandr;
-import oshi.util.tuples.Pair;
 import oshi.util.tuples.Triplet;
 
 /**
@@ -28,7 +27,8 @@ public final class UnixDisplay extends AbstractDisplay {
 
     private final String devicePort;
     private final int connectorId;
-    private final Supplier<Map<String, Pair<Integer, byte[]>>> xrandrData;
+    private final boolean primary;
+    private final Supplier<Map<String, Triplet<Integer, byte[], Boolean>>> xrandrData;
 
     /**
      * Constructor for UnixDisplay.
@@ -36,7 +36,7 @@ public final class UnixDisplay extends AbstractDisplay {
      * @param edid a byte array representing a display EDID
      */
     public UnixDisplay(byte[] edid) {
-        this(edid, Constants.UNKNOWN, -1);
+        this(edid, Constants.UNKNOWN, -1, false);
     }
 
     /**
@@ -47,7 +47,19 @@ public final class UnixDisplay extends AbstractDisplay {
      * @param connectorId the DRM connector ID ({@code -1} if not available)
      */
     public UnixDisplay(byte[] edid, String devicePort, int connectorId) {
-        this(edid, devicePort, connectorId, memoize(Xrandr::getDisplayData));
+        this(edid, devicePort, connectorId, false);
+    }
+
+    /**
+     * Constructor for UnixDisplay with device port, DRM connector ID, and primary status.
+     *
+     * @param edid        a byte array representing a display EDID
+     * @param devicePort  the DRM connector name (e.g. {@code HDMI-A-1})
+     * @param connectorId the DRM connector ID ({@code -1} if not available)
+     * @param primary     whether this display is the primary display
+     */
+    public UnixDisplay(byte[] edid, String devicePort, int connectorId, boolean primary) {
+        this(edid, devicePort, connectorId, primary, memoize(Xrandr::getDisplayData));
     }
 
     /**
@@ -56,13 +68,15 @@ public final class UnixDisplay extends AbstractDisplay {
      * @param edid        a byte array representing a display EDID
      * @param devicePort  the DRM connector name (e.g. {@code HDMI-A-1})
      * @param connectorId the DRM connector ID ({@code -1} if not available)
+     * @param primary     whether this display is the primary display
      * @param xrandrData  the display's source of xrandr data, expected to be memoized or already realized
      */
-    private UnixDisplay(byte[] edid, String devicePort, int connectorId,
-            Supplier<Map<String, Pair<Integer, byte[]>>> xrandrData) {
+    private UnixDisplay(byte[] edid, String devicePort, int connectorId, boolean primary,
+            Supplier<Map<String, Triplet<Integer, byte[], Boolean>>> xrandrData) {
         super(edid);
         this.devicePort = devicePort;
         this.connectorId = connectorId;
+        this.primary = primary;
         this.xrandrData = xrandrData;
     }
 
@@ -76,18 +90,24 @@ public final class UnixDisplay extends AbstractDisplay {
         return Xrandr.findOutputName(this.xrandrData.get(), this.connectorId, this.getDisplayInfo().getEdid());
     }
 
+    @Override
+    public boolean isPrimary() {
+        return this.primary;
+    }
+
     /**
      * Gets Display Information from xrandr. Used as a fallback when DRM sysfs is not available.
      *
      * @return An array of Display objects representing monitors, etc.
      */
     public static List<Display> getDisplays() {
-        Map<String, Pair<Integer, byte[]>> data = Xrandr.getDisplayData();
+        Map<String, Triplet<Integer, byte[], Boolean>> data = Xrandr.getDisplayData();
         List<Display> displays = new ArrayList<>(data.size());
         // The data is already in hand, so these displays need no further xrandr query
-        Supplier<Map<String, Pair<Integer, byte[]>>> sharedData = () -> data;
-        for (Map.Entry<String, Pair<Integer, byte[]>> entry : data.entrySet()) {
-            displays.add(new UnixDisplay(entry.getValue().getB(), entry.getKey(), entry.getValue().getA(), sharedData));
+        Supplier<Map<String, Triplet<Integer, byte[], Boolean>>> sharedData = () -> data;
+        for (Map.Entry<String, Triplet<Integer, byte[], Boolean>> entry : data.entrySet()) {
+            displays.add(new UnixDisplay(entry.getValue().getB(), entry.getKey(), entry.getValue().getA(),
+                    entry.getValue().getC(), sharedData));
         }
         return displays;
     }
@@ -105,22 +125,25 @@ public final class UnixDisplay extends AbstractDisplay {
     }
 
     /**
-     * Builds a batch of displays sharing one query for the xrandr data behind {@link #getOutputName()}.
+     * Builds a batch of displays sharing one query for the xrandr data behind {@link #getOutputName()} and
+     * {@link #isPrimary()}.
      * <p>
      * The query is memoized indefinitely, because a {@link Display} is an immutable snapshot: the output name matching
-     * its connector cannot change over the object's lifetime. The hardware abstraction layer re-queries displays on its
-     * own schedule, building a new batch with a new supplier, so a topology change is picked up there.
+     * its connector and primary status cannot change over the object's lifetime. The hardware abstraction layer
+     * re-queries displays on its own schedule, building a new batch with a new supplier, so a topology change is
+     * picked up there.
      *
      * @param drmData     the DRM sysfs data to build displays from
      * @param xrandrQuery the query for xrandr display data, run at most once for the whole batch
      * @return An array of Display objects representing monitors, etc.
      */
     static List<Display> getDisplays(List<Triplet<String, Integer, byte[]>> drmData,
-            Supplier<Map<String, Pair<Integer, byte[]>>> xrandrQuery) {
+            Supplier<Map<String, Triplet<Integer, byte[], Boolean>>> xrandrQuery) {
         List<Display> displays = new ArrayList<>(drmData.size());
-        Supplier<Map<String, Pair<Integer, byte[]>>> sharedData = memoize(xrandrQuery);
+        Supplier<Map<String, Triplet<Integer, byte[], Boolean>>> sharedData = memoize(xrandrQuery);
         for (Triplet<String, Integer, byte[]> drm : drmData) {
-            displays.add(new UnixDisplay(drm.getC(), drm.getA(), drm.getB(), sharedData));
+            boolean primary = Xrandr.findPrimaryStatus(sharedData.get(), drm.getB(), drm.getC());
+            displays.add(new UnixDisplay(drm.getC(), drm.getA(), drm.getB(), primary, sharedData));
         }
         return displays;
     }
--- a/oshi-common/src/main/java/oshi/util/EdidUtil.java
+++ b/oshi-common/src/main/java/oshi/util/EdidUtil.java
@@ -115,6 +115,41 @@ public final class EdidUtil {
                 getAlphaNumericOrHex(edid[SERIAL_NUMBER_OFFSET + 1]), getAlphaNumericOrHex(edid[SERIAL_NUMBER_OFFSET]));
     }
 
+    /**
+     * Gets the packed 16-bit manufacturer/vendor number from EDID bytes 8 and 9, as a numeric value matching the macOS
+     * {@code CGDisplayVendorNumber} and the IOKit {@code kDisplayVendorID} property.
+     *
+     * @param edid The EDID byte array
+     * @return The packed vendor number (unsigned 16-bit value in the low 16 bits)
+     */
+    public static int getVendorNumber(byte[] edid) {
+        return ((edid[MANUFACTURER_ID_OFFSET] & 0xFF) << 8) | (edid[MANUFACTURER_ID_OFFSET + 1] & 0xFF);
+    }
+
+    /**
+     * Gets the 16-bit product/model number from EDID bytes 10 and 11 (little-endian), as a numeric value matching the
+     * macOS {@code CGDisplayModelNumber} and the IOKit {@code kDisplayProductID} property.
+     *
+     * @param edid The EDID byte array
+     * @return The product/model number (unsigned 16-bit value in the low 16 bits)
+     */
+    public static int getProductNumber(byte[] edid) {
+        return (edid[PRODUCT_ID_OFFSET] & 0xFF) | ((edid[PRODUCT_ID_OFFSET + 1] & 0xFF) << 8);
+    }
+
+    /**
+     * Gets the 32-bit serial number from EDID bytes 12-15 (little-endian), as a numeric value matching the macOS
+     * {@code CGDisplaySerialNumber} and the IOKit {@code kDisplaySerialNumber} property. The bit pattern is preserved
+     * in a signed Java {@code int}; equality comparisons against the corresponding CoreGraphics value are valid.
+     *
+     * @param edid The EDID byte array
+     * @return The serial number (bit pattern of the unsigned 32-bit value)
+     */
+    public static int getSerialNumber(byte[] edid) {
+        return (edid[SERIAL_NUMBER_OFFSET] & 0xFF) | ((edid[SERIAL_NUMBER_OFFSET + 1] & 0xFF) << 8)
+                | ((edid[SERIAL_NUMBER_OFFSET + 2] & 0xFF) << 16) | ((edid[SERIAL_NUMBER_OFFSET + 3] & 0xFF) << 24);
+    }
+
     private static String getAlphaNumericOrHex(byte b) {
         return Character.isLetterOrDigit((char) b) ? String.format(Locale.ROOT, "%s", (char) b)
                 : String.format(Locale.ROOT, "%02X", b);
--- a/oshi-common/src/main/java/oshi/util/driver/unix/Xrandr.java
+++ b/oshi-common/src/main/java/oshi/util/driver/unix/Xrandr.java
@@ -4,18 +4,12 @@
  */
 package oshi.util.driver.unix;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
 
 import oshi.annotation.concurrent.ThreadSafe;
 import oshi.util.ExecutingCommand;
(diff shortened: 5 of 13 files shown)


oshi-common/src/main/java/oshi/hardware/Display.java
```diff
@@ -88,4 +88,15 @@ default String getDevicePort() {
...
     }
+
+    /**
+     * Whether this display is the primary display. Returns {@code true} only when the platform positively identifies
+     * this display as primary; returns {@code false} when the primary display cannot be determined (e.g. under Wayland,
+     * on headless systems, or on platforms that do not expose primary-display information).
+     *
+     * @return {@code true} if this display is the primary display, {@code false} otherwise
+     */
+    default boolean isPrimary() {
```
dbwiddis, line 99: Heads up on release mechanics, not on this code: `default` does not keep this to a patch release. japicmp classifies a new default interface method as change type `METHOD_NEW_DEFAULT`, whose built-in semantic version level is MINOR, so `breakBuildBasedOnSemanticVersioning` fails the build against a `7.6.1` version:

```
Requires semantic version level MINOR: ... oshi.hardware.Display.isPrimary()
  getCompatibilityChanges()=[JApiCompatibilityChange{type=METHOD_NEW_DEFAULT}]
```

(Adding a method to a *class* — including an enum — is `METHOD_ADDED_TO_PUBLIC_CLASS`, which japicmp rates PATCH. That asymmetry is why a previous enum addition sailed through.) This will ship in 7.7.0.

On the javadoc itself: the `false`-means-either-"not-primary"-or-"unknown" conflation is documented clearly here, which is the right call for a `default`. Worth being aware that any third-party `Display` implementation, and the `oshi.nativefree` displays, will silently report `false` forever unless they opt in.

oshi-common/src/main/java/oshi/util/driver/unix/Xrandr.java
```diff
@@ -4,18 +4,12 @@
...
 package oshi.util.driver.unix;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
dbwiddis, line 7: `import java.util.*;` replaced the seven explicit imports and fails Checkstyle `AvoidStarImport`:

```
Xrandr.java:7:17: Using the '.*' form of import should be avoided.
```

Please restore the explicit imports.

oshi-common/src/main/java/oshi/util/driver/unix/Xrandr.java
```diff
@@ -156,14 +153,14 @@ static Map<String, Pair<Integer, byte[]>> getDisplayData(List<String> xrandr) {
      * @return an {@link Optional} containing the xrandr output name, or empty if no X server is available or no match
      *         is found
      */
-    public static Optional<String> findOutputName(Map<String, Pair<Integer, byte[]>> xrandrData, int connectorId,
+    public static Optional<String> findOutputName(Map<String, Triplet<Integer, byte[], Boolean>> xrandrData, int connectorId,
```
dbwiddis, line 156: This line and three in `XrandrTest.java` now exceed 120 characters after the `Pair` -> `Triplet` widening, so `spotless:check` fails. `./mvnw spotless:apply` fixes all four.

oshi-common/src/main/java/oshi/util/driver/unix/Xrandr.java
```diff
@@ -111,11 +107,12 @@ static Map<String, Pair<Integer, byte[]>> getDisplayData(List<String> xrandr) {
             if (!s.isEmpty() && !Character.isWhitespace(s.charAt(0))) {
                 // Flush the previous output before starting a new one
                 if (currentConnected && currentEdid != null) {
-                    results.put(currentPort, new Pair<>(currentConnectorId, currentEdid));
+                    results.put(currentPort, new Triplet<>(currentConnectorId, currentEdid, currentPrimary));
                 }
                 String[] words = ParseUtil.whitespaces.split(s.trim(), -1);
                 currentPort = words[0];
                 currentConnected = words.length > 1 && "connected".equals(words[1]);
+                currentPrimary = words.length > 2 && "primary".equals(words[2]);
```
dbwiddis, line 115: `words[2]` assumes `primary` is always the third field of the output header line. xrandr emits it there for connected outputs, but the surrounding fields vary (`connected`/`disconnected`, and the geometry token is absent on a connected-but-unconfigured output), so a positional index is more brittle than the rest of this parser.

Scanning the tokens for a `primary` word would be equally short and wouldn't depend on position.

oshi-common/src/main/java/oshi/util/driver/unix/Xrandr.java
```diff
@@ -67,42 +61,44 @@ public static List<byte[]> getEdidArrays() {
...
      *
      * @param xrandr output of {@code xrandr --verbose}
-     * @return an ordered map of xrandr port name to {@link Pair} of connector ID ({@code -1} if not available) and EDID
-     *         byte array (at least 128 bytes). Only connected outputs with a valid EDID are included.
+     * @return an ordered map of xrandr port name to {@link Triplet} of connector ID ({@code -1} if not available), EDID
+     *         byte array (at least 128 bytes), and primary status ({@code true} if the output is marked primary). Only
+     *         connected outputs with a valid EDID are included.
      */
-    static Map<String, Pair<Integer, byte[]>> getDisplayData(List<String> xrandr) {
+    static Map<String, Triplet<Integer, byte[], Boolean>> getDisplayData(List<String> xrandr) {
```
dbwiddis, line 94: `Pair<Integer, byte[]>` -> `Triplet<Integer, byte[], Boolean>` makes these signatures quite hard to read, and it repeats across nine call sites plus the tests. A fourth attribute would mean a `Quartet`.

I'm ok with this (it's the reason the tuple classes exist) but a small `XrandrOutput` value class (connectorId, edid, primary) would read better and make future additions free.

oshi-common/src/main/java/oshi/util/driver/unix/Xrandr.java
```diff
@@ -182,6 +179,42 @@ public static Optional<String> findOutputName(Map<String, Pair<Integer, byte[]>>
...
+     * Finds the primary status for a display identified by its DRM connector ID and/or EDID, matching by
+     * {@code CONNECTOR_ID} first (Linux 6.5+) and falling back to EDID comparison.
+     *
+     * @param xrandrData  xrandr display data as returned by {@link #getDisplayData()}, which the caller is expected to
+     *                    share among the displays it is naming rather than querying per display
+     * @param connectorId the DRM connector ID ({@code -1} if not available)
+     * @param edid        the EDID byte array from DRM sysfs
+     * @return {@code true} if the matched xrandr output is marked as primary, {@code false} otherwise
+     */
+    public static boolean findPrimaryStatus(Map<String, Triplet<Integer, byte[], Boolean>> xrandrData, int connectorId,
```
dbwiddis, line 192: `findPrimaryStatus` duplicates `findOutputName`'s matching strategy almost line for line — connector-ID scan first, then the 128-byte EDID comparison fallback. Two copies of "how do I correlate a DRM connector with an xrandr output" will drift.

Suggest one private matcher that returns the matched entry (or `Optional.empty()`), with `findOutputName` and `findPrimaryStatus` reading their respective field off it. Per AGENTS.md's dedup guidance this is the kind of thing that should exist once.

oshi-common/src/main/java/oshi/hardware/common/platform/unix/UnixDisplay.java
```diff
@@ -105,22 +125,25 @@ public static List<Display> getDisplays(List<Triplet<String, Integer, byte[]>> d
...
      */
     static List<Display> getDisplays(List<Triplet<String, Integer, byte[]>> drmData,
-            Supplier<Map<String, Pair<Integer, byte[]>>> xrandrQuery) {
+            Supplier<Map<String, Triplet<Integer, byte[], Boolean>>> xrandrQuery) {
         List<Display> displays = new ArrayList<>(drmData.size());
-        Supplier<Map<String, Pair<Integer, byte[]>>> sharedData = memoize(xrandrQuery);
+        Supplier<Map<String, Triplet<Integer, byte[], Boolean>>> sharedData = memoize(xrandrQuery);
         for (Triplet<String, Integer, byte[]> drm : drmData) {
-            displays.add(new UnixDisplay(drm.getC(), drm.getA(), drm.getB(), sharedData));
+            boolean primary = Xrandr.findPrimaryStatus(sharedData.get(), drm.getB(), drm.getC());
```
dbwiddis, line 145: This is the main design concern in the PR: `sharedData.get()` here runs `xrandr --verbose` eagerly, during `getDisplays()`, for every enumeration. The whole point of the memoized-supplier plumbing was that the subprocess only runs if a caller actually asks for something that needs it — a caller that only reads EDID, width, and height previously spawned nothing.

Making `isPrimary()` lazy the same way `getOutputName()` already is restores that, and has the pleasant side effect of deleting the `primary` field and the new public 4-arg constructor below:

```java
@Override
public boolean isPrimary() {
    return Xrandr.findPrimaryStatus(xrandrData.get(), connectorId, edid);
}
```

I do see the argument in the javadoc above for snapshotting at construction, but `getOutputName()` already accepts the same staleness window and resolves it lazily, so the two should behave alike.

oshi-common/src/main/java/oshi/hardware/common/platform/unix/UnixDisplay.java
```diff
@@ -47,7 +47,19 @@ public UnixDisplay(byte[] edid) {
...
+
+    /**
+     * Constructor for UnixDisplay with device port, DRM connector ID, and primary status.
+     *
+     * @param edid        a byte array representing a display EDID
+     * @param devicePort  the DRM connector name (e.g. {@code HDMI-A-1})
+     * @param connectorId the DRM connector ID ({@code -1} if not available)
+     * @param primary     whether this display is the primary display
+     */
+    public UnixDisplay(byte[] edid, String devicePort, int connectorId, boolean primary) {
```
dbwiddis, line 61: This new public constructor is permanent API surface (`UnixDisplay` is public in `oshi-common`). If `isPrimary()` becomes lazy as suggested at the batch builder below, it isn't needed at all — worth avoiding the commitment if it's avoidable.

oshi-common/src/test/java/oshi/hardware/common/platform/unix/UnixDisplayTest.java
```diff
@@ -57,13 +56,38 @@ void testBatchSharesOneXrandrQuery() {
...
-    void testBatchQueryIsNotRunUntilOutputNameIsRequested() {
+    void testBatchQueryIsRunOnceForPrimaryStatus() {
         AtomicInteger queries = new AtomicInteger();
         UnixDisplay.getDisplays(drmData(), () -> {
             queries.incrementAndGet();
             return xrandrData();
         });
-        assertThat(queries.get(), is(0));
+        // Primary status is computed eagerly during display construction
+        assertThat(queries.get(), is(1));
```
dbwiddis, line 66: This assertion is the eager-query regression stated out loud. On master the test was `testBatchQueryIsNotRunUntilOutputNameIsRequested` and asserted `is(0)` — that zero *was* the behavior under test, not an incidental detail.

If `isPrimary()` becomes lazy, please restore the original test alongside the new ones rather than replacing it; it's the only thing guarding the "no subprocess unless asked" property.

oshi-common/src/main/java/oshi/util/EdidUtil.java
```diff
@@ -115,6 +115,41 @@ public static String getSerialNo(byte[] edid) {
...
     }
 
+    /**
+     * Gets the packed 16-bit manufacturer/vendor number from EDID bytes 8 and 9, as a numeric value matching the macOS
+     * {@code CGDisplayVendorNumber} and the IOKit {@code kDisplayVendorID} property.
+     *
+     * @param edid The EDID byte array
+     * @return The packed vendor number (unsigned 16-bit value in the low 16 bits)
+     */
+    public static int getVendorNumber(byte[] edid) {
```
dbwiddis, line 125: These three read bytes 8-15 with no length check, and the mac call path that feeds them guards only `length > 0`, so a short or truncated EDID throws `ArrayIndexOutOfBoundsException` out of `getDisplays()`.

The pre-existing `EdidUtil` methods are equally unguarded, so this is consistent with the file — but those are reached from code paths that have already validated a 128-byte block, whereas this is the first path to parse an array it only checked for non-emptiness. Either tighten the call-site guard to a full 128 bytes or have these return a sentinel on a short array.

The byte order and the CoreGraphics equivalences in the javadoc are correct, for what it's worth: vendor big-endian at 8-9, product little-endian at 10-11, serial little-endian at 12-15.

oshi-core/src/main/java/oshi/hardware/platform/mac/MacDisplayJNA.java
```diff
@@ -209,7 +299,7 @@ private static List<Display> getAppleSiliconBuiltInDisplay() {
     // Synthesizes a display for the built-in panel from its DisplayAttributes dictionary. External framebuffer nodes
     // (marked with "external" = true) are skipped, as are idle pipes with no DisplayAttributes.
     private static void addBuiltInDisplay(IORegistryEntry fb, CFStringRef cfExternal, CFStringRef cfAttrs,
-            List<Display> displays) {
+            List<Display> displays, int mainDisplayId) {
```
dbwiddis, line 302: `mainDisplayId` is never read in this method, which fails the Error Prone build at ERROR:

```
MacDisplayJNA.java:[301,41] [UnusedVariable] The parameter 'mainDisplayId' is never read.
```

The body re-derives the id with `findBuiltInDisplayId()` instead. Using the parameter fixes the build error and the redundant query below in one go. Note the Error Prone job isn't in the PR check set, so this only surfaces under `-Perror-prone`.

oshi-core/src/main/java/oshi/hardware/platform/mac/MacDisplayJNA.java
```diff
@@ -242,7 +332,16 @@ private static void addBuiltInDisplay(IORegistryEntry fb, CFStringRef cfExternal
         try {
             DisplayInfo info = synthesize(fb, new CFDictionaryRef(attrsRaw.getPointer()), devicePort);
             if (info != null) {
-                displays.add(new MacDisplayJNA(info, devicePort));
+                int builtInId = findBuiltInDisplayId();
```
dbwiddis, line 335: `findBuiltInDisplayId()` makes two `CGGetActiveDisplayList` calls, and this sits inside the framebuffer iteration in `getAppleSiliconBuiltInDisplay`, so it repeats per framebuffer. `mainDisplayId` was already computed once in `getDisplays()` and passed all the way down to this method — using it here removes the repeated native calls and resolves the `UnusedVariable` error above.

The built-in display is the main display exactly when `mainDisplayId` equals the built-in id, so I think this reduces to a comparison against the parameter.

oshi-core/src/main/java/oshi/hardware/platform/mac/MacDisplayJNA.java
```diff
@@ -53,14 +53,34 @@ final class MacDisplayJNA extends AbstractDisplay {
     private static final CFIndex K_CF_NUMBER_SINT64 = new CFIndex(4);
 
     private final String devicePort;
+    private final boolean primary;
+
+    /**
+     * Value object holding the CoreGraphics main display identity for correlating IOKit-enumerated external displays.
+     */
+    private static final class MainDisplayIdentity {
```
dbwiddis, line 61: `MainDisplayIdentity` is duplicated verbatim in `MacDisplayFFM` (same five fields, same constructor), as is `getMainDisplayIdentity` and `countDisplaysWithIdentity` modulo the native call syntax. None of it touches JNA or FFM types.

Per AGENTS.md this belongs in `oshi-common` — the value class and the ambiguity logic hoisted, leaving only the three CoreGraphics calls in each twin. That's the difference between fixing the ambiguity bug below once versus twice.

oshi-core/src/main/java/oshi/hardware/platform/mac/MacDisplayJNA.java
```diff
@@ -150,7 +231,15 @@ private static List<Display> getDisplaysFromService(String serviceName, String e
                                             : propertySource.getStringProperty(portKeyName);
                                     String devicePort = ParseUtil
                                             .getStringValueOrUnknown(ParseUtil.getStringBefore(transport, '/'));
-                                    displays.add(new MacDisplayJNA(p.getByteArray(0, length), devicePort));
+                                    byte[] edidBytes = p.getByteArray(0, length);
+                                    // Correlate EDID identity with the CoreGraphics main display identity.
+                                    boolean primary = false;
+                                    if (mainIdentity.valid && !mainIdentity.ambiguous) {
+                                        primary = EdidUtil.getVendorNumber(edidBytes) == mainIdentity.vendor
```
dbwiddis, line 238: This three-way EDID-identity comparison is duplicated exactly in `MacDisplayFFM`. Since the operands are all `EdidUtil` calls and plain ints, a single `EdidUtil.matchesIdentity(edid, vendor, product, serial)` would remove the twin copy and give the length-guard concern one place to live.

oshi-core/src/main/java/oshi/hardware/platform/mac/MacDisplayJNA.java
```diff
@@ -369,4 +468,24 @@ private static int findBuiltInDisplayId() {
         }
         return null;
     }
+
+    // Counts how many active CoreGraphics displays share the given (vendor, product, serial) identity.
+    private static int countDisplaysWithIdentity(CoreGraphics cg, int vendor, int product, int serial) {
```
dbwiddis, line 473: Two gaps in the ambiguity guard, both of which can produce a false `isPrimary() == true`:

1. **A failed query is indistinguishable from a count of zero.** Both `CGGetActiveDisplayList` failure paths `return 0`, so the caller computes `ambiguous = 0 > 1` -> `false` and proceeds to trust the identity match. A CoreGraphics failure should make the identity *ambiguous* (or invalid), not unambiguous. CodeRabbit flagged this one too.

2. **The population counted isn't the population enumerated.** This counts CoreGraphics *active* displays, while the displays being labelled come from IOKit. A second identical monitor that IOKit sees but CoreGraphics does not report as active — asleep, mirrored away, or attached to an inactive path — never enters the count, so ambiguity is missed and both monitors match the identity and both report primary.

Worth deciding how much this matters. Two displays with identical vendor+product+serial is the pathological case (some vendors do ship zero serials), and the consequence is two displays claiming primary rather than a crash. But if that's an acceptable risk it should be a comment saying so.

oshi-core-ffm/src/main/java/oshi/hardware/platform/mac/MacDisplayFFM.java
```diff
@@ -151,7 +221,7 @@ private static List<Display> getAppleSiliconBuiltInDisplay() {
     // Synthesizes a display for the built-in panel from its DisplayAttributes dictionary. External framebuffer nodes
     // (marked with "external" = true) are skipped, as are idle pipes with no DisplayAttributes.
     private static void addBuiltInDisplay(IORegistryEntry fb, CFStringRef cfExternal, CFStringRef cfAttrs,
-            List<Display> displays) {
+            List<Display> displays, int mainDisplayId) {
```
dbwiddis, line 224: Same dead `mainDisplayId` parameter as the JNA twin, with `findBuiltInDisplayId()` re-derived at line 253. This file wasn't compiled in the failing Error Prone run — the build stopped at `oshi-core` first — so expect the same `UnusedVariable` error here once `MacDisplayJNA` is fixed.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsDisplayJNA.java
```diff
@@ -56,14 +58,29 @@ final class WindowsDisplayJNA extends AbstractDisplay {
     private static final int QDC_ATTEMPTS = 3;
 
     private final String devicePort;
+    private final boolean primary;
+
+    /**
+     * Value object holding the results of a CCD display configuration query: the connector-name map and the set of
+     * normalized monitor device paths that belong to the Windows primary display (source mode position 0,0).
+     */
+    private static final class DisplayConfig {
```
dbwiddis, line 67: `DisplayConfig` is duplicated verbatim in `WindowsDisplayFFM`, and so is `isSourceAtOrigin` below (identical logic, `Memory.getInt` vs `MemorySegment.get`). The struct offsets were correctly hoisted to `DisplayConnector` — the value class and the origin test can follow them into `oshi-common`, leaving only the memory-read call in each twin.

oshi-core/src/main/java/oshi/hardware/platform/windows/WindowsDisplayJNA.java
```diff
@@ -216,15 +252,36 @@ private static Map<String, String> queryConnectorPorts() {
...
+                    addConnector(portMap, primaryPaths, u32, adapterId, targetId, primaryPath);
                 }
             }
         }
-        return map;
+        return new DisplayConfig(portMap, primaryPaths);
+    }
+
+    // Returns true if the source mode for the given path has desktop position (0, 0).
+    private static boolean isSourceAtOrigin(Memory paths, long pathBase, Memory modes, int modeCount) {
```
dbwiddis, line 266: The offset arithmetic is right, and the `modeIdx` bounds check is a good guard. Two behavioral notes worth a comment in the code:

- Reading `modeInfoIdx` as a flat `UINT32` is only valid because OSHI calls `QueryDisplayConfig` without `QDC_VIRTUAL_MODE_AWARE`. With that flag the field becomes a union of `{cloneGroupId, sourceModeInfoIdx}` bitfields and this silently reads the wrong value. A one-line comment would save the next person the header dive.
- Cloned displays share one source mode, so every target on the primary source reports `isPrimary() == true`. Defensible, but it means `isPrimary()` is not guaranteed to be unique across the returned list — which callers will assume unless the interface javadoc says otherwise.

oshi-common/src/main/java/oshi/driver/common/windows/DisplayConnector.java
```diff
@@ -64,6 +64,19 @@ private DisplayConnector() {
     /** Offset of {@code monitorDevicePath} (a {@code WCHAR[128]}) within a {@code DISPLAYCONFIG_TARGET_DEVICE_NAME}. */
     public static final int TDN_MONITOR_DEVICE_PATH_OFFSET = 164;
 
+    // Offsets for determining the primary display from the DISPLAYCONFIG_PATH_INFO and DISPLAYCONFIG_MODE_INFO arrays.
+    // The Windows primary display is the active path whose source mode desktop position is (0, 0).
+    /** Offset of {@code sourceInfo.modeInfoIdx} (a {@code UINT32}) within a {@code DISPLAYCONFIG_PATH_INFO}. */
+    public static final int PATH_SOURCE_MODE_IDX_OFFSET = 12;
```
dbwiddis, line 70: Checked all five of these against the Win32 headers and they're correct: `sourceInfo` is the first member of `DISPLAYCONFIG_PATH_INFO` at offset 0, its `LUID adapterId` occupies 8 bytes and `UINT32 id` the next 4, putting `modeInfoIdx` at 12; `DISPLAYCONFIG_MODE_INFO` leads with `infoType` at 0, and `DISPLAYCONFIG_SOURCE_MODE` at offset 20 has `width`/`height`/`pixelFormat` before `POINTL position`, giving 28 and 32.

Nice that these landed in `oshi-common` next to `MODE_INFO_SIZE` rather than being duplicated in both twins.

dbwiddis (commented):
Thanks for taking this on, @ayonization — the feature is a genuinely useful addition and the platform work is thorough. The Windows CCD offsets are all correct against the headers (I checked each one), and the order-independent xrandr parsing and the connector-ID-then-EDID fallback are the right shapes.

Three things need to happen before this can merge, and then a handful of design points I'd like to work through.

**Release timing.** This can't ship as a patch. `default` prevents a *major* bump but not a minor one: japicmp classifies a new default interface method as `METHOD_NEW_DEFAULT`, rated MINOR, so the build fails against a 7.6.1 version. That's on me to sort out, not you — it means this goes into 7.7.0. No code change needed, I'll probalby do the version bump in a separate commit and rebase against it.  But recognize that this review/release cycle will probably be longer.

**Failing CI.**
- `spotless:check` — four lines over 120 characters from the `Pair` -> `Triplet` widening. `./mvnw spotless:apply`.
- Checkstyle — `AvoidStarImport` on the new `import java.util.*;` in `Xrandr.java`.
- "Lint (ForbiddenAPIs, Javadoc, SLF4J)" — this is actually the japicmp failure in that job's `install` step, i.e. the release-timing item above. forbiddenapis and javadoc both pass on their own.

**A build error CI didn't run.** Error Prone isn't in the PR check set, but `-Perror-prone` fails at ERROR on an unused `mainDisplayId` parameter in `MacDisplayJNA` — the value is plumbed three levels down and then ignored in favor of re-deriving it. The FFM twin has the same shape and will report the same thing once the first one is fixed. Details inline.

**Design points**, in rough priority order:
1. `isPrimary()` is computed eagerly, so `getDisplays()` now always spawns `xrandr --verbose` on Linux. That undoes the laziness the memoized-supplier plumbing exists for; the replaced `testBatchQueryIsNotRunUntilOutputNameIsRequested` was guarding exactly that property. Making `isPrimary()` lazy like `getOutputName()` also removes the new field and the new public constructor.
2. The macOS ambiguity guard treats a failed CoreGraphics query as "unambiguous," and counts CG-active displays while labelling IOKit-enumerated ones. Both can yield a false positive primary.
3. Several pieces are duplicated across the JNA/FFM twins — `MainDisplayIdentity`, `DisplayConfig`, `isSourceAtOrigin`, the EDID-identity comparison — none of which touch JNA or FFM types. These should be in `oshi-common` so the JNA/FFM implementations can't diverge.
4. `Xrandr.findPrimaryStatus` duplicates `findOutputName`'s matching logic; and the `Triplet` growth is at the point where a small value class would read better.
5. The new `EdidUtil` numeric getters read bytes 8-15 behind a call-site guard of only `length > 0`.

Also needs a `CHANGELOG.md` entry under *Next Release* — this is user-visible.

I have not yet verified the behavior on macOS, can you update where you've actually tested this code?

